PR #5469, opened 2019-03-15
Remove Haystack reference in Docs

saadmk11:
closes #5462

PR #5473, opened 2019-03-16
Remove Haystack reference in Docs thus addressing issue #5462

rshrc:
Addressing issue #5462. Haystack in the docs can be confusing because it can be confusing. Since the dependency was removed a time ago #4039


dojutsu-user:
@rshrc 
There's already an open PR for this #5469

rshrc:
> @rshrc
> There's already an open PR for this #5469

Overlooked. Gonna close it.

PR #5157, opened 2019-01-23
Sort versions smartly everywhere

dojutsu-user:
Closes #5151 

It wasn't possible to use the `sort_version_compare` only on the template level. So, I have also used it in view and forms level.

PR #5475, opened 2019-03-17
Refactor sitemap sort order 

### LINKED ISSUE #5447: Sitemap sort order may prioritise unstable/development documentation
In `sitemap.xml`, the sort order should prioritise the user-selected default version (Admin > Versions > Default Version) in the backend instead of setting `latest` to highest priority. In many cases (and also by [definition](https://docs.readthedocs.io/en/stable/versions.html)), `latest` points at development documentation while `stable` is the version most people should be using, and should appear first in search results.

Example from: https://docs.dash.org/sitemap.xml
`stable` should be higher priority than `latest`, but the reverse is true.
```
<url>
  <loc>http://docs.dash.org/en/latest/</loc>
  <changefreq>daily</changefreq>
  <priority>1</priority>
</url>

<url>
  <loc>http://docs.dash.org/en/stable/</loc>
  <changefreq>weekly</changefreq>
  <priority>0.9</priority>
</url>
```

mashrikt:
This is an initial fix.

Closes #5447

### DIFF
--- a/readthedocs/projects/version_handling.py
+++ b/readthedocs/projects/version_handling.py
@@ -43,8 +43,8 @@ def comparable_version(version_string):
     """
     Can be used as ``key`` argument to ``sorted``.
 
-    The ``LATEST`` version shall always beat other versions in comparison.
-    ``STABLE`` should be listed second. If we cannot figure out the version
+    The ``STABLE`` version shall always beat other versions in comparison.
+    ``LATEST`` should be listed second. If we cannot figure out the version
     number then we sort it to the bottom of the list.
 
     :param version_string: version as string object (e.g. '3.10.1' or 'latest')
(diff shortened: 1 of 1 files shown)


readthedocs/projects/version_handling.py
```diff
@@ -56,9 +56,9 @@ def comparable_version(version_string):
     """
     comparable = parse_version_failsafe(version_string)
     if not comparable:
-        if version_string == LATEST_VERBOSE_NAME:
+        if version_string == STABLE_VERBOSE_NAME:
```
dojutsu-user: I don't think that we should do it like this because -

https://github.com/rtfd/readthedocs.org/blob/93bd03b4a8ed42c3a0b3633f114f21a0eac9ea11/readthedocs/projects/templatetags/projects_tags.py#L13-L20

is using `comparable_version` and changing this means that the order in which versions are rendered everywhere will get disturbed. #5157.

PR #5491, opened 2019-03-19
Added note warning about using sqlite 3.26.0 in development

rshrc:
Added a note in the developer guide which asks the developer to use python version 3.6.7 instead of python 3.6.8 which would use a higher version of sqlite3 giving rise to issur #5477

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -11,6 +11,12 @@ First, obtain `Python 3.6`_ and virtualenv_ if you do not already have them.
 Using a virtual environment is strongly recommended,
 since it will help you to avoid clutter in your system-wide libraries.
 
+.. note::
+
+    Currently the recommended version of python is Python 3.6.7. Which uses a 
+    lower version of sqlite3. The next version of Python can be used when the 
+    project starts using the latest Django version.
+
 Additionally Read the Docs depends on:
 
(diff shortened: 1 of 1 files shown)


docs/install.rst
```diff
@@ -11,6 +11,12 @@ First, obtain `Python 3.6`_ and virtualenv_ if you do not already have them.
 Using a virtual environment is strongly recommended,
 since it will help you to avoid clutter in your system-wide libraries.
 
+.. note::
```
humitos: This should probably be a `warning` instead.

ericholscher: I'd really like to keep these edge cases out of the default flow for the install guide, otherwise it just gets messy. We should add them to a troubleshooting section at the bottom of the guide, with the error that they fix.

humitos: The thing is that it's not an edge case.

All users installing Python 3.6 with an up to date system will hit this problem. The issue is fixed on Django 2.x but we are using 1.11.x. So, all new contributors will hit it until we upgrade our django version.

ericholscher: OK, thanks sense then.

ericholscher: OK, thanks sense then.

PR #5509, opened 2019-03-20
Use ValueError instead of InvalidParamsException

dojutsu-user:
Ref: https://github.com/rtfd/readthedocs.org/pull/5452#discussion_r267412964


humitos:
:100: Thanks!

PR #5522, opened 2019-03-22
 Update contributing docs for RTD's own docs

davidjb:
This makes a few improvements to the documentation for Read the Docs itself:

* Better explains how to perform a local install for working on the docs
* Explains there's a separate page for guidelines for docs on the contributing page
* Updates the URL in GitHub's contributing doc to use https + new readthedocs.io domain


davidjb:
All good, the docs install instructions are a little janky (eg stop half way through the install process and come back) so perhaps that could be improved with substitutions or the like, but this seems the lightest touch for now!

PR #5530, opened 2019-03-22
Optimize database performance of the footer API

### LINKED ISSUE #3712: Improve URL resolver performance
Our URL resolver (https://github.com/rtfd/readthedocs.org/blob/4d10f2e/readthedocs/core/resolver.py) which turns a `Project` and/or `Version` into a URL to their documentation needs some performance improvements to reduce or eliminate the number of database queries it makes.

This is particularly noticeable on the `footer_html` API call (https://github.com/rtfd/readthedocs.org/blob/ba71aa1/readthedocs/restapi/templates/restapi/footer.html#L43) which is called on every docs pageview to inject the version selector. In order to get the URL to each version in the version selector, four database queries are needed (to make sure this isn't a subproject or translation and get the domain). So for a project with 10 active versions, we are looking at 40 database queries. Most of this effort is wasted since the data is only needed once for the project.

davidfischer:
Currently, the number of database queries in the footer API depends on the number of versions of the project in question. This optimizes the API to always be a constant number of queries.

This PR partially fixes #3712 in that it fixes the major performance issues with the resolver when generating a number of URLs for different versions of the same project. Currently `project.ordered_active_versions` is only used by the footer API but anywhere else looking to generate URLs for a list of versions could use that function to get the same performance benefits.

The test case does seem a little bit brittle but I'm not sure what the best way to improve that is.

### DIFF
--- a/readthedocs/core/resolver.py
+++ b/readthedocs/core/resolver.py
@@ -1,11 +1,14 @@
 """URL resolver for documentation."""
 
+import logging
 from urllib.parse import urlunparse
 
 from django.conf import settings
 
 from readthedocs.core.utils.extend import SettingsOverrideObject
 from readthedocs.projects.constants import PRIVATE, PUBLIC
+
+log = logging.getLogger(__name__)
 
 
 class ResolverBase:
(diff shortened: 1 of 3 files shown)


davidfischer:
> Will be curios to see how this improves performance on the footer API

This should be a large improvement but it's hard to say exactly how large it will be because footer API calls with 1-2 versions won't be improved as much.

> I wonder if it's worth another PR to add these improvements to the version listing page, at the least?

That's a good idea.

ericholscher:
Gonna ship this today.

PR #5535, opened 2019-03-25
Sitemap assumes that all versions are translated Fix.

### LINKED ISSUE #5365: Sitemap assumes that all versions are translated
When the project has a translation, URLs for _all the versions_ from the original one are generated and this is not true in most of the cases.

Example,

    http http://docs.python-requests.org/sitemap.xml

`latest` is the only one translated for `es`. So, the rest of alternate rel link should not be shown in the `sitemap.xml` file.

At this point we need to check that the version of the translation exists, https://github.com/rtfd/readthedocs.org/blob/1615242365ee0663fa0b744a78af8b5cb195d056/readthedocs/core/views/serve.py#L403-L412

saadmk11:
closes #5365

### DIFF
--- a/readthedocs/core/views/serve.py
+++ b/readthedocs/core/views/serve.py
@@ -379,7 +379,6 @@ def sitemap_xml(request, project):
             only_active=True,
         ),
     )
-
     versions = []
     for version, priority, changefreq in zip(
             sorted_versions,
@@ -401,15 +400,18 @@ def sitemap_xml(request, project):
 
         if project.translations.exists():
             for translation in project.translations.all():
-                href = project.get_docs_url(
-                    version_slug=version.slug,
-                    lang_slug=translation.language,
-                    private=version.privacy_level == constants.PRIVATE,
-                )
-                element['languages'].append({
-                    'hreflang': translation.language,
-                    'href': href,
-                })
+                translation_versions = translation.versions.public(
+                    ).values_list('slug', flat=True)
+                if version.slug in translation_versions:
+                    href = project.get_docs_url(
+                        version_slug=version.slug,
+                        lang_slug=translation.language,
+                        private=version.privacy_level == constants.PRIVATE,
+                    )
+                    element['languages'].append({
+                        'hreflang': translation.language,
+                        'href': href,
+                    })
 
             # Add itself also as protocol requires
             element['languages'].append({
--- a/readthedocs/core/views/serve.py
+++ b/readthedocs/core/views/serve.py
@@ -406,7 +406,7 @@ def sitemap_xml(request, project):
                     href = project.get_docs_url(
                         version_slug=version.slug,
                         lang_slug=translation.language,
-                        private=version.privacy_level == constants.PRIVATE,
+                        private=False,
                     )
                     element['languages'].append({
                         'hreflang': translation.language,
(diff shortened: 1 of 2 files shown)


readthedocs/rtd_tests/tests/test_doc_serving.py
```diff
@@ -260,3 +274,11 @@ def test_sitemap_xml(self):
                 private=True,
             ),
         )
+        self.assertNotContains(
```
humitos: This test looks tricky to me.

We are adding a new public translated version of `stable`, which I expect to be listed in the response, but we are asserting to not contains. Why it does not appear in the response?

humitos: I think the test we want to is a project with 3 public versions and only 1 version translated. So, in the response 2 should not appear and 1 should.

I'm not sure if I'm confused or missing something here.

readthedocs/core/views/serve.py
```diff
@@ -401,15 +400,18 @@ def changefreqs_generator():
...
-                    'hreflang': translation.language,
-                    'href': href,
-                })
+                translation_versions = translation.versions.public(
+                    ).values_list('slug', flat=True)
+                if version.slug in translation_versions:
+                    href = project.get_docs_url(
+                        version_slug=version.slug,
+                        lang_slug=translation.language,
+                        private=version.privacy_level == constants.PRIVATE,
```
humitos: This can be `False` directly, we are already filtering by `.public`.

readthedocs/rtd_tests/tests/test_doc_serving.py
saadmk11: @humitos Here the public project has 2 public versions `latest` and `stable` (ie: Latest version is created automatically). but the translation project has only one version which is `latest`. so, I'm checking that as the translation project does not have a `stable` version, `stable` should not have an URL for that Translation in the sitemap.

readthedocs/core/views/serve.py
saadmk11: okay :+1:

readthedocs/rtd_tests/tests/test_doc_serving.py
humitos: Makes sense.

Can you add this explanation as a comment above the assert line?

When creating this tests, I like to use descriptive names on things when possible. For example, the slug for the `stable` version in this case could be `not-translated-version` (`stable` is confusing since it's also created automatically).

Besides, the slug for the translated project could be something similar to `translation-es` and similar.

humitos: Then, when checking for the assert everything is clearer:

```
          self.public.get_docs_url(
                version_slug=not_translated_version.slug,
                lang_slug=translation.language,
                private=False,
            ),
```

It easier to note that the "not translated version" should not be shown on the sitemap xml because it's not translated :)

saadmk11: @humitos Thanks for Pointing it out. I'll keep that in mind. I have updated the PR have a Look.

saadmk11:
@humitos Thank you for you guidance! :100:

PR #5552, opened 2019-03-30
Specify python3 in installation instructions.

shivanshu1234:
Closes https://github.com/rtfd/readthedocs.org/issues/5550

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -92,7 +92,7 @@ Clone the repository somewhere on your disk and enter to the repository::
 
 Create a virtual environment and activate it::
 
-    virtualenv venv
+    virtualenv --python=python3 venv
     source venv/bin/activate
 
 Next, install the dependencies using ``pip``


jasonaowen:
> Thanks, I wonder if can just use `python -m venv venv`

That would work (I just tested it), but we'd also need to update at least the Ubuntu required packages to include `python3-venv`:

```
$ python3 -m venv venv 
The virtual environment was not created successfully because ensurepip is not
available.  On Debian/Ubuntu systems, you need to install the python3-venv
package using the following command.

    apt-get install python3-venv

You may need to use sudo with that command.  After installing the python3-venv
package, recreate your virtual environment.

Failing command: ['/tmp/venv/bin/python3', '-Im', 'ensurepip', '--upgrade', '--default-pip']
```

shivanshu1234:
Yes, can confirm this. [Ubuntu Bug #1532231](https://bugs.launchpad.net/ubuntu/+source/python3.4/+bug/1532231)

jasonaowen:
It looks like the `python3-venv` packages exists on [Xenial (16.04)](https://packages.ubuntu.com/xenial/python3-venv) and later. After installing it on my 18.04 machine, I was able to run `python -m venv venv` successfully, and it created an environment where `python` was linked to the system `python3.6`.

humitos:
Thanks!

PR #5571, opened 2019-04-04
Fix for Firefox to close the ad correctly

davidfischer:
Firefox handles `<a href="javascript:">` links differently.


humitos:
I didn't really tested it but looks accurate.

PR #5588, opened 2019-04-12
Unify usage of settings and remove the usage of getattr for settings

### LINKED ISSUE #2140: Unify usage of settings
There are a number of place that we are using `getattr(settings, <setting>, <default>)`, with varying defaults for the same setting. We should either:
- Trust the base configuration as canonical and stop trying to redefine settings defaults everywhere
- Use a secondary module to import settings into and set defaults in a common area
- Break out defaults for use in various calls to `getattr` -- though this duplicates the default in base config, unless we remove settings there

Raised in #1829

## Resolution

All settings that we use should be implemented with a default in `settings.base`. `getattr` usage for settings will be phased out and can be refactored.

saadmk11:
I wasn't quite sure what to with this so I left it as it is:
https://github.com/rtfd/readthedocs.org/blob/7fa9002586b2c5c2fae5c78e5ad736d9d46f7ad4/readthedocs/core/utils/extend.py#L29

closes #2140

### DIFF
--- a/readthedocs/builds/syncers.py
+++ b/readthedocs/builds/syncers.py
@@ -5,7 +5,6 @@ Classes to copy files between build and web servers.
 local machine.
 """
 
-import getpass
 import logging
 import os
 import shutil
--- a/readthedocs/builds/syncers.py
+++ b/readthedocs/builds/syncers.py
@@ -64,8 +64,8 @@ class RemoteSyncer(BaseSyncer):
 
         Respects the ``MULTIPLE_APP_SERVERS`` setting when copying.
         """
-        sync_user = getattr(settings, 'SYNC_USER', getpass.getuser())
-        app_servers = getattr(settings, 'MULTIPLE_APP_SERVERS', [])
+        sync_user = settings.SYNC_USER
+        app_servers = settings.MULTIPLE_APP_SERVERS
         if app_servers:
             log.info('Remote Copy %s to %s on %s', path, target, app_servers)
             for server in app_servers:
@@ -103,8 +103,8 @@ class DoubleRemotePuller(BaseSyncer):
 
         Respects the ``MULTIPLE_APP_SERVERS`` setting when copying.
         """
-        sync_user = getattr(settings, 'SYNC_USER', getpass.getuser())
-        app_servers = getattr(settings, 'MULTIPLE_APP_SERVERS', [])
+        sync_user = settings.SYNC_USER
+        app_servers = settings.MULTIPLE_APP_SERVERS
         if not is_file:
             path += '/'
         log.info('Remote Copy %s to %s', path, target)
@@ -143,7 +143,7 @@ class RemotePuller(BaseSyncer):
 
         Respects the ``MULTIPLE_APP_SERVERS`` setting when copying.
         """
-        sync_user = getattr(settings, 'SYNC_USER', getpass.getuser())
+        sync_user = settings.SYNC_USER
         if not is_file:
             path += '/'
         log.info('Remote Pull %s to %s', path, target)
--- a/readthedocs/config/config.py
+++ b/readthedocs/config/config.py
@@ -66,18 +66,12 @@ SUBMODULES_INVALID = 'submodules-invalid'
 INVALID_KEYS_COMBINATION = 'invalid-keys-combination'
 INVALID_KEY = 'invalid-key'
 
-DOCKER_DEFAULT_IMAGE = getattr(
-    settings, 'DOCKER_DEFAULT_IMAGE', 'readthedocs/build'
-)
-DOCKER_DEFAULT_VERSION = getattr(settings, 'DOCKER_DEFAULT_VERSION', '2.0')
+DOCKER_DEFAULT_IMAGE = settings.DOCKER_DEFAULT_IMAGE
+DOCKER_DEFAULT_VERSION = settings.DOCKER_DEFAULT_VERSION
 # These map to corresponding settings in the .org,
 # so they haven't been renamed.
-DOCKER_IMAGE = getattr(
-    settings,
-    'DOCKER_IMAGE',
-    '{}:{}'.format(DOCKER_DEFAULT_IMAGE, DOCKER_DEFAULT_VERSION),
-)
-DOCKER_IMAGE_SETTINGS = getattr(settings, 'DOCKER_IMAGE_SETTINGS', {})
+DOCKER_IMAGE = settings.DOCKER_IMAGE
+DOCKER_IMAGE_SETTINGS =settings.DOCKER_IMAGE_SETTINGS
 
 LATEST_CONFIGURATION_VERSION = 2
 
--- a/readthedocs/config/config.py
+++ b/readthedocs/config/config.py
@@ -71,7 +71,7 @@ DOCKER_DEFAULT_VERSION = settings.DOCKER_DEFAULT_VERSION
 # These map to corresponding settings in the .org,
 # so they haven't been renamed.
 DOCKER_IMAGE = settings.DOCKER_IMAGE
-DOCKER_IMAGE_SETTINGS =settings.DOCKER_IMAGE_SETTINGS
+DOCKER_IMAGE_SETTINGS = settings.DOCKER_IMAGE_SETTINGS
 
 LATEST_CONFIGURATION_VERSION = 2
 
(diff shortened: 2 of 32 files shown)


readthedocs/settings/base.py
```diff
@@ -270,13 +296,19 @@ def USE_PROMOS(self):  # noqa
...
+    MULTIPLE_APP_SERVERS = [CELERY_DEFAULT_QUEUE]
+    MULTIPLE_BUILD_SERVERS = [CELERY_DEFAULT_QUEUE]
 
     # Sentry
     SENTRY_CELERY_IGNORE_EXPECTED = True
 
     # Docker
     DOCKER_ENABLE = False
+    DOCKER_SOCKET = 'unix:///var/run/docker.sock'
+    DOCKER_BUILD_IMAGES = None
```
stsewd: Maybe we can put a comment that this is deprecated in favor or `DOCKER_IMAGE_SETTINGS`

readthedocs/settings/base.py
```diff
@@ -270,13 +296,19 @@ def USE_PROMOS(self):  # noqa
...
+    DOCKER_LIMITS = {}
```
stsewd: We are missing some defaults here

readthedocs/settings/base.py
```diff
@@ -42,11 +44,18 @@ class CommunityBaseSettings(Settings):
     PUBLIC_DOMAIN = None
     PUBLIC_DOMAIN_USES_HTTPS = False
     USE_SUBDOMAIN = False
+    USE_SUBDOMAINS = None
```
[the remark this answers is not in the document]
saadmk11: I was a bit confused about this one is `USE_SUBDOMAIN` and the other one is `USE_SUBDOMAINS` Which is called here:
https://github.com/rtfd/readthedocs.org/blob/3d36585aca80e7f79b0c5bfeb1e4f6f1e38b8f06/readthedocs/core/context_processors.py#L13

readthedocs/settings/base.py
saadmk11: Okay :100:

readthedocs/builds/syncers.py
```diff
@@ -143,7 +143,7 @@ def copy(cls, path, target, host, is_file=False, **kwargs):  # pylint: disable=a
 
         Respects the ``MULTIPLE_APP_SERVERS`` setting when copying.
         """
-        sync_user = getattr(settings, 'SYNC_USER', getpass.getuser())
+        sync_user = settings.SYNC_USER
```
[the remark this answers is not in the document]
saadmk11: Oh! I missed that one. Thanks

readthedocs/settings/base.py
saadmk11: As the `getattr` only had an empty dict as the default so, I added this. Should I add this as the default?
https://github.com/rtfd/readthedocs.org/blob/3d36585aca80e7f79b0c5bfeb1e4f6f1e38b8f06/readthedocs/doc_builder/constants.py#L36

readthedocs/settings/base.py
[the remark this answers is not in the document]
stsewd: oh, I just realized of the final `s`. We should investigate a liitle more here.

readthedocs/settings/base.py
stsewd: Yeah

readthedocs/settings/base.py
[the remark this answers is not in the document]
saadmk11: @stsewd I did some investigation on my own. I found that this settings is only being used by `context_processors.py `https://github.com/rtfd/readthedocs.org/blob/7fa9002586b2c5c2fae5c78e5ad736d9d46f7ad4/readthedocs/core/context_processors.py#L13

saadmk11:
@stsewd Updated the PR with the changes you suggested :)

readthedocs/config/config.py
```diff
@@ -66,18 +66,12 @@
...
 # These map to corresponding settings in the .org,
 # so they haven't been renamed.
-DOCKER_IMAGE = getattr(
-    settings,
-    'DOCKER_IMAGE',
-    '{}:{}'.format(DOCKER_DEFAULT_IMAGE, DOCKER_DEFAULT_VERSION),
-)
-DOCKER_IMAGE_SETTINGS = getattr(settings, 'DOCKER_IMAGE_SETTINGS', {})
+DOCKER_IMAGE = settings.DOCKER_IMAGE
+DOCKER_IMAGE_SETTINGS = settings.DOCKER_IMAGE_SETTINGS
```
humitos: Why we are re-defining these here instead of using them directly from the `settings.` object in the code?

readthedocs/settings/base.py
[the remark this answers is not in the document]
humitos: I suppose this was a typo and the intended usage was `USE_SUBDOMAIN`.

readthedocs/config/config.py
saadmk11: @humitos I did not change it as it was already like this. And also they are being called in many parts of the `config.py` file. should I use them directly from the settings in the places of `config.py`?

stsewd: I vote to do that in another PR

readthedocs/settings/base.py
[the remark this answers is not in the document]
saadmk11: @humitos I thought that too. Should I Change this to `USE_SUBDOMAIN` and Update the PR?

davidfischer:
I just wanted to chime in that I'm excited for this change. This will simplify a lot of things and centralize all our settings in one place so they are easier to understand!

readthedocs/settings/base.py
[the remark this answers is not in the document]
stsewd: Looks like most of the code and tests we use `USE_SUBDOMAIN`, also I search the whole project (and the corporate site), it's safe to rename it in the context processor file.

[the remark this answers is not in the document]
saadmk11: Cool :100:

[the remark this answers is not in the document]
saadmk11: @stsewd  Updated the PR

readthedocs/core/context_processors.py
```diff
@@ -10,7 +10,7 @@ def readthedocs_processor(request):
     exports = {
         'PUBLIC_DOMAIN': settings.PUBLIC_DOMAIN,
         'PRODUCTION_DOMAIN': settings.PRODUCTION_DOMAIN,
-        'USE_SUBDOMAINS': settings.USE_SUBDOMAINS,
+        'USE_SUBDOMAINS': settings.USE_SUBDOMAIN,
```
stsewd: We should rename the key too (I just checked, and there is not other places using this)

saadmk11: sure!

saadmk11: @stsewd Done.:)

stsewd:
> I wasn't quite sure what to with this so I left it as it is:
https://github.com/rtfd/readthedocs.org/blob/7fa9002586b2c5c2fae5c78e5ad736d9d46f7ad4/readthedocs/core/utils/extend.py#L29

That should remain as it is.

saadmk11:
> We still to change more code

@stsewd I might have missed these two from my search. Updated the PR!

humitos:
This PR touches many different files. I'm merging it now to avoid conflicts and also to start testing it locally considering that there are many settings re-defined. Although, the defaults on `base.py` looks accurate and the same as the ones used by `getattr`. I'm not expecting to fail on prod, but we should be careful when deploying it.

ericholscher:
I think we should probably do a followup here to remove places where we're using `sync_user = settings.SYNC_USER` to use the setting directly. We're doing this a number of places in the code, and I think it's a bad pattern.

ericholscher:
This is fine for now though 👍

saadmk11:
> I think we should probably do a followup here to remove places where we're using `sync_user = settings.SYNC_USER` to use the setting directly. We're doing this a number of places in the code, and I think it's a bad pattern.

@ericholscher Okay! I will start working on it:)

humitos:
> I think we should probably do a followup here to remove places where we're using `sync_user = settings.SYNC_USER` to use the setting directly. We're doing this a number of places in the code, and I think it's a bad pattern.

@ericholscher @saadmk11 I already mentioned this at https://github.com/rtfd/readthedocs.org/pull/5588#discussion_r275119621

I think the new PR should include all the changes for this pattern.

PR #5593, opened 2019-04-15
Add meta descriptions to top documentation

davidfischer:
Initially my goal here is to just have slightly better blurbs on search engine result pages (SERPs).

Screen Shot 2019-04-15 at 11 41 09 AM
Screen Shot 2019-04-15 at 11 42 02 AM

### DIFF
--- a/docs/intro/import-guide.rst
+++ b/docs/intro/import-guide.rst
@@ -2,7 +2,7 @@ Importing Your Documentation
 ============================
 
 .. meta::
-   :description: Import your existing technical documentation from source control into Read the Docs.
+   :description lang=en: Import your existing technical documentation from source control into Read the Docs.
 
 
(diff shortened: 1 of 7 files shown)


docs/intro/import-guide.rst
```diff
@@ -1,6 +1,10 @@
 Importing Your Documentation
 ============================
 
+.. meta::
+   :description lang=en: Import your existing technical documentation from source control into Read the Docs.
```
ericholscher: I wonder if we should choose "source control" or "version control" -- we should pick one.

davidfischer: Version control is probably better and I changed it to that one. The one positive of source control is that it doesn't overload the term "version" which RTD uses.

PR #5596, opened 2019-04-16
Notify the user when deleting a superproject

### LINKED ISSUE #3742: Notify the user when deleting a superproject
Now, when the user deletes a superproject (a `Project` that has subproject(s)) that Project is deleted but all the subprojects are "disconnected" from its parent and now become just regular Project (change the hierarchy from sub to project).

This means that the URL for those subproject changes also and probably other things are changed also.

We need to find a good pattern to handle this in a better way, but our first step will be to alert the user that this will happen when they try to delete the superproject.

The flow should be:

1. user tries to remove the superproject
1. user is informed in a very well detailed way what will happen and how the subprojects will be affected
1. user accepts it
1. project is deleted

For this issue we need:

- [ ] know exactly what is happening now when a superproject is deleted
- [ ] write a good message to communicate this to the user
- [ ] implement the flow proposed in this issue
- [ ] write some tests for this flow

Ref: #3649

dojutsu-user:
Closes #3742

### DIFF
--- a/readthedocs/rtd_tests/tests/test_project_views.py
+++ b/readthedocs/rtd_tests/tests/test_project_views.py
@@ -414,6 +414,20 @@ class TestPrivateViews(MockBuildTestCase):
                 args=[(project.doc_path,)],
             )
 
+    def test_delete_superproject(self):
+        super_proj = get(Project, slug='pip', users=[self.user])
+        sub_proj = get(Project, slug='test-sub-project', users=[self.user])
+
+        self.assertFalse(super_proj.subprojects.all().exists())
+        super_proj.add_subproject(sub_proj)
+
+        response = self.client.get('/dashboard/pip/delete/')
+        self.assertEqual(response.status_code, 200)
+        self.assertIn(
+            'This project has subprojects under it. Deleting this will make them as regular projects. This will effect the URLs of the subprojects and they will be served normally as other projects.',
+            response.content.decode('utf-8')
+        )
+
     def test_subproject_create(self):
         project = get(Project, slug='pip', users=[self.user])
         subproject = get(Project, users=[self.user])
--- a/readthedocs/rtd_tests/tests/test_project_views.py
+++ b/readthedocs/rtd_tests/tests/test_project_views.py
@@ -426,8 +426,8 @@ class TestPrivateViews(MockBuildTestCase):
         self.assertContains(
             response,
             'This project <a href="/dashboard/pip/subprojects/">has subprojects</a> under it. '
-            'Deleting this will make them as regular projects. '
-            'This will effect the URLs of the subprojects and they will be served normally as other projects.',
+            'Deleting this project will make them to become regular projects. '
+            'This will break the URLs of all its subprojects and they will be served normally as other projects.',
             count=1,
             html=True,
         )
--- a/readthedocs/rtd_tests/tests/test_project_views.py
+++ b/readthedocs/rtd_tests/tests/test_project_views.py
@@ -423,9 +423,12 @@ class TestPrivateViews(MockBuildTestCase):
 
         response = self.client.get('/dashboard/pip/delete/')
         self.assertEqual(response.status_code, 200)
+        resp_content = response.content.decode('utf-8')
         self.assertIn(
-            'This project <a href="/dashboard/pip/subprojects/">has subprojects</a> under it. Deleting this will make them as regular projects. This will effect the URLs of the subprojects and they will be served normally as other projects.',
-            response.content.decode('utf-8')
+            'This project <a href="/dashboard/pip/subprojects/">has subprojects</a> under it. '
+            'Deleting this will make them as regular projects. '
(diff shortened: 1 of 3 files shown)


dojutsu-user:
@stsewd 
Here is the screenshot.
Screenshot from 2019-04-24 23-34-54

Also, I have updated the code.

readthedocs/templates/projects/project_delete.html
```diff
@@ -6,6 +6,11 @@
 {% block content-header %}<h1>{% blocktrans with project.name as name %}Delete {{ name }}?{% endblocktrans %}</h1>{% endblock %}
 
 {% block content %}
+
+  {% if is_superproject %}
+    <p>This project has subprojects under it. Deleting this will make them as regular projects. This will effect the URLs of the subprojects and they will be served normally as other projects.</p>
```
stsewd: We should put a link to the subprojects page here `has subprojects...`

Also, the paragraph needs to be marked for translation

readthedocs/templates/projects/project_delete.html
```diff
@@ -6,6 +6,16 @@
 {% block content-header %}<h1>{% blocktrans with project.name as name %}Delete {{ name }}?{% endblocktrans %}</h1>{% endblock %}
 
 {% block content %}
+
+  {% if is_superproject %}
+    {% url "projects_subprojects" project.slug as subproject_url %}
+    <p>
+      {% blocktrans %}
+        This project <a href="{{ subproject_url }}">has subprojects</a> under it. Deleting this will make them as regular projects. This will effect the URLs of the subprojects and they will be served normally as other projects.
+      {% endblocktrans %}
```
stsewd: You should use the blocktrans like this https://github.com/rtfd/readthedocs.org/blob/30583646e470b1bdcf606c52e851835d6495df20/readthedocs/templates/wipe_version.html#L16-L18

(also, splitting the line will make this easier to read in the code)

dojutsu-user: @stsewd 
we cannot do the url lookup in this way here.
 
> Reverse URL lookups cannot be carried out within the blocktrans

Source: https://docs.djangoproject.com/en/2.2/topics/i18n/translation/#blocktrans-template-tag

readthedocs/rtd_tests/tests/test_project_views.py
```diff
@@ -423,9 +423,12 @@ def test_delete_superproject(self):
...
         response = self.client.get('/dashboard/pip/delete/')
         self.assertEqual(response.status_code, 200)
+        resp_content = response.content.decode('utf-8')
         self.assertIn(
-            'This project <a href="/dashboard/pip/subprojects/">has subprojects</a> under it. Deleting this will make them as regular projects. This will effect the URLs of the subprojects and they will be served normally as other projects.',
-            response.content.decode('utf-8')
+            'This project <a href="/dashboard/pip/subprojects/">has subprojects</a> under it. '
+            'Deleting this will make them as regular projects. '
+            'This will effect the URLs of the subprojects and they will be served normally as other projects.',
+            ' '.join(resp_content.split())  # this is done to remove all escapte sequences and new lines characters.
```
stsewd: I think you can avoid this by using the `trimmed` option in the blocktrans

dojutsu-user: Thank you.
I have updated the PR.

readthedocs/rtd_tests/tests/test_project_views.py
```diff
@@ -414,6 +414,22 @@ def test_delete_project(self):
...
+    def test_delete_superproject(self):
+        super_proj = get(Project, slug='pip', users=[self.user])
+        sub_proj = get(Project, slug='test-sub-project', users=[self.user])
+
+        self.assertFalse(super_proj.subprojects.all().exists())
+        super_proj.add_subproject(sub_proj)
+
+        response = self.client.get('/dashboard/pip/delete/')
+        self.assertEqual(response.status_code, 200)
+        self.assertIn(
```
stsewd: assertContains is a shorcut for this https://docs.djangoproject.com/en/2.2/topics/testing/tools/#django.test.SimpleTestCase.assertContains

dojutsu-user: Thank you.
This was a nice shortcut.
I have updated the PR.

stsewd: status code is also checked by assertContains ;)

dojutsu-user: Default is already 200.
Should we write it here explictly?

stsewd: Default is fine

humitos:
OK. I tried to do a change suggestion and I did it wrong and a proper commit was created. I updated the copy on the test as well.

I think this is ready to be merged. @stsewd do you agree?

PR #5607, opened 2019-04-18
Remove ProxyMiddleware

### LINKED ISSUE #5605: ProxyMiddleware could be removed
We have installed `ProxyMiddleware` (forked from old Django) at:

https://github.com/rtfd/readthedocs.org/blob/a7e27120e34e87ebccc691008c5d25a0192847b4/readthedocs/settings/base.py#L137

and it only adds the `REMOTE_ADDR` based on `HTTP_X_FORWARDED_FOR`.

The only place where `REMOTE_ADDR` is used is at `readthedocs.analytics.utils.get_client_ip`:

https://github.com/rtfd/readthedocs.org/blob/a7e27120e34e87ebccc691008c5d25a0192847b4/readthedocs/analytics/utils.py#L19-L28

which has a fallback that does exactly the same than the middleware.

I think we could make another checks just in case, and remove it completely.

dojutsu-user:
Closes #5605

### DIFF
--- a/readthedocs/analytics/utils.py
+++ b/readthedocs/analytics/utils.py
@@ -21,19 +21,19 @@ def get_client_ip(request):
     Gets the real client's IP address.
 
     It returns the real IP address of the client based on ``HTTP_X_FORWARDED_FOR``
-    header. If the header is not found, it returns ``None``.
+    header. If ``HTTP_X_FORWARDED_FOR`` is not found, it returns the value of
+    ``REMOTE_ADDR`` header and returns ``None`` if both the headers are not found.
     """
-
-    x_forwarded_for = request.META.get('HTTP_X_FORWARDED_FOR', '')
-
+    x_forwarded_for = request.META.get('HTTP_X_FORWARDED_FOR', None)
     if x_forwarded_for:
         # HTTP_X_FORWARDED_FOR can be a comma-separated list of IPs.
         # The client's IP will be the first one.
         # (eg. "X-Forwarded-For: client, proxy1, proxy2")
-        real_ip = x_forwarded_for.split(',')[0].strip()
-        return real_ip
+        client_ip = x_forwarded_for.split(',')[0].strip()
+    else:
+        client_ip = request.META.get('REMOTE_ADDR', None)
 
-    return None
+    return client_ip
 
 
 def anonymize_ip_address(ip_address):
--- a/readthedocs/analytics/utils.py
+++ b/readthedocs/analytics/utils.py
@@ -30,6 +30,9 @@ def get_client_ip(request):
         # The client's IP will be the first one.
         # (eg. "X-Forwarded-For: client, proxy1, proxy2")
         client_ip = x_forwarded_for.split(',')[0].strip()
+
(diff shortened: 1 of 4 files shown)


readthedocs/analytics/utils.py
```diff
@@ -17,15 +17,23 @@
...
     if x_forwarded_for:
-        ip_address = x_forwarded_for.rsplit(':')[0]
+        # HTTP_X_FORWARDED_FOR can be a comma-separated list of IPs.
+        # The client's IP will be the first one.
+        # (eg. "X-Forwarded-For: client, proxy1, proxy2")
+        real_ip = x_forwarded_for.split(',')[0].strip()
+        return real_ip
 
-    return ip_address
+    return None
```
davidfischer: There's two cases here and these changes neglect one of them:

* If `X-Forwarded-For` is present, this function should return the first value from the comma separated list. Arguably, it could validate that it is a real IP address, but that isn't that important. 
* If `X-Forwarded-For` is not present, this function should return the value from `REMOTE_ADDR`. This implementation returns `None`.

This function is used in advertising code for geo-targeting as well as being used for server side analytics (currently used in advertising but in the future might replace Google Analytics JS) but the middleware can be safely removed.

dojutsu-user: @davidfischer 
Thank you for the information.
I have updated the PR.

Will there be any case in which both of these headers are not found? Currently the implementation returns `None` in that case.

readthedocs/analytics/utils.py
```diff
@@ -17,15 +17,23 @@
...
 
-    # Get the original IP address (eg. "X-Forwarded-For: client, proxy1, proxy2")
-    x_forwarded_for = request.META.get('HTTP_X_FORWARDED_FOR', '').split(',')[0]
+    It returns the real IP address of the client based on ``HTTP_X_FORWARDED_FOR``
+    header. If ``HTTP_X_FORWARDED_FOR`` is not found, it returns the value of
+    ``REMOTE_ADDR`` header and returns ``None`` if both the headers are not found.
+    """
+    x_forwarded_for = request.META.get('HTTP_X_FORWARDED_FOR', None)
     if x_forwarded_for:
-        ip_address = x_forwarded_for.rsplit(':')[0]
```
davidfischer: This logic was removed and also shouldn't be. The X-Forwarded-For header is not exactly a standard but some implementations include a port number. This line strips that port number.

This function is correct and working as-is. Is there a reason to change it? I do appreciate comments which capture the reasoning though.

dojutsu-user: @davidfischer 

> The X-Forwarded-For header is not exactly a standard but some implementations include a port number.

Thank you for this information. I didn't found this while searching and removed this line thinking that it can produce bugs.
I have updated the code.

> Is there a reason to change it?

The only reason was the readability and to improve comments.

ericholscher: We should add tests for this also.

dojutsu-user: @ericholscher 
I have added the tests.

PR #5612, opened 2019-04-22
Add explicit egg versoin to unicode-slugify

ericholscher:


ericholscher:
Merged without review, because I'm about to do a deploy and it's not a code change.

PR #5618, opened 2019-04-22
Revert "Merge pull request #5540 from dojutsu-user/env-var-json"

ericholscher:
This reverts commit c5e705e21cea0866dfb6ec62ef0e96aabe25b256, reversing
changes made to a992ad1a2695d6d6f2396f67af2163abac2a22d0.

### DIFF
--- a/readthedocs/doc_builder/python_environments.py
+++ b/readthedocs/doc_builder/python_environments.py
@@ -1,7 +1,6 @@
 """An abstraction over virtualenv and Conda environments."""
 
 import copy
-import hashlib
 import itertools
 import json
 import logging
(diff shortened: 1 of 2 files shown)


humitos:
We had to revert this commit because this line

https://github.com/rtfd/readthedocs.org/pull/5540/files#diff-31f8391d8643c47c6caedf53930c0b7eR239

is hitting the DB and builders don't have access to the database in production.

We should use get this value from `version.project.environment_variables` instead (this value comes from the API call that the builder does).

dojutsu-user:
@humitos 
I did some research and we did have `version.project.environment_variables` as a dictionary. Thanks for the information.
I have made another PR to fix this. #5622 

Can we somehow catch these errors in development environment while testing?

humitos:
> Can we somehow catch these errors in development environment while testing?

We have been thinking on a way to reproduce that architecture we have in production, but we don't have a real solution yet :(

PR #5656, opened 2019-05-02
Use latexmk if Sphinx > 1.6

### LINKED ISSUE #5655: Use latexmk to build PDF when Sphinx > 1.6
Some time ago I added a flag called `USE_PDF_LATEXMK` to support non-ASCII languages and Unicode characters on docs. This flag is enabled by request at the moment.

We want to change this behavior to "enable that flag" automatically, but still support building PDF on Sphinx =< 1.6 (before `latexmk` was used). So, to do this, we want to decide what command to call depending on the Sphinx version installed in the virtual environment used to build the docs.

humitos:
- Checks for the Sphinx version installed in the user virtualenv to decide if `latexmk` or `pdflatex` should be called to build the PDF.
- Removes the USE_PDF_LATEXMK feature flag since it's not needed anymore.

Closes #5655

### DIFF
--- a/readthedocs/doc_builder/backends/sphinx.py
+++ b/readthedocs/doc_builder/backends/sphinx.py
@@ -149,9 +149,6 @@ class BaseSphinx(BaseBuilder):
             'dont_overwrite_sphinx_context': self.project.has_feature(
                 Feature.DONT_OVERWRITE_SPHINX_CONTEXT,
             ),
-            'use_pdf_latexmk': self.project.has_feature(
-                Feature.USE_PDF_LATEXMK,
-            ),
         }
 
         finalize_sphinx_context_data.send(
--- a/readthedocs/doc_builder/backends/sphinx.py
+++ b/readthedocs/doc_builder/backends/sphinx.py
@@ -242,16 +242,17 @@ class BaseSphinx(BaseBuilder):
         In this case, the output will be ``2.0.0``.
         """
 
-
         command = [
             self.python_env.venv_bin(filename='python'),
             '-c'
-            '"import sphinx;print(sphinx.__version__)"',
+            '"import sphinx; print(sphinx.__version__)"',
         ]
+
         cmd_ret = self.run(
             *command,
-            cwd=self.project.checkout_path(self.version.slug),
             bin_path=self.python_env.venv_bin(),
+            cwd=self.project.checkout_path(self.version.slug),
+            escape_command=False,
         )
         return cmd_ret.output
 
--- a/readthedocs/doc_builder/backends/sphinx.py
+++ b/readthedocs/doc_builder/backends/sphinx.py
@@ -10,7 +10,6 @@ import itertools
 import logging
 import os
 import shutil
-import sys
 import zipfile
 from glob import glob
 from pathlib import Path
@@ -18,6 +17,7 @@ from pathlib import Path
 from django.conf import settings
 from django.template import loader as template_loader
 from django.template.loader import render_to_string
+from packaging.version import Version
 
 from readthedocs.builds import utils as version_utils
 from readthedocs.projects.exceptions import ProjectConfigurationError
@@ -232,6 +232,32 @@ class BaseSphinx(BaseBuilder):
         )
         return cmd_ret.successful
 
+    def venv_sphinx_version(self):
+        """
+        Execute Python from the venv to query the ``sphinx`` version.
+
+        Example,
+
+            >>> import sphinx
+            >>> print(sphinx.__version__)
+            '2.0.0'
+
+        In this case, the output will be ``2.0.0``.
+        """
+
+
+        command = [
+            self.python_env.venv_bin(filename='python'),
+            '-c'
+            '"import sphinx;print(sphinx.__version__)"',
+        ]
+        cmd_ret = self.run(
+            *command,
+            cwd=self.project.checkout_path(self.version.slug),
+            bin_path=self.python_env.venv_bin(),
+        )
+        return cmd_ret.output
+
 
 class HtmlBuilder(BaseSphinx):
     type = 'sphinx'
@@ -395,7 +421,10 @@ class PdfBuilder(BaseSphinx):
             raise BuildEnvironmentError('No TeX files were found')
 
         # Run LaTeX -> PDF conversions
-        if self.project.has_feature(Feature.USE_PDF_LATEXMK):
+        # Build PDF with ``latexmk`` if Sphinx > 1.6, otherwise fallback to
+        # ``pdflatex`` to support old versions
+        sphinx_version = Version(self.venv_sphinx_version())
+        if sphinx_version > Version('1.6'):
             return self._build_latexmk(cwd, latex_cwd)
 
         return self._build_pdflatex(tex_files, latex_cwd)
--- a/readthedocs/doc_builder/backends/sphinx.py
+++ b/readthedocs/doc_builder/backends/sphinx.py
@@ -259,7 +259,7 @@ class BaseSphinx(BaseBuilder):
             shell=True,  # used on BuildCommand
             record=False,
         )
-        return True if cmd_ret.exit_code == 0 else False
+        return cmd_ret.exit_code == 0
 
 
 class HtmlBuilder(BaseSphinx):
--- a/readthedocs/doc_builder/backends/sphinx.py
+++ b/readthedocs/doc_builder/backends/sphinx.py
@@ -17,7 +17,6 @@ from pathlib import Path
 from django.conf import settings
 from django.template import loader as template_loader
 from django.template.loader import render_to_string
-from packaging.version import Version
 
 from readthedocs.builds import utils as version_utils
 from readthedocs.projects.exceptions import ProjectConfigurationError
--- a/readthedocs/doc_builder/backends/sphinx.py
+++ b/readthedocs/doc_builder/backends/sphinx.py
@@ -229,23 +229,26 @@ class BaseSphinx(BaseBuilder):
         )
         return cmd_ret.successful
 
-    def venv_sphinx_version(self):
+    def venv_sphinx_supports_latexmk(self):
         """
-        Execute Python from the venv to query the ``sphinx`` version.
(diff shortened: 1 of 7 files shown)


humitos:
Running the command manually from inside the Docker image I get the output that I want:

    docs@f7bc3ebeec08:~$ /bin/sh -c 'cd /home/docs/readthedocs.org/user_builds/test-builds/checkouts/datetime && /home/docs/readthedocs.org/user_builds/test-builds/envs/datetime/bin/python -c "import sphinx;print(sphinx.__version__)"'
    1.8.5

I don't understand why it does not work when calling it from Python.

humitos:
I found the issue: our `get_wrapped_command()` makes the command invalid.

```
(Pdb++) self.get_wrapped_command()
'/bin/sh -c \'cd /home/humitos/rtfd/code/readthedocs.org/user_builds/test-builds/checkouts/datetime && /home/humitos/rtfd/code/readthedocs.org/user_builds/test-builds/envs/datetime/bin/python -c\\"import\\ sphinx\\;print\\(sphinx.__version__\\)\\"\''

(Pdb++) command = '/bin/sh -c \'cd /home/humitos/rtfd/code/readthedocs.org/user_builds/test-builds/checkouts/datetime && /home/humitos/rtfd/code/readthedocs.org/user_builds/test-builds/envs/datetime/bin/python -c"import sphinx;print(sphinx.__version__)"\''

(Pdb++) exec = client.exec_create(container=self.build_env.container_id,cmd=self.get_wrapped_command(),stderr=True)
(Pdb++) output = client.exec_start(exec_id=exec['Id'], stream=False)
(Pdb++) output
b''

(Pdb++) exec = client.exec_create(container=self.build_env.container_id,cmd=command,stderr=True)
(Pdb++) output = client.exec_start(exec_id=exec['Id'], stream=False)
(Pdb++) output
b'1.8.5\n'
(Pdb++) 
```

Note that I'm removing all the unneeded `\` bars.

stsewd:
We do that to prevent command injection. But if the code is entirely called from our code, we could add an option `allow_insecure` or something.

humitos:
Yes! I pushed some changes adding a `escape_command` attribute to the `DockerBuildCommand` and I was able to make it work.

humitos:
Works!

With 1.8.5 calls `latexmk`.

Screenshot_2019-05-02_16-32-15

With 1.5.2 calls `pdflatex`.

Screenshot_2019-05-02_16-32-21

readthedocs/doc_builder/environments.py
```diff
@@ -296,14 +298,20 @@ class DockerBuildCommand(BuildCommand):
...
 
-        :param cmd_input: input to pass to command in STDIN
-        :type cmd_input: str
-        :param combine_output: combine STDERR into STDOUT
+        :param escape_command: whether escape special chars the command before
+            executing it in the container. This should only be disabled on
+            trusted or internal commands.
+        :type escape_command: bool
         """
+        self.escape_command = escape_command
```
stsewd: Feels like something that can be moved to the parent class. I think (not sure) that the same behavior for this in local command is `shell=True`

https://docs.python.org/3.6/library/subprocess.html

> On POSIX with shell=True, the shell defaults to /bin/sh. If args is a string, the string specifies the command to execute through the shell. This means that the string must be formatted exactly as it would be when typed at the shell prompt. This includes, for example, quoting or backslash escaping filenames with spaces in them. If args is a sequence, the first item specifies the command string, and any additional items will be treated as additional arguments to the shell itself...

readthedocs/doc_builder/backends/sphinx.py
```diff
@@ -232,6 +229,34 @@ def build(self):
...
+            >>> import sphinx
+            >>> print(sphinx.__version__)
+            '2.0.0'
+
+        In this case, the output will be ``2.0.0``.
+        """
+
+        command = [
+            self.python_env.venv_bin(filename='python'),
+            '-c'
```
stsewd: You are missing a comma here

readthedocs/doc_builder/backends/sphinx.py
```diff
@@ -232,6 +229,34 @@ def build(self):
...
+            '"import sphinx; print(sphinx.__version__)"',
```
stsewd: when `DOCKER_ENABLED` is false, this will break. Because of the double quotes

```python
>>> proc = subprocess.Popen(['python3', '-c', 'import sphinx; print(sphinx.__version__)'], stdou
t=subprocess.PIPE)
>>> proc.communicate()
(b'1.8.5\n', None)
>>> proc = subprocess.Popen(['python3', '-c', '"import sphinx; print(sphinx.__version__)"'], std
out=subprocess.PIPE)
>>> proc.communicate()
(b'', None)

```

agjohnson:
I'd probably vote that we shouldn't be relying on command output in our build process. It's a little awkward and it's the only case we have for reading from inside the environment so far. Is there a reason we shouldn't use normal sh operators to chain? Something like:

```
python -c "import sys; import sphinx; sys.exit(0 if sphinx.version_info >= (1, 7) else 1)" \
  && latexmk ... \
  || pdflatex ...
```

humitos:
> I'd vote that we shouldn't be relying on command output in our build process. No-op commands like this are already awkward, doing a text comparison against the command output seems pretty fragile.

@agjohnson I don't see this as a problem. The output of the command that we are relying here is a command that we control and users can't modify. I don't see any potential problem with it.

Also, we are not comparing text directly, but using `packaging` module to parse the version properly as an object and then comparing it using its helpers.

> Is there a reason we shouldn't use normal sh operators to chain?

Using operators chain has some problems. We need execute ~5 different commands depending on which pdf builder we use, so we would need to add the same check on each of those commands which is repeated code in the end.

> This does make it more clear why we're executing latexmk or pdflatex to the user, rather than just having a version number in the output, and it works regardless of the builder type.

Also, I don't want to log this to users in the build output. It's hard to read and confusing to users in my opinion. This should be transparent to users, similar to what happen locally --they don't care about this, Sphinx makes the decisions.

Just to be clear, the Sphinx version is not logged in the build output. That was just a test locally to _see_ what was happening and debug it properly. The command is executed with `record=False`.

On the other hand, from a developer perspective, I find the code easy to read by having the different chunk of commands wrapped up into different methods.

readthedocs/doc_builder/backends/sphinx.py
humitos: Thanks for raising this.

I made it work by using `shell=True` and flattening the command when `shell=True` using the helper that we had there.

I like how the ended now. On `BuildCommand` it uses the `shell=True` attribute and flatten the command to be compatible. On `DockerBuildCommand` it uses `escape_command=False` to not escape the command.

It works in both builder backends.

agjohnson:
@humitos If it's too much to wrap the commands with a shell and use shell pipes to execute the command we want, I still think string parsing is the wrong approach, as it is more fragile than it needs to be.

A more robust version of your implementation would use the exit code instead of coercing the Sphinx version multiple times. The command would be similar to the one i posted, but we don't use pipes and would instead use the exit code as a boolean for our decision on whether to run latexmk or not. If there are problems running the current PR outside docker, this avoids those issues as well, as we can expect an exit code from either execution strategy.

I agree that in this case, the command is a noop command that is not helpful to users and should be hidden.

humitos:
> If there are problems running the current PR outside docker, this avoids those issues as well, as we can expect an exit code from either execution strategy.

The problem is not running the command inside or outside docker. The issue is because the command string is _sanitized_ in a way that breaks depending how that string is used internally to run the command. Although, this problem is fixed using `shell=True` in one case and `escape_command=False` in the other.

humitos:
> A more robust version of your implementation would use the exit code instead of coercing the Sphinx version multiple times. The command would be similar to the one i posted, but we don't use pipes and would instead use the exit code as a boolean for our decision on whether to run latexmk or not

I updated the PR to use the exit code instead.

readthedocs/doc_builder/backends/sphinx.py
```diff
@@ -232,6 +229,38 @@ def build(self):
...
+
+        cmd_ret = self.run(
+            *command,
+            bin_path=self.python_env.venv_bin(),
+            cwd=self.project.checkout_path(self.version.slug),
+            escape_command=False,  # used on DockerBuildCommand
+            shell=True,  # used on BuildCommand
+            record=False,
+        )
+        return True if cmd_ret.exit_code == 0 else False
```
stsewd:
```suggestion
        return cmd_ret.exit_code == 0
```

readthedocs/doc_builder/environments.py
```diff
@@ -166,8 +168,13 @@ def run(self):
             environment['PATH'] = ':'.join(env_paths)
 
         try:
+            # When using ``shell=True`` the command should be flatten
+            command = self.command
+            if self.shell:
+                command = self.get_command()
```
stsewd: Just a note that this would make some problems to people using paths with spaces and reserved bash symbols (not our case). Flatten commands are not escaped by default.

readthedocs/doc_builder/backends/sphinx.py
```diff
@@ -10,14 +10,14 @@
...
 import shutil
-import sys
 import zipfile
 from glob import glob
 from pathlib import Path
 
 from django.conf import settings
 from django.template import loader as template_loader
 from django.template.loader import render_to_string
+from packaging.version import Version
```
agjohnson: Looks unused now.

readthedocs/doc_builder/backends/sphinx.py
humitos: Haha! I don't know what I was thinking when I wrote this :)

humitos:
@agjohnson @stsewd both commented about the same, and both are right. `shell=True` should not be used by default or exposed to the user at all. It's intended only for internal usage and purposes like this one where we write the command by ourselves and now how it will behave.

I didn't added the `shell=` attribute in the class, I just made it work when used (flattening the command). Although, I think this is the only place where it's used. I didn't find another one.

humitos:
I'll merge the PR once the tests pass.

PR #5696, opened 2019-05-14
Enable auth validate passwords

### LINKED ISSUE #5551: Local instance allows weak passwords
## Details

The [installation instructions](https://docs.readthedocs.io/en/stable/install.html) say to run `python manage.py createsuperuser`, which prompts for a password; self-sign-up users are also prompted for a password. By default, `createsuperuser` requires a non-empty password, and self-sign-up requires a password that is six characters or longer. Neither impose any further password complexity requirements, including checking for well-known passwords like `password` or `123456`.

## Expected Result

An error message - or at least a warning - about using insecure passwords.

## Actual Result

"Superuser created successfully." and successful sign-up.

humitos:
Define some simple validators to use when creating a new user or
changing the password.

https://docs.djangoproject.com/en/1.11/topics/auth/passwords/#module-django.contrib.auth.password_validation

Closes #5551

### DIFF
--- a/readthedocs/settings/base.py
+++ b/readthedocs/settings/base.py
@@ -181,6 +181,24 @@ class CommunityBaseSettings(Settings):
         'allauth.account.auth_backends.AuthenticationBackend',
     )
 
+    AUTH_PASSWORD_VALIDATORS = [
+        {
+            'NAME': 'django.contrib.auth.password_validation.UserAttributeSimilarityValidator',
+        },
+        {
+            'NAME': 'django.contrib.auth.password_validation.MinimumLengthValidator',
+            'OPTIONS': {
+                'min_length': 9,
+            }
+        },
+        {
+            'NAME': 'django.contrib.auth.password_validation.CommonPasswordValidator',
+        },
+        {
+            'NAME': 'django.contrib.auth.password_validation.NumericPasswordValidator',
+        },
+    ]
+
     MESSAGE_STORAGE = 'readthedocs.notifications.storages.FallbackUniqueStorage'
 
     NOTIFICATION_BACKENDS = [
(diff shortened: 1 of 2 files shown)


readthedocs/settings/base.py
```diff
@@ -181,6 +181,24 @@ def USE_PROMOS(self):  # noqa
         'allauth.account.auth_backends.AuthenticationBackend',
     )
 
+    AUTH_PASSWORD_VALIDATORS = [
+        {
+            'NAME': 'django.contrib.auth.password_validation.UserAttributeSimilarityValidator',
+        },
+        {
+            'NAME': 'django.contrib.auth.password_validation.MinimumLengthValidator',
```
ericholscher: I really dislike these defaults in dev. I think these should likely only be prod settings.

davidfischer:
I'm in favor of these settings being present in the base settings file as they are in this PR. This helps others who take our code to do the right thing as well as ensuring that all the different instances we control (community site, corporate site, etc.) have sound security defaults.

I can understand @ericholscher's position and we could also set `AUTH_PASSWORD_VALIDATOR=[]` in `dev.py` to ease the development settings.

readthedocs/settings/base.py
```diff
@@ -181,6 +181,24 @@ def USE_PROMOS(self):  # noqa
...
+            'OPTIONS': {
+                'min_length': 9,
```
davidfischer: Was 9 characters a recommendation from somewhere? I believe 8 is more common.

davidfischer: I see. It was copied from the Django docs.

humitos:
> I can understand @ericholscher's position and we could also set `AUTH_PASSWORD_VALIDATOR=[]` in `dev.py` to ease the development settings.

I like this. I will update the PR for this.

PR #5698, opened 2019-05-14
Storage updates

### LINKED ISSUE #5691: Deleting a project does not delete media in storage
As we write more build output to cloud storage we need to make sure that those remote files are removed when versions or projects are deleted. Currently this does not appear to be happening.

davidfischer:
Since #5549 was merged (although it is not yet deployed) a few issues emerged. This is an attempt to fix them:

- Delete from storage when a project or version is deleted (Fixes #5691)
- Project delete now handled in a method on the model (`Project.delete()`)
- Define media type constants (next step: use them everywhere)
- A few places still using default storage instead of `RTD_BUILD_MEDIA_STORAGE`

### DIFF
--- a/readthedocs/projects/models.py
+++ b/readthedocs/projects/models.py
@@ -42,9 +42,10 @@ from readthedocs.search.parse_json import process_file
 from readthedocs.vcs_support.backends import backend_cls
 from readthedocs.vcs_support.utils import Lock, NonBlockingLock
 
+from .constants import MEDIA_TYPES
+
 
 log = logging.getLogger(__name__)
-storage = get_storage_class()()
 
 
 class ProjectRelationship(models.Model):
@@ -463,6 +464,29 @@ class Project(models.Model):
         except Exception:
             log.exception('Error creating default branches')
 
+    def delete(self, *args, **kwargs):  # pylint: disable=arguments-differ
+        from readthedocs.projects import tasks
+
+        # Remove local FS build artifacts on the web servers
+        broadcast(
+            type='app',
+            task=tasks.remove_dirs,
+            args=[(self.doc_path,)],
+        )
+
+        # Remove build artifacts from storage
+        storage_paths = []
+        for type_ in MEDIA_TYPES:
+            storage_paths.append(
+                '{}/{}'.format(
+                    type_,
+                    self.slug,
+                )
+            )
+        tasks.remove_build_storage_paths.delay(storage_paths)
+
+        super().delete(*args, **kwargs)
+
     def get_absolute_url(self):
         return reverse('projects_detail', args=[self.slug])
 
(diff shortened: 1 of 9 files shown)


davidfischer:
One thing I didn't do in this PR that needs to be done before we turn off all storage on the local webs is  to make sure that the code that handles `ImportedFile`s and search indexing uses storage instead of the local disk.

readthedocs/projects/models.py
```diff
@@ -463,6 +464,29 @@ def save(self, *args, **kwargs):  # pylint: disable=arguments-differ
...
+            type='app',
+            task=tasks.remove_dirs,
+            args=[(self.doc_path,)],
+        )
+
+        # Remove build artifacts from storage
+        storage_paths = []
+        for type_ in MEDIA_TYPES:
+            storage_paths.append(
+                '{}/{}'.format(
```
humitos: I think it's better to use either `os.path.join` or `pathlib.Path` here.

readthedocs/projects/models.py
```diff
@@ -750,28 +774,49 @@ def has_pdf(self, version_slug=LATEST):
...
+        if settings.RTD_BUILD_MEDIA_STORAGE:
+            storage = get_storage_class(settings.RTD_BUILD_MEDIA_STORAGE)()
+            storage_path = self.get_storage_path(
+                type_='epub', version_slug=version_slug
+            )
+            return storage.exists(storage_path)
+
+        return False
 
     def has_htmlzip(self, version_slug=LATEST):
```
humitos: nit: these 3 methods are exactly the same; the only change is `type_` value. They could be refactored to use the same chunk of code.

Not needed to be done in this PR, though.

readthedocs/projects/tasks.py
```diff
@@ -1680,6 +1680,20 @@ def remove_dirs(paths):
         shutil.rmtree(path, ignore_errors=True)
 
 
+@app.task()
+def remove_build_storage_paths(paths):
```
humitos: I'm not sure what we decided here, but I think that build servers don't have the credentials (or the permissions) to make this operation. I think this task may be only executed by webs. In that case, `queue='web'` should be passed in the decorator.

readthedocs/projects/models.py
davidfischer: This is not a path on the local disk. This is a storage path which is different. It always uses forward slashes regardless of platform.

readthedocs/projects/tasks.py
davidfischer: It should only be executed by webs typically in response to a project or version being deleted completely. I'll update it.

davidfischer:
I would like to find a better way to test this. It can be a little tricky because many of the functions check both the local disk *and* storage which in test should probably be the same thing. With that said, there still might be a better approach to testing.

readthedocs/projects/models.py
davidfischer: That seems like a worthwhile refactor and it seems useful as part of this PR.

readthedocs/projects/models.py
humitos: If we want to go fancy, what we need here is a [`PurePosixPath`](https://docs.python.org/3/library/pathlib.html#pathlib.PurePosixPath): it's platform independent and does not access the filesystem when manipulating paths.

readthedocs/projects/tasks.py
humitos: I think you forgot to push this change.

readthedocs/projects/models.py
davidfischer: The Django storage API does not implement a posix standard either. I believe that using pathlib is not the correct course of action because these are not filesystem paths.

PR #5706, opened 2019-05-17
Move version "Clean" button to details page

humitos:
Instead of showing the "Clean" button next to a version in the "Inactive Versions" listing, this button is moved to the detail page with an explanation about what it does.

Changed from,

Screenshot_2019-05-17_13-10-27

to,

Screenshot_2019-05-17_13-18-35

This change is made because we found that multiple buttons in a row breaks in other languages when translated (because of margin/padding). Also, this button is [not shown at all in the corporate site](https://github.com/rtfd/readthedocs-corporate/issues/534) and it should be --but we don't want to show it in the listing.

### DIFF
--- a/readthedocs/templates/projects/project_version_detail.html
+++ b/readthedocs/templates/projects/project_version_detail.html
@@ -21,7 +21,7 @@
       {% if not version.active and version.built %}
         <p class="empty">
           {% url "project_version_delete_html" project.slug version.slug as version_delete_url %}
-          {% blocktrans %}
+          {% blocktrans trimmed %}
             This version is inactive but its documentation is still available online.
             You can <a href="{{ version_delete_url }}">delete this version's documentation</a> if you want to remove it completely.
           {% endblocktrans %}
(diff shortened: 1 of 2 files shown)


humitos:
As a side note, I think that ending up with a `active=False` and `built=True` version may not possible anymore after #5600 since it was the only place where those fields were not updated together.

Now, when we [update a version from the dashboard we do these all together](https://github.com/rtfd/readthedocs.org/blob/616a3038f5ad6b541a59b80f50077b15a8f180f5/readthedocs/projects/views/private.py#L177-L190):

1. mark is as `active=False`
1. remove the documentation's file
1. mark it as `built=False`

On the other hand, our new automation rules and APIv3 version edit probably does not behaves in the same way. This is outside the scope of this PR, but may worth to consider to unify the behavior when the same action is performed by the user.

readthedocs/templates/projects/project_version_detail.html
```diff
@@ -15,6 +16,20 @@
 
 {% block content %}
     <h2> Editing {{ version.slug }} </h2>
+
+    {% if request.user|is_admin:project %}
+      {% if not version.active and version.built %}
+        <p class="empty">
+          {% url "project_version_delete_html" project.slug version.slug as version_delete_url %}
+          {% blocktrans %}
```
agjohnson: Use `trimmed` on the blocktrans call to improve the text whitespace handling in gettext

PR #5742, opened 2019-05-30
Add init.py under readthedocs/api

davidfischer:


davidfischer:
This actually was causing some issues on production servers because the `__init__.pyc` file is still there from before we upgraded.

PR #5746, opened 2019-05-30
Fix several outstanding differences between master -> relcorp

agjohnson:
Something got hosed on an auto conflict resolve or something probably. Looks
minor.


agjohnson:
Uhh. no. Get it right, github.

PR #5760, opened 2019-06-03
Try to fix Elastic connection pooling issues

ericholscher:
This creates a new connection on every request,
hopefully working around the issues with the connection pool.

This will go out on 1 web and see if it resolves the issues at all.

This is branched from `rel`, which seems to have caused some weirdness :/

### DIFF
--- a/readthedocs/search/documents.py
+++ b/readthedocs/search/documents.py
@@ -3,6 +3,8 @@ import logging
 from django.conf import settings
 from django_elasticsearch_dsl import DocType, Index, fields
 
+from elasticsearch import Elasticsearch
+
 from readthedocs.projects.models import HTMLFile, Project
 from readthedocs.sphinx_domains.models import SphinxDomain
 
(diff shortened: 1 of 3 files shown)


safwanrahman:
@ericholscher When did you pushed it in production?

readthedocs/search/documents.py
```diff
@@ -22,8 +24,19 @@
...
 
 
+class RTDDocTypeMixin:
+
+    def update(self, *args, **kwargs):
+        # Hack a fix to our broken connection pooling
+        # This creates a new connection on every request,
+        # but actually works :)
+        log.info('Hacking Elastic indexing to fix connection pooling')
+        self.using = Elasticsearch(**settings.ELASTICSEARCH_DSL['default'])
```
agjohnson: What exactly is the fix here? I'm assuming `Elasticsearch` is normally instantiated with a class or instance for the connection pooling?

readthedocs/search/tasks.py
```diff
@@ -60,7 +60,11 @@ def delete_objects_in_es(app_label, model_name, document_class, objects_id):
     queryset = doc_obj.get_queryset()
     queryset = queryset.filter(id__in=objects_id)
     log.info("Deleting model: %s, '%s' objects", model.__name__, queryset.count())
-    doc_obj.update(queryset.iterator(), action='delete')
+    try:
+        # This is a common case that we should be handling a better way
+        doc_obj.update(queryset.iterator(), action='delete')
+    except Exception:
```
agjohnson: We should catch a more specific exception here to prevent gobbling up any important exception that might be raised upstream.

PR #5763, opened 2019-06-03
Try to fix Elastic connection pooling issues

ericholscher:
This creates a new connection on every request,
hopefully working around the issues with the connection pool.

This is a version of #5760 that merges into master


ericholscher:
Merging this into `master` since #5760 was already reviewed. 👍

PR #5777, opened 2019-06-07
Project advertising page/form update

davidfischer:
- Remove the forms to contribute to Read the Docs. We don't really get any contributions this way
- Build value for the commercial product a bit more
- More prominent get in touch for alternatives
- Remove mentions of gold. Gold doesn't remove ads from projects anymore. However, projects can get in touch.

Screen Shot 2019-06-07 at 1 54 21 PM

### DIFF
--- a/readthedocs/templates/projects/project_advertising.html
+++ b/readthedocs/templates/projects/project_advertising.html
@@ -30,18 +30,18 @@
   <ul class="project-ads-guidelines">
     <li>
       {% blocktrans trimmed %}
-        We don't give sponsors access to user information
+        We don't give sponsors access to user information.
       {% endblocktrans %}
     </li>
     <li>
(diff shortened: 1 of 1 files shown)


davidfischer:
> I suggest to change the gray box/table for the "Opt-Out" section to a regular bullet list.

Unfortunately our CSS doesn't support bullet lists very well. I am attempting to not write any more custom CSS until our theme work is complete.

davidfischer:
Maybe I can reuse the bullet list above in the page. It requires a special class as just using a `<ul>` won't have bullets.

humitos:
OK, OK. It's fine as it is. I was just saying because that gray box is not used in any other place as far as I know, and I thought that a bullet list would be easy to add :)

PR #5787, opened 2019-06-11
Remove access to DB from clean_build function

ericholscher:
This is a hack to make our builders work for now.


ericholscher:
Replaced with https://github.com/rtfd/readthedocs.org/pull/5788

PR #5797, opened 2019-06-12
Use a real SessionBase object on FooterNoSessionMiddleware

humitos:
Instead of forcing to be an empty dict `{}`, we force it to be a real SessionBase object which it's what Django uses internally. This avoid issues where other pieces expects the `SessionBase`.

Django docs: https://docs.djangoproject.com/en/1.11/topics/http/sessions/#django.contrib.sessions.backends.base.SessionBase

### DIFF
--- a/readthedocs/core/middleware.py
+++ b/readthedocs/core/middleware.py
@@ -1,6 +1,7 @@
 import logging
 
 from django.conf import settings
+from django.contrib.sessions.backends.base import SessionBase
 from django.contrib.sessions.middleware import SessionMiddleware
 from django.core.exceptions import MultipleObjectsReturned, ObjectDoesNotExist
 from django.http import Http404, HttpResponseBadRequest
@@ -205,7 +206,7 @@ class FooterNoSessionMiddleware(SessionMiddleware):
                 settings.SESSION_COOKIE_NAME not in request.COOKIES
             ):
                 # Hack request.session otherwise the Authentication middleware complains.
-                request.session = {}
+                request.session = SessionBase()  # create an empty session
                 return
         super().process_request(request)
 
--- a/readthedocs/rtd_tests/tests/test_footer.py
+++ b/readthedocs/rtd_tests/tests/test_footer.py
@@ -1,4 +1,5 @@
 import mock
+from django.contrib.sessions.backends.base import SessionBase
 from django.test import TestCase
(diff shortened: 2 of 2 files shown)


readthedocs/core/middleware.py
```diff
@@ -205,7 +206,7 @@ def process_request(self, request):
                 settings.SESSION_COOKIE_NAME not in request.COOKIES
             ):
                 # Hack request.session otherwise the Authentication middleware complains.
-                request.session = {}
+                request.session = SessionBase()  # create an empty session
```
humitos: We could instantiate an empty instance of the same class that we are using in the backend.

```
from importlib import import_module
SessionStore = import_module(settings.SESSION_ENGINE).SessionStore
request.session = SessionStore()
```

I suppose that won't be any difference since the `session_key` is `None`, though.

ericholscher:
I'm pretty sure this isn't the solution to our problem. We don't want a session on the request, anywhere downstream that _thinks_ it's editing a session, that data won't be saved. We want that to error, and to stop trying to edit the session, not fake out users who think they are really editing it.

humitos:
Hrm... I implemented this solution because we are already faking it with a dict (allowing get/set values). The comment says that otherwise Auth backend fails.

On the other hand, using this general `SessionBase` will behave similar to a `dict` (it's a dict-like object) but actually it will fail with a proper error (`NotImplementedError`) when calling the methods `.save`, `.delete`, `.create`, etc instead saying that the `dict` does not have this method.

Finally, for the use case I'm trying to fix solve, it will work since `request.session.session_key` will return `None` which it's expected when there is no session.

I'm not 100% sure, but seems to be the right way of having the "same behavior than a regular session object without saving anything and hard-failing if trying to save it".

humitos:
I know that this affects _a lot_ of requests. So, if we are in doubt we can just go in another direction that we consider safer. Although, at some point this will bite us again.

ericholscher:
Good point. I mostly have 2 reservations:

* I tried something similar when I wrote this code, but I don't think it was quite this
* Like you mentioned, the scale of the change. Though it should be pretty obvious if this breaks, so we can just roll it back. 

So I guess I'm +0 on this proposed change, then.

humitos:
OK. I think we can merge and deploy carefully being prepared to roll it back, as you said. Although, we can have another person reviewing this before merging.

PR #5805, opened 2019-06-14
Remove PR Versions from Elasticsearch Indexing and HTMLFile Queryset Added

saadmk11:
This is a continuation from #5750

### DIFF
--- a/readthedocs/search/documents.py
+++ b/readthedocs/search/documents.py
@@ -158,9 +158,9 @@ class PageDocument(RTDDocTypeMixin, DocType):
 
         # Do not index files that belong to non sphinx project
         # Also do not index certain files
-        queryset = queryset.filter(
+        queryset = queryset.internal().filter(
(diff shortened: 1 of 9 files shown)


readthedocs/search/documents.py
```diff
@@ -159,7 +160,7 @@ def get_queryset(self):
         # Also do not index certain files
         queryset = queryset.filter(
             project__documentation_type__contains='sphinx'
-        )
+        ).exclude(version__type=PULL_REQUEST)
```
ericholscher: can this not use the manager? If not, we should probably add a queryset method that does the same thing, so we can use it in cases like this.

saadmk11: @ericholscher Added Queryset for this

ericholscher:
This can now be updated against current `gsoc-19-pr-builder` branch.

PR #5807, opened 2019-06-14
 PR Builds tab added to project dashboard

saadmk11:
This is a continuation of #5779

This will add a tab for PR Builds in the project dashboard.

Screenshot from 2019-06-18 17-39-50

### DIFF
--- a/readthedocs/templates/core/project_bar_base.html
+++ b/readthedocs/templates/core/project_bar_base.html
@@ -45,7 +45,9 @@
 
         <li class="{{ builds_active }}"><a href="{{ project.get_builds_url }}">{% trans "Builds" %}</a></li>
 
-        <li class="{{ external_builds_active }}"><a href="{{ project.get_external_builds_url }}">{% trans "External Builds" %}</a></li>
+        {% if project.has_external_builds %}
+            <li class="{{ external_builds_active }}"><a href="{{ project.get_external_builds_url }}">{% trans "External Builds" %}</a></li>
+        {% endif %}
 
         <li class="{{ versions_active }}"><a href="{% url "project_version_list" project.slug %}">{% trans "Versions" %}</a></li>
 
--- a/readthedocs/templates/core/project_bar_base.html
+++ b/readthedocs/templates/core/project_bar_base.html
@@ -45,6 +45,8 @@
 
         <li class="{{ builds_active }}"><a href="{{ project.get_builds_url }}">{% trans "Builds" %}</a></li>
 
+        <li class="{{ pr_builds_active }}"><a href="{{ project.get_pr_builds_url }}">{% trans "PR Builds" %}</a></li>
+
         <li class="{{ versions_active }}"><a href="{% url "project_version_list" project.slug %}">{% trans "Versions" %}</a></li>
 
         {% comment %}
--- a/readthedocs/templates/core/project_bar_base.html
+++ b/readthedocs/templates/core/project_bar_base.html
@@ -45,7 +45,7 @@
 
         <li class="{{ builds_active }}"><a href="{{ project.get_builds_url }}">{% trans "Builds" %}</a></li>
 
-        <li class="{{ pr_builds_active }}"><a href="{{ project.get_pr_builds_url }}">{% trans "PR Builds" %}</a></li>
+        <li class="{{ external_builds_active }}"><a href="{{ project.get_external_builds_url }}">{% trans "External Builds" %}</a></li>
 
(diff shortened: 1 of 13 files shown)


humitos:
Isn't "Pull Request" a term fixed to GitHub itself? I think that GitLab uses "Merge Request" for example (https://docs.gitlab.com/ee/user/project/merge_requests/). Can we come up with a general and descriptive name for this?

saadmk11:
@humitos Github and Bitbucket uses `Pull Request` as the name. Do you have any cool name Idea for this ?

ericholscher:
I wonder if we should use `External versions` -- to keep it consistent with the code.

readthedocs/templates/core/project_bar_base.html
```diff
@@ -45,6 +45,8 @@ <h1>
 
         <li class="{{ builds_active }}"><a href="{{ project.get_builds_url }}">{% trans "Builds" %}</a></li>
 
+        <li class="{{ pr_builds_active }}"><a href="{{ project.get_pr_builds_url }}">{% trans "PR Builds" %}</a></li>
```
ericholscher: This should only show up if the user _has_ any PR builders.

saadmk11: @ericholscher Updated the PR

ericholscher:
This can now be updated against current `gsoc-19-pr-builder` branch.

agjohnson:
Long term, why can't this all be under the `Builds` tab? I'd like to reduce confusion on this page with the template rework project and we'll likely mostly be removing tabbed interfaces from this page.

saadmk11:
> Long term, why can't this all be under the `Builds` tab? I'd like to reduce confusion on this page with the template rework project and we'll likely mostly be removing tabbed interfaces from this page.

Should this be done now or after the `Template Project` ? we can have sub sections under `Builds` in the new template

agjohnson:
I guess why do we need subsections or a separate page at all? Could the normal and PR builds be intermingled in a single view (eventually)? Having the two joined seems like it will be better UX for the user.

saadmk11:
If we do that then we don't need this PR at all. We can just update the `BuildList` view to show all version type builds like before.

agjohnson:
Well, i have been outside any planning for this feature, but from the outside, it seems like a lot of duplicated UI and code. It's not immediately apparent why having a separate section for PR builds is better than combining the two build types under a single listing. If we don't have a good answer to why these build types need to be separated in our UI, then I think combining them is the best answer. We can apply state and filtering based on the build type if we need to.

Is there a strong reason to split these up? Why are we moving away from the original implementation that used a single listing view?

ericholscher:
I think the original thinking was because external builds will come from other users, and will be much more likely to fail. We don't want it to confuse users who see them in their normal build listings. I do think we could get away with a filter on the page, but having a separate top-level listing seemed to make the most sense for now. 

I was also taking inspiration from Travis, which splits them out:

Screenshot 2019-06-21 06 47 35

I don't feel too strongly though, this just seemed easiest and least confusing for users to start.

agjohnson:
Gotcha. It's a little better when named "Pull Requests", as that communicates the page's purpose fairly well, but I'm still not sure separating is the best UX for us however. For comparison, CircleCI and GitLab have singular list of builds.

If we consider the perspective of the maintainer:

* If I'm relying on pull requests to figure out the merge status of a PR with docs, I don't really care much about the status of the build in `master` branch anymore. We went through the process of figuring out build status in the pull request before merge to master. Therefore, polluting the build list might not be a concern
* "Builds" vs "External builds" doesn't distinguish what's different between the types, and I'm a little lost on where to go both for normal builds and for pull request builds
* I wouldn't normally go to a listing to see the pull request build status, I interact with GitHub for this and then follow a link through to our page our the build output. With PR metadata on the listing, the view is at least useful, but the primary point of interaction with these builds will still be through GitHub.

If our main concern is polluting the build list, we can add filtering later (probably defaulting to filtering PR builds out on the build list). By merging the two lists, we worry about optimizing later, and we can avoid complicating the implementation now, so we don't have to rip out redundant code and templates later.

If they truly need to be separate, "External builds" should be renamed. We have the repo type, so we can match "Pull requests" vs "Merge requests" etc on the tab based on what provider the project uses. The function of the page would at least be clearer to maintainers this way. I'd still be -0 on this path though, it seems like we're making more UI for us to maintain without a clear benefit.

ericholscher:
I think we've decided not to do this for now. We can come back to this PR if we need to in the future, but I'm closing it for now.

PR #5814, opened 2019-06-17
Enable Travis-ci for PR Builder branch

saadmk11:


humitos:
I'm not sure if this is what we want here. Branches are not triggered by Travis CI by default, but PRs are. I'd say that on Travis we do care about those main three branches to pass all the tests all the time and notify us on failure --the other branches are not that important.

I think what you need/want is to create a PR for your branch and travis will run the tests automatically.

ericholscher:
Let's not spend time on this for now. This is a temporary fix, we don't need to do this.

Issue #5675, opened 2019-05-07
Guide users more on top level documentation page


agjohnson:
We determined that the top level documentation page should probably be restructured and help guide users better. Looking at the types of users that we have viewing our documentation, and the context with which they arrive at the top level documentation page, there were a few changes that we discussed:

* The marketing material is important, but can be reduced. The top level documentation page isn't the best place to sell users -- the main product pages are a better place for this
* The toctree directive just plops all of our headings out in a list, we should instead curate them, group them, and make the flow of our documentation easier for users to digest

humitos:
Kolibri project does this in an interesting way:

![Screenshot_2019-06-06_13-03-22](https://user-images.githubusercontent.com/244656/59028297-8964d900-885b-11e9-88ad-5d525ab3b535.png)

> https://kolibri.readthedocs.io/en/latest/

davidfischer:
> The marketing material is important, but can be reduced. The top level documentation page isn't the best place to sell users -- the main product pages are a better place for this

I agree with this in theory but not strictly in practice. The big issue with docs.readthedocs.io is that it is the top level page. Clicking the logo or "home" goes to that page rather than to readthedocs.org. That is the reason why we need some marketing material ("what is RTD?" and "Why RTD") right at the top of that page.

Comparing our documentation with vaguely similar projects like [Travis CI](https://docs.travis-ci.com/), [Circle CI](https://circleci.com/docs/2.0/) both pages link back to the home page for those respective projects/companies and that's where the bulk of the marketing material lives. Until we do the same, I believe that we need a bit more marketing material.

PR #5819, opened 2019-06-17
Rework documentation index page

davidfischer:
## Goals

* Make sure the documentation is primarily focused on **users** (developers of RTD are secondary or lower)
* Explain the various sections in a bit more detail and guide users better than a simple toctree
* At this stage, I was not attempting to significantly restructure, rewrite, or remove any docs.

This structure was influenced heavily by the Django documentation but I definitely referred to other documentation including for Travis, CircleCI, and Heroku.

Fixes: #5675
Fixes: #2667


docs/index.rst
```diff
@@ -43,80 +43,185 @@ to help you create fantastic documentation for your project.
...
+
    user-defined-redirects
    automatic-redirects
+
+   guides/index
+
+   api/index
+   embed
+
+   vcs
```
davidfischer: These four sections (vcs, conda, etc.) seem to fit better in `docs/guides`. I propose moving them there. I also think `docs/features/embed` should be just a section of `docs/embed` rather than its own page.

docs/index.rst
```diff
@@ -43,80 +43,185 @@ to help you create fantastic documentation for your project.
...
+   sponsors
+
+   commercial/index
+
+   custom_installs/index
+
 
 .. toctree::
    :maxdepth: 1
-   :caption: Developer Documentation
```
davidfischer: Other than the security docs and changelog which I moved out of this section, I think that the developer docs should **not** be on the front docs page. Instead, I propose expanding `docs/contribute` and putting this toctree into a section there. I also think all of these RST files should be under `docs/development`.

docs/index.rst
```diff
@@ -43,80 +43,185 @@ to help you create fantastic documentation for your project.
...
+
+* **Multi-language documentation**:
+  :doc:`Translations and localization <localization>`
+
+* **Redirects**:
+  :doc:`User defined redirects <user-defined-redirects>` |
+  :doc:`Automatic redirects <automatic-redirects>`
+
+* **Topic specific guides**:
+  :doc:`How-to guides <guides/index>`
```
stsewd: What about putting this in the Troubleshooting section?

davidfischer: These guides (eg. "managing translations" or "adding custom JS/CSS to sphinx") are not exactly troubleshooting. That's not to say we can't come up with something better than "how-to guides".

Maybe we could just have a full sentence giving more details about the guides. Django does something like that in their documentation:

Screen Shot 2019-06-17 at 4 45 12 PM

Travis has a whole section on their homepage for "Language specific guides":

Screen Shot 2019-06-17 at 4 44 57 PM

Our guides are a little too varied for what Travis does but maybe something similar to what Django does.

humitos:
> At this stage, I was not attempting to significantly restructure, rewrite, or remove any docs.

I think this stage will be covered by #5771 and now we can care about the first steps that you are doing here.

docs/index.rst
ericholscher: Makes sense. We really need to create a git-mapped redirect app, to make creating redirects for ourselves simpler.

docs/index.rst
ericholscher: They might need a bit of a rewrite, but I agree the "Feature" documentation has always felt a bit weird, and should probably just be a list of features and then guides for how to use them well.

davidfischer:
Before this is merged, I'm going to try to do a couple things. I should have them done today:

- Move a few things (vcs, conda, etc.) under guides
- Move developer documentation under contributing

davidfischer:
I did some minor moving things around but we should probably setup a few redirects once this is deployed (not just merged but deployed to `/en/stable`):

## Guides that moved
- `/canonical.html` -> `/guides/canonical.html`
- `/conda.html` -> `/guides/conda.html`
- `/vcs.html` -> `/guides/vcs.html`

## Development docs that moved
- `/design.html` -> `/development/design.html`
- `/design/apiv3.html` -> `/development/design/apiv3.html`
-  `/design/in-doc-search-ui.html` -> `/development/design/in-doc-search-ui.html`
- `/design/index.html` -> `/development/design/index.html`
- `/design/pr-builder.html` -> `/development/design/pr-builder.html`
- `/design/theme-context.html` -> `/development/design/theme-context.html`
- `/design/yaml-file.html` -> `/development/design/yaml-file.html`
- `/docs.html` -> `/development/docs.html`
- `/i18n.html` -> `/development/i18n.html`
- `/install.html` -> `/development/install.html`
- `/issue-labels.html` -> `/development/issue-labels.html`
- `/settings.html` -> `/development/settings.html`
- `/symlinks.html` -> `/development/symlinks.html`
- `/tests.html` -> `/development/tests.html`

davidfischer:
This also closes https://github.com/rtfd/readthedocs.org/issues/2667 because it removed the embed API docs which were broken.

ericholscher:
We also should redirect `/en/latest/` as well :/

davidfischer:
> We also should redirect /en/latest/ as well :/

Do you mean to redirect `/en/latest` to `/en/stable`?

ericholscher:
No, we need to also add those redirects to the `latest` version.

davidfischer:
Currently, when setting a page redirect, it applies to all versions:

Screen Shot 2019-06-18 at 2 56 46 PM

ericholscher:
Sounds good -- I'm happy to ship this once the redirects are in place. 👍

PR #5822, opened 2019-06-18
Sync External versions to our database

ericholscher:
This is just an initial proof of concept.
We still need to update the External type properly,
and do lots of other work.
This is just a start to show how we can sync and build them.


ericholscher:
This is taken up elsewhere 👍

PR #5828, opened 2019-06-19
Auto Build Documentation on Pull Request (PR Builder)

saadmk11:
This is the PR which will have all the initial code for the PR Builder.
There are lots of stuff that need to be done.
ref: #1340 #5684

### DIFF
--- a/readthedocs/api/v2/utils.py
+++ b/readthedocs/api/v2/utils.py
@@ -13,7 +13,9 @@ from readthedocs.builds.constants import (
     STABLE,
     STABLE_VERBOSE_NAME,
     TAG,
+    EXTERNAL,
 )
+from readthedocs.core.utils import trigger_build
 from readthedocs.builds.models import Version
 
 
@@ -77,6 +79,15 @@ def sync_versions(project, versions, type):  # pylint: disable=redefined-builtin
                     version_name,
                     version_id,
                 )
+        elif type == EXTERNAL:
+            created_version = Version.objects.create(
+                project=project,
+                type=type,
+                identifier=version_id,
+                verbose_name=version_name,
+            )
+            added.add(created_version.slug)
+
         else:
             # New Version
             created_version = Version.objects.create(
@@ -135,6 +146,9 @@ def delete_versions(project, version_data):
     versions_tags = [
         version['verbose_name'] for version in version_data.get('tags', [])
     ]
+    external_versions = [
+        version['verbose_name'] for version in version_data.get('external_branches', [])
+    ]
     versions_branches = [
         version['identifier'] for version in version_data.get('branches', [])
     ]
@@ -146,6 +160,10 @@ def delete_versions(project, version_data):
     to_delete_qs = to_delete_qs.exclude(
         type=BRANCH,
         identifier__in=versions_branches,
+    )
+    to_delete_qs = to_delete_qs.exclude(
+        type=EXTERNAL,
+        verbose_name__in=external_versions,
     )
     to_delete_qs = to_delete_qs.exclude(uploaded=True)
     to_delete_qs = to_delete_qs.exclude(active=True)
@@ -174,6 +192,21 @@ def run_automation_rules(project, versions_slug):
     rules = project.automation_rules.all()
     for version, rule in itertools.product(versions, rules):
         rule.run(version)
+
+
+def trigger_external_build(project, version_list):
+    """
+    Trigger Builds for all external versions provided.
+
+    The rules are sorted by priority.
+
+    """
+    for version_slug in version_list:
+        version = project.versions(manager=EXTERNAL).get(slug=version_slug)
+        version.active = True
+        version.save()
+
+        trigger_build(project=project, version=version)
 
 
 class RemoteOrganizationPagination(PageNumberPagination):
--- a/readthedocs/api/v2/utils.py
+++ b/readthedocs/api/v2/utils.py
@@ -13,9 +13,7 @@ from readthedocs.builds.constants import (
     STABLE,
     STABLE_VERBOSE_NAME,
     TAG,
-    EXTERNAL,
 )
-from readthedocs.core.utils import trigger_build
 from readthedocs.builds.models import Version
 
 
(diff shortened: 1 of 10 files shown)


readthedocs/api/v2/utils.py
```diff
@@ -77,6 +79,15 @@ def sync_versions(project, versions, type):  # pylint: disable=redefined-builtin
...
                 )
+        elif type == EXTERNAL:
+            created_version = Version.objects.create(
+                project=project,
+                type=type,
+                identifier=version_id,
+                verbose_name=version_name,
+            )
+            added.add(created_version.slug)
+
```
ericholscher: Isn't this line the same as the line below it? Why do we need this extra logic?

readthedocs/api/v2/utils.py
```diff
@@ -176,6 +194,22 @@ def run_automation_rules(project, versions_slug):
...
+    This will mark all the provided external versions as active
+    and trigger build for those versions
+
+    """
+    for version_slug in version_list:
+        version = project.versions(manager=EXTERNAL).get(slug=version_slug)
+        version.active = True
+        version.save()
+
+        trigger_build(project=project, version=version)
```
ericholscher: We definitely don't want to trigger all builds for external versions. This will create hundreds of builds on first import. We need a better mechanism for this I think.

readthedocs/api/v2/views/integrations.py
```diff
@@ -271,6 +274,10 @@ def handle_webhook(self):
                 raise ParseError('Parameter "ref" is required')
         if event in (GITHUB_CREATE, GITHUB_DELETE):
             return self.sync_versions(self.project)
+
+        if event == GITHUB_PULL_REQUEST:
+            return self.sync_versions(self.project)
```
ericholscher: This might be the place where we want to trigger a build after syncing. I'm not 100% sure though.

ericholscher: We might need to think a bit more about how we're doing version syncing actually. I think maybe in this case we can just create the version via the incoming PR and trigger the build. That way we don't need to clone the repo 2x.

readthedocs/api/v2/views/integrations.py
```diff
@@ -218,6 +229,13 @@ def get_data(self):
                 pass
         return super().get_data()
 
+    def get_external_version_data(self):
+        """Get Commit Sha and pull request number from payload"""
+        identifier = self.data['pull_request']['head']['sha']
+        verbose_name = str(self.data['number'])
```
saadmk11: here I'm considering `PR_Number` as verbose name.

readthedocs/api/v2/views/integrations.py
```diff
@@ -271,6 +289,21 @@ def handle_webhook(self):
...
+        if (
+            event == GITHUB_PULL_REQUEST and
+            self.data['action'] in [GITHUB_PULL_REQUEST_OPEN, GITHUB_PULL_REQUEST_SYNC]
+        ):
+            try:
+                identifier, verbose_name = self.get_external_version_data()
+                external_version = get_or_create_external_version(
+                    self.project, identifier, verbose_name
+                )
+                return self.get_response_push(self.project, external_version.verbose_name)
```
saadmk11: To use the pre-written `self.get_response_push()` we can not pass a version instance so here I passed the verbose name but I think it will be much better if we pass the instance. to do this we need to create another function and not use `self.get_response_push()` we might use that function to send `triggered` status on the PR using the `Github Status API`. I would like to get you thought on this :)

readthedocs/api/v2/views/integrations.py
```diff
@@ -271,6 +289,21 @@ def handle_webhook(self):
...
+
+            except KeyError:
+                raise ParseError('Parameters "sha" and "number" are required')
```
saadmk11: The Error message can be changed

readthedocs/core/views/hooks.py
```diff
@@ -88,3 +90,21 @@ def sync_versions(project):
     except Exception:
         log.exception('Unknown sync versions exception')
     return None
+
+
+def get_or_create_external_version(project, identifier, verbose_name):
+    external_version = project.versions(manager=EXTERNAL).filter(verbose_name=verbose_name).first()
+    if external_version:
+        if external_version.identifier != identifier:
```
saadmk11: This will be updated when there is a new commit in the PR

readthedocs/projects/tasks.py
```diff
@@ -138,7 +138,7 @@ def sync_repo(self):
                 }
             )
             version_repo = self.get_vcs_repo()
-            version_repo.update()
+            version_repo.update(version=self.version)
```
saadmk11: we need to pass the version here so that we can determine which PR should be fetched. there previous implementation fetched all the PR's all together regardless it was `open` or `closed`

readthedocs/api/v2/views/integrations.py
```diff
@@ -271,6 +289,21 @@ def handle_webhook(self):
                 raise ParseError('Parameter "ref" is required')
         if event in (GITHUB_CREATE, GITHUB_DELETE):
             return self.sync_versions(self.project)
+
+        if (
+            event == GITHUB_PULL_REQUEST and
+            self.data['action'] in [GITHUB_PULL_REQUEST_OPEN, GITHUB_PULL_REQUEST_SYNC]
```
saadmk11: This will handle both PR `opened` and `synchronized` (new commit in the PR) events.

ericholscher:
This approach looks good to me, so far!

readthedocs/api/v2/views/integrations.py
ericholscher: Yea, we can use whatever makes sense. If it's doing something different than existing code, don't feel bad coming up with a new method.

readthedocs/core/views/hooks.py
ericholscher: This is probably better in a code comment, not a PR comment that will be lost :)

readthedocs/projects/tasks.py
ericholscher: Seems like a good and small change 👍

readthedocs/api/v2/views/integrations.py
ericholscher: Tested it locally and it worked great 👍

PR #5831, opened 2019-06-19
Reduce the number of queries we do on search indexing

ericholscher:
This does two things:

* Removes a `count` which wasn't useful
* Does a select_related on the `project` and `version` which we were querying for each object.

In my testing this vastly reduced query counts during indexing from 2+N*2 to 2

### DIFF
--- a/readthedocs/settings/base.py
+++ b/readthedocs/settings/base.py
@@ -539,6 +539,12 @@ class CommunityBaseSettings(Settings):
                 'filename': os.path.join(LOGS_ROOT, 'debug.log'),
                 'formatter': 'default',
             },
+            'db': {
+                'level': 'DEBUG',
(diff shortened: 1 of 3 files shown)


readthedocs/settings/base.py
```diff
@@ -539,6 +539,12 @@ def USE_PROMOS(self):  # noqa
                 'filename': os.path.join(LOGS_ROOT, 'debug.log'),
                 'formatter': 'default',
             },
+            'db': {
+                'level': 'DEBUG',
```
davidfischer: We have to be pretty careful that this won't happen in production. That could end up being a large file.

humitos: Maybe these settings can go into the `dev.py` file.

ericholscher:
This was done elsewhere and better. 👍

PR #5833, opened 2019-06-20
Hide "Protected" privacy level from users

humitos:
Protected privacy level causes a lot of confusions to users. I'm hiding it for now as a temporarily solution while we implement more Version states and this behavior can be easily managed. See https://github.com/rtfd/readthedocs.org/issues/5321

All the logic around Protect versions keeps working as is. Although, Projects and Versions are not being able to be added as `Protected` once this PR gets merged.

> This is just a proposal to ease the problem, but it's not a solution.


readthedocs/projects/constants.py
```diff
@@ -80,7 +80,10 @@
 
 PRIVACY_CHOICES = (
     (PUBLIC, _('Public')),
-    (PROTECTED, _('Protected')),
+    # Remove Protected for now since it cause confusions to users.
+    # There is a better way to manage this by using Version states
+    # See: https://github.com/rtfd/readthedocs.org/issues/5321
+    # (PROTECTED, _('Protected')),
```
ericholscher: What do people with `Protected` in the DB currently see for this listing?

humitos: Weird stuffs :)

Some places it will say `protected` (version listing) and some other `Public` (version details).

It may be better to remove it from the Form itself by overriding the field instead of from this global constant.

agjohnson: Yeah, i'd have the same reservations around just removing it from the database full stop. I think we might need to be more gentle here, or perhaps even use a feature flag to grandfather projects into `protected`.

We also discussed maybe researching how projects are using the feature now, and probably messaging the users before making a hard change like this one.

humitos: I'm not too worry about asking users how they use `protected` because it's definitely a weird concept and it's currently broken.

I don't want to remove it from the DB either, I just want to hide it from the users so they can not use `Protected` anymore in new projects/versions and get confused with it. The old ones should remain as they are: "in a Protected privacy level, but broken"

humitos:
I update the PR. This is better than what we have. Version/Project listing will say `Protected` if they have that level in DB. Although, if you go into "Edit" one of this object you will see `Public` (because `Protected` is not allowed in the form). It kind of makes sense if we want people moving out from this privacy level.

humitos:
If this "partial solution" does not fulfill what we want, we should just close the PR and spend time on the real solution.

readthedocs/core/mixins.py
```diff
@@ -22,3 +25,29 @@ class LoginRequiredMixin:
...
+    https://github.com/rtfd/readthedocs.org/issues/5321
+    """
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        if any([
+            not self.instance,
+            self.instance and self.instance.privacy_level != PROTECTED,
+        ]):
```
ericholscher: This logic feels quite complex. Can we not just do something like:

`if self.instance is None or self.instance.privacy_level != PROTECTED`

humitos:
I added a test case for this and simplify the if logic as you mentioned. We can merge after the tests pass.

PR #5845, opened 2019-06-25
Re-work around design document for APIv3

humitos:
This commit,

* describes main Goals/Non-Goals
* includes implementation plan based on stages
* updates the document mentioning what's already implemented (Version 1)
* proposes a plan for iteration 2, 3 and 4
* ~~mentions open questions for further discussion~~
* includes potential feature out of the current roadmap

### DIFF
--- a/docs/development/design/apiv3.rst
+++ b/docs/development/design/apiv3.rst
@@ -20,15 +20,31 @@ from where most of the endpoint will be based on it.
 Goals
 -----
 
-* Easy to use for our users
+* Easy to use for our users (access most of resources by ``slug``)
 * Useful to perform read and write operations
-* Cover most useful cases
+* Authentication/Authorization
+
+  * Authentication based on scoped-tokens
+  * Handle Authorization nicely using an abstraction layer
+
+* Cover most useful cases:
+
+  * Integration on CI (check build status, trigger new build, etc)
+  * Usage from public Sphinx/MkDocs extensions
+  * Allow creation of flyout menu client-side
+  * Simplify migration from other services (import projects, create multiple redirects, etc)
 
 
 Non-Goals
 ---------
 
 * Filter by arbitrary and non-useful fields
+
+  * "Builds with ``exit_code=1``"
+  * "Builds containing ``ERROR`` on their output"
+  * "Projects created after X datetime"
+  * "Versions with tag ``python``"
+
 * Cover *all the actions* available from the WebUI
 
 
--- a/docs/development/design/apiv3.rst
+++ b/docs/development/design/apiv3.rst
@@ -141,18 +141,6 @@ These are some features that we may want to build but they are not in the roadma
 * Internal Build process
 
 
-Open questions
---------------
-
-There are some questions that we still have.
-These will need more discussion before making a decision on where,
-when and how to implement them.
-
-* Do we want to add ``/api/v2/sustainability/`` to APIv3?
-  Should be part of the new "Ad Server" that we are building?
-* Should we make our search endpoint at ``/api/v2/search`` publicly on APIv3?
-
-
 Nice to have
 ------------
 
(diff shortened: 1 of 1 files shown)


docs/development/design/apiv3.rst
```diff
@@ -1,15 +1,34 @@
-Design of APIv3
-===============
+APIv3
```
ericholscher: I think we should include design in the title still, so it's easier to find in search, etc.

docs/development/design/apiv3.rst
```diff
@@ -1,15 +1,34 @@
...
+=====
+
+This document describes the design,
+some decisions already made and an implementation plan for APIv3 in stages.
```
ericholscher: This sentence needs something connecting the 2 halves.

docs/development/design/apiv3.rst
```diff
@@ -1,15 +1,34 @@
...
 It will be based on Resources as APIv2 but considering the ``Project`` resource as the main one,
 from where most of the endpoint will be based on it.
 
 
+Goals
+-----
+
+* Easy to use for our users
+* Useful to perform read and write operations
+* Cover most useful cases
```
ericholscher: This needs to be clearer. We should define these.

docs/development/design/apiv3.rst
```diff
@@ -1,15 +1,34 @@
...
+
+
+Non-Goals
+---------
+
+* Filter by arbitrary and non-useful fields
```
ericholscher: Again, this isn't clear, we need to define this more.

docs/development/design/apiv3.rst
```diff
@@ -20,185 +39,117 @@ Problems with APIv2
...
+--------------
 
-Considering some useful cases for the corporate site:
+These are some features that we may want to build but they are not in the roadmap at this moment:
 
-* Give access to a doc page (``objects.inv``, ``/design/core.html``)
+* CRUD for Domain
+* Add User as maintainer
+* Give access to a documentation page (``objects.inv``, ``/design/core.html``)
+* Internal Build process
```
ericholscher: We should mention why all of these are out of the roadmap.

docs/development/design/apiv3.rst
```diff
@@ -20,185 +39,117 @@ Problems with APIv2
...
 
-* Random filtering over a whole and not useful Resource
+There are some questions that we still have.
+These will need more discussion before making a decision on where,
+when and how to implement them.
 
-  * "All the ``stable`` versions"
-  * "Builds with ``exit_code`` equal to 257"
+* Do we want to add ``/api/v2/sustainability/`` to APIv3?
+  Should be part of the new "Ad Server" that we are building?
```
ericholscher: Definitely not on both of these. We can remove them.

docs/development/design/apiv3.rst
```diff
@@ -20,185 +39,117 @@ Problems with APIv2
...
+* Should we make our search endpoint at ``/api/v2/search`` publicly on APIv3?
```
ericholscher: I don't believe this exists. Are you talking about `docsearch`?

humitos: Yes, `docsearch`

docs/development/design/apiv3.rst
humitos: I put them out of the roadmap for now because, in case we build them, we need to decide in which Version they will be assigned. Also, they do not affect too many users or are only for internal usage only.

I'm updating the paragraph mentioning this.

docs/development/design/apiv3.rst
humitos: I'm not sure to understand what you refers to here. I expanded a little more the sentence, but I don't think that is what you want though :)

docs/development/design/apiv3.rst
humitos: I expand the list. Looks better now and more concise.

humitos:
@ericholscher I worked on the feedback you gave me. It looks better to me now. Let me know if there are more things to tackle.

humitos:
@ericholscher thanks you for the good feedback :wink:

PR #5848, opened 2019-06-26
Add a relation from SphinxDomain -> HTMLFile

ericholscher:
This will help with search indexing,
as it will let us get this data through a relation on the model instead of in naive Python code.

Refs #5821


readthedocs/sphinx_domains/admin.py
```diff
@@ -6,10 +6,10 @@
 
 class SphinxDomainAdmin(admin.ModelAdmin):
     list_filter = ('type',)
-    list_display = ('docs_url', 'version', 'build')
-    raw_id_fields = ('project', 'version')
+    list_display = ('__str__', 'version', 'build')
```
ericholscher: This uses _way_ fewer queries than doing a `resolve`

dojutsu-user: How did you actually find that out?

ericholscher: django-debug-toolbar shows the query count 👍

readthedocs/sphinx_domains/models.py
```diff
@@ -32,6 +32,11 @@ class SphinxDomain(TimeStampedModel):
         verbose_name=_('Version'),
         related_name='sphinx_domains',
     )
+    html_file = models.ForeignKey(
+        HTMLFile,
+        related_name='sphinx_domains',
+        null=True,
```
davidfischer: I would also suggest `blank=True` unless we're planning to backfill this.

ericholscher: The migration just makes it `null`, and it should be filled going forward. Any reason to allow `blank`?

humitos: If we want to allow `null` on already existing objects, but force on new ones, it makes sense as it is to me.

ericholscher:
Gonna merge this 🎉

PR #5850, opened 2019-06-26
Improve performance of SphinxDomains in the Django Admin

davidfischer:
This page is currently timing out in the Django admin and this change should fix it.

Put very simply, calling `resolve` in a loop is too expensive right now.


humitos:
I think @ericholscher did the same at #5848

davidfischer:
I always miss these things.

ericholscher:
Thanks! Great minds, etc :)

PR #5854, opened 2019-06-27
Search indexing with storage

davidfischer:
## Changes

* Indexes `HTMLFile`s and `ImportedFile`s from storage rather than from local disk
* Reads intersphinx data from storage (over HTTP) rather than local disk
* This change (as currently written) would begin to require `settings.RTD_BUILD_MEDIA_STORAGE`. This has implications for RTD corporate as well as development.

Note: "storage" could be a filesystem backed storage rather than a remote cloud storage but basically this uses the Django storage abstraction.


davidfischer:
There are going to be some performance implications of this when storage is cloud storage as opposed to local storage. If, however, we stop doing any processing on non-HTML files, the performance difference probably won't matter.

readthedocs/projects/models.py
```diff
@@ -1231,26 +1231,36 @@ def get_processed_json(self):
...
+            fjson_paths.append(basename + '.fjson')
+            if basename.endswith('/index'):
+                new_basename = re.sub(r'\/index$', '', basename)
+                fjson_paths.append(new_basename + '.fjson')
+
+            storage_path = self.project.get_storage_path(
+                type_='json', version_slug=self.version.slug, include_file=False
+            )
+            try:
+                for fjson_path in fjson_paths:
```
ericholscher: I believe we should be able to get away from this soon. We should be starting to store the proper path of a file after https://github.com/rtfd/readthedocs-sphinx-ext/pull/62 is merged.

readthedocs/projects/tasks.py
```diff
@@ -1419,29 +1431,36 @@ def clean_build(version_pk):
...
     # Re-create all objects from the new build of the version
-    for root, __, filenames in os.walk(path):
+    storage_path = version.project.get_storage_path(
+        type_='html', version_slug=version.slug, include_file=False
+    )
+    for root, __, filenames in storage.walk(storage_path):
         for filename in filenames:
             if filename.endswith('.html'):
                 model_class = HTMLFile
             else:
```
ericholscher: I'd like to add the `elif project.cdn_enabled` here with an `else: continue`. 👍

davidfischer: What does `project.cdn_enabled` do exactly and what would it do differently for search?

ericholscher: It means we store _all_ ImportedFile's not just HTMLFile's, so we can purge them from the CDN properly when they change.

davidfischer:
This should be good for a full review. The test failure seems to be in master and is unrelated.

readthedocs/projects/tasks.py
```diff
@@ -1424,29 +1436,40 @@ def clean_build(version_pk):
...
     Update imported files for version.
 
     :param version: Version instance
-    :param path: Path to search
     :param commit: Commit that updated path
     :param build: Build id
     """
 
+    if not settings.RTD_BUILD_MEDIA_STORAGE:
+        return
```
ericholscher: We should likely log something here, so we don't get confused if indexing isn't working.

readthedocs/settings/base.py
```diff
@@ -234,7 +234,7 @@ def USE_PROMOS(self):  # noqa
 
     # Optional Django Storage subclass used to write build artifacts to cloud or local storage
     # https://docs.readthedocs.io/en/stable/settings.html#build-media-storage
```
ericholscher: This link needs to be updated, and the default changed.

davidfischer:
I think this is good for a full review. For .org and for development, this should just work without issue. For .com, I think `settings.RTD_BUILD_MEDIA_STORAGE` needs to be unset on the build machines (they shouldn't copy files anywhere, the syncers will do that) and should be the default `BuildMediaFileSystemStorage` on the web machines where indexing happens.

davidfischer:
It looks like some of the test directories changed in the last week and that's causing the test failures. I'll get them fixed up.

davidfischer:
Ok, I believe the tests should pass. The added tests did actually uncover a bug in this implementation. The paths generated were not quite right.

ericholscher:
Great, I'll get this shipped today 👍

PR #5857, opened 2019-06-27
APIv3 "Import Project" endpoint

humitos:
This PR allows `POST` at `/api/v3/projects` to Import a Project.

It follows the same idea that importing from the Dashboard:

1. Create the Project object
1. Assign the user who makes the request as maintainer
1. Send `project_import` Django signal
1. Trigger an initial build 

> Note that I had to refactor the function that trigger the initial build to be able to call the same logic from two different places.

### DIFF
--- a/readthedocs/api/v3/permissions.py
+++ b/readthedocs/api/v3/permissions.py
@@ -16,6 +16,7 @@ class PublicDetailPrivateListing(IsAuthenticated):
         if is_authenticated:
             if view.basename == 'projects' and any([
                     view.action == 'list',
+                    view.action == 'create',  # used to create Form in BrowsableAPIRenderer
                     view.action is None,  # needed for BrowsableAPIRenderer
             ]):
                 # hitting ``/projects/``, allowing
--- a/readthedocs/api/v3/serializers.py
+++ b/readthedocs/api/v3/serializers.py
@@ -394,14 +394,16 @@ class ProjectCreateSerializer(FlexFieldsModelSerializer):
     """Serializer used to Import a Project."""
 
     repository = RepositorySerializer(source='*')
+    homepage = serializers.URLField(source='project_url', required=False)
 
     class Meta:
         model = Project
         fields = (
             'name',
             'language',
+            'programming_language',
             'repository',
-            'project_url',  # project_homepage
+            'homepage',
         )
 
 
--- a/readthedocs/api/v3/serializers.py
+++ b/readthedocs/api/v3/serializers.py
@@ -9,7 +9,7 @@ from rest_flex_fields.serializers import FlexFieldsSerializerMixin
 from rest_framework import serializers
 
 from readthedocs.builds.models import Build, Version
-from readthedocs.projects.constants import LANGUAGES, PROGRAMMING_LANGUAGES
+from readthedocs.projects.constants import LANGUAGES, PROGRAMMING_LANGUAGES, REPO_CHOICES
 from readthedocs.projects.models import Project
 from readthedocs.redirects.models import Redirect, TYPE_CHOICES as REDIRECT_TYPE_CHOICES
 
@@ -313,7 +313,10 @@ class ProjectURLsSerializer(serializers.Serializer):
 class RepositorySerializer(serializers.Serializer):
 
     url = serializers.CharField(source='repo')
-    type = serializers.CharField(source='repo_type')
+    type = serializers.ChoiceField(
+        source='repo_type',
+        choices=REPO_CHOICES,
+    )
 
 
 class ProjectLinksSerializer(BaseLinksSerializer):
--- a/readthedocs/api/v3/serializers.py
+++ b/readthedocs/api/v3/serializers.py
@@ -389,6 +389,22 @@ class ProjectLinksSerializer(BaseLinksSerializer):
         return self._absolute_url(path)
 
 
+class ProjectCreateSerializer(FlexFieldsModelSerializer):
+
+    """Serializer used to Import a Project."""
+
+    repository = RepositorySerializer(source='*')
+
+    class Meta:
+        model = Project
+        fields = (
+            'name',
+            'language',
+            'repository',
+            'project_url',  # project_homepage
+        )
+
+
 class ProjectSerializer(FlexFieldsModelSerializer):
 
     language = LanguageSerializer()
--- a/readthedocs/api/v3/views.py
+++ b/readthedocs/api/v3/views.py
@@ -1,6 +1,7 @@
 import django_filters.rest_framework as filters
 from django.utils.safestring import mark_safe
 from rest_flex_fields.views import FlexFieldsMixin
+from rest_framework import status
 from rest_framework.authentication import TokenAuthentication
 from rest_framework.decorators import action
 from rest_framework.metadata import SimpleMetadata
@@ -174,7 +175,7 @@ class ProjectsViewSet(APIv3Settings, NestedViewSetMixin, ProjectQuerySetMixin,
             data = self.get_serializer(superproject).data
             return Response(data)
         except Exception:
-            return Response(status=404)
+            return Response(status=status.HTTP_404_NOT_FOUND)
 
 
 class SubprojectRelationshipViewSet(APIv3Settings, NestedViewSetMixin,
@@ -263,7 +264,7 @@ class VersionsViewSet(APIv3Settings, NestedViewSetMixin, ProjectQuerySetMixin,
         # ``httpOnly`` on our cookies and the ``PUT/PATCH`` method are triggered
         # via Javascript
         super().update(request, *args, **kwargs)
-        return Response(status=204)
+        return Response(status=status.HTTP_204_NO_CONTENT)
 
 
 class BuildsViewSet(APIv3Settings, NestedViewSetMixin, ProjectQuerySetMixin,
@@ -303,11 +304,11 @@ class BuildsCreateViewSet(BuildsViewSet, CreateModelMixin):
 
         if build:
             data.update({'triggered': True})
-            status = 202
+            code = status.HTTP_202_ACCEPTED
         else:
             data.update({'triggered': False})
-            status = 400
-        return Response(data=data, status=status)
+            code = status.HTTP_400_BAD_REQUEST
+        return Response(data=data, status=code)
 
 
 class RedirectsViewSet(APIv3Settings, NestedViewSetMixin, ProjectQuerySetMixin,
--- a/readthedocs/api/v3/views.py
+++ b/readthedocs/api/v3/views.py
@@ -145,7 +145,7 @@ class ProjectsViewSet(APIv3Settings, NestedViewSetMixin, ProjectQuerySetMixin,
             # ProjectViewSet that returns the superproject of a project.
             return ProjectSerializer
 
-        if self.action in ('create',):
+        if self.action == 'create':
             return ProjectCreateSerializer
 
     def get_queryset(self):
--- a/readthedocs/api/v3/views.py
+++ b/readthedocs/api/v3/views.py
@@ -22,6 +22,7 @@ from readthedocs.builds.models import Build, Version
 from readthedocs.core.utils import trigger_build
 from readthedocs.projects.models import Project
 from readthedocs.projects.views.mixins import ProjectImportMixin
+from readthedocs.redirects.models import Redirect
 
 
 from .filters import BuildFilter, ProjectFilter, VersionFilter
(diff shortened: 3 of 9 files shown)


readthedocs/api/v3/permissions.py
```diff
@@ -16,6 +16,7 @@ def has_permission(self, request, view):
         if is_authenticated:
             if view.basename == 'projects' and any([
                     view.action == 'list',
+                    view.action == 'create',  # used to create Form in BrowsableAPIRenderer
```
ericholscher: Do we want to support that? I feel like creating a project via the web form is probably not going to be the best UX.

readthedocs/api/v3/views.py
```diff
@@ -126,6 +129,20 @@ class ProjectsViewSet(APIv3Settings, NestedViewSetMixin, ProjectQuerySetMixin,
         'active_versions.last_build.config',
     ]
 
+    def get_serializer_class(self):
+        """
+        Return correct serializer depending on the action.
+
+        For GET it returns a serializer with many fields and on PUT/PATCH/POST,
+        it return a serializer to validate just a few fields.
```
ericholscher: Does this not allow us to update the entire project via API?

readthedocs/api/v3/views.py
```diff
@@ -126,6 +129,20 @@ class ProjectsViewSet(APIv3Settings, NestedViewSetMixin, ProjectQuerySetMixin,
...
+        """
+        if self.action in ('list', 'retrieve', 'superproject'):
```
ericholscher: what is `superproject` here? They should probably be constants.

readthedocs/api/v3/views.py
```diff
@@ -161,6 +178,33 @@ def get_view_description(self, *args, **kwargs):  # pylint: disable=arguments-di
...
+        """
+        serializer = self.get_serializer(data=request.data)
+        serializer.is_valid(raise_exception=True)
+        project = serializer.save()
+        headers = self.get_success_headers(serializer.data)
+
+        # TODO: these lines need to be adapted for Corporate
+        project.users.add(request.user)
+        project_import.send(sender=project, request=request)
+        trigger_initial_build(project, request.user)
```
ericholscher: I wonder if this should all be abstracted into a single function called `trigger_build(initial)` and used in all the builds, so we only have to change this logic in one place.

readthedocs/api/v3/views.py
```diff
@@ -161,6 +178,33 @@ def get_view_description(self, *args, **kwargs):  # pylint: disable=arguments-di
...
+        * Sent project_import signal
+        * Trigger an initial Build
+        """
+        serializer = self.get_serializer(data=request.data)
+        serializer.is_valid(raise_exception=True)
+        project = serializer.save()
+        headers = self.get_success_headers(serializer.data)
+
+        # TODO: these lines need to be adapted for Corporate
+        project.users.add(request.user)
```
ericholscher: Where does this happen in the normal import workflow? I can't find it in the code.

ericholscher: Ah, was on the subclass: https://github.com/rtfd/readthedocs.org/blob/master/readthedocs/projects/forms.py#L50-L53

ericholscher: So perhaps we should be inheriting form this form for the Form we're using?

humitos:
> This looks good, but I do think it might be useful to keep all the logic together from the view and code

Like a `import_project(validated_data, request)` helper that, create the project, assign the user, trigger the initial build, etc? I think this can be handy, I need to think a little on the implementation since we probably will need a class so it can be override from corporate (that has some differencies).

readthedocs/api/v3/permissions.py
humitos: I've been thinking on making the browsable API consistent with what the API offers. Besides consistency, it helps to explore it and have a better understanding of what's available and what can be done: not "self-documented" but in that direction.

readthedocs/api/v3/views.py
humitos: No, in theory only the fields from `ProjectCreateSerializer.Meta.fields` are allowed (name, language, repository and project_url). If there are extra fields passed, they are ignored.

We may want to add more, though.

readthedocs/api/v3/views.py
humitos: `superproject` is the `@action` defined in the `ProjectViewSet` that returns the superproject of a project.

I'm not sure if it makes sense to make it a constant. Although, maybe it's good to add a comment in this line to avoid confusions.

readthedocs/api/v3/views.py
humitos: I'm not sure I like to inherit a Serializer from a Wizard Form. Although, maybe these common methods could be extracted into a Mixin that it's usage by both. Actually, in that case, there is nothing to change in corporate since it's already done the override.

humitos: This involves a bigger refactor than I thought. All the process to import a project is very tied to forms methods. `ProjectBasicForm.save` links the remote repository.

`ImportWizardView` manipulates the Basic and the Extra form to make it all fit in one project before saving it plus triggering a new build (`done` method in the form)

readthedocs/api/v3/permissions.py
humitos: On the other hand, removing the `create` from this check it fails with 404 when accessing to `/api/v3/projects/` because of an internal DRF check. Actually, that's how I realized that I needed it :)

humitos:
@ericholscher I refactored the Import Project code by extracting the shared bits into a Mixin. Looks better than the first option that I proposed, but still it's not perfect.

All the manipulation of the fields (from the Form or Serializer) happen on each view and once those things are done, the data is passed to the common method that will finish the import.

readthedocs/projects/views/mixins.py
```diff
@@ -44,3 +47,46 @@ def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
         context[self.project_context_object_name] = self.get_project()
         return context
+
+
+class ProjectImportMixin:
+
+    """Helpers to import a Project."""
+
+    def import_project(self, project, tags, request):
```
humitos: Maybe `finish_import_project` is a better name here, what do you think @ericholscher? Considering that the Project object is created outside this function and this one does the final steps.

readthedocs/api/v3/views.py
humitos: Added.

readthedocs/api/v3/serializers.py
```diff
@@ -375,6 +378,24 @@ def get_translations(self, obj):
...
+    homepage = serializers.URLField(source='project_url', required=False)
+
+    class Meta:
+        model = Project
+        fields = (
+            'name',
+            'language',
+            'programming_language',
+            'repository',
+            'homepage',
```
humitos: We can add more fields here. Maybe these are also useful:

- default_branch
- default_version
- tags
- analytics_code
- show_version_warning

readthedocs/api/v3/views.py
```diff
@@ -110,14 +115,13 @@ class ProjectsViewSet(APIv3Settings, NestedViewSetMixin, ProjectQuerySetMixin,
     * Subprojects of a project: ``/api/v3/projects/{project_slug}/subprojects/``
     * Superproject of a project: ``/api/v3/projects/{project_slug}/superproject/``
 
-    Go to [https://docs.readthedocs.io/en/stable/api/v3.html](https://docs.readthedocs.io/en/stable/api/v3.html)
+    Go to [https://docs.readthedocs.io/page/api/v3.html](https://docs.readthedocs.io/page/api/v3.html)
```
ericholscher: I think this should auto-link, no? Or you can put `<>` around it I believe (depending on markdown implementation, 😦)

readthedocs/api/v3/views.py
ericholscher: Yea, we should add a comment at least, since it's not like the others.

readthedocs/api/v3/views.py
```diff
@@ -126,6 +130,19 @@ class ProjectsViewSet(APIv3Settings, NestedViewSetMixin, ProjectQuerySetMixin,
...
+        """
+        Return correct serializer depending on the action.
+
+        For GET it returns a serializer with many fields and on PUT/PATCH/POST,
+        it return a serializer to validate just a few fields.
+        """
+        if self.action in ('list', 'retrieve', 'superproject'):
+            return ProjectSerializer
+
+        if self.action in ('create',):
```
ericholscher: any reason not to just use a list here? Or check for the value explicitly?

readthedocs/api/v3/views.py
```diff
@@ -161,6 +178,32 @@ def get_view_description(self, *args, **kwargs):  # pylint: disable=arguments-di
...
+
+    def perform_create(self, serializer):
+        """
+        Import Project.
+
+        Trigger our internal mechanism to import a project after it's saved in
+        the database.
+        """
+        project = serializer.save()
+        self.import_project(project, [], self.request)
```
ericholscher: What is the empty list for? We should use `kwargs` to make it more explicit.

readthedocs/projects/views/mixins.py
ericholscher: That seems better.

readthedocs/projects/views/mixins.py
```diff
@@ -44,3 +47,46 @@ def get_context_data(self, **kwargs):
...
+        """
+        Import a Project into Read the Docs.
```
ericholscher: As you noted in the PR comment above, we should note what work it assumes has already been done in the code comment as well.

readthedocs/projects/views/mixins.py
ericholscher: I might put the `request` first, as that is the normal order in Django in my experience.

readthedocs/projects/views/mixins.py
```diff
@@ -44,3 +47,46 @@ def get_context_data(self, **kwargs):
...
+
+        update_docs, build = prepare_build(project)
+        if (update_docs, build) == (None, None):
+            return None
+
+        from readthedocs.oauth.tasks import attach_webhook
+        task_promise = chain(
+            attach_webhook.si(project.pk, user.pk),
+            update_docs,
+        )
```
ericholscher: Do we need to chain these? Seems like we still want to update the docs even if webhook setup fails, and they don't depend on each other in any real way.

readthedocs/projects/views/private.py
```diff
@@ -255,42 +257,28 @@ def done(self, form_list, **kwargs):
...
-            project.tags.add(tag)
+
         for field, value in list(form_data.items()):
             if field in extra_fields:
                 setattr(project, field, value)
         project.save()
 
-        # TODO: this signal could be removed, or used for sync task
-        project_import.send(sender=project, request=self.request)
+        self.import_project(project, tags, self.request)
```
ericholscher: kwargs here too would be good.

readthedocs/api/v3/views.py
humitos: Yes. It's rendered as Markdown using "some implementation" and `[Text](Link)` is the format that works.

readthedocs/api/v3/views.py
humitos: Just for consistency.

humitos:
> We also need to add public-facing docs here, that is the biggest missing piece for me.

I was thinking on finish all the docs in another PR so we can deploy this without blocking on docs, which are deployed just with a merge. It's on my list to write them.

readthedocs/api/v3/views.py
humitos: I put `request` as first argument and make `tags` an optional with `None` as default. I think it's better now.

readthedocs/projects/views/mixins.py
humitos: We do a chain just to follow an order. If `attach_webhook` fails, it handles it by itself, set a message to the user and the build process continue. It does not block the update of the docs.

Anyways, this logic is not something I want to change/refactor in this PR because it's something bigger than this.

humitos:
@ericholscher I think I made all the changes needed from your comments. I had some issues when rebasing with master, hopefully I didn't break anything :)

ericholscher:
Why not just merge master into the branch?

ericholscher:
now I can't really see what commits have changed since my review :/

ericholscher:
Looking at it though, the changes look good to me 👍

humitos:
@ericholscher I'm sorry about that. I usually do rebase against master but, I may need to change that flow because of this problem

PR #5859, opened 2019-06-27
API v3: Endpoint to trigger build on default version

humitos:
Making a `POST` request on `/api/v3/projects/<project_slug>/builds/`
will trigger a build for the default version of that project.

### DIFF
--- a/readthedocs/api/v3/views.py
+++ b/readthedocs/api/v3/views.py
@@ -263,7 +263,7 @@ class VersionsViewSet(APIv3Settings, NestedViewSetMixin, ProjectQuerySetMixin,
 
 
 class BuildsViewSet(APIv3Settings, NestedViewSetMixin, ProjectQuerySetMixin,
-                    FlexFieldsMixin, ReadOnlyModelViewSet):
+                    FlexFieldsMixin, CreateModelMixin, ReadOnlyModelViewSet):
     model = Build
     lookup_field = 'pk'
     lookup_url_kwarg = 'build_pk'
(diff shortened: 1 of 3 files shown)


readthedocs/api/v3/views.py
```diff
@@ -304,3 +319,14 @@ def create(self, request, **kwargs):  # pylint: disable=arguments-differ
             data.update({'triggered': False})
             status = 400
         return Response(data=data, status=status)
+
+
+class BuildsCreateViewSet(BuildsViewSet):
```
ericholscher: I don't quite follow this. What is this class for?

humitos: This view is registered at URLs like `/api/v3/projects/pip/versions/v3.6.2/builds/` which allow "listing of all builds for that version" or "trigger a build for that version"

The view inherit `BuildsViewSet` which is registered for URLs like `/api/v3/projects/pip/builds/` that allow "listing of all builds for that project" or "trigger a build for the default version". The only different between these views is the way how they get the version to trigger a build.

humitos:
> I do wonder whether actually supporting an endpoint that is "build the default version" is necessary. It does mean we'll have two endpoints that essentially do the same thing.

I see your point.

I was thinking this endpoint as a way to automate a process. Although, the same can be achieved by first hitting the `/projects/slug/` and get `default_version` from there and then using the build endpoint for a specific version.

You made me realize that in the end it's just a shortcut that we don't strictly need.

Maybe we can hold off in merging this PR.

humitos:
I'm closing this PR for now. We can revisit it in case we need this specific endpoint.

PR #5866, opened 2019-06-29
README link fix

dojutsu-user:


dojutsu-user:
@stsewd Updated the codecov link too.

PR #5868, opened 2019-07-01
APIv3 refactor some fields

humitos:
* Move `urls.project_homepage` to the root of the Project object
* Add Project, Builds, Versions and Documentation (all URLs under Read the Docs) in `urls`

These changes make the object detail more clear, although I'm not sold on the name `urls` --I'd like to have something more related to "Read the Docs URLs" or "URLs under Read the Docs platform". Any suggestion here?

Besides the above concern, I'd like to rename `project_homepage` to just `homepage` (once I get a better name for `urls`) since it's a field of the project itself.

### DIFF
--- a/readthedocs/api/v3/serializers.py
+++ b/readthedocs/api/v3/serializers.py
@@ -303,11 +303,6 @@ class ProgrammingLanguageSerializer(serializers.Serializer):
 
 class ProjectURLsSerializer(serializers.Serializer):
     documentation = serializers.CharField(source='get_docs_url')
-    project_homepage = serializers.SerializerMethodField()
-
-    def get_project_homepage(self, obj):
-        # Overridden only to return ``None`` when the description is ``''``
-        return obj.project_url or None
 
 
 class RepositorySerializer(serializers.Serializer):
@@ -388,6 +383,7 @@ class ProjectLinksSerializer(BaseLinksSerializer):
 
 class ProjectSerializer(FlexFieldsModelSerializer):
 
+    project_homepage = serializers.SerializerMethodField()
     language = LanguageSerializer()
     programming_language = ProgrammingLanguageSerializer()
     repository = RepositorySerializer(source='*')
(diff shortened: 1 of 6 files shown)


readthedocs/api/v3/serializers.py
```diff
@@ -438,6 +450,10 @@ class Meta:
             '_links',
         ]
 
+    def get_project_homepage(self, obj):
+        # Overridden only to return ``None`` when the description is ``''``
```
davidfischer:
```suggestion
        # Overridden only to return ``None`` when the project homepage is ``''``
```

readthedocs/api/v3/serializers.py
```diff
@@ -438,6 +450,10 @@ class Meta:
...
+        return obj.project_url or None
+
     def get_description(self, obj):
```
davidfischer: This should be removed right?

readthedocs/api/v3/serializers.py
```diff
@@ -300,13 +300,26 @@ def get_name(self, programming_language):
...
 
 
-class ProjectURLsSerializer(serializers.Serializer):
+class ProjectURLsSerializer(BaseLinksSerializer, serializers.Serializer):
+
+    """Serializer with all the user-facing URLs under Read the Docs."""
+
     documentation = serializers.CharField(source='get_docs_url')
-    project_homepage = serializers.SerializerMethodField()
+    homepage = serializers.SerializerMethodField()
```
davidfischer: Given that we have both a "project homepage" (something set by the user) and this "homepage" (the URL on RTD) I think the latter needs a better name. I think "home" is better than "homepage" although I'm blanking on something even better than that.

Screen Shot 2019-07-03 at 9 28 23 AM

humitos: `home` is better than `homepage`, as you said. Still not perfect, though. At least, is a little more clear and consistent with the WebUI.

I will use your suggestion.

humitos:
Updated the PR with the feedback given. It's ready to re-review and merge.

humitos:
Lint issues are fixed in #5857

PR #5886, opened 2019-07-06
Build only HTML and Save external version artifacts in different directory

saadmk11:


readthedocs/projects/models.py
```diff
@@ -543,8 +543,13 @@ def get_storage_path(self, type_, version_slug=LATEST, include_file=True):
         :return: the path to an item in storage
             (can be used with ``storage.url`` to get the URL)
         """
+        type_dir = type_
+        # Add `external/` prefix for external versions
+        if self.versions(manager=EXTERNAL).filter(slug=version_slug).exists():
```
ericholscher: I believe this runs on the builders, so we can't query the DB for security reasons. We might need to think more about how we can pass this data to the builder. Can we set `type_` differently coming in?

saadmk11: @ericholscher We call this function in some other places also. if we change how `_type` works then we have to check everywhere if the version is external or not then send the type accordingly and also we have to change how we get file `extension` in that function. but a better way would be that we pass `version_type` to the function and check it there (thought we have to handle it in all other places). What do you think we should do?

ericholscher: Sure, we just need to capture that info somewhere and pass it along. It's already passed into the build process somewhere right?

saadmk11: Yes it is in available the build process. but I have to check other places we call this function :)

readthedocs/projects/models.py
```diff
@@ -533,18 +533,30 @@ def get_subproject_urls(self):
...
+            version_slug=LATEST,
+            include_file=True,
+            version_type=None
+    ):
         """
         Get a path to a build artifact for use with Django's storage system.
 
         :param type_: Media content type, ie - 'pdf', 'htmlzip'
         :param version_slug: Project version slug for lookup
+        :param version_type: Project version type
```
ericholscher: Wrong order. This should be last.

saadmk11: Done!

PR #5891, opened 2019-07-08
Fix spaces in email subject link

davidfischer:
Some email clients that aren't gmail apparently don't handle this well as I've gotten a couple emails with the actual subject "Alternatives+to+advertising" instead of "Alternatives to advertising".


davidfischer:
The lint errors are unrelated. I'm going to merge this.

PR #5893, opened 2019-07-08
Rework on documentation guides

davidfischer:
- Significant improvements to the custom CSS/JS guide
- Broke out the guides into Read the Docs and Sphinx/MkDocs guides
- Small grammatical and copy improvements

I'll probably do more work on more of the guides slowly but surely. The key things to improve on are to make sure that the right words that people would search in google for are present on the page and it make sure to explain why in addition to how.

## Screenshot

Screen Shot 2019-07-08 at 4 33 11 PM

Here's the [current guides page](https://docs.readthedocs.io/en/stable/guides/index.html) for comparison.

### DIFF
--- a/docs/guides/index.rst
+++ b/docs/guides/index.rst
@@ -1,11 +1,44 @@
 Guides
 ======
 
-These guides will help walk you through the usage of Read the Docs.
+These guides will help walk you through specific use cases
+related to Read the Docs itself, documentation tools like Sphinx and MkDocs
+and how to write successful documentation.
 
+
+Sphinx & MkDocs how-to guides
+-----------------------------
+
+These guides will help you get the most out of your documentation authoring tool
+whether that is Sphinx or MkDocs.
 
 .. toctree::
-   :maxdepth: 1
-   :glob:
+    :maxdepth: 1
 
-   *
+    adding-custom-css
+    custom-404-page
+    manage-translations
+    mkdocs-old-versions
+    pdf-non-ascii-languages
+    remove-edit-buttons
+    vcs
+
+
+Read the Docs how-to guides
+---------------------------
+
+These guides will help you customize or tune aspects of the Read the Docs build environment.
+
+.. toctree::
+    :maxdepth: 1
+
+    build-notifications
+    build-using-too-many-resources
+    canonical
+    conda
+    environment-variables
+    feature-flags
+    google-analytics
+    sitemaps
+    specifying-dependencies
+    wipe-environment
--- a/docs/guides/pdf-non-ascii-languages.rst
+++ b/docs/guides/pdf-non-ascii-languages.rst
@@ -1,8 +1,8 @@
-Sphinx PDFs for non-ASCII languages
-===================================
+Sphinx PDFs with Unicode
+========================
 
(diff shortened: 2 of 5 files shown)


davidfischer:
Actually, I'm going to update this as it looks like Sphinx made this quite a bit easier in Sphinx 1.8.

humitos:
Side comment: "Sitemaps: An SEO tool for documentation" is not a guide since the user can't do anything with that. It's more like a Read the Docs Feature. Maybe we should move it there.

davidfischer:
> Side comment: "Sitemaps: An SEO tool for documentation" is not a guide since the user can't do anything with that. It's more like a Read the Docs Feature. Maybe we should move it there.

That's the next guide to refresh on the calendar. There's no longer a directory of "features" in the documentation. I think we should build it out into an SEO guide that also has other things like how to set `<meta>` tags and setting the canonical URL. If you want to work on it, let me know 😄. Otherwise I think that's next up in a few weeks.

docs/guides/pdf-non-ascii-languages.rst
```diff
@@ -1,5 +1,5 @@
-Build PDF format for non-ASCII languages
-========================================
+Sphinx PDFs for non-ASCII languages
```
ericholscher: non-ASCII should probably be something else. 
```suggestion
Sphinx PDFs for languages other than English
```

docs/guides/index.rst
```diff
@@ -1,11 +1,44 @@
 Guides
 ======
 
-These guides will help walk you through the usage of Read the Docs.
+These guides will help walk you through specific use cases
+related to Read the Docs itself, documentation tools like Sphinx and MkDocs
+and how to write successful documentation.
 
 
+Sphinx & MkDocs how-to guides
```
ericholscher: I do feel like we're making the guides harder to find when we add more nesting to the TOC's. I can almost never find the development docs now, since it's nested under 2 levels and really hard to ctrl+f to find. I think we might hit a similar issue here with folks wanting to go to our docs and do `ctrl+f` then search for `PDF`. I don't know the best way to solve this though :/

davidfischer: Since the guides were already not on the home page (anymore) I don't know if `ctrl+f` would really help in this instance.

In general, I believe that most users are finding pages by searching google and getting a direct link rather than going to our docs homepage first.

docs/guides/pdf-non-ascii-languages.rst
davidfischer: Technically it's non-latin languages. How about "Sphinx PDFs for Unicode Languages" or "Sphinx PDFs for Non-European Languages"?

ericholscher: Yea, I didn't know the best way of putting it. I'm fine with Unicode, as it seems the most correct. 👍

docs/guides/index.rst
ericholscher: Yea, definitely -- but I'd like to support both flows reasonably well. Perhaps promoting search into a landing page-style box in the main content would help here. I need to think more on it.

PR #5910, opened 2019-07-11
Add Feature Flag to Enable External Version Building

saadmk11:


readthedocs/api/v2/views/integrations.py
```diff
@@ -367,21 +367,22 @@ def handle_webhook(self):
...
-                ]
-            ):
-                # Handle opened, synchronize, reopened pull_request event.
-                return self.get_external_version_response(self.project)
-
-            if action == GITHUB_PULL_REQUEST_CLOSED:
-                # Handle closed pull_request event.
-                return self.get_delete_external_version_response(self.project)
+        if self.project.has_feature(Feature.ENABLE_EXTERNAL_VERSION_BUILD):
+            if event == GITHUB_PULL_REQUEST and action:
```
ericholscher: these can be joined together with an `and`, instead of a nested `if`.

saadmk11: Done! :)

PR #5912, opened 2019-07-11
Refactor to use our own QuerySet helpers

humitos:
Small refactor to share more code instead.


humitos:
Ups. I'm reopening this. I want to make this change but I had not have the time just yet.

humitos:
I tried debugging why this error happens but I wasn't able to find it out yet. I do not see any difference regarding the other QuerySet classes where we have custom methods and they work. I checked Django docs and for the case we want and I didn't find the issue. I'm confused about why this does not find the method :confused:

ericholscher:
@humitos would be worth getting this merged or closed

PR #5916, opened 2019-07-11
Update build list and detail page UX

saadmk11:
#### TODO:

- [x] Add Tests


readthedocs/builds/models.py
```diff
@@ -756,6 +775,20 @@ def is_stale(self):
...
+        return self.version.type == EXTERNAL
+
+    @property
+    def external_version_name(self):
+        if self.is_external:
+            try:
+                if self.project.remote_repository.account.provider == 'github':
+                    return GITHUB_EXTERNAL_VERSION_NAME
+            except Exception:
+                return None
```
ericholscher: Shouldn't this return a generic `External Version` or something?

readthedocs/templates/builds/build_detail.html
```diff
@@ -79,10 +79,17 @@
 
     <div class="build-version">
       <span class="build-version-slug">
-        {% if request.user|is_admin:project %}
+        {% if request.user|is_admin:project and not build.is_external %}
           <a href="{% url "project_version_detail" build.version.project.slug build.version.slug %}">{{ build.version.slug }}</a>
         {% else  %}
-          {{ build.version.slug }}
+          {% if build.is_external %}
```
ericholscher: shouldn't this just be an `{% else if %}` on the top-level `if`? This is more complex to read.

readthedocs/templates/core/build_list_detailed.html
```diff
@@ -6,7 +6,7 @@
...
                 <a href="{{ build.get_absolute_url }}">{% if build.is_stale %}<span class="icon-warning" title="{% trans 'This build is still waiting to be built' %}"></span>{% endif %}<span id="build-state">{% if build.state != 'finished' %}{{ build.get_state_display }} {% else %} {% if build.success %}{% trans "Passed" %}{% else %}{% trans "Failed" %}{% endif %}{% endif %}</span>
                   <img src="{% static 'core/img/loader.gif' %}" class="build-loading hide">
-              <span class="quiet">{% if build.version %}{% blocktrans with build.version.slug as slug and build.type as type %}version {{ slug }} ({{ type }}){% endblocktrans %}{% endif %}</span><span class="quiet right">{% blocktrans with build.date|timesince as date %}{{ date }} ago{% endblocktrans %}</span>
+              <span class="quiet">{% if build.version %}{% if build.is_external %}{% blocktrans %}External {% endblocktrans %}{% endif %}{% blocktrans with build.version.slug as slug and build.type as type %}version {{ slug }} ({{ type }}){% endblocktrans %}{% endif %}</span><span class="quiet right">{% blocktrans with build.date|timesince as date %}{{ date }} ago{% endblocktrans %}</span>
```
ericholscher: This should also use the `external_version_name` here.

readthedocs/builds/models.py
saadmk11: Yeah. that's what we were talking about. I will add that

readthedocs/templates/builds/build_detail.html
```diff
@@ -79,10 +79,17 @@
...
+              {% blocktrans with build.external_version_name as external_version_name %}
+                  <b>{{ external_version_name }}</b> #
+              {% endblocktrans %}
+              <a href="{{ build.version.get_external_version_url }}">{{ build.version.verbose_name }}</a>
```
ericholscher: We should put the `#` here, so it looks better in the UI.

readthedocs/templates/builds/build_detail.html
```diff
@@ -79,10 +79,17 @@
...
+          {% else %}
+             {{ build.version.slug }}
+          {% endif %}
         {% endif %}
       </span>
       <span class="build-commit"
```
ericholscher: I'd also love to link the commit as well, using the same pattern.

readthedocs/projects/constants.py
```diff
@@ -325,6 +325,14 @@
     'https://github.com/{user}/{repo}/'
     '{action}/{version}{docroot}{path}{source_suffix}'
 )
+GITHUB_PULL_REQEST_URL = (
+    'https://github.com/{user}/{repo}/'
+    'pull/{number}'
+)
+GITHUB_PULL_REQEST_COMMIT_URL = (
```
ericholscher: Typos here on `REQUEST`

readthedocs/builds/models.py
```diff
@@ -162,41 +162,36 @@ def vcs_url(self):
...
-        elif self.slug == LATEST:
-            slug_url = self.project.default_branch or self.project.vcs_repo().fallback_branch
-        else:
-            slug_url = self.slug
-
         if self.type == EXTERNAL:
             if 'github' in self.project.repo:
                 user, repo = get_github_username_repo(self.project.repo)
-                return GITHUB_PULL_REQEST_COMMIT_URL.format(
+                # Get Github Pull Request Commit URL.
```
ericholscher: This comment isn't useful :)

readthedocs/builds/models.py
```diff
@@ -162,41 +162,36 @@ def vcs_url(self):
...
                     user=user,
                     repo=repo,
                     number=self.verbose_name,
                     commit= self.commit_name
-
                 )
-            if 'gitlab' in self.project.repo:
-                slug_url = self.identifier
-                url = f'/merge_requests/{slug_url}/'
+            # TODO: Add VCS ULR for other Git Providers
```
ericholscher: URL mistype

readthedocs/builds/models.py
```diff
@@ -558,11 +553,12 @@ def get_external_version_url(self):
                 return ''
 
             repo = repo.rstrip('/')
-            return GITHUB_PULL_REQEST_URL.format(
+            return GITHUB_PULL_REQUEST_URL.format(
                 user=user,
                 repo=repo,
                 number=self.verbose_name
             )
+        # TODO: Add External Version ULR for other Git Providers
```
ericholscher: Mistype here too

readthedocs/builds/models.py
```diff
@@ -756,6 +778,20 @@ def is_stale(self):
...
+    def is_external(self):
+        return self.version.type == EXTERNAL
+
+    @property
+    def external_version_name(self):
+        if self.is_external:
+            try:
+                if self.project.remote_repository.account.provider == 'github':
+                    return GITHUB_EXTERNAL_VERSION_NAME
+            except Exception:
```
ericholscher: We should log this exception, otherwise it will hide real issues. What exception are we really trying to catch?

readthedocs/builds/models.py
```diff
@@ -158,9 +162,22 @@ def vcs_url(self):
         """
         Generate VCS (github, gitlab, bitbucket) URL for this version.
 
-        Branch/Tag Example: https://github.com/rtfd/readthedocs.org/tree/3.4.2/.
-        Pull/merge Request Example: https://github.com/rtfd/readthedocs.org/pull/9999/.
+        Example: https://github.com/rtfd/readthedocs.org/tree/3.4.2/.
+        External Version Example: https://github.com/rtfd/readthedocs.org/pull/99/commits/b630b630/.
```
ericholscher: Don't we want this URL to point to the PR still? Otherwise we are changing the meaning of this function based on whether the version is external or internal.

We only want specific commits to link to a commit, not the version itself. We need another function to add a specific URL to the commit I think.

readthedocs/builds/models.py
```diff
@@ -767,6 +751,60 @@ def get_full_url(self):
...
+                    return ''
+
+                repo = repo.rstrip('/')
+                return BITBUCKET_COMMIT_URL.format(
+                    user=user,
+                    repo=repo,
+                    commit=self.commit
+                )
+
+        return ''
```
ericholscher: I do wonder if we should return `None` here, instead of `''`. I don't think it matters too much though.

saadmk11: for URL's I think its better to return empty string. as it will show `href=''` not `href=None` in the template.

ericholscher: We should not output the link if we don't have an `href` attribute.

ericholscher: Merging this, but we should fix it ^^

saadmk11: @ericholscher Updated please see this https://github.com/readthedocs/readthedocs.org/pull/5933/commits/763b857626004deb987878d80f6e4513ea59922f

PR #5918, opened 2019-07-12
Add tests for section-linking

dojutsu-user:
Tests


ericholscher:
Is this still a WIP?

dojutsu-user:
Just some last commits.

readthedocs/search/tests/test_api.py
```diff
@@ -20,40 +25,112 @@ def setup_class(cls):
         # installed
         cls.url = reverse('doc_search')
 
-    @pytest.mark.parametrize('data_type', ['content', 'headers', 'title'])
+    @pytest.mark.parametrize('data_type', ['title'])
```
ericholscher: We don't need param's here if there's only 1.

readthedocs/search/tests/test_views.py
```diff
@@ -79,38 +82,121 @@ def _get_search_result(self, url, client, search_params):
...
+        # searching without the filter
+        results, page = self._get_search_result(
+            url=self.url,
+            client=client,
+            search_params=search_params
+        )
+        assert len(results) >= 2  # there are > 1 results without the filter
+
+        # checking if `std:confval` filter is present
+        content = page.find('.navigable .language-list').text()
```
ericholscher: Also we should fix the HTML, `language-list` isn't the class we should be using, it should be `role-list`.

ericholscher: I also worry about depending on the HTML structure. This will break all the tests when we do our site redesign. Can we not capture this data from the query result itself? I believe the Django test client gives us template data directly, which we could use: https://docs.djangoproject.com/en/2.2/topics/testing/tools/#django.test.Response.context

dojutsu-user:
@ericholscher 
I have updated the tests to be independent of templates as per your suggestion in the [comment above](https://github.com/readthedocs/readthedocs.org/pull/5918#discussion_r303584789).

ericholscher:
Great, thanks!

PR #5921, opened 2019-07-13
Remove external version files from storage on 'pr_close' webhook event

saadmk11:


ericholscher:
I don't think we actually need to do this. It will break all the old links from users in the PR's, which I don't think we want to do. I think we can basically keep this content forever, since it costs very little to store.

saadmk11:
Okay Got it :+1:

PR #5932, opened 2019-07-16
Fix spacing between the results and add highlight url param

dojutsu-user:


readthedocs/core/static-src/core/js/doc-embed/search.js
```diff
@@ -170,7 +169,12 @@ function attach_elastic_search_query(data) {
 
                             contents.find('em').addClass('highlighted');
                             list_item.append(contents);
-                            list_item.append($("<br>"));
+
+                            // Create some spacing between the results.
+                            // Also, don't add this spacing in the last hit.
+                            if (j !== inner_hits.length - 1) {
+                                list_item.append($("<div style='margin: 10px 0'></div>"));
+                            }
```
ericholscher: I don't like inline styles here, but it might be necessary since it needs to work across all themes. I feel like we need to make this a classname and have CSS though to make it be user-configurable.

ericholscher: We could include a CSS class addition here, which we add to all docs: https://assets.readthedocs.org/static/css/readthedocs-doc-embed.css

dojutsu-user: @ericholscher 
I have changed it use a css file and have added css styles in `readthedocs-doc-embed.css`.

PR #5933, opened 2019-07-16
Refine PR Builder Code

saadmk11:
This PR Will work on improving the code for PR builder by updating the code suggested in code reviews

### DIFF
--- a/readthedocs/core/views/hooks.py
+++ b/readthedocs/core/views/hooks.py
@@ -113,9 +113,8 @@ def get_or_create_external_version(project, identifier, verbose_name):
 
     if created:
         log.info(
-            '(Create External Version) Added Version: [%s] ', ' '.join(
-                external_version.slug
-            )
+            '(Create External Version) Added Version: [%s] ',
+            external_version.slug
         )
     else:
         # identifier will change if there is a new commit to the Pull/Merge Request
@@ -124,9 +123,8 @@ def get_or_create_external_version(project, identifier, verbose_name):
             external_version.save()
 
             log.info(
-                '(Update External Version) Updated Version: [%s] ', ' '.join(
-                    external_version.slug
-                )
+                '(Update External Version) Updated Version: [%s] ',
+                external_version.slug
             )
     return external_version
 
@@ -151,9 +149,8 @@ def delete_external_version(project, identifier, verbose_name):
         # Delete External Version
         external_version.delete()
         log.info(
-            '(Delete External Version) Deleted Version: [%s]', ' '.join(
-                external_version.slug
-            )
+            '(Delete External Version) Deleted Version: [%s]',
+            external_version.slug
         )
 
         return external_version.verbose_name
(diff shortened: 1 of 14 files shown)


readthedocs/core/views/hooks.py
```diff
@@ -105,30 +105,23 @@ def get_or_create_external_version(project, identifier, verbose_name):
...
-        manager=EXTERNAL
-    ).filter(verbose_name=verbose_name).first()
+    external_version, created = project.versions.get_or_create(
+        verbose_name=verbose_name,
+        type=EXTERNAL,
+        defaults={'identifier': identifier, 'active': True},
+    )
 
-    if external_version:
+    if not created:
```
ericholscher: I'd switch this logic around, and do `if created`. It just reads a bit cleaner.

readthedocs/core/views/hooks.py
```diff
@@ -105,30 +105,23 @@ def get_or_create_external_version(project, identifier, verbose_name):
...
         # identifier will change if there is a new commit to the Pull/Merge Request
         if external_version.identifier != identifier:
             external_version.identifier = identifier
             external_version.save()
```
ericholscher: We should probably log in this case too (eg. updated external version)

readthedocs/templates/builds/build_detail.html
```diff
@@ -91,7 +91,11 @@
         {% endif %}
       </span>
       <span class="build-commit" data-bind="visible: commit">
+        {% if build.get_commit_url %}
           (<a data-bind="attr: {href: commit_url}"><span data-bind="text: commit">{{ build.commit }}</span></a>)
+        {% else %}
+            (<span data-bind="text: commit">{{ build.commit }}</span>)
```
ericholscher: Does this break the JS code that tries to update `commit_url` if it doesn't exist? I don't know how knockout.js handles this, but we might also need to check for it on that side.

saadmk11: Okay I'll check this :)

saadmk11: I have checked. there is no error on the browsers console window. and `commit_url` does always exist. but set to `None` when conditions are not met.

readthedocs/projects/tasks.py
```diff
@@ -578,18 +578,21 @@ def run_build(self, docker, record):
 
         if self.build_env.failed:
             self.send_notifications(self.version.pk, self.build['id'])
-            # send build failure status to git Status API
+            # if the build failed, send build failure status to git Status API
+            # to show status report on the providers pull/merge request UI.
```
ericholscher: This is still just explaining what the code does. We can just remove these comments, we don't need to explain anything here.

readthedocs/core/views/hooks.py
```diff
@@ -105,30 +105,29 @@ def get_or_create_external_version(project, identifier, verbose_name):
...
+        verbose_name=verbose_name,
+        type=EXTERNAL,
+        defaults={'identifier': identifier, 'active': True},
+    )
 
-    if external_version:
+    if created:
+        log.info(
+            '(Create External Version) Added Version: [%s] ', ' '.join(
+                external_version.slug
```
ericholscher: Why are we calling `join` on a slug?

saadmk11: Opps! I was doing something else forgot to remove it sorry.

PR #5944, opened 2019-07-17
Properly return None when there's no highlight on a hit.

ericholscher:


dojutsu-user:
This also needs to change -- https://github.com/readthedocs/readthedocs.org/blob/0f47f05b2e48822161ee85cb647bac9c1a72f42e/readthedocs/search/views.py#L120-L134

I think we should move `_get_inner_hits_highlights` to `search.utils` and then use it from there?

dojutsu-user:
I think it would be better to move the entire logic of sorting the search results to `search.utils` and call from there.

dojutsu-user:
@ericholscher 
I have opened PR #5950

ericholscher:
Great, I'm going to merge this since it's already hotfixed, and then we can update that branch to not conflict 👍

PR #5950, opened 2019-07-17
Hotfix: Return empty dict when no highlight dict is present

dojutsu-user:
I have moved the sorting logic to `search.utils`.

### DIFF
--- a/readthedocs/search/api.py
+++ b/readthedocs/search/api.py
@@ -1,6 +1,5 @@
 import itertools
 import logging
-from pprint import pformat
 
 from rest_framework import generics, serializers
 from rest_framework.exceptions import ValidationError
@@ -38,7 +37,7 @@ class PageSearchSerializer(serializers.Serializer):
         highlight = getattr(obj.meta, 'highlight', None)
         if highlight:
             ret = highlight.to_dict()
-            log.debug('API Search highlight [Page title]: %s', pformat(ret))
+            log.debug('API Search highlight [Page title]: %s', ret)
             return ret
 
     def get_inner_hits(self, obj):
@@ -47,11 +46,13 @@ class PageSearchSerializer(serializers.Serializer):
             sections = inner_hits.sections or []
             domains = inner_hits.domains or []
             all_results = itertools.chain(sections, domains)
-            sorted_results = list(utils._get_sorted_results(
+
+            sorted_results = utils._get_sorted_results(
                 results=all_results,
                 source_key='_source',
-                logging=True,
-            ))
+            )
+
+            log.debug('[API] Sorted Results: %s', sorted_results)
             return sorted_results
 
 
--- a/readthedocs/search/api.py
+++ b/readthedocs/search/api.py
@@ -1,6 +1,5 @@
 import itertools
 import logging
-from operator import attrgetter
 from pprint import pformat
 
 from rest_framework import generics, serializers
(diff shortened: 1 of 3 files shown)


dojutsu-user:
Ohhh... I just noticed.
We have modified the parse_json to remove all the new-line characters.
If we don't have any new line characters in our index -- then there won't be any new line characters in the highlight.
I think the func `_remove_newlines_from_dict()` is no longer required.

dojutsu-user:
This should also increase the speed as that func is going through every highlight dict recursively and that too -- doing nothing.

readthedocs/search/utils.py
```diff
@@ -153,24 +155,27 @@ def _indexing_helper(html_objs_qs, wipe=False):
...
-        else:
-            # elastic returns the contents of the
-            # highlighted field in a list.
-            if isinstance(v, list):
-                v_new_list = [res.replace('\n', '. ') for res in v]
-                highlight[k] = v_new_list
-
-    return highlight
+        if logging:
+            log.debug('API Search highlight: %s', pformat(highlight_dict))
```
dojutsu-user: I also wanted to know that if we can do something for `pformat`.
@humitos pointed out that it is doing unnecessary work ([comment](https://github.com/readthedocs/readthedocs.org/pull/5919#discussion_r303579977)).
There are other places also where we are using `pformat` for logging purposes.
This must be reducing some speed and taking memory.

ericholscher: We can just output it unformatted.

readthedocs/search/api.py
```diff
@@ -48,27 +47,13 @@ def get_inner_hits(self, obj):
...
-                    '_source': hit._source.to_dict(),
-                    'highlight': self._get_inner_hits_highlights(hit),
-                }
-                for hit in sorted(all_results, key=attrgetter('_score'), reverse=True)
-            ]
-
+            sorted_results = list(utils._get_sorted_results(
+                results=all_results,
+                source_key='_source',
+                logging=True,
```
ericholscher: optional logging feels weird. Do we ever not want logs?

ericholscher: I think we should probably just log outside this function, and remove the logging from this function call.

readthedocs/search/utils.py
```diff
@@ -153,24 +155,27 @@ def _indexing_helper(html_objs_qs, wipe=False):
...
+
+        return highlight_dict
+    return {}
+
+
+def _get_sorted_results(results, source_key='_source', logging=False):
+    """Sort results according to their score and return a generator expression."""
+    sorted_results = (
```
ericholscher: Do we really need this to be a generator? What is the goal of it versus just using a list?  It looks like we're just calling `list` on it both times we use it, which means it should probably just return a list :)

readthedocs/search/utils.py
```diff
@@ -153,24 +155,27 @@ def _indexing_helper(html_objs_qs, wipe=False):
                 delete_objects_in_es.delay(**kwargs)
 
 
-def _remove_newlines_from_dict(highlight):
-    """
-    Recursively change results to turn newlines into periods.
+def _get_inner_hits_highlights(hit, logging=False):
```
ericholscher: It seems like maybe we don't even need this function. It's mostly just checking if the hit has a highlight and returning it.

dojutsu-user: Remvoed. :+1:

readthedocs/search/api.py
dojutsu-user: Done.

readthedocs/search/utils.py
dojutsu-user: I thought that returning a generator expression would be efficient than returning a huge list of results?

ericholscher:
Looks good once we fix the conflict 👍

PR #5952, opened 2019-07-17
APIv3 endpoint: allow to modify a Project once it's imported

humitos:
* allow users to modify existing projects using the API
* only allow to change fields that are not possible to change via the YAML
* users should only be able to change projects where they are maintainers


readthedocs/api/v3/tests/test_projects.py
```diff
@@ -181,3 +181,92 @@ def test_import_project_with_extra_fields(self):
...
+        response = self.client.patch(
+            reverse(
+                'projects-detail',
+                kwargs={
+                    'project_slug': self.others_project.slug,
+                },
+            ),
+            data,
+        )
+        self.assertEqual(response.status_code, 403)
```
humitos: This is returning 204, allowing a user to modify other's people project. There is a bug where we check for permissions that I need to fix before merging this.

humitos:
This PR should be ready to review and all the permissions should be fixed and working as they should.

readthedocs/api/v3/tests/test_projects.py
humitos: This is fixed in last commits.

docs/api/v3.rst
```diff
@@ -227,6 +227,35 @@ Project create
     :statuscode 400: Some field is invalid
 
 
+Project update
+++++++++++++++
+
+.. http:patch:: /api/v3/projects/(string:project_slug)/
+
+    Edit an existent project.
```
ericholscher:
```suggestion
    Edit an existing project.
```

ericholscher: We should also use `edit` or `update` here. The title says `update` but this says `edit`.

docs/api/v3.rst
```diff
@@ -227,6 +227,35 @@ Project create
...
+    The content of ``body.json`` is like,
+
+    .. sourcecode:: json
+
+        {
+            "name": "New name for the project",
+            "default_version": "v0.27.0"
+        }
+
+    :statuscode 204: Updated successfully
```
ericholscher: I assume it will also raise an error in some cases?

readthedocs/api/v3/mixins.py
```diff
@@ -92,3 +100,20 @@ def get_queryset(self):
 
         # List view are only allowed if user is owner of parent project
         return self.listing_objects(queryset, self.request.user)
+
+
+class UpdatePartialUpdateMixin:
```
ericholscher: This is a super confusing name. Why does it have `Update` in the name twice?

readthedocs/api/v3/permissions.py
```diff
@@ -22,7 +22,9 @@ def has_permission(self, request, view):
                 # hitting ``/projects/``, allowing
                 return True
 
-            if view.detail:
+            # detail view is only allowed on list/retrieve actions (not
+            # ``update`` or ``partial_update``)
+            if view.detail and view.action in ('list', 'retrieve', 'superproject'):
```
ericholscher: These feel like they should be constants? What is a `superproject`?  It still feels weird that it is a method name to me.

readthedocs/api/v3/serializers.py
```diff
@@ -427,6 +435,41 @@ class Meta:
         )
 
 
+class ProjectUpdateSerializer(FlexFieldsModelSerializer):
+
+    """Serializer used to modify a Project once imported."""
+
+    repository = RepositorySerializer(source='*')
```
ericholscher: What is `source='*'`? It should likely have a comment.

readthedocs/api/v3/views.py
```diff
@@ -271,7 +276,8 @@ class TranslationRelationshipViewSet(APIv3Settings, NestedViewSetMixin,
 # of ``ProjectQuerySetMixin`` to make calling ``super().get_queryset()`` work
 # properly and filter nested dependencies
 class VersionsViewSet(APIv3Settings, NestedViewSetMixin, ProjectQuerySetMixin,
-                      FlexFieldsMixin, UpdateModelMixin, ReadOnlyModelViewSet):
+                      FlexFieldsMixin, UpdatePartialUpdateMixin,
+                      UpdateModelMixin, ReadOnlyModelViewSet):
```
ericholscher: Does this also allow us to `PATCH` to Versions?

readthedocs/api/v3/permissions.py
humitos: `superproject` is an action name, defined by the class method under `ProjectViewSet`. We should apply the same permissions restrictions than for a detail action (since it only returns _one_ superproject if exists).

`list` and `retrieve` are DRF standard action names (same as `update` or `partial_update`).

We already talked about this at https://github.com/readthedocs/readthedocs.org/pull/5857#discussion_r299172350

(I'm adding this comment as a code comment)

readthedocs/api/v3/serializers.py
humitos: `*` it's standard for DRF but has a special meaning. It means that the whole object will be passed to the `RepositorySerializer`, not just a specific field of it.

(see docs https://www.django-rest-framework.org/api-guide/fields/#source)

readthedocs/api/v3/mixins.py
humitos: I should probably rename this to `UpdateMixin` only. This class makes PUT to return 204 on success as PATCH.

docs/api/v3.rst
humitos: On the "Ship APIv3" PR I was thinking to remove all the redundant docs or un-useful list of fields: avoid things like `slug: the slug of the Project` and just document that ones that are useful to avoid.

I applied the same concept for status codes, mentioning only the important ones (based on David's comment as well: https://github.com/readthedocs/readthedocs.org/pull/4863#discussion_r230834643)

I did a commit for this in that PR: https://github.com/readthedocs/readthedocs.org/pull/6169/commits/f7e168f929bff3a33b48e066786cfada653536ee

What do you think?

humitos:
> where is the logic that actually makes it so that PATCH works now?Is it the overridden `update` method?

Yes.

Basically, this is the line that allows the PATCH/PUT: https://github.com/readthedocs/readthedocs.org/pull/5952/files#diff-e7d0914abe4a0641d49078f90f75e695R77

`UpdateModelMixin` defines the `update` method (PUT) and `partial_update` (PATCH) and makes DRF to "listen" to these verbs as well. The rest of the PR are permission handling, refactor, tests and docs.

docs/api/v3.rst
ericholscher: Seems reasonable 👍

PR #5954, opened 2019-07-18
Show APIv3 Token under Profile settings

humitos:
Some comments here:

- Currently we support one Token per User (it's a OneToOne relation on the model that comes with DRF) --although, the Django view it's a ListView
- I added this token under the user's settings since it's where I think it makes the most sense
- Once we support Scoped Tokens, we can refactor this page to show them all here mentioning their limitations/allowed permissions.
- In case the Scoped Tokens would be Project-based, we can add a similar page under Project's Admin.

I'm not sold on this page and I accept comments and probably a complete re-work or re-structure (*)

Screenshot_2019-07-18_14-16-44

(*) my other idea was to just add a message like: "Hey there! We are testing our APIv3 and you were given early access to help us with your feedback. Here you have the token to start testing it out: {token}." in a gray small box similar to what we shown under Project Admin Settings.

Example,

Screenshot_2019-07-18_14-27-34

### DIFF
--- a/readthedocs/profiles/urls/private.py
+++ b/readthedocs/profiles/urls/private.py
@@ -25,3 +25,13 @@ urlpatterns = [
         name='account_advertising',
     ),
 ]
+
+tokens_urls = [
+    url(
+        r'^tokens/$',
+        views.TokenList.as_view(),
+        name='profiles_tokens',
+    ),
+]
+
+urlpatterns += tokens_urls
--- a/readthedocs/profiles/urls/private.py
+++ b/readthedocs/profiles/urls/private.py
@@ -7,8 +7,12 @@ from django.conf.urls import url
 from readthedocs.core.forms import UserProfileForm
 from readthedocs.profiles import views
 
+# Split URLs into different lists to be able to selectively import them from a
+# another application (like Read the Docs Corporate), where we may don't need to
+# define Token URLs, for example.
+urlpatterns = []
 
-urlpatterns = [
+account_urls = [
     url(
         r'^edit/',
         views.edit_profile,
@@ -25,6 +29,8 @@ urlpatterns = [
         name='account_advertising',
     ),
 ]
+
+urlpatterns += account_urls
 
 tokens_urls = [
     url(
--- a/readthedocs/profiles/views.py
+++ b/readthedocs/profiles/views.py
@@ -10,6 +10,8 @@ from django.http import HttpResponseRedirect
 from django.shortcuts import get_object_or_404, redirect, render
 from django.urls import reverse
 from django.utils.translation import ugettext_lazy as _
+from django.views.generic import ListView
+from rest_framework.authtoken.models import Token
 
 from readthedocs.core.forms import UserAdvertisingForm, UserDeleteForm
 
@@ -207,3 +209,26 @@ def account_advertising(request):
             'user': profile_obj.user,
         },
     )
+
+
+class TokenMixin:
+
+    """Environment Variables to be added when building the Project."""
+
+    model = Token
+    lookup_url_kwarg = 'token_pk'
+    template_name = 'profiles/private/token_list.html'
+
+    def get_queryset(self):
+        # Token has a OneToOneField relation with User
+        return Token.objects.filter(user__in=[self.request.user])
+
+    def get_success_url(self):
+        return reverse(
+            'projects_token',
+            args=[self.get_project().slug],
+        )
+
+
+class TokenList(TokenMixin, ListView):
+    pass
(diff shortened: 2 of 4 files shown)


readthedocs/profiles/views.py
```diff
@@ -207,3 +209,26 @@ def account_advertising(request):
...
+
+    """Environment Variables to be added when building the Project."""
+
+    model = Token
+    lookup_url_kwarg = 'token_pk'
+    template_name = 'profiles/private/token_list.html'
+
+    def get_queryset(self):
+        # Token has a OneToOneField relation with User
+        return Token.objects.filter(user__in=[self.request.user])
```
stsewd: Why can't this be just `Token.objects.filter(user=self.request.user)`?

readthedocs/profiles/views.py
```diff
@@ -207,3 +209,26 @@ def account_advertising(request):
             'user': profile_obj.user,
         },
     )
+
+
+class TokenMixin:
+
+    """Environment Variables to be added when building the Project."""
```
stsewd: This docstring doesn't look correct

humitos:
> I think tokens should be treated like we do with env variables, only see them in the creation step (github does this).

I agree this is the best pattern, although we don't allow users to create these tokens right now.

> We could allow users token creation under a feature flag instead of us doing it manually.

How we can use feature flags based on users and not on projects?

readthedocs/profiles/urls/private.py
```diff
@@ -25,3 +25,13 @@
         name='account_advertising',
     ),
 ]
+
+tokens_urls = [
```
ericholscher: Why did we add a new list here instead of just putting it in the `urlpatterns`? Could use a comment if there's a reason.

readthedocs/templates/profiles/private/token_list.html
```diff
@@ -0,0 +1,34 @@
...
+
+{% load i18n %}
+
+{% block title %}{% trans "Personal Access Tokens" %}{% endblock %}
+
+{% block profile-admin-tokens %}active{% endblock %}
+
+{% block edit_content_header %} {% trans "Personal Access Tokens" %} {% endblock %}
+
+{% block edit_content %}
```
ericholscher: We definitely need something saying that users can't set them up yet, and to contact us to create one.

readthedocs/templates/profiles/base_profile_edit.html
```diff
@@ -49,6 +49,7 @@ <h2>
           <li class="{% block profile-admin-social-accounts %}{% endblock %}"><a href="{% url 'socialaccount_connections' %}">{% trans "Connected Services" %}</a></li>
           <li class="{% block profile-admin-change-password %}{% endblock %}"><a href="{% url 'account_change_password' %}">{% trans "Change Password" %}</a></li>
           <li class="{% block profile-admin-change-email %}{% endblock %}"><a href="{% url 'account_email' %}">{% trans "Change Email" %}</a></li>
+          <li class="{% block profile-admin-tokens %}{% endblock %}"><a href="{% url 'profiles_tokens' %}">{% trans "Personal Access Tokens" %}</a></li>
```
davidfischer: Is there a reason this is "Personal Access Tokens". I think "API Tokens" seems a bit clearer.

humitos: No. I wasn't sure how to name it and I took the name from GitHub.

readthedocs/profiles/urls/private.py
humitos: I just followed the pattern that we have in other urls.py file.

Although, I think I need to use a different name for the first list and add it to urlpatterns later. This way allow us to decide which URLs to import and define from corporate. Actually, we don't want to publish this new URL in Corporate since there is not going to be API Tokens yet.

readthedocs/templates/profiles/private/token_list.html
humitos: I added a message. Let me know.

Screenshot_2019-07-22_13-55-38

readthedocs/templates/profiles/private/token_list.html
```diff
@@ -0,0 +1,42 @@
...
+  <p class="empty">
+    {% blocktrans trimmed with contact_email="support@readthedocs.org" %}
+      Access Tokens are currently in Beta state and are not possible to create by yourself.
```
ericholscher:
```suggestion
      Access Tokens are currently an invite-only Beta feature. 
```

readthedocs/templates/profiles/private/token_list.html
```diff
@@ -0,0 +1,42 @@
...
+      In case you want to test APIv3 and help giving us feedback,
```
ericholscher:
```suggestion
      In case you want to test APIv3 and give us feedback on it,
```

readthedocs/templates/profiles/private/token_list.html
```diff
@@ -0,0 +1,42 @@
...
+      please <a href="mailto:{{ contact_email }}">contact us</a> and let us know.
```
ericholscher:
```suggestion
      please <a href="mailto:{{ contact_email }}">email us</a>.
```

readthedocs/templates/profiles/base_profile_edit.html
ericholscher: 👍 on API tokens.

humitos:
Merging once the tests pass.

PR #5631, opened 2019-04-25
Use `--satisfied-skip-solve` on `conda install`

humitos:
Skip (do not upgrade) already satisfied dependencies that were
installed by the user when creating the environment, when installing
our own dependencies.

> I'm building the images locally to do some QA before merging it.

Closes #3829
Related https://github.com/rtfd/readthedocs-docker-images/pull/102

Issue #3829, opened 2018-03-21
Ability to pin Sphinx and friends in a conda environment


jakirkham:
## Details

* Project URL: https://readthedocs.org/projects/dask-image/
* Build URL (if applicable): https://readthedocs.org/projects/dask-image/builds/6925299/
* Read the Docs username (if applicable): (same as GitHub)

## Expected Result

We are [currently pinning Sphinx]( https://github.com/dask/dask-image/blob/3d68897828e7b234b80d85548fb82ca8791af098/environment_doc.yml#L9 ) with the expectation that RTD honors this pinning. 

## Actual Result

However it [appears that RTD overrides our Sphinx pinning later]( https://readthedocs.org/projects/dask-image/builds/6925299/ ), which is causing a different problem ( https://github.com/rtfd/readthedocs.org/issues/3769 ).

cc @humitos

humitos:
In my comment https://github.com/rtfd/readthedocs.org/issues/3769#issuecomment-375014874 I mentioned that the problem is the order of the commands executed.

RTD first:

1. create the env using the `yaml` provided by the user
1. install the needed packages to build the documentation (`mock` `pillow` `sphinx` `sphinx_rtd_theme`)

The last step that installs these packages will UPGRADE them if they are already installed in the conda env.

I've been checking the option `--no-update-dependencies` (from `conda install`) but it seems that doesn't work as I want. So, at the moment I don't know how to do something like "ensure installed but not upgrade it".

```
/tmp ⌚ 12:06:29
$ conda install --no-update-dependencies sphinx          

The following NEW packages will be INSTALLED:

    packaging:                17.1-py27_0                     
    pyparsing:                2.2.0-py27hf1513f8_1            

[...]

The following packages will be UPDATED:

    sphinx:                   1.5.1-py27_0         conda-forge --> 1.7.1-py27_0    

The following packages will be SUPERSEDED by a higher-priority channel:

    asn1crypto:               0.24.0-py27_0        conda-forge --> 0.24.0-py27_0   
[...]

Proceed ([y]/n)? n

Exiting
```

(output cropped)

I don't have too much knowledge about `conda`, but it would be great if someone who knows it more help us here saying if there is an option that solves this.

Otherwise, we will need to think in another solution. For example, "remove that command" from the flow, which involves force the user to always add those default packages (sphinx, mock, pillow, rtd_sphinx_theme) to their YAML file, which I think it's not a good "solution".

[...]

PR #5956, opened 2019-07-18
Append core requirements to Conda environment file

humitos:
We run 3 steps when a project depends on conda.

1. create the whole environment based on user's YAML file
2. run `conda install` with our own dependencies
3. run `pip install` with some of our dependencies that are not
published on conda repositories.

This commit changes this to make it in just one step (at environment
creation). To do this, it appends our own requirements to the
`environment.yml` file under `dependencies` and `dependencies.pip`
config (see https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/manage-environments.html#create-env-file-manually)

It also shows the resulting `environment.yml` in the build output.

This behavior is added behind a feature flag so we can test it first.

This allow users to be able to pin their dependencies as they want
without us upgrading them in the second step. Also, this should
improve the build time, since dependencies resolution is done just
once.

Related to

* https://github.com/readthedocs/readthedocs.org/issues/3829
* https://github.com/readthedocs/readthedocs.org/pull/5631

### DIFF
--- a/readthedocs/doc_builder/python_environments.py
+++ b/readthedocs/doc_builder/python_environments.py
@@ -1,14 +1,16 @@
 """An abstraction over virtualenv and Conda environments."""
 
 import copy
+import codecs
 import hashlib
 import itertools
 import json
 import logging
 import os
 import shutil
+import yaml
 
-from readthedocs.config import PIP, SETUPTOOLS
+from readthedocs.config import PIP, SETUPTOOLS, ParseError, parse as parse_yaml
 from readthedocs.config.models import PythonInstall, PythonInstallRequirements
 from readthedocs.doc_builder.config import load_yaml_config
 from readthedocs.doc_builder.constants import DOCKER_IMAGE
@@ -442,6 +444,10 @@ class Conda(PythonEnvironment):
         if self.project.has_feature(Feature.UPDATE_CONDA_STARTUP):
             self._update_conda_startup()
 
+        if self.project.has_feature(Feature.CONDA_APPEND_CORE_REQUIREMENTS):
+            self._append_core_requirements()
+            self._show_environment_yaml()
+
         self.build_env.run(
             'conda',
             'env',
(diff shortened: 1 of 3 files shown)


ericholscher:
I believe we tried to do this in the past and it caused conflicts with version resolution in conda.

readthedocs/doc_builder/python_environments.py
```diff
@@ -472,8 +543,19 @@ def install_core_requirements(self):
             pip_requirements.append('mkdocs')
         else:
             pip_requirements.append('readthedocs-sphinx-ext')
-            requirements.extend(['sphinx', 'sphinx_rtd_theme'])
+            conda_requirements.extend(['sphinx', 'sphinx_rtd_theme'])
```
ericholscher: I believe we explicitly don't add versions here just to confirm that they are installed without specifying a version or having them be upgraded. Has this changed? I don't know how this logic would be upgrading users versions.

humitos:
> I believe we tried to do this in the past and it caused conflicts with version resolution in conda.

Conda has changed a lot on how they solve the version resolution. I think it worth to give a try to this approach.

All the tests that I've done locally worked properly, and at the moment, is the only working solution that we have.

readthedocs/doc_builder/python_environments.py
humitos: Once the environment is created with `conda env create` if the user pin `sphinx==1.6` it will be upgraded (in the current production code) in the step `conda install sphinx` --even if we don't specify any version.

There was an attempt to add an attribute in `conda` itself called `--satisfied-skip-solve` but does not work as we need unfortunately :( . See #5631

ericholscher:
> All the tests that I've done locally worked properly, and at the moment, is the only working solution that we have.

OK. Though I'm pretty sure this will blow up in some random new way. Is there someone in the conda community we can ask if this is the right approach?

humitos:
> Though I'm pretty sure this will blow up in some random new way.

This is probably true. Although, this will give us some context on where it fails and we can research how to fix it from there if we want.

I'm not _too worry_ about this because it's behind a feature flag, though. So, we can enable on request and if it does not work, we just disable it.

> Is there someone in the conda community we can ask if this is the right approach?

I don't personally known anyone. There were already some attempts to add this feature on the conda community, but there is no solution yet. As I mention, `--satisfied-skip-solve` was one attempt, but it only works when _all the dependencies are satisfied_ and it does not work for our use case. See https://github.com/conda/conda/issues/6845#issuecomment-444656635

As far as I know, unfortunately, what we need _it's not currently possible_ on conda.

ericholscher:
If this is a better approach, I'd like to remove the feature flag eventually and just ship it. I don't fully understand it though :/

humitos:
@ericholscher this is a better approach, yes. It solves a problem that we have for what there is not current workaround. `conda` does not provide a "correct way of doing it" as far as I know.

I personally prefer to ship this under a feature flag, though. Why? Because it's easier to start testing this out and be able to go one step back if we found that it does not work for some cases that I may not considered in my tests. Then, after some time we can remove the flag and make it the default behavior, which won't involve a lot of work to remove the flag.

PR #5970, opened 2019-07-20
Fix: `parse_json` also including html in titles

dojutsu-user:
Some titles have html code in them:
- https://docs.readthedocs.io/en/latest/?rtd_search=404%20not%20found
- http://docs.celeryproject.org/en/latest/search.html?q=utilities&check_keywords=yes&area=default

`parse_json` was not yielding the text if title has `backticks` content.


readthedocs/rtd_tests/tests/test_search_json_parsing.py
```diff
@@ -19,6 +19,9 @@ def test_h2_parsing(self):
         )
         self.assertEqual(data['path'], 'api')
         self.assertEqual(data['sections'][1]['id'], 'a-basic-api-client-using-slumber')
+
+        # In api.fjson, title is in the form: A basic API client ``using slumber``
+        self.assertEqual(data['sections'][1]['title'], 'A basic API client using slumber')
```
humitos: In my test the quotes will still be there:

    >>> pyquery.PyQuery('A basic API client ``using slumber``').text().replace('¶', '').strip()
    'A basic API client ``using slumber``'
    >>>

dojutsu-user: When a text like this: `Include ``404`` page` is converted to html by sphinx. It is something like this: `Include <code class=\"docutils literal notranslate\"><span class=\"pre\">404</span></code> page`.

We were indexing this line as it is, because of which -- we are getting results with the html code - http://docs.celeryproject.org/en/latest/search.html?q=utilities&check_keywords=yes&area=default

It is something like this:
```python3
>>> import pyquery
>>> temp = 'Include <code class=\"docutils literal notranslate\"><span class=\"pre\">404</span></code> page'
>>> pyquery.PyQuery(temp).text().replace('¶', '').strip()
'Include 404 page' 
```

humitos: Thanks for clarifying this!

PR #5973, opened 2019-07-21
Community only ads for more themes

davidfischer:
- Request community only ads on themes where we don't support regular ads
- This should not be merged until the relevant server side changes are merge in https://github.com/readthedocs/readthedocs-ext/pull/247

### DIFF
--- a/readthedocs/core/static-src/core/js/doc-embed/rtd-data.js
+++ b/readthedocs/core/static-src/core/js/doc-embed/rtd-data.js
@@ -22,7 +22,7 @@ var configMethods = {
         return constants.ALABASTER_LIKE_THEMES.indexOf(this.get_theme_name()) > -1;
     },
 
-    theme_supports_promo: function () {
+    theme_supports_paid_promo: function () {
         return this.is_rtd_like_theme() || this.is_alabaster_like_theme();
     },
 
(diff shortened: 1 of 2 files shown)


dojutsu-user:
Does this also targets the error here -- http://2.python-requests.org/en/v2.2.1/search/?q=get ?

davidfischer:
> Does this also targets the error here

The problems with search are unrelated to this change.

> I'm not sure to understand what it means that the theme does not support regular Ads but it supports Community Ads.

As part of the rollout of the new fixed footer ads, we [stated](https://blog.readthedocs.com/fixed-footer-ad-all-themes/):

    At first, we will show only community and house ads on these custom themes.

readthedocs/core/static-src/core/js/doc-embed/rtd-data.js
```diff
@@ -40,8 +40,9 @@ var configMethods = {
 
     show_promo: function () {
         return (
-            this.api_host !== 'https://readthedocs.com' &&
-            this.theme_supports_promo());
+            this.api_host === 'https://readthedocs.org' ||
+            this.api_host === 'http://127.0.0.1:8000'
```
ericholscher: Should this also check for `localhost`? Also feels a bit hacky, but I don't know a better way.

davidfischer: The default in `settings/base.py` is `127.0.0.1:8000`. It is a bit hacky but the above change covers the defaults.

PR #5979, opened 2019-07-23
Index more domain data into elasticsearch

dojutsu-user:
PR https://github.com/readthedocs/readthedocs-sphinx-ext/pull/74 will dump a json file -- `readthedocs-sphinx-domain-data.json` with the following content:

```json
{
    "domain-id-1": "domain-docstring-1",
    "domain-id-2": "domain-docstring-2",
    "domain-id-3": "domain-docstring-3"
}
```
We already have values for `domain.anchor` which is same as that of keys in `readthedocs-sphinx-domain-data.json`
During reindexing/indexing of  project, that file will be read and docstrings for a particular domain will available from its `domain.anchor` value.

### DIFF
--- a/readthedocs/search/documents.py
+++ b/readthedocs/search/documents.py
@@ -94,6 +94,7 @@ class PageDocument(RTDDocTypeMixin, DocType):
             # For showing in the search result
             'type_display': fields.TextField(),
             'doc_display': fields.TextField(),
+            'docstrings': fields.TextField(),
 
             # Simple analyzer breaks on `.`,
             # otherwise search results are too strict for this use case
@@ -123,6 +124,7 @@ class PageDocument(RTDDocTypeMixin, DocType):
                 'anchor': domain.anchor,
                 'type_display': domain.type_display,
                 'doc_display': domain.doc_display,
+                'docstrings': html_file.domain_data_json.get(domain.anchor, ''),
                 'name': domain.name,
                 'display_name': domain.display_name if domain.display_name != '-' else '',
             }
--- a/readthedocs/search/parse_json.py
+++ b/readthedocs/search/parse_json.py
@@ -3,6 +3,7 @@
 import codecs
 import json
 import logging
+import os
 
 from pyquery import PyQuery
 
@@ -107,3 +108,13 @@ def parse_content(content):
     # converting newlines to ". "
     content = '. '.join([text.strip().rstrip('.') for text in content])
     return content
+
+
+def process_domain_file(file_path):
+    """Read the ``readthedocs-sphinx-domain-data.json`` file and return its data."""
+    if not os.path.exists(file_path):
+        return {}
+
+    with open(file_path, 'r') as file:
+        data = json.load(file)
+    return data
--- a/readthedocs/projects/models.py
+++ b/readthedocs/projects/models.py
@@ -38,7 +38,7 @@ from readthedocs.projects.validators import (
     validate_repository_url,
 )
 from readthedocs.projects.version_handling import determine_stable_version
-from readthedocs.search.parse_json import process_file
+from readthedocs.search.parse_json import process_file, process_domain_file
 from readthedocs.vcs_support.backends import backend_cls
 from readthedocs.vcs_support.utils import Lock, NonBlockingLock
 
@@ -1276,6 +1276,33 @@ class HTMLFile(ImportedFile):
     @cached_property
     def processed_json(self):
         return self.get_processed_json()
+
+    def get_domain_data(self):
+        """Get sphinx domain data from ``readthedocs-sphinx-domain-data.json`` file for indexing."""
+        domain_data_dir = self.project.get_production_media_path(
+            type_='json',
+            version_slug=self.version.slug,
+            include_file=False
+        )
+
+        domain_data_json = os.path.join(
+            domain_data_dir,
+            'readthedocs-sphinx-domain-data.json'
+        )
+
+        domain_data = {}
+        try:
+            domain_data = process_domain_file(domain_data_json)
+        except Exception as e:
+            log.warning(
+                'Unhandled exception during search process domain data file: %s',
+                domain_data_json
+            )
+        return domain_data
+
+    @cached_property
+    def domain_data_json(self):
+        return self.get_domain_data()
 
 
 class Notification(models.Model):
--- a/readthedocs/search/faceted_search.py
+++ b/readthedocs/search/faceted_search.py
@@ -107,6 +107,7 @@ class PageSearchBase(RTDFacetedSearch):
         'domains.type_display',
         'domains.name^2',
         'domains.display_name',
+        'domains.docstrings',
     ]
     fields = _outer_fields
 
@@ -161,6 +162,7 @@ class PageSearchBase(RTDFacetedSearch):
                         'domains.type_display': {},
                         'domains.name': {},
                         'domains.display_name': {},
+                        'domains.docstrings': {},
                     }
                 }
             }


readthedocs/search/parse_json.py
```diff
@@ -107,3 +110,40 @@ def parse_content(content):
...
+    for dl_tag in dl_tags:
+
+        dt = dl_tag.findall('dt')  
+        dd = dl_tag.findall('dd')
+
+        # len(dt) should be equal to len(dd)
+        # because these tags go together.
+        for title, desc in zip(dt, dd):
+            id_ = title.attrib.get('id')
+            if id_:
```
ericholscher: We should log if this fails for some reason, as we'd be losing data.

readthedocs/search/parse_json.py
```diff
@@ -107,3 +110,40 @@ def parse_content(content):
     # converting newlines to ". "
     content = '. '.join([text.strip().rstrip('.') for text in content])
     return content
+
+
+def generate_domains_data_from_pyquery(body):
+
```
ericholscher: Need a docstring explaining the data that comes back from this.

readthedocs/search/parse_json.py
```diff
@@ -107,3 +110,40 @@ def parse_content(content):
...
+                # remove the 'dl', 'dt' and 'dd' tags from it
+                # because all the 'dd' and 'dt' tags are inside 'dl'
+                # and all 'dl' tags are already captured.
+                desc_contents.remove('dl')
+                desc_contents.remove('dt')
+                desc_contents.remove('dd')
+
+                docstrings = desc_contents.text().replace('¶', '').strip()
+                docstrings = '. '.join(
```
ericholscher: This could likely use a comment, what's the purpose of this logic?

dojutsu-user:
@ericholscher 
I have updated the PR.
Also, this PR also index sphinx domain signature content (Inside the `<dd>` tag).
I am not sure if it is needed or not.

I found some domains which don't have proper html strucutre.
Example: https://2.python-requests.org/en/master/api/#requests.cookies.RequestsCookieJar.popitem
In this.
Content inside `<dt>` tag: `popitem() → (k, v), remove and return some (key, value) pair`
Content inside `<dd>` tag: `as a 2-tuple; but raise KeyError if D is empty.`

So there's no proper domain signature and no proper domain doc strings.
What should we do in this case?

dojutsu-user:
@ericholscher 
> I'd like to see some tests

I will add the tests, once we are sure of the working.

readthedocs/search/parse_json.py
```diff
@@ -101,9 +104,71 @@ def parse_content(content):
 
     # removing the starting text of each
     content = content.split('\n')
-    if len(content) > 1:  # there were \n
+    if remove_first_line and len(content) > 1:
```
ericholscher: Was this the issue we were having with not getting the full docstring?

readthedocs/search/parse_json.py
```diff
@@ -101,9 +104,71 @@ def parse_content(content):
...
+                "signature": "domain_signature_2(*args, **kwargs),
+                "docstrings": "docstrings for the domain-id-2",
+            }
+        }
+    """
+
+    domain_data = {}
+    try:
+        dl_tags = body('dl')
+        for dl_tag in dl_tags:
```
ericholscher: I do wonder if we should put the `try` a bit lower here, so that we can continue parsing if we hit an issue. Here if we hit an exception in one tag, it will stop all others from processing.

readthedocs/search/parse_json.py
dojutsu-user: Nope.
This is done because section content contains the section title as its first line.
Section title is already parsed and we don't need same thing twice.

dojutsu-user:
@ericholscher 
Turns out that not every sphinx domain is inside `<dl>` tag.
For example: http://docs.celeryproject.org/en/latest/userguide/configuration.html#beat-scheduler
Therefore, the docstrings for these sphinx domains are not getting indexed. However these docstrings will be there in `sections.content`.

ericholscher:
Those aren't in Sphinx Domain's: https://raw.githubusercontent.com/celery/celery/241d2e8ca85a87a2a6d01380d56eb230310868e3/docs/userguide/configuration.rst

dojutsu-user:
@ericholscher 
They are getting indexed as sphinx domains.

Screenshot from 2019-07-30 02-26-41

Screenshot from 2019-07-30 02-29-32

readthedocs/core/static-src/core/js/doc-embed/search.js
```diff
@@ -49,10 +50,9 @@ function attach_elastic_search_query(data) {
                             }
                         }
 
-                        // Creating the result from elements
-                        var link = doc.link + DOCUMENTATION_OPTIONS.FILE_SUFFIX + "?highlight=" + $.urlencode(query);
```
ericholscher: I do worry about not using the local `DOCUMENTATION_OPTIONS.FILE_SUFFIX` -- are we sure that our API is returning the proper file path?

readthedocs/search/documents.py
```diff
@@ -70,7 +70,6 @@ class PageDocument(RTDDocTypeMixin, DocType):
     # Metadata
     project = fields.KeywordField(attr='project.slug')
     version = fields.KeywordField(attr='version.slug')
-    path = fields.KeywordField(attr='processed_json.path')
```
ericholscher: Why are we removing this?

readthedocs/search/documents.py
```diff
@@ -88,17 +87,14 @@ class PageDocument(RTDDocTypeMixin, DocType):
             'role_name': fields.KeywordField(),
 
             # For linking to the URL
-            'doc_name': fields.KeywordField(),
             'anchor': fields.KeywordField(),
 
             # For showing in the search result
-            'type_display': fields.TextField(),
-            'doc_display': fields.TextField(),
```
ericholscher: Why are we removing this?

readthedocs/templates/search/elastic_search.html
```diff
@@ -179,61 +179,64 @@ <h3>
...
-                    </a>
-
-                    {% for inner_hit in result.meta.inner_hits %}
-                      {% if inner_hit.type == 'domains' %}
-                        <p>
-                          <a href="{% doc_url result.project|get_project result.version inner_hit.source.doc_name %}?highlight= {{ query }}#{{ inner_hit.source.anchor }}">
-
-                            {% if inner_hit.source.display_name|length >= 1 %}
-                              ({{ inner_hit.source.role_name }}) {{ inner_hit.source.display_name}}
+                    {% with "100" as MAX_SUBSTRING_LIMIT %}
```
ericholscher: I'm feeling the pain of maintaining multiple sets of result HTML. We should consider trying to get the site search using the search as you type code, hopefully with minimal changes from the theme code. That way we don't have 3 places to change HTML. 

The goal will be to ship search as you type across all of RTD so we only have to maintain one set of results, but baby steps :)

readthedocs/core/static-src/core/js/doc-embed/search.js
dojutsu-user: I have undo this change. Earlier I got confused because of the `Http404: Directory indexes are not allowed here.` error because I am not serving the static files in local with nginx.

readthedocs/search/documents.py
dojutsu-user: We don't need `doc_display` -- because the main title of each result contains the `doc_display`.
But I have undoed the change for `type_display` -- we can make use of this.

dojutsu-user: :confused: 
I can't seem to find a good way to display every information.
`type_display` looks unncessary when we are displaying `role_name`.

readthedocs/templates/search/elastic_search.html
dojutsu-user: Yes... maintaining multiple sets of html is definitely a pain.

> The goal will be to ship search as you type across all of RTD so we only have to maintain one set of results, but baby steps :)

I think that's quite a big refactor and we can target that after this PR?

readthedocs/search/documents.py
dojutsu-user: Got confused the nginx static files service...
and changed everything with the `full_path`.
Undoed this change.

readthedocs/search/documents.py
ericholscher: Well, sure. They are showing the same information. I believe the `type_display` is meant for human readable output, and `role_name` is not, no?

dojutsu-user: Both are quite readable:
py:function
std:confval
py:class

ericholscher:
This looks like a great approach. Let's see if we can get it deployed this week after we fix up tests. :+1:

readthedocs/search/documents.py
ericholscher: I think `configuration value` over `confval` is much better for non-technical users. We should definitely try to use the display value, that's the whole reason it exists if for this use.

dojutsu-user: @ericholscher 
I realised (while testing) that some sphinx domains don't have values for type_display.
Should we still show for those who have? or just show role_name?

PR #5984, opened 2019-07-24
Use try...catch block with underscore.js template.

### LINKED ISSUE #5978: Search not working
## Details

* Read the Docs project URL: https://toolz.readthedocs.io/

## Expected Result
Search should show a list of matching doc chunks

## Actual Result
A search page shows, but no search result is displayed

dojutsu-user:
#5980 doesn't fixes #5978 
Turns out that the problem was not of variable declaration.
But the error was due to syntax changes in underscore.js templating engine.

See: https://stackoverflow.com/a/25881231/8601393


readthedocs/core/static-src/core/js/doc-embed/search.js
```diff
@@ -7,6 +7,33 @@ var xss = require('xss/lib/index');
...
+ * @param {Object} contents html element on which additional html is be appended
+ * @param {String} template underscore.js template string
+ * @param {Object} data template vars and their values
+ */
+function append_html_to_contents(contents, template, data) {
+    // underscore.js throws variable not defined error
+    // because of change of syntax in new versions.
+    // See: https://stackoverflow.com/a/25881231/8601393
+    try {
+        // this is the old syntax
```
davidfischer: Let's be a little more specific in the comment here. Sphinx bundles Underscore.js v1.3.1 (and has for many years) and it looks like this syntax changed in Underscore v1.7.0 (not exactly how semver is supposed to work...). So Sphinx should always use the old syntax and almost never the new syntax. How was this throwing a syntax error exactly? Were people somehow bundling newer versions of Underscore?

davidfischer: Ok, I figured it out. Sphinx has bundled Underscore.js v1.3.1 as I mentioned. However, doc builds before ~2019 used centrally located versions of jQuery and Underscore.js (due to https://github.com/readthedocs/readthedocs-sphinx-ext/pull/52) as an optimization. This was removed precisely because it caused version mismatch issues like this!

davidfischer:
```suggestion
        // this is the pre-1.7 syntax from Underscore.js
```

agjohnson:
Is it okay that we're assuming that underscore is loaded by the page? Is underscore a new implicit dependency with this change?

The only case I can think of where a user does not have `underscore.js` is if they have a heavily customized theme -- which is more of an issue on our commercial hosting.

dojutsu-user:
@agjohnson 
I think that it is okay to assume this.
I am not sure of this, but I believe sphinx adds underscore.js in every docs.
I have one example -- https://docs.signalfx.com/en/latest/
This is a heavily customised theme and it does includes underscore.js.

agjohnson:
My point was I thought we needed to verify this, we've postulated about it a good deal. I inspected our commercial hosting, where themes are more heavily customized, and the only projects that did not have `underscore.js` were mkdocs projects. Sphinx adds these files if they are referenced, but a heavily customized theme might not trigger this reference. An implicit dependency seems okay here. :+1:

dojutsu-user:
@ericholscher 
This is mergeable now.

davidfischer:
> The only case I can think of where a user does not have underscore.js is if they have a heavily customized theme -- which is more of an issue on our commercial hosting.

While it is possible, Sphinx search doesn't work without Underscore.js. So if they've customized so heavily that they've removed Underscore, they've likely also removed search.

PR #5985, opened 2019-07-24
Fix github build status reporting bug

saadmk11:
This will fix the bug that was sending status reports for wrong commits.This happens when multiple commits are added to a pull request in a short amount of time and versions get updated with new identifiers every time so, that's why when sending the finish status report (`success/failure`) it might send it for a different commit sha and give some inconsistent result.


saadmk11:
@ericholscher Build object does not have the commit sha when the build is triggered. Its added while building. so to send the triggered/pending status we need to send the version identifier

ericholscher:
We can change the code to send the commit when triggering the build though, right? We have the data at least for external versions.

saadmk11:
Yes we can do that. do you think that is a better approach then this?

ericholscher:
Yes. This can still hit bugs when a commit will switch between triggering and building. We should be explicitly passing a commit along if we have it.

readthedocs/core/utils/__init__.py
```diff
@@ -133,7 +133,10 @@ def prepare_build(
 
     if build:
         # Send pending Build Status using Git Status API for External Builds.
-        send_external_build_status(version=version, build_pk=build.id, status=BUILD_STATUS_PENDING)
+        send_external_build_status(
+            version_type=version.type, build_pk=build.id,
+            commit=version.identifier, status=BUILD_STATUS_PENDING
```
ericholscher: I feel like this isn't quite right. We're still just getting the commit _from_ the VCS that is getting checked out. This is still coming from the `Version` object. We also don't have the VCS-based commit when we trigger the build.

We need to pass this `commit` along to the actual build, so that we're triggering the specific build with this commit, not the random commit that's stored in the build object.

readthedocs/core/utils/__init__.py
```diff
@@ -102,6 +104,7 @@ def prepare_build(
     kwargs = {
         'record': record,
         'force': force,
+        'commit': commit,
```
ericholscher: I think we should also be passing `commit` into the Build object here when created, if we have it.

saadmk11: Done!

readthedocs/projects/tasks.py
```diff
@@ -130,27 +130,23 @@ def sync_repo(self):
...
+                'project': self.project.slug,
+                'version': self.version.slug,
+                'msg': msg,
+            }
+        )
+        version_repo = self.get_vcs_repo()
+        version_repo.update()
+        self.sync_versions(version_repo)
+        identifier = self.commit or self.version.identifier
+        version_repo.checkout(self.version.identifier)
```
saadmk11: I think we should pass this here. `identifier` from the line above. might have changes while resolving conflict :)

ericholscher: Ah yea, that was a failure in the conflict resolution on my side.

PR #5987, opened 2019-07-24
Update GitHub build status details URL

saadmk11:


readthedocs/oauth/services/github.py
```diff
@@ -335,9 +336,14 @@ def send_build_status(self, build, state):
         github_build_state = SELECT_BUILD_STATUS[state]['github']
         description = SELECT_BUILD_STATUS[state]['description']
 
+        target_url = build.get_full_url()
+
+        if state == BUILD_STATUS_SUCCESS:
+            target_url = build.version.get_absolute_url()
```
ericholscher: `get_absolute_url` also needs to be changed here to point to the new place we're storing external builds, instead of our existing docs URL.

saadmk11: This PR can be merged as i'm working on serving the PR docs here: https://github.com/readthedocs/readthedocs.org/pull/6002

PR #6002, opened 2019-07-26
serve external version files through media

saadmk11:
I don't feel like this is a good way at all. I think we should not update `get_absolute_url` and serve from media as it looks like a hack not a proper solution. also it does not work when we go to the mail docs page it needs filenames to be added. also if we do this we wont be using the `serve_docs()` view.
Maybe we can create a different view or something to handle this but not sure. trying out different approaches :)

### DIFF
--- a/readthedocs/core/resolver.py
+++ b/readthedocs/core/resolver.py
@@ -75,12 +75,12 @@ class ResolverBase:
         else:
             url = '/docs/{project_slug}/'
 
+        if subproject_slug:
+            url += 'projects/{subproject_slug}/'
+
         if version_type == EXTERNAL:
             # we serve external version media files from media storage
             url = '/media/external/html/{project_slug}/'
-
-        if subproject_slug:
-            url += 'projects/{subproject_slug}/'
 
         if single_version:
             url += '{filename}'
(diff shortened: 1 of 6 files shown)


saadmk11:
I will update a different approach soon. Which will try to solve the issues mentioned above :)

readthedocs/core/urls/__init__.py
```diff
@@ -38,6 +38,17 @@
         serve.serve_docs,
         name='docs_detail',
     ),
+    # Serving external version docs
```
ericholscher: We don't want to hook this up to `serve_docs`, we can serve it already directly with the media serving code.

readthedocs/core/views/serve.py
```diff
@@ -264,17 +266,28 @@ def _serve_symlink_docs(request, project, privacy_level, filename=''):
     files_tried = []
 
     if (settings.DEBUG or constants.PUBLIC in settings.SERVE_DOCS) and privacy_level != constants.PRIVATE:  # yapf: disable  # noqa
-        public_symlink = PublicSymlink(project)
-        basepath = public_symlink.project_root
+        if version_type == EXTERNAL:
```
ericholscher: We shouldn't change any of this code.

readthedocs/core/resolver.py
```diff
@@ -63,23 +64,28 @@ def base_resolve_path(
...
+            if subproject_slug:
+                url += 'projects/{subproject_slug}/'
 
-        if single_version:
-            url += '{filename}'
+            if single_version:
+                url += '{filename}'
+            else:
+                url += '{language}/{version_slug}/{filename}'
         else:
```
ericholscher: We should keep the `if` statement smaller, not have all the existing logic inside an if. Flip the logic around, and it makes sense.

readthedocs/core/resolver.py
```diff
@@ -63,23 +64,28 @@ def base_resolve_path(
...
-            url += '{language}/{version_slug}/{filename}'
+            # we serve external version media files from media storage
+            url = '/media/external/html/{project_slug}/{version_slug}/{filename}'
```
ericholscher: This is the part we need to logic around. We _still_ don't have a setting for the domain hosting the external builds. That's the entire point of this PR, to link properly to `external-builds.readthedocs.io` in prod and something similar in dev.

readthedocs/core/urls/__init__.py
saadmk11: @ericholscher Yes we can serve it directly through the django media serving but then we won't be able to do this
`https://docs.readthedocs.io/32/` we have to have a file name and extension `https://docs.readthedocs.io/32/index.html`
thats why I did it like this.

readthedocs/core/resolver.py
saadmk11: are you planing to add the domain:host here?
something like this:
`settings.EXTERNAL_DOC_URL = 'external-builds.readthedocs.io'`
`{external_doc_url}/media/external/html/{project_slug}/{version_slug}/{filename}`

ericholscher:
Another approach here: https://github.com/readthedocs/readthedocs.org/pull/6020

ericholscher:
This might be useful if we want to add the PR builder to the .com, but for now we can close this.

PR #6004, opened 2019-07-27
Link to Read the Docs for Business docs from relevant sections

davidfischer:
- This PR exclusively updates docs
- Link to the RTD for Business docs from relevant pages (eg. "Need commercial support? See RTD.com")
- I did only a very minor update to the privacy levels docs which probably need a more major overhaul to explain what RTD for Business offers there relative to the community site. This is on the roadmap already, I believe.

### DIFF
--- a/docs/builds.rst
+++ b/docs/builds.rst
@@ -14,6 +14,9 @@ Our current build limits are:
 
 We can increase build limits on a per-project basis,
 sending an email to support@readthedocs.org providing a good reason why your documentation needs more resources.
+If your business is hitting build limits hosting documentation on Read the Docs,
+please consider :doc:`Read the Docs for Business </commercial/index>`
+which has much higher build resources.
 
 You can see the current Docker build images that we use in our `docker repository <https://github.com/readthedocs/readthedocs-docker-images>`_.
 `Docker Hub <https://hub.docker.com/r/readthedocs/build/>`_ also shows the latest set of images that have been built.
--- a/docs/commercial/custom_domains.rst
+++ b/docs/commercial/custom_domains.rst
@@ -1,12 +1,16 @@
 Custom Domains
 ==============
 
+.. note:: These directions are for projects hosted on Read the Docs for Business.
+          For setting up a custom domain on a project hosted on readthedocs.org,
+          please read our :doc:`community documentation </custom_domains>`.
+
 Subdomain support
 -----------------
 
-Once a project is imported under Read the Docs,
+Once a project is imported into Read the Docs,
 by default it's hosted under a subdomain on one of our domains.
-If you need a custom domain, continue on custom domain setup.
+If you need a custom domain, continue on to custom domain setup.
 
 
 Serving documentation with a custom domain
--- a/docs/custom_domains.rst
+++ b/docs/custom_domains.rst
@@ -1,11 +1,9 @@
 Custom Domains
 ==============
 
-.. note:: These setup directions are for our community site.
-          If you want to setup a custom domain under our `commercial hosting`_,
-          please read our :ref:`commercial documentation <commercial/custom_domains:Custom Domains>`.
-
-.. _commercial hosting: https://readthedocs.com
+.. note:: These directions are for projects hosted on our community site.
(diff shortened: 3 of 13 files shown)


docs/builds.rst
```diff
@@ -14,6 +14,9 @@ Our current build limits are:
 
 We can increase build limits on a per-project basis,
 sending an email to support@readthedocs.org providing a good reason why your documentation needs more resources.
+If your business is hitting build limits hosting documentation on Read the Docs,
```
humitos: This is tricky to me. I mean, "if you have a _business_, you should not be using .org" I understand.

Although, maybe _business_ has a more expand meaning and I'm not getting it. I immediately related it to _company_, so.

docs/custom_domains.rst
```diff
@@ -1,11 +1,9 @@
 Custom Domains
 ==============
 
-.. note:: These setup directions are for our community site.
-          If you want to setup a custom domain under our `commercial hosting`_,
-          please read our :ref:`commercial documentation <commercial/custom_domains:Custom Domains>`.
-
-.. _commercial hosting: https://readthedocs.com
+.. note:: These directions are for projects hosted on our community site.
+          If you want to setup a custom domain on `Read the Docs for Business <https://readthedocs.com/>`_,
```
humitos: Sometimes we link "Read the Docs for Business" to readthedocs.com and some others to `<commercial/index>`. Is that on purpose?

docs/custom_installs/index.rst
```diff
@@ -1,9 +1,18 @@
...
 
-.. warning:: Read the Docs developers do not support custom installs of our software. These documents are maintained by the community, and might not be up to date.
+If your main reasons for considering a custom installation
+are to connect to private repositories or to have fine grained access control over your documentation,
+please consider :doc:`Read the Docs for Business </commercial/index>`.
+It has those features and more!
+
+.. warning::
+    Read the Docs developers do not support custom installs of our software.
+    These documents are maintained by the community, and might not be up to date.
```
agjohnson: We probably already link to our policy on open source and support of custom installs, but this could be a good place to re-iterate.

docs/commercial/custom_domains.rst
```diff
@@ -1,12 +1,16 @@
 Custom Domains
 ==============
 
+.. note:: These directions are for projects hosted on Read the Docs for Business.
+          For setting up a custom domain on a project hosted on readthedocs.org,
+          please read our :doc:`community documentation </custom_domains>`.
```
agjohnson: Instead of a separate page for these docs, the feature we discussed adding with this PR is to use `sphinx_tabs` extension to make it really easy to click between .org and .com directions for areas where docs deviated. This seems like a more digestible pattern for the reader. Do we want to address that now?

docs/custom_domains.rst
davidfischer: In general, I tried to link directly to readthedocs.com from subpages of `commercial` and linked to `commercial/index` on other pages. This page was the only special one where I linked to both (actually it already did that).

docs/commercial/custom_domains.rst
davidfischer: I think that's a separate issue. This PR is purely marketing as opposed to actually editing much in the way of documentation and merging two docs into one.

davidfischer:
> One addition that might make sense to play with here would be to use the sphinx_tabs extension. I think we talked earlier about using sphinx_tabs extension to house .org/.com competing content on the same doc page.

I'm on board with this, but this PR is purely marketing changes. I think that's a separate PR.

docs/custom_installs/index.rst
davidfischer: You're talking about this, correct? https://readthedocs.com/services/#open-source-support

davidfischer: I added a link to that.

docs/builds.rst
davidfischer: Well, I think the bigger issue here is businesses asking for free support (additional resources, specific issues in the issue tracker, etc.) for running their critical business processes. If they want that support, there's a way to get it: paid plans. For a business that is just using Read the Docs and doesn't really need anything, have at it!

PR #6007, opened 2019-07-29
Add troubleshooting instructions

tapaswenipathak:
Fix https://github.com/readthedocs/readthedocs.org/issues/5888.

### DIFF
--- a/docs/development/install.rst
+++ b/docs/development/install.rst
@@ -206,16 +206,20 @@ Please follow the guidelines for installation given here:
   `http://localhost:<number>/admin`
 
 - EDIT the SITES section:
-  In the sites section of the Django Admin UI change the example.com domain as
+  In the sites section of the Django Admin panel change the example.com domain as
   the localhost admint/sites domain with: `http://localhost:<number>/admin/sites`
 
 **Note:** The ID of the site should be of the `SITE_ID` key in `readthedocs/settings/base.py`.
 
 - Connect GitHub account
 
-- Add a social account in the readthedocs installation via Django admin UI
+- You should create a test [GitHub OAuth Apps](https://github.com/settings/applications/new).
+
+- Add a social account in the readthedocs installation via Django admin panel
   - visit `http://localhost/admin/socialaccount/ and tab on a social account`.
   - Write in the tab saying `GitHub`
   - Add `availble site` and `chosen site`.
 
-- Verify by pressing connect GitHub account in readthedocs UI.
+- You might require using services like `ngrok` for setting GitHub connection.
+
+- Verify by pressing connect GitHub account in readthedocs panel.
--- a/docs/development/install.rst
+++ b/docs/development/install.rst
@@ -194,11 +194,6 @@ Error: `SocialApp matching query does not exist.`
 
 Please follow the guidelines for installation given here:
 
-- Edit `readthedocs/settings/base.py`
-  Add entry to the authentication backend section:
-  `'allauth.account.auth_backends.AuthenticationBackend'`
-  ``AUTHENTICATION_BACKENDS = ( # Needed to login by username in Django admin, regardless ofallauth'django.contrib.auth.backends.ModelBackend', #allauth specific authentication methods, such as login by e-mail 'allauth.account.auth_backends.AuthenticationBackend', 'guardian.backends.ObjectPermissionBackend' )``
-
 - Start the installation
   Run the installation: `sudo python3 manage.py runserver 0.0.0.0:<number>`
 
--- a/docs/development/install.rst
+++ b/docs/development/install.rst
@@ -184,3 +184,38 @@ For building this documentation,
 see :doc:`docs`.
 
 And for setting up for the front end development, see :doc:`standards`.
+
+Troubleshooting
+---------------
+
+You might face issues while installing.
+
+Error: `SocialApp matching query does not exist.`
+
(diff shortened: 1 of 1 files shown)


docs/development/install.rst
```diff
@@ -184,3 +184,38 @@ For building this documentation,
...
+
+You might face issues while installing.
+
+Error: `SocialApp matching query does not exist.`
+
+Please follow the guidelines for installation given here:
+
+- Edit `readthedocs/settings/base.py`
+  Add entry to the authentication backend section:
+  `'allauth.account.auth_backends.AuthenticationBackend'`
```
saadmk11: I think this is already added to the base settings

docs/development/install.rst
```diff
@@ -184,3 +184,38 @@ For building this documentation,
...
+
+- EDIT the SITES section:
+  In the sites section of the Django Admin UI change the example.com domain as
+  the localhost admint/sites domain with: `http://localhost:<number>/admin/sites`
+
+**Note:** The ID of the site should be of the `SITE_ID` key in `readthedocs/settings/base.py`.
+
+- Connect GitHub account
+
+- Add a social account in the readthedocs installation via Django admin UI
```
saadmk11: I think we should tell the users about creating a test GitHub app and also adding  the GitHub App to `Social Applications` through  the RTD admin panel. Also they need to use services like `ngrok` to connect to GItHub or other providers.

docs/development/install.rst
tapaswenipathak: Are you sure?

saadmk11: Yes :)
https://github.com/readthedocs/readthedocs.org/blob/4b38e8228f2df43cd81823139f525d999f6e4467/readthedocs/settings/base.py#L176-L181

tapaswenipathak: Cool.

tapaswenipathak:
@saadmk11 do you suggest any more enhancements?  or mind merging the pr?

docs/development/install.rst
```diff
@@ -184,3 +184,37 @@ For building this documentation,
 see :doc:`docs`.
 
 And for setting up for the front end development, see :doc:`standards`.
+
+Troubleshooting
+---------------
+
+You might face issues while installing.
```
saadmk11: Can we be more explicit here and say where and when exactly this issue might occur?

docs/development/install.rst
```diff
@@ -184,3 +184,37 @@ For building this documentation,
...
+
+Please follow the guidelines for installation given here:
+
+- Start the installation
+  Run the installation: `sudo python3 manage.py runserver 0.0.0.0:<number>`
+
+- Visit the readthedocs DJANGO Admin site:
+  `http://localhost:<number>/admin`
+
+- EDIT the SITES section:
```
saadmk11: Not sure we actually need to do anything with the `sites`

docs/development/install.rst
```diff
@@ -184,3 +184,37 @@ For building this documentation,
...
+  the localhost admint/sites domain with: `http://localhost:<number>/admin/sites`
+
+**Note:** The ID of the site should be of the `SITE_ID` key in `readthedocs/settings/base.py`.
+
+- Connect GitHub account
+
+- You should create a test [GitHub OAuth Apps](https://github.com/settings/applications/new).
+
+- Add a social account in the readthedocs installation via Django admin panel
+  - visit `http://localhost/admin/socialaccount/ and tab on a social account`.
```
saadmk11: we can point to this url `http://localhost:8000/admin/socialaccount/socialapp/add/`

docs/development/install.rst
```diff
@@ -184,3 +184,37 @@ For building this documentation,
...
+  - Write in the tab saying `GitHub`
+  - Add `availble site` and `chosen site`.
```
saadmk11: We may want to tell them to collect the Client id and Secret key from the OAuth App and add it here.

tapaswenipathak:
@saadmk11 could you take the effort of doing the nit behaviour fixes from your side? I super appreciate your thoughts and characteristically described enhancements, knowledge exposure. I don't think I will adding anything more on the PR. Feel free closing out, if isn't useful or requires lot of effort tackling.

docs/development/install.rst
tapaswenipathak: Could you read the line described as `Error` below? In a Troubleshooting section in a document, you should be less explicit in first one line. Though you should describe `error`.

docs/development/install.rst
tapaswenipathak: Could you run the setup and see if you get any errors?

tapaswenipathak:
For more -

Could you let me know if screenshots would be useful -
Screen Shot 2019-08-09 at 2 54 33 PM

docs/development/install.rst
tapaswenipathak: Woah, now that's a find.

docs/development/install.rst
saadmk11: `Error: `SocialApp matching query does not exist.` this error does not occur while installing. So, I was saying It we can tell users that it may occur when they try to login through `GitHub/GitLab/BitBucket`

tapaswenipathak:
As I haven't received response for all my comments, closing the PR. Please feel free finishing keeping the PR as base PR.

PR #6009, opened 2019-07-29
Add reCaptcha on import page

dojutsu-user:
I have used [django-recaptcha](https://github.com/praekelt/django-recaptcha).
**This PR does not includes settings for production.**
To enable this in production -- these instructions are to be followed: https://github.com/praekelt/django-recaptcha#installation

These settings should do the work in Google Admin Console.

Screenshot from 2019-07-29 22-19-08

### DIFF
--- a/readthedocs/projects/forms.py
+++ b/readthedocs/projects/forms.py
@@ -5,6 +5,7 @@ from urllib.parse import urlparse
 
 from crispy_forms.helper import FormHelper
 from crispy_forms.layout import Fieldset, Layout, HTML, Submit
+from captcha.fields import ReCaptchaField
 from django import forms
 from django.conf import settings
 from django.contrib.auth.models import User
@@ -82,6 +83,8 @@ class ProjectBackendForm(forms.Form):
 class ProjectBasicsForm(ProjectForm):
 
     """Form for basic project fields."""
+
+    recaptcha = ReCaptchaField()
 
     class Meta:
         model = Project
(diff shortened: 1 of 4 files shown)


agjohnson:
I'm generally -1 on captcha use and would rather this PR starts off in a standard design decision. When I've asked friends in the accessibility space about how to use captchas and maintain site accessibility, their response has been "don't use captchas."

https://www.w3.org/TR/turingtest/#the-google-recaptcha for more. It's not clear what is current in their remarks, but recaptcha service at the time did not seem to consistently support audio or keyboard fallback assistive technologies.

Finally, captcha support is not a panacea. If the spammers introducing spam projects are not part of denied or known spam operation network block, or they aren't fully automating the form itself, the captcha might not do anything besides present the spammer with a check box.

I still say that heuristics and statistical analysis on import, hidden form fields, and more consistently flagging spam projects are all better options before we introduce possible accessibility issues and risks for Google data mining users.

dojutsu-user:
@agjohnson 
While I completely agree with you, I do want to have something to automate the process of preventing spam projects from being created.
I have this vague idea involving machine learning stuff.
- We have many projects that are already identified as SPAM.
- We take those projects, train a machine learning model with them.
- And use that ML model to identify future spam projects.

I am not into ML/AI field and so I'm not sure how to implement this or will this a good approach or is this even possible.

agjohnson:
We already have form fields in place on the import form for this, but need to return to making classification work better for the form. We have this on our roadmap for later this year.

dojutsu-user:
@agjohnson 
Ohh..

> We have this on our roadmap for later this year.

Didn't knew that.
I'm closing this PR.

PR #6012, opened 2019-07-29
Handle .x in version sorting

davidfischer:
* Handle `.x` in version sorting (eg. `1.0.x` or `2.x`). A few projects use this in their branch names like Pallets and Django.
* Added tests for version sorting

Fixes: https://github.com/readthedocs/readthedocs.org/issues/6010

### DIFF
--- a/readthedocs/rtd_tests/tests/projects/test_version_sorting.py
+++ b/readthedocs/rtd_tests/tests/projects/test_version_sorting.py
@@ -1,0 +1,67 @@
+from django.test import TestCase
+from django_dynamic_fixture import get
+
+from readthedocs.builds.constants import BRANCH
+from readthedocs.builds.models import Version
+from readthedocs.projects.models import Project
+from readthedocs.projects.templatetags.projects_tags import sort_version_aware
+
+
+class SortVersionsTest(TestCase):
+
+    def setUp(self):
+        self.project = get(Project)
+
+    def test_basic_sort(self):
+        identifiers = ['1.0', '2.0', '1.1', '1.9', '1.10']
+        for identifier in identifiers:
+            get(
+                Version,
+                project=self.project,
+                type=BRANCH,
+                identifier=identifier,
+                verbose_name=identifier,
+                slug=identifier,
+            )
+
+        versions = list(Version.objects.filter(project=self.project))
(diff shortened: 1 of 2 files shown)


readthedocs/rtd_tests/tests/projects/test_version_sorting.py
```diff
@@ -0,0 +1,67 @@
...
+                project=self.project,
+                type=BRANCH,
+                identifier=identifier,
+                verbose_name=identifier,
+                slug=identifier,
+            )
+
+        versions = list(Version.objects.filter(project=self.project))
+        self.assertEqual(
+            ['latest', 'carrot', 'banana', 'apple'],
```
humitos: I know that this is not a behavior that you changed, but I'd like to sort alphanumeric versions from A to Z instead at some point. Feels more human-readable when you have a many of these.

davidfischer: Do we have any examples of projects that use something like this? I can think of Ubuntu off the top of my head, but that's about it.

davidfischer:
A side comment that isn't exactly related to this PR is that the `sort_versions()` function ignores "uncomparable" versions like "latest" and "stable". This is by design it seems but perhaps a comment is warranted. This wasn't immediately clear to be before testing it out. Thoughts on adding a comment?

https://github.com/readthedocs/readthedocs.org/blob/a370adec0d40185703f6972b00b3fc421808a283/readthedocs/projects/version_handling.py#L76-L101

readthedocs/rtd_tests/tests/projects/test_version_sorting.py
humitos: test-builds, which may not apply as a real project but shows the sorting problem when building alphanumeric branch names

humitos:
> This is by design it seems but perhaps a comment is warranted. This wasn't immediately clear to be before testing it out. Thoughts on adding a comment?

I'm :+1: adding a comment explaining this where you consider necessary.

davidfischer:
I added the comments explaining comparable versions. This is ready to merge, I believe.

ericholscher:
I wish this was present in the upstream, but that's a fight for another day :)

PR #6019, opened 2019-07-31
Search analytics

### LINKED ISSUE #5967: Provide search analytics to users.
It would be good if the users have access to some Search analytics. It would help them to know what users are actually searching in their docs.

dojutsu-user:
Closes #5967 
WIP


readthedocs/search/utils.py
```diff
@@ -168,3 +169,38 @@ def _get_sorted_results(results, source_key='_source'):
     ]
 
     return sorted_results
+
+
+def record_search_query(project_slug, version_slug, query, total_results):
+    """Record search query in database."""
+    if not project_slug or not version_slug or not query or not total_results:
```
dojutsu-user: Should this function be a celery task?
Every search has to create/update object in database before showing results to the user.
I am not sure on how much this will impact the speed.
But I think that it will be better to have this as a celery task -- so that it can create the objects asynchronously and we have the results at the same speed.

ericholscher: Yea, having it be out of the direct query response will be best.

readthedocs/search/utils.py
```diff
@@ -168,3 +169,38 @@ def _get_sorted_results(results, source_key='_source'):
...
+        return
```
ericholscher: This should debug log probably.

readthedocs/settings/base.py
```diff
@@ -326,6 +326,16 @@ def USE_PROMOS(self):  # noqa
             'schedule': crontab(minute=0, hour='*/3'),
             'options': {'queue': 'web'},
         },
+        'every-three-month-delete-old-search-queries': {
+            'task': 'readthedocs.search.tasks.delete_old_search_queries_from_db',
+            'schedule': crontab(
+                minute=0,
+                hour=0,
+                day_of_month=1,
+                month_of_year='*/4'
```
ericholscher: We want to run this more frequently, but only delete the last 3 months of data. I imagine it running daily.

readthedocs/search/models.py
```diff
@@ -0,0 +1,76 @@
...
+        Version,
+        verbose_name=_('Version'),
+        related_name='search_queries',
+        on_delete=models.CASCADE,
+    )
+    query = models.CharField(
+        _('Query'),
+        max_length=4092,
+    )
+    count = models.PositiveIntegerField(
```
ericholscher: I'm wondering if we want more data here. Should we be storing an object _each_ time a search happens? That way we can show the frequency of a search over time. Currently, this only tells us how many times a search has happened.

I think if we plan to delete the data every 3 months, we can probably store _every_ search query with it's own timestamp. I'm fine with shipping this initially though, before we start storing a lot more data.

dojutsu-user: I am not sure what you meant be `more data`.

Storing search object everytime a search was made is a better idea. I realised that the graphs were wrong before. And going this way makes them correct and easier.

> That way we can show the frequency of a search over time.

Can you exapand this more?
* Do we want this to be selected by user? Like the user can select a date and we show him the frequency of searches made vs time for that day.
* Or we just show this for today/yesterday?

dojutsu-user:
> I'd like to keep to the same libraries we're already using, so if you can use the same stuff as the ad code, that would be great. I don't feel strongly though.

Which libraries? I can't find any.
Also, I have added the chart.js files only on search-analytics page, so it shouldn't have any problems with some other pages.

readthedocs/search/models.py
ericholscher: Just that we will be able to see when a specific search was done each time it was done. The current modeling only shows the number of times a search was done, but no time data about each search.

readthedocs/search/tasks.py
```diff
@@ -118,3 +122,51 @@ def index_missing_objects(app_label, model_name, document_class, index_generatio
...
+
+    This is run by celery beat every day.
+    """
+    last_3_months = timezone.now().date() - timezone.timedelta(days=90)
+    search_queries_qs = SearchQuery.objects.filter(
+        created__date__lte=last_3_months,
+    )
+
+    if search_queries_qs.exists():
+        log.info('Deleting search queries for last 3 months. Total: %s', search_queries_qs.count())
```
ericholscher: Good logging :)

readthedocs/search/tasks.py
```diff
@@ -118,3 +122,51 @@ def index_missing_objects(app_label, model_name, document_class, index_generatio
...
+        search_queries_qs.delete()
+
+
+@app.task(queue='web')
+def record_search_query(project_slug, version_slug, query, total_results):
```
ericholscher: Why are we passing `total_results` then not saving it?

readthedocs/templates/projects/project_edit_base.html
```diff
@@ -24,6 +24,7 @@
         <li class="{% block project-integrations-active %}{% endblock %}"><a href="{% url "projects_integrations" project.slug %}">{% trans "Integrations" %}</a></li>
         <li class="{% block project-environment-variables-active %}{% endblock %}"><a href="{% url "projects_environmentvariables" project.slug %}">{% trans "Environment Variables" %}</a></li>
         <li class="{% block project-notifications-active %}{% endblock %}"><a href="{% url "projects_notifications" project.slug %}">{% trans "Notifications" %}</a></li>
+        <li class="{% block project-searcn-analytics-active %}{% endblock %}"><a href="{% url "projects_search_analytics" project.slug %}">{% trans "Search Analytics" %}</a></li>
```
ericholscher:
```suggestion
        <li class="{% block project-search-analytics-active %}{% endblock %}"><a href="{% url "projects_search_analytics" project.slug %}">{% trans "Search Analytics" %}</a></li>
```

readthedocs/templates/projects/projects_search_analytics.html
```diff
@@ -0,0 +1,141 @@
+{% extends "projects/project_edit_base.html" %}
+
+{% load i18n %}
+{% load static %}
+
+{% block title %}{% trans "Search Analytics" %}{% endblock %}
+
+{% block nav-dashboard %} class="active"{% endblock %}
+
+{% block project-searcn-analytics-active %}active{% endblock %}
```
ericholscher:
```suggestion
{% block project-search-analytics-active %}active{% endblock %}
```

readthedocs/projects/constants.py
```diff
@@ -355,3 +355,19 @@
...
 
 GITHUB_PR_PULL_PATTERN = 'pull/{id}/head:external-{id}'
+
+SEARCH_ANALYTICS_PARAMS = {
+    'period': (
+        ('recent', 'Recent Queries'),
+        ('last-24-hrs', 'Top Queries of Past 24 Hrs'),
+        ('last-48-hrs', 'Top Queries of Past 48 Hrs'),
+        ('last-7-days', 'Top Queries of Past 7 Days'),
+        ('last-1-month', 'Top Queries of Past 1 Month'),
```
ericholscher: Should we show a `last-3-months` option, if we are storing the data?

readthedocs/projects/utils.py
```diff
@@ -42,3 +44,33 @@ def safe_write(filename, contents):
...
+    count_data = (
+        queryset.values('query')
+        .annotate(count=Count('id'))
+    )
+
+    if sort:
+        # If sort is true, order the results by ``count``.
+        count_data = count_data.order_by('-count')
+
+    final_values = count_data.values_list('query', flat=True)
```
ericholscher: Feels a bit weird to be doing a `values` and a `values_list` on the same queryset. I think this can probably be simplified.

readthedocs/search/tasks.py
ericholscher: This will also cause us to not store search queries with 0 results, which feels wrong.

ericholscher:
> Which libraries? I can't find any.

We're using `nvd3` -- I'd prefer to keep to the same set of libraries, if possible. But not if it's too much work to change them.

readthedocs/search/tasks.py
dojutsu-user: My motive was to not save the query if no results.
If we save every queries, we might end up having queries which don't show anything.
Like, with the readthedocs-sphinx-search, a search query is made after some time the user stops typing (~500ms).
So, for typing -- 'advertising',
we might end up having multiple queries: `adve`, `advertis`, `advertising`.

readthedocs/projects/constants.py
dojutsu-user: Ohh yeah...
I forgot to update this part.
With the previous modelling, it was having some problems.
But I have changed it according to your suggestion.
I will update the PR

readthedocs/projects/utils.py
dojutsu-user: > to be doing a values and a values_list on the same queryset

I am not sure that I followed.

dojutsu-user:
@ericholscher 
I'd would prefer to use chart.js because it is more interactive and have better animations, if it is okay with you?
You can compare: 
- http://nvd3.org/examples/pie.html
- https://www.chartjs.org/docs/latest/charts/doughnut.html

readthedocs/projects/constants.py
dojutsu-user: Done.

ericholscher:
Yea that's fine 👍

readthedocs/projects/utils.py
dojutsu-user: Ohh.. I get it.
I think it's okay.
Last `values_list` is just to extract the data in list format without iterating over manually.

dojutsu-user:
@ericholscher 
I have updated the PR.
And also added a download button which allows proj admins to download all data in csv format.

readthedocs/search/tasks.py
```diff
@@ -118,3 +122,51 @@ def index_missing_objects(app_label, model_name, document_class, index_generatio
...
+            'Not recording the search query. Passed arguments: '
+            'project_slug: %s, version_slug: %s, query: %s, total_results: %s' % (
+                project_slug, version_slug, query, total_results
+            )
+        )
+        return
+
+    project_qs = Project.objects.filter(slug=project_slug)
+
+    if not project_qs.exists():
```
ericholscher: This should probably log a warning.

readthedocs/templates/projects/search_analytics/csv_data_template.txt
```diff
@@ -0,0 +1,3 @@
+serial_no,date_time,query
+{% for row in data %}{{ forloop.counter }},"{{ row.0|addslashes }}","{{ row.1|addslashes }}"
+{% endfor %}
```
ericholscher: Definitely we should use the `csv` library for this.

PR #6020, opened 2019-07-31
Update get_absolute_url for External Versions

ericholscher:
This allows us to properly set a new domain in production for the external version hosting.
It's a bit hacky for now,
mostly because we don't have a way to specify HTML or PDF versions of the external versions.

We will need to think more about the UX around multiple types,
but that's also an existing issue with our current doc builds (we only link to HTML).


ericholscher:
Yea, the local serving is a bit weak, but it isn't too much of a worry. We will be changing up how all this works soon, so that's why I'm trying to keep it simple.

PR #6028, opened 2019-08-01
Remove project count on homepage

davidfischer:
This query is causing a lot of load, currently.


davidfischer:
I'll remove it entirely. Seems not very useful and rather expensive.

PR #6033, opened 2019-08-02
Send site notification on Build status reporting failure and follow DRY

saadmk11:
ref: #6014

### DIFF
--- a/readthedocs/oauth/notifications.py
+++ b/readthedocs/oauth/notifications.py
@@ -62,9 +62,9 @@ class SendBuildStatusFailureNotification(SiteNotification):
     context_object_name = 'project'
     failure_level = ERROR_PERSISTENT
     failure_message = _(
-        'Could not send {{ provider.name }} build status report for {{ project.name }}. '
-        'Make sure you have the correct {{ provider.name }} repository permissions</a> and '
-        'your <a href="{{ url_connect_account }}">{{ provider.name }} account</a> '
+        'Could not send {{ provider_name }} build status report for {{ project.name }}. '
+        'Make sure you have the correct {{ provider_name }} repository permissions</a> and '
+        'your <a href="{{ url_connect_account }}">{{ provider_name }} account</a> '
         'is connected to ReadtheDocs.'
     )  # noqa
 
--- a/readthedocs/oauth/notifications.py
+++ b/readthedocs/oauth/notifications.py
@@ -64,7 +64,8 @@ class SendBuildStatusFailureNotification(SiteNotification):
     failure_message = _(
         'Could not send {{ provider.name }} build status report for {{ project.name }}. '
         'Make sure you have the correct {{ provider.name }} repository permissions</a> and '
-        'your <a href="{{ url_connect_account }}"> {{ provider.name }} account</a> is connected to ReadtheDocs.'
+        'your <a href="{{ url_connect_account }}">{{ provider.name }} account</a> '
+        'is connected to ReadtheDocs.'
     )  # noqa
 
     def get_context_data(self):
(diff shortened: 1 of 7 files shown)


saadmk11:
As I was Adding the User communication part to the task I saw that we were using some of the same code more than once. So, I tried to follow DRY and moved some of the redundant code to `projects.models.Project` as methods. I think this approach is much cleaner and reusable for the future when we add other providers. :)

readthedocs/oauth/notifications.py
```diff
@@ -55,3 +55,22 @@ def get_context_data(self):
             ),
         })
         return context
+
+
+class SendBuildStatusFailureNotification(SiteNotification):
+
+    context_object_name = 'project'
+    failure_level = ERROR_PERSISTENT
```
ericholscher: I don't know if we need it to persist. I don't know the best pattern for this, but we should probably err on the side of something that disappears, so we don't overwhelm users.

readthedocs/projects/tasks.py
```diff
@@ -1889,47 +1889,55 @@ def send_build_status(build_pk, commit, status):
     :param status: build status failed, pending, or success to be sent.
     """
     build = Build.objects.get(pk=build_pk)
+    provider_name = build.project.git_provider_name
 
-    try:
-        if build.project.remote_repository.account.provider == 'github':
-            service = GitHubService(
+    if provider_name == 'GitHub':
```
ericholscher: This should be a constant. Also, why are we checking the name instead of `build.project.remote_repository.account.provider` like previously?

readthedocs/oauth/notifications.py
```diff
@@ -55,3 +55,22 @@ def get_context_data(self):
             ),
         })
         return context
+
+
+class SendBuildStatusFailureNotification(SiteNotification):
```
ericholscher: This shouldn't have `Send` in the name.

readthedocs/projects/tasks.py
```diff
@@ -1889,47 +1889,55 @@ def send_build_status(build_pk, commit, status):
...
+            )
+            return False
 
-        log.info(
-            'No social account or repository permission available for %s',
-            build.project
-        )
-        return False
+        except Exception:
+            log.exception('Send build status task failed for %s', build.project)
```
ericholscher:
```suggestion
            log.exception('Send build status task failed for %s', build.project.slug)
```

readthedocs/oauth/notifications.py
saadmk11: @ericholscher `ERROR_PERSISTENT` Disappears when clicked on the link and error `ERROR_NON_PERSISTENT` disappears on page reload. which do you think we should have. and also both of then will re appear every time on failure. Do we want that or do we want something different?

readthedocs/projects/tasks.py
saadmk11: previously we were just getting the provider name by using `build.project.remote_repository.account.provider` but now we have a better way `project.git_provider_name` as this actually checks which provider the project is from and `project.git_service_class()` get the actual class the projects needs (e.g: GitHubService) so we don't need that anymore.

readthedocs/oauth/notifications.py
ericholscher: I think the one that disappers on page-reload makes sense. We can adjust over time.

readthedocs/projects/constants.py
```diff
@@ -355,3 +355,6 @@
 )
 
 GITHUB_PR_PULL_PATTERN = 'pull/{id}/head:external-{id}'
+
+# Git provider names
+GITHUB = 'GitHub'
```
ericholscher: This isn't a good name, since it has specialized caps. It should likely be `GITHUB_BRAND` or something,

saadmk11: `GITHUB_PROVIDER_NAME` maybe?

saadmk11:
Updated the Changes :)

PR #6034, opened 2019-08-02
Do not show wipe version message on build details page for External versions

saadmk11:
We should not show wipe version message on the build details page for external versions as we disabled wipe version functionality for external versions.
https://github.com/readthedocs/readthedocs.org/blob/93bcf8d19c367e1088893a4902b15a05d1ad190f/readthedocs/core/views/__init__.py#L74-L79


saadmk11:
Yeah we can do that :)

saadmk11:
@ericholscher Updated the PR.

PR #6050, opened 2019-08-08
Search analytics improvements

dojutsu-user:
This is a follow up PR from #6019


dojutsu-user:
@ericholscher 
I have made the suggested changes.

> Not sure if we really even want to cascade these deletes. Is there a reason we don't want to store Version here as a string, so we can keep them forever even if a version is deleted? ([comment](https://github.com/readthedocs/readthedocs.org/pull/6019#discussion_r311712452))

We can definitely do these. But we are keeping nothing for `forever` -- we delete the old queries every day. So I think that won't be of much benefit?

dojutsu-user:
We do have some queries to optimize if they break in the prod. :crossed_fingers:

readthedocs/projects/views/base.py
```diff
@@ -115,3 +115,17 @@ def post(self, request, *args, **kwargs):
...
+    """
+    A class that implements just the write method of the file-like interface.
+
+    This class can be used for generating StreamingHttpResponse.
+    See: https://docs.djangoproject.com/en/2.2/howto/outputting-csv/#streaming-large-csv-files
+    """
+
+    def write(self, value):
+        """Write the value by returning it, instead of storing in a buffer."""
+        return value
```
ericholscher: I'd probably put this in a `utils.py` class, not a base View.

readthedocs/templates/projects/projects_search_analytics.html
```diff
@@ -13,13 +13,16 @@
 {% block project_edit_content %}
   {% if show_analytics %}
 
-    <h3>{% trans "Top Queries:" %}</h3>
+    <h3>{% trans "Top queries of all time" %}</h3>
```
ericholscher: I don't "all time" is correct. I think the previous text is fine.

readthedocs/projects/views/base.py
dojutsu-user: Done. :+1:

ericholscher:
Looks good, will ship this out today. 👍

PR #6062, opened 2019-08-09
IPv6 in X-Forwarded-For fix

davidfischer:
It's unclear to me whether our production proxies would ever include an IPv6 with port. In the errors I'm seeing in prod, I don't see that. X-Forwarded-For is not exactly a standard so there isn't something to look up here.

### DIFF
--- a/readthedocs/analytics/tests.py
+++ b/readthedocs/analytics/tests.py
@@ -34,12 +34,18 @@ class UtilsTests(TestCase):
         )
 
     def test_get_client_ip_with_x_forwarded_for(self):
-        
         # only client's ip is present
         request = RequestFactory().get('/')
         request.META['HTTP_X_FORWARDED_FOR'] = '203.0.113.195'
         client_ip = get_client_ip(request)
         self.assertEqual(client_ip, '203.0.113.195')
+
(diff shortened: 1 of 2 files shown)


readthedocs/analytics/tests.py
```diff
@@ -34,13 +34,19 @@ def test_anonymize_ua(self):
...
-        
         # only client's ip is present
         request = RequestFactory().get('/')
         request.META['HTTP_X_FORWARDED_FOR'] = '203.0.113.195'
         client_ip = get_client_ip(request)
         self.assertEqual(client_ip, '203.0.113.195')
 
+        # only client's ip is present
+        request = RequestFactory().get('/')
+        ip = '2001:abc:def:012:345:6789:abcd:ef12'
```
dojutsu-user: Where did you get the sample ip?

ericholscher:
Hotfixed to `rel`.

readthedocs/analytics/tests.py
davidfischer: I just created it. You can always test if it's a valid IP (it is) with the `ipaddress` module.

PR #6063, opened 2019-08-09
Send media downloads to analytics

davidfischer:
While data will be sampled, this will give us an idea of how frequently media (zips, PDFs, ePubs) are downloaded.

### DIFF
--- a/readthedocs/projects/views/public.py
+++ b/readthedocs/projects/views/public.py
@@ -21,6 +21,8 @@ from django.views.decorators.cache import never_cache
 from django.views.generic import DetailView, ListView
 from taggit.models import Tag
 
+from readthedocs.analytics.tasks import analytics_event
+from readthedocs.analytics.utils import get_client_ip
 from readthedocs.builds.constants import LATEST
 from readthedocs.builds.models import Version
 from readthedocs.builds.views import BuildTriggerMixin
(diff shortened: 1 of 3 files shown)


readthedocs/projects/views/public.py
```diff
@@ -220,6 +222,15 @@ def project_download_media(request, project_slug, type_, version_slug):
         slug=version_slug,
     )
 
+    # Send media download to analytics - sensitive data is anonymized
+    analytics_event.delay(
+        event_category='Build Media',
+        event_action=f'Download {type_}',
+        event_label=str(version),
+        ua=request.META.get('HTTP_USER_AGENT'),
+        uip=get_client_ip(request),
```
ericholscher: Do we want to send this?

davidfischer: If we don't care about it, we could definitely drop it. It is anonymized but it also isn't really useful. I'd be fine dropping it.

ericholscher: Up to you. I'm 👍 on dropping it if we don't need it though.

davidfischer: I'll update. Let's drop it. Seems cleaner that way especially since we don't need it. I'll just verify that GA doesn't silently drop requests without them.

davidfischer: In my testing, without sending *something* for the user agent and IP address, the event doesn't show up in the realtime reporting. I'll give it a day to see if it does show up in the aggregated reporting but we may need to send some value for those settings. That could be the server's IP address.

ericholscher: Alright, let's just ship it and let GA anonymize it, we're already doing it elsewhere.

davidfischer: After looking at GA's regular reporting, only the events with an IP and UA were counted.

ericholscher:
👍

PR #6066, opened 2019-08-11
PR Builder GitLab Integration

saadmk11:
ref: https://github.com/readthedocs/readthedocs.org/issues/6052

### DIFF
--- a/readthedocs/api/v2/views/integrations.py
+++ b/readthedocs/api/v2/views/integrations.py
@@ -42,6 +42,12 @@ GITHUB_PULL_REQUEST_REOPENED = 'reopened'
 GITHUB_PULL_REQUEST_SYNC = 'synchronize'
 GITHUB_CREATE = 'create'
 GITHUB_DELETE = 'delete'
+GITLAB_MERGE_REQUEST = 'merge_request'
+GITLAB_MERGE_REQUEST_CLOSE = 'close'
+GITLAB_MERGE_REQUEST_MERGE = 'merge'
+GITLAB_MERGE_REQUEST_OPEN = 'open'
+GITLAB_MERGE_REQUEST_REOPEN = 'reopen'
+GITLAB_MERGE_REQUEST_UPDATE = 'update'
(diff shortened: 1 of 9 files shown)


readthedocs/api/v2/views/integrations.py
```diff
@@ -441,6 +463,17 @@ def is_payload_valid(self):
             return False
         return token == secret
 
+    def get_external_version_data(self):
+        """Get Commit Sha and merge request number from payload."""
```
ericholscher:
```suggestion
        """Get commit SHA and merge request number from payload."""
```

readthedocs/api/v2/views/integrations.py
```diff
@@ -441,6 +463,17 @@ def is_payload_valid(self):
...
+        try:
+            identifier = self.data['object_attributes']['last_commit']['id']
+            verbose_name = str(self.data['object_attributes']['iid'])
+
+            return identifier, verbose_name
+
+        except KeyError:
+            raise ParseError('Parameters "id" and "iid" are required')
```
ericholscher: Where is this `ParseError` handled?

readthedocs/rtd_tests/tests/test_project.py
```diff
@@ -193,6 +195,21 @@ def test_get_latest_build_excludes_external_versions(self):
         # Test that External Version is not considered for get_latest_build.
         self.assertEqual(self.pip.get_latest_build(), None)
 
+    def test_git_provider_name_github(self):
+        self.assertEqual(self.pip.git_provider_name, GITHUB_BRAND)
```
ericholscher: These should probably explicitly assign the `repo`, like the `gitlab` tests, so they won't fail if we change something elsewhere.

readthedocs/api/v2/views/integrations.py
saadmk11: When there is a `ParseError` we return `400` to the API. We do the same thing for the `push` events.

readthedocs/rtd_tests/tests/test_project.py
saadmk11: Done!

Issue #3932, opened 2018-04-10
No way back after deleting the webhook integration


humitos:
Steps to reproduce it:

1. Import GitHub project (do not use Manual Import)
1. Webhook is created sucessfully
1. Go to your GitHub project and remove the RTD integration from settings
1. Go to your Admin -> Integrations
1. Select your integration
1. Remove/Delete it

Now, at this point, you have a repository completely disconnected from GitHub: none of them knows the other part.

At this point, 

1. go to Admin -> Integrations
1. Click Add Integration
1. Select Github
1. Click Add Integration

![captura de pantalla_2018-04-10_13-27-02](https://user-images.githubusercontent.com/244656/38576136-e33a4b6c-3cc2-11e8-873e-b0fd9915a0f4.png)

This integration lives under RTD but GitHub knows nothing about it yet. To make it work, you have to go to GitHub settings and add it manually.

I'd expect this to be done automatically by RTD if I'm adding a known integration (not a Generic one).

agjohnson:
Ultimately, the number of folks hitting this is probably very low, so this will be low priority. Also, if the user intervened to disconnect the webhook, they can be responsible for manually fixing this as well -- and we have some directions here at least for now. Having a resync function on the webhook also addresses this issue, as that is the actual functionality you're going for.

dplanella:
A couple of additional things for folks being affected with this bug:

- The "_this integration was created automatically from an existing webhook configured on your repository_" message is misleading. If you see it, the webhook on your github/gitlab/etc. repo is probably no longer there and you'll have to recreate it upstream.
- To recreate the webhook on the upstream repository, use the URL from the readthedocs message as the "Payload" (e.g. on GitHub) input.
- Even after doing this, the "Resync webhook" button will no longer appear in the readthedocs UI 

@agjohnson  @humitos happy to file separate bugs for these issues if it helps.

humitos:
Yesterday, I was bitten by this again. This time in a different case.

1. I imported my project from GH using the auto-import (not manually)
1. the webhook failed to be created because I didn't have admin permissions on the repo

At this point, I request admin permissions to the organization the project is under, and I was granted.

Now, I don't have a way (without deleting and recreating the project) to make RTD to re-create the webhook on GH automatically.

ericholscher:
Can we not just call `attach_webhook` when we create the Integration?

saadmk11:
[...]

PR #6071, opened 2019-08-13
Auto Sync and Re-Sync for Manually Created Integrations

saadmk11:
A lot of work needs to be done and a lot of cases needs to be handled before its ready for review. :)
closes https://github.com/readthedocs/readthedocs.org/issues/3932

related to the first step of #6052

### DIFF
--- a/readthedocs/projects/views/private.py
+++ b/readthedocs/projects/views/private.py
@@ -781,6 +781,11 @@ class IntegrationList(IntegrationMixin, ListView):
 
 class IntegrationCreate(IntegrationMixin, CreateView):
 
+    def form_valid(self, form):
+        self.object = form.save()
+        update_webhook(self.get_project(), self.object, request=self.request)
+        return HttpResponseRedirect(self.get_success_url())
+
     def get_success_url(self):
         return reverse(
             'projects_integrations_detail',
--- a/readthedocs/projects/views/private.py
+++ b/readthedocs/projects/views/private.py
@@ -783,11 +783,12 @@ class IntegrationCreate(IntegrationMixin, CreateView):
 
     def form_valid(self, form):
         self.object = form.save()
-        attach_webhook(
-            project_pk=self.get_project().pk,
-            user_pk=self.request.user.pk,
-            integration=self.object
-        )
+        if self.object.integration_type != Integration.API_WEBHOOK:
+            attach_webhook(
+                project_pk=self.get_project().pk,
+                user_pk=self.request.user.pk,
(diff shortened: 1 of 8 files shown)


readthedocs/templates/projects/integration_webhook_detail.html
```diff
@@ -19,15 +19,11 @@
   {% if integration.has_sync and integration.can_sync %}
     <p>
       {% blocktrans trimmed %}
-        This webhook was configured when this project was imported. If this
-        integration is not functioning correctly, try resyncing the webhook:
+        This webhook was configured when this project was imported
+          or it was manually created with correct configuration. If this
+        integration is not functioning correctly, try re-syncing the webhook:
```
ericholscher: This looks good, but I'd like to see us _always_ displaying the webhook URL information, so that users can confirm that the automated webhook is the one they see in their GitHub admin. We should always show this data.

readthedocs/templates/projects/integration_webhook_detail.html
```diff
@@ -19,15 +19,11 @@
   {% if integration.has_sync and integration.can_sync %}
     <p>
       {% blocktrans trimmed %}
-        This webhook was configured when this project was imported. If this
-        integration is not functioning correctly, try resyncing the webhook:
+        This webhook was configured when this project was imported
+          or it was manually created with correct configuration. If this
```
ericholscher:
```suggestion
          or it was automatically created with the correct configuration. If this
```

ericholscher:
BTW, this is _so much better_ than making users login to GitHub and create the webhook manually. This is a huge win!

saadmk11:
@ericholscher ready for a full review :)

readthedocs/projects/views/private.py
```diff
@@ -783,11 +783,12 @@ class IntegrationCreate(IntegrationMixin, CreateView):
 
     def form_valid(self, form):
         self.object = form.save()
-        attach_webhook(
-            project_pk=self.get_project().pk,
-            user_pk=self.request.user.pk,
-            integration=self.object
-        )
+        if self.object.integration_type != Integration.API_WEBHOOK:
```
ericholscher: Do we not have some kind of `has_sync` check here? I feel like this might break in the future as we add new types of Integration.

saadmk11: Thanks. I think what you said would be better. :)

saadmk11: @ericholscher Updated the PR!

PR #6076, opened 2019-08-16
GitLab Build Status Reporting for PR Builder

saadmk11:
This PR is created on top of #6066. PR: #6066 should be merged first.

ref: #6052


ericholscher:
Was hoping that would rebase the commits, but we need to update this now that #6066 is merged.

saadmk11:
@ericholscher Updated the PR

readthedocs/rtd_tests/tests/test_celery.py
```diff
@@ -356,8 +356,8 @@ def test_send_build_status_task_with_remote_repo(self, send_build_status):
         )
         self.assertEqual(Message.objects.filter(user=self.eric).count(), 0)
 
-    @patch('readthedocs.projects.tasks.GitHubService.send_build_status')
-    def test_send_build_status_task_with_social_account(self, send_build_status):
+    @patch('readthedocs.oauth.services.github.GitHubService.send_build_status')
```
ericholscher: Why did this change? It seems like it was wrong before?

saadmk11: Because we are not directly importing `GitHubService` to the tasks we are using `project.git_service_class()`

saadmk11: It was working before because I forgot to remove an unused import of `GitHubService`

PR #6080, opened 2019-08-19
Use `sphinx-hoverxref` extension to show tooltips in our docs

humitos:
Example of how our docs will look like with the `sphinx-hoverxref` extension installed and applied to all `:ref:` roles.

Result can be seen at https://external-builds.readthedocs.io/html/docs/6080/index.html


docs/conf.py
```diff
@@ -39,9 +39,14 @@ def get_version():
     'recommonmark',
     'notfound.extension',
     'sphinx_search.extension',
+    'hoverxref.extension',
 ]
 templates_path = ['_templates']
 
+hoverxref_auto_ref = True
+# hardcode version because external versions does not work properly
```
ericholscher: Why don't they work properly?

humitos: Because the version slug is the number of the PR. In this case, `6080`.

So, the URL https://readthedocs.org/api/v2/embed/?project=docs&version=6080&doc=config-file/v2&section=Migrating%20from%20v1 returns a 404

but using `latest` works.

https://readthedocs.org/api/v2/embed/?project=docs&version=latest&doc=config-file/v2&section=Migrating%20from%20v1

I didn't check the backend yet to know if this can be fixed.

humitos:
This makes the extension works on our docs.

screenshot-external-builds readthedocs io-2019 10 09-12_51_42

Although, there is still a bug that makes relative links inside the tooltip content to break: https://github.com/readthedocs/readthedocs.org/pull/6082

I'm closing this PR for now. We can enable the extension once we have that other issue fixed.

PR #6087, opened 2019-08-21
Remove 'highlight' URL param from search results

dojutsu-user:


humitos:
I'm lacking context to review this PR. Why are we doing this?

dojutsu-user:
@humitos 
It was kind of annoying.
For example: https://sphinx-notfound-page.readthedocs.io/en/latest/autoapi/notfound/extension/index.html?highlight=Point%20media%20files%20at%20our%20media%20server#notfound.extension.finalize_media

humitos:
Gotcha! I think the problem to solve here is different. Instead of removing the highlighting, I think we should make to highlight _only_ the full text searched.
In your example, `Point media files at our media server`.

dojutsu-user:
@humitos 

> highlight only the full text searched

How can we achieve this?
I think, currently we are using default highlighting mechanism provided by Sphinx.

humitos:
I have no idea :grin:  --maybe solving it upstream? is it even reported?

My point is that we are removing a good feature because of a small bug in only one case and I think we need to solve that bug instead.

stsewd:
@humitos there was a conversation in slack about this. Other suggestion was to have an ui element to deactivate the highlight. Also, I think that now we link to specific sections, is more easy to find the word you were looking for.

Personally I always modify the url to remove the highlight, and just use the firefox search if I need something like that

stsewd:
The other PR https://github.com/readthedocs/readthedocs-sphinx-search/pull/43

ericholscher:
I think there are some theme updates we can do to make it easier to remove the highlighting, some patches to Sphinx in general, but I also always remove it when I hit the page. I think it should likely be opt-in with default off, which would make it much more user-friendly IMO.

humitos:
OK. I like the highlight feature. I understand that there are cases where it's broken and I'm fine with this immediate solution for this.

I think the right solution is to fix these cases instead (I don't know how much work it means, though) and make this opt-in.

dojutsu-user:
@ericholscher 

> I think it should likely be opt-in with default off, which would make it much more user-friendly IMO.

I remembered that I saw this feature in `requests` docs.

Peek 2019-08-23 12-07

Link: http://2.python-requests.org/en/master/user/quickstart/?highlight=payload#passing-parameters-in-urls

dojutsu-user:
Opened: https://github.com/readthedocs/sphinx_rtd_theme/issues/816

agjohnson:
I am a strong -1 on this change, I think we should revert this and figure out a better way to address whatever problems we were having with the results. Highlighting in doc is 100% a useful feature, though I always have to remove the `highlight=...` from the URL when linking to a user. This PR is lacking context as to why this change was needed or why we made this change, so I'd like to revisit this and consider reverting.

dojutsu-user:
@agjohnson 
I apologize for not writing the underlying reason and need for this PR.
The main reason for removal of highlight param was that highlighted keywords were mostly seems like distractions, like in this case: https://sphinx-notfound-page.readthedocs.io/en/latest/autoapi/notfound/extension/index.html?highlight=Point%20media%20files%20at%20our%20media%20server#notfound.extension.finalize_media

I always find it removing from the url.

agjohnson:
I raised some better fixes for this case in #6305

PR #6106, opened 2019-08-27
Update our Elastic Search shard/replica settings

ericholscher:
This is to make our search function a bit better. The comment in the code
explains why.

Our current .org prod `page_index` is 12GB, so we're well within the range where this config makes sense 👍


safwanrahman:
From my experience, I think it will work well. But I think if we run a query in single shard index, it will run through all the amount of data it have. But if we have 2 shard, it get split between two shards and the query can run parallelly in half of the amount of data.

We can try this and can tune it for other performance improvement. r+

PR #6121, opened 2019-08-29
Store Pageviews in DB

dojutsu-user:
Related: #5968


readthedocs/api/v2/views/footer_views.py
```diff
@@ -210,6 +211,13 @@ def get(self, request, format=None):
             'version_supported': version.supported,
         }
 
+        # increase the page view count
+        increase_page_view_count.delay(
+            project=context['project'],
```
ericholscher: We can't pass the actual model objects, we have to pass the PK or slug. You should probably be running Celery in development to catch these issues: https://docs.readthedocs.io/en/latest/development/standards.html#core-team-standards

readthedocs/search/models.py
```diff
@@ -137,3 +137,60 @@ def generate_distribution_of_top_queries(cls, project_slug, n):
             final_data['int_data'].append(count_of_other)
 
         return final_data
+
+
+class PageView(TimeStampedModel):
```
ericholscher: I don't think this should use `TimeStampedModel` -- we only want a `date` field on it, not a created & updated DateTime field.

readthedocs/search/models.py
```diff
@@ -137,3 +137,60 @@ def generate_distribution_of_top_queries(cls, project_slug, n):
...
+        143 total page views respectively.
+        """
+        qs = (
+            cls.objects
+            .filter(project=project)
+            .values_list('path')
+            .annotate(total_views=Sum('view_count'))
+            .values_list('path', 'total_views')
+            .order_by('-total_views')[:10]
+        )
```
ericholscher: I'm pretty worried about this query. It seems like we're going to be doing something that is definitely not indexed in the DB across a whole lot of fields. 

Can you legally have two `values_list` calls? How does that work?

dojutsu-user: I think it is correct.
This generates the following SQL query.

```sql
SELECT "search_pageview"."path", SUM("search_pageview"."view_count") AS "total_views" 
FROM "search_pageview"
WHERE "search_pageview"."project_id" = 62
GROUP BY "search_pageview"."path"
ORDER BY "total_views" DESC
LIMIT 10
```

readthedocs/search/tasks.py
```diff
@@ -212,3 +212,27 @@ def record_search_query(project_slug, version_slug, query, total_results, time_s
...
+    today_date = timezone.now().date()
+    page_view_obj = PageView.objects.filter(
+        project__slug=project_slug,
+        version__slug=version_slug,
+        path=path,
+        date=today_date,
+    ).first()
+
+    if page_view_obj:
+        page_view_obj.view_count += 1
```
davidfischer: This should use an [F expression](https://docs.djangoproject.com/en/1.11/ref/models/expressions/#f-expressions) to avoid a race condition.

readthedocs/api/v2/views/footer_views.py
```diff
@@ -210,6 +211,14 @@ def get(self, request, format=None):
             'version_supported': version.supported,
         }
 
+        # increase the page view count
+        page_slug = request.GET.get('page', 'index')
```
ericholscher: This should likely not default to `index`, no? If we don't have a `page`, we probably shouldn't be storing anything, or some other string like `rtd-unknown-page`, not `index`.

readthedocs/projects/views/private.py
```diff
@@ -1002,3 +1002,41 @@ def _search_analytics_csv_data(self):
...
+
+        top_viewed_pages = PageView.get_top_viewed_pages(project)
+        top_viewed_pages_iter = zip(
+            top_viewed_pages['pages'],
+            top_viewed_pages['view_counts']
+        )
+
+        all_pages = PageView.objects.filter(project=project).values_list('path', flat=True)
+        if all_pages.exists():
+            all_pages = sorted(list(set(all_pages)))
```
ericholscher: Why are we calling `sorted(list(set` here? Should we not just sort it in the query?

readthedocs/projects/views/private.py
```diff
@@ -1002,3 +1002,41 @@ def _search_analytics_csv_data(self):
...
+            page_path = self.request.GET.get('page', all_pages[0])
```
ericholscher: Why is this defaulting to `all_pages[0]`?

readthedocs/rtd_tests/tests/test_footer.py
```diff
@@ -249,30 +249,36 @@ def render(self):
...
+                response = self.render()
+                self.assertContains(response, '0.99.0')
 
     def test_domain_queries(self):
         # Setting up a custom domain shouldn't impact the number of queries
-        self.pip.domains.create(
-            domain='http://docs.foobar.com',
-            canonical=True,
-        )
+        with mock.patch('readthedocs.api.v2.views.footer_views.increase_page_view_count') as mocked:
```
ericholscher: Why do we need these changes?

readthedocs/search/models.py
```diff
@@ -81,10 +82,109 @@ def generate_queries_count_of_one_month(cls, project_slug):
...
+        verbose_name=_('Version'),
+        related_name='page_views',
+        on_delete=models.CASCADE,
+    )
+    path = models.CharField(max_length=4096)
+    view_count = models.PositiveIntegerField(default=1)
+    date = models.DateField()
+
+    def __str__(self):
+        return f'[{self.project.slug}:{self.version.slug}]: {self.path}'
```
ericholscher: This should probably mention it's a pageview.

readthedocs/search/models.py
```diff
@@ -81,10 +82,109 @@ def generate_queries_count_of_one_month(cls, project_slug):
...
+        Sample returned data::
+            {
+                'labels': ['01 Jul', '02 Jul', '03 Jul'],
+                'int_data': [150, 200, 143]
+            }
+        This data shows that there were 150 page views on 01 July,
+        200 page views on 02 July and 143 page views on 03 July for a particular `page_path`.
+        """
+        today = timezone.now().date()
+        last_30th_day = timezone.now().date() - timezone.timedelta(days=30)
```
ericholscher:
```suggestion
        30_days_ago = timezone.now().date() - timezone.timedelta(days=30)
```

readthedocs/search/utils.py
```diff
@@ -168,3 +169,23 @@ def _get_sorted_results(results, source_key='_source'):
...
+    """Returns iterator for last 31 days (including today)."""
+    today = timezone.now().date()
+    last_30th_day = timezone.now().date() - timezone.timedelta(days=30)
+
+    # this includes the current day also
+    last_31_days_iter = [last_30th_day + timezone.timedelta(days=n) for n in range(31)]
+    return last_31_days_iter
+
+
+def _get_last_31_days_str(date_format):
```
ericholscher: We should probably have a default string here, which is the one used everywhere currently.

ericholscher:
This is still useful, but I want to wait until our DB load calms down a bit before adding additional load to it.

readthedocs/search/models.py
davidfischer: A variable can't start with a number!

readthedocs/rtd_tests/tests/test_footer.py
davidfischer: I'm guessing this is because otherwise that celery task will be called (with `CELERY_ALWAYS_EAGER`) and increase the number of queries.

readthedocs/search/tasks.py
```diff
@@ -212,3 +213,27 @@ def record_search_query(project_slug, version_slug, query, total_results, time_s
...
+        page_view_obj.save()
+    else:
+        project = Project.objects.get(slug=project_slug)
+        version = Version.objects.get(project=project, slug=version_slug)
+        PageView.objects.create(
+            project=project,
+            version=version,
+            path=path,
+            date=today_date,
+        )
```
davidfischer: This whole function could be simplified to something like this:

```python
# This assumes that the default `view_count` is 0 instead of 1 as it is currently
page_view, _ = PageView.objects.get_or_create(
    # If we passed project/version IDs to this function instead of slugs, we could save 2 queries too!
    project=Project.objects.get(slug=project_slug),
    version=Version.objects.get(project=project, slug=version_slug),
    path=path,
    date=timezone.now().date(),
)
PageView.objects.filter(pk=page_view.pk).update(
    view_count=F('view_count') + 1
)
```

readthedocs/search/models.py
```diff
@@ -81,10 +82,109 @@ def generate_queries_count_of_one_month(cls, project_slug):
...
+
+        final_data = {
+            'labels': last_31_days_str,
+            'int_data': count_data,
+        }
+
+        return final_data
+
+
+class PageView(models.Model):
```
davidfischer: We should consider having a [unique index](https://docs.djangoproject.com/en/1.11/ref/models/options/#unique-together) for this model although it could result in a lot of space used. Nonetheless, it would prevent a whole class of problem:

```python
class Meta:
    unique_together = ("project", "version", "path", "date")
```

readthedocs/search/models.py
davidfischer: This query is probably ok since it's limited by the project.

readthedocs/api/v2/views/footer_views.py
davidfischer: I haven't seen any real examples where the page parameter is blank. It is `index` for the root and even with the dirhtml builder it uses the subdirectory name. I'll add a simple `if page_slug` check.

davidfischer:
I refactored this and brought it to a merge-ready state.

- There's a periodic task to remove old page views
- The screen (see screenshot) now shows aggregated pageviews by day. That's a total traffic graph.

I am interested to see what this does to our database as it will add a (simple) write query in a celery task on every Footer API call.

## Screenshot

Screen Shot 2020-05-08 at 2 12 57 PM

readthedocs/projects/views/private.py
```diff
@@ -977,15 +975,6 @@ class RegexAutomationRuleUpdate(RegexAutomationRuleMixin, UpdateView):
     pass
 
 
-@login_required
-def search_analytics_view(request, project_slug):
```
davidfischer, line 981: I have no idea what this view is. It doesn't appear to be used or valid.

readthedocs/search/tasks.py
```diff
@@ -209,3 +210,29 @@ def record_search_query(project_slug, version_slug, query, total_results, time_s
...
+
+
+@app.task(queue='web')
+def delete_old_page_counts():
+    """
+    Delete page counts older than 30 days.
+
+    This is intended to run from a periodic task daily.
+    """
+    thirty_days_ago = timezone.now().date() - timezone.timedelta(days=30)
```
ericholscher, line 237: This should probably log something, so we know it's running properly.

readthedocs/search/tasks.py
```diff
@@ -209,3 +210,29 @@ def record_search_query(project_slug, version_slug, query, total_results, time_s
         version=version,
         query=query,
     )
+
+
+@app.task(queue='web')
+def increase_page_view_count(project_slug, version_slug, path):
```
ericholscher, line 216: I wonder if running this as a task is actually a good deal slower then running it in the footer itself? We already have the Project & Version object in the footer, and not we're adding additional queries for them here. They should be fast queries, but definitely additional load. 

I guess the proper solution is to cache these lookups, which we can do later if it's an issue.

davidfischer:
> This looks good to me. I'm curious if we should be keeping the pageview data in the search app, but I think it's fine for now?

I had this thought myself. I don't have a strong opinion either way. I'd be fine moving it into the Project app or a case could even be made for the Build app as it is Version specific. The case for the search app is that that is where the data is most useful: for ordering search results. However, as we have a page in the project admin, maybe that isn't the right place.

I'd rather get it into the right place the first time it's deployed. Moving it could be a bit of a pain. If you have an opinion, let's get that in.

ericholscher:
Well, arguably the Versions should be in the Project app instead of Builds, so we're already in a confusing place to start :) 

I think we put them in Search originally because we wanted to use them to weight search results by pageview count. I think it probably makes sense for them to be in their own app, something like `analytics` or `pageviews`, but again, I don't have super clear idea of whats the best.

ericholscher:
Ah, we already have an `analytics` app, so that probably makes sense :)

ericholscher:
@davidfischer it would be good to get this patched up and shipped this week if possible.

madwaxamillion:
I'll take a look at it  But I can't make any promises I might need  Some
help

davidfischer:
I moved the pageview modeling from the search app to the analytics app. This is ready for a full review.

ericholscher:
Looks good 🎉

PR #6127, opened 2019-09-01
Add PR Title to The Build Details Page

saadmk11:
This Pr will add the Pull request Title to the Build details page.
Screenshot from 2019-09-01 18-08-22

### DIFF
--- a/readthedocs/builds/models.py
+++ b/readthedocs/builds/models.py
@@ -1,6 +1,7 @@
 """Models for the builds app."""
 
 import datetime
+import json
 import logging
 import os.path
 import re
@@ -8,6 +9,7 @@ from shutil import rmtree
 
 from django.conf import settings
 from django.db import models
+from django.db.models import Q
 from django.urls import reverse
 from django.utils import timezone
 from django.utils.translation import ugettext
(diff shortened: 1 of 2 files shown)


readthedocs/builds/models.py
```diff
@@ -879,6 +881,28 @@ def external_version_name(self):
...
+        """Returns the title of the pull/merge request by parsing HttpExchange object."""
+        if self.is_external:
+            sha_str = f'"sha": "{self.commit}",'
+            number_str = f'"number": {self.version.verbose_name},'
+            integration_list = self.project.integrations.all().values_list('id', flat=True)
+            http_exchange = HttpExchange.objects.filter(
+                Q(object_id__in=integration_list) &
+                Q(request_body__contains=sha_str) &
+                Q(request_body__contains=number_str)
+            ).first()
```
ericholscher: I'm worried about this query. It feels really heavy for a page that is loaded so much. Is there not an easier way to do this? We should cache the data on the build if we want to display it.

saadmk11: Maybe its better to add a field where we will store this. It will be cleaner and less complex.

humitos: I'm :+1: on adding a field in the db. Although, I don't like the idea of executing this query with `contains` on these objects because that data could be really big.

I'm echoing Eric's question here, "Is there not an easier way to do this?"

saadmk11: Another Approach can be converting `TextField` to `JSONField` which will make it much easier to query
ref: https://docs.djangoproject.com/en/2.2/ref/contrib/postgres/fields/#key-index-and-path-lookups

PR #6135, opened 2019-09-04
Fix link to html_extra_path

ericholscher:
This should probably use intersphinx, but keeping it small for now.


ericholscher:
I've updated this to be a bit more correct, and mention it needs to come from the default version.

PR #6139, opened 2019-09-05
Allow to extend DomainCreate view

humitos:
This has no affect in Community. The change is to allow us to do something different in Corporate (redirect to Domain details after creation).


ericholscher:
Is there a reason we don't want to have the same logic on the community site?

humitos:
@ericholscher yes.

Basically, what I'm adding in the corporate site is a redirect to the Domain details once it's created, because the user needs to know there are some extra steps they need to setup.

This is not exactly the same case in community. We could show the message saying "Add a CNAME pointing to readthedocs.io" if we want, but we will need more work in the community template as well.

agjohnson:
I think that the logic should be the same, and there is a PR open for starting to make this UI use the correct patterns, but I'm fine accepting this as a solution for now.

I've pushed up 500f4a2 to also extend the update view, which we use to inject context data into the template. Re-requesting review.

PR #6142, opened 2019-09-06
Serialize time in search queries properly

ericholscher:


readthedocs/search/tasks.py
```diff
@@ -153,6 +153,7 @@ def record_search_query(project_slug, version_slug, query, total_results, time):
         )
         return
 
+    time = timezone.datetime.strptime(time_string, "%Y-%m-%dT%H:%M:%S.%f+00:00")
```
davidfischer:
```suggestion
    time = timezone.utc.localize(timezone.datetime.strptime(time_string, "%Y-%m-%dT%H:%M:%S.%f+00:00"))
```

PR #6148, opened 2019-09-07
Add footer template file caching

tapaswenipathak:
Added caching for footer html file. Writing one PR for memcache implementation.

Fixes - https://github.com/readthedocs/readthedocs.org/issues/6126.


tapaswenipathak:
Yeah, I require input from your end. I suggest:

- Add template file caching for unchanging header https://github.com/readthedocs/readthedocs.org/pull/6148
- File based cache for the file https://github.com/readthedocs/readthedocs.org/pull/6149/files
- memcache instead of local cache

?

davidfischer:
We use redis for caching in production.

tapaswenipathak:
Hi @davidfischer: Ok, does everything else works w/ you? or you have more ideas?

tapaswenipathak:
@davidfischer: have any thoughts?

davidfischer:
@stsewd's comment still stands, I believe. This would cache the footer for one project across all projects.

tapaswenipathak:
@stsewd, @davidfischer: I require inputs/thoughts for: https://github.com/readthedocs/readthedocs.org/pull/6148#issuecomment-529690680.

readthedocs/api/v2/templates/restapi/footer.html
```diff
@@ -1,4 +1,6 @@
 <!-- Inserted RTD Footer -->
+{% load cache %}
+{% cache 900 footer %}
```
humitos: This will cache the footer and use _the same footer for all the projects_. That's not correct. We need to add more keywords where.

Also, adding cache here, even if we use the proper variables/keywords, could cause confusions on users when a new version is activated, a translation is added, etc. So, we may want to invalidate the cache in those cases as well.

tapaswenipathak:
@humitos hi: You can read the ticket, the footer view should be cached properly. Why are we going back and forth in discussing what the pr does

> would cache the footer for one project across all projects

can I please get few notes for https://github.com/readthedocs/readthedocs.org/pull/6148#issuecomment-529690680?

humitos:
Hi @tapaswenipathak! This PR can't be merged as it is because it will return the same footer for all the projects on Read the Docs as I explained at https://github.com/readthedocs/readthedocs.org/pull/6148#discussion_r335552833.

Also, we did not settle on a proper solution and there should be more discussion around it before continue with an implementation. Unfortunately, the original issue is under "Needed: design decision" and it's not "Approved".

I'm closing this PR for now and we can come back later once we have a better idea of how to cache the footer.

PR #6172, opened 2019-09-12
Document connected account permissions

davidfischer:
This is a documentation only pull request that just documents why we ask for the permissions from GitHub, Bitbucket, and GitLab. We've had a few questions about this in the past.

There is a meta question of whether this is the right place for this information. Another question is whether we should link to this from the login and signup screens so users are prepared for permission prompts.

### DIFF
--- a/docs/connected-accounts.rst
+++ b/docs/connected-accounts.rst
@@ -1,5 +1,5 @@
 Connecting Your Account
------------------------
+=======================
 
 If you are going to import repositories from GitHub, Bitbucket, or GitLab,
 you should connect your Read the Docs account to your repository host first.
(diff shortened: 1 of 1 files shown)


saadmk11:
Should we talk about the GitLab Event we are accessing (e.g: `merge_requests_events`) ?

https://github.com/readthedocs/readthedocs.org/blob/e50345c70b69adc5a72ff5363683e52d9febef8b/readthedocs/oauth/services/gitlab.py#L261-L267

davidfischer:
If we're good with this change, let's go forward with it. Once this has "been released" an is available in the stable version, we'll add those links to it.

docs/connected-accounts.rst
```diff
@@ -18,3 +18,64 @@ and select `Connected Services <https://readthedocs.org/accounts/social/connecti
...
+
+Permissions for connected accounts
+----------------------------------
+
+Read the Docs does **not** ask for write permission to your repositories' code
+and since we only connect to public repositories we don't need special permissions to read them.
+However, we do need permissions for authorizing your account
+so that you can login to Read the Docs with your connected account credentials
+and to setup :doc:`webhooks`
+which allow us to build your documentation on every change to your repository.
```
agjohnson: Unfortunately, the commercial side *does* ask for admin (and therefore write) permissions, as this is the lowest granularity that GitHub has/had in order to add private ssh keys. Moving to a GitHub App instead does probably give us finer permission control, if i remember correctly.

docs/connected-accounts.rst
```diff
@@ -18,3 +18,64 @@ and select `Connected Services <https://readthedocs.org/accounts/social/connecti
...
+    Repository statuses allow Read the Docs to report the status
+    (eg. passed, failed, pending) of pull requests to GitHub.
+    This is used for a feature currently in beta testing
+    that builds documentation on each pull request similar to a continuous integration service.
+
+.. note::
+
+    :doc:`Read the Docs for Business </commercial/index>`
+    asks for one additional permission (``repo``) to allow access to private repositories
+    and to allow us to setup SSH keys to clone your private repositories.
```
agjohnson: This is the permission noted above. We have to ask for repo admin permissions, which does give us write permission on the repository

docs/connected-accounts.rst
```diff
@@ -18,3 +18,64 @@ and select `Connected Services <https://readthedocs.org/accounts/social/connecti
...
+
+Bitbucket
+~~~~~~~~~
+
+For similar reasons to those above for GitHub, we request permissions for:
+
+* Reading your account information including your email address
+* Read access to your team memberships
+* Read access to your repositories
+* Read and write access to webhooks
```
agjohnson: It would be worth checking if our bitbucket/gitlab permissions are different on the commercial side as well. I don't believe we have to ask for special permissions like GitHub.

davidfischer: I did look and I didn't see anything different.

docs/connected-accounts.rst
davidfischer: Ya, it does. I'll update this to be a little more explicit. I don't know if the new GitHub App does give us finer permission control. I don't think it does.

PR #6174, opened 2019-09-12
Add terms of service

davidfischer:
**As this changes the terms of service for Read the Docs, it should not be merged without approval of either @agjohnson, @ericholscher, or both**

This PR adds a terms of service for Read the Docs, Read the Docs for Business, and Documentation Sites. It updates the Privacy Policy to link to the terms.

* I wrote the terms with the intention that they apply to Read the Docs for Business
* I followed [sentence case](https://docs.readthedocs.io/en/stable/development/docs.html#titles) for non-H1 headings except when a term was from the definitions section
* These terms were heavily influenced (see: borrowed or lifted) from [GitHub's CC-0 licensed site policies](https://github.com/github/site-policy/)
* GitHub's terms had a whole sub-document on serving legal notice. Since we don't have that in place, I removed much of it.
* The terms are "governed by the federal laws of the United States of America and the laws of the State of Oregon" 
* If you are casually reviewing this, I think a second set of eyes on the Payments/Refunds section specifically wouldn't be bad
* I tentatively put the date on the terms/privacy policy for the end of the month but this can be adjusted if and when this is merged.

## Future work

- Link to this terms of service from the footer of readthedocs.org. Since our links usually link to the "stable" version of the docs, we should wait until this is released to add that link.
- Link to this ToS from the footer of readthedocs.com and replace the [existing terms](https://readthedocs.com/terms/).
- Consider linking to this from the flyout menu on Documentation Sites

### DIFF
--- a/docs/terms-of-service.rst
+++ b/docs/terms-of-service.rst
@@ -288,7 +288,7 @@ Intellectual property notice
 ----------------------------
 
 **Short version:** *We own the Service and all of our Content.
-In order for you to use our content, we give you certain rights to it,
+In order for you to use our Content, we give you certain rights to it,
 but you may only use our content in the way we have allowed.*
 
 Read the Docs' rights to content
--- a/docs/terms-of-service.rst
+++ b/docs/terms-of-service.rst
@@ -348,12 +348,12 @@ and may not be appropriate for some high-bandwidth uses or other prohibited uses
 Read the Docs reserves the right at all times to reclaim any Read the Docs subdomain without liability.
 
 
-Third Party Applications
+Third party applications
 ---------------------------
 
 **Short version:** *You need to follow certain rules if you create an application for other Users.*
 
-Creating Applications
+Creating applications
 ~~~~~~~~~~~~~~~~~~~~~
 
 If you create a third-party application or other developer product that collects User Personal Information
@@ -497,7 +497,7 @@ Communications with Read the Docs
 
 **Short version:** *We use email and other electronic means to stay in touch with our users.*
 
-Electronic Communication Required
+Electronic communication required
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 For contractual purposes, you:
--- a/docs/terms-of-service.rst
+++ b/docs/terms-of-service.rst
@@ -61,7 +61,7 @@ readthedocs.org ("Read the Docs Community")
     writing and distributing technical documentation.
 
 readthedocs.com ("Read the Docs for Business")
-    This website is a commercial hosted offering for hosting private
+    This Website is a commercial hosted offering for hosting private
     documentation for corporate clients.
 
 readthedocs.io, readthedocs-hosted.com, and other domains ("Documentation Sites")
--- a/docs/terms-of-service.rst
+++ b/docs/terms-of-service.rst
@@ -652,4 +652,5 @@ including any confidentiality or nondisclosure agreements.
 
 Questions
 ~~~~~~~~~
-Questions about the Terms of Service? `Contact us <mailto:support@readthedocs.org>`_.
+
+Questions about the Terms of Service? `Get in touch <mailto:support@readthedocs.org>`_.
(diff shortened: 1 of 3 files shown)


docs/terms-of-service.rst
```diff
@@ -0,0 +1,656 @@
...
+There may be legal consequences for sending a false or frivolous takedown notice.
+Before sending a takedown request, you must consider legal uses such as fair use and licensed uses.
+
+We will terminate the Accounts of repeat infringers of this policy.
+
+
+Intellectual property notice
+----------------------------
+
+**Short version:** *We own the service and all of our content.
```
agjohnson: The "short versions" seem to waffle whether they are including legal terms defined above, but it seems the trend is to include them in these sections. These next few can be skipped if not.

```suggestion
**Short version:** *We own the Service and all of our Content.
```

docs/terms-of-service.rst
```diff
@@ -0,0 +1,656 @@
...
+In order for you to use our content, we give you certain rights to it,
```
agjohnson:
```suggestion
In order for you to use our Content, we give you certain rights to it,
```

docs/terms-of-service.rst
```diff
@@ -0,0 +1,656 @@
...
+but you may only use our content in the way we have allowed.*
```
agjohnson:
```suggestion
but you may only use our Content in the way we have allowed.*
```

docs/terms-of-service.rst
```diff
@@ -0,0 +1,656 @@
...
+in addition to the rest of the Terms.*
+
+Documentation Sites
+~~~~~~~~~~~~~~~~~~~
+
+Each Read the Docs Account comes with the ability to host Documentation Sites.
+This hosting service is intended to host static web pages for All Users.
+Documentation Sites are subject to some specific bandwidth and usage limits,
+and may not be appropriate for some high-bandwidth uses or other prohibited uses.
+Read the Docs reserves the right at all times to reclaim any Read the Docs subdomain without liability.
```
agjohnson: "Read the Docs subdomain" might not be clear enough here, it's more of an internal concept. Perhaps it's best to use a higher level concept? We're communicating here that we have the right to alter or remove a project at any time. We don't define what "Project" is, but perhaps this is a helpful addition. We can say "we have the right to remove User Generated Content," but this has a slightly different meaning.

docs/terms-of-service.rst
```diff
@@ -0,0 +1,656 @@
...
+
+Third party applications
+---------------------------
+
+**Short version:** *You need to follow certain rules if you create an application for other Users.*
+
+Creating applications
+~~~~~~~~~~~~~~~~~~~~~
+
+If you create a third-party application or other developer product that collects User Personal Information
```
agjohnson: "User Personal Information" was never defined and it might not be clear enough what this is.

davidfischer: This term is described in detail in our privacy policy: https://docs.readthedocs.io/en/stable/privacy-policy.html#information-from-users-with-accounts

docs/terms-of-service.rst
davidfischer: This is true. GitHub defines subdomains in their [definitions section](https://github.com/github/site-policy/blob/master/Policies/github-terms-of-service.md#a-definitions). We have already say we "have the right to remove content or close Accounts if we need to" in our User-Generated Content section. Maybe this can just be removed.

The core concept that I wanted to get across and that I think GitHub wanted to get across in their terms is that if they need to reclaim something like github.com/store (because there's a user named "store") that they can. The same applies to subdomains (eg. `*.readthedocs.io` for us and `*.github.io` for github).

agjohnson: It looks like they don't really defined this sort of scenario in super clear terms either. Perhaps the blob should just clarify what the subdomain is -- "subdomains on readthedocs.io and readthedocs-hosted.com, prefixed with your project name" etc -- but also maybe describe that we have the right to alter project names? For instance, if your project is abandoned, or in conflict with another project, we reserve the right to rename your project.

I only mention the second case here as this is the net effect the user would notice -- the project slug would change. Feel free to reword things, I'm only using examples here. I think we're on the same on what needs to be clarified

davidfischer:
I think all the issues are addressed if this could get another (final?) review.

agjohnson:
Changes look good! The unresolved bit on my side was the subdomain removal note, but that looks to have been removed. That should be fine I think, we've discussed project removal already.

PR #6191, opened 2019-09-17
Remove if storage blocks

davidfischer:
Storage is now required. By default, storage uses local storage on the filesystem so it doesn't have to be some cloud storage.


davidfischer:
I had to put back one check because [build servers for RTD for Business](https://github.com/readthedocs/readthedocs-corporate-ops/pull/231) are going to temporarily use it.

PR #6192, opened 2019-09-17
Remove pie-chart from search analytics page

dojutsu-user:
Remove piechart and few refactoring.

Demo:
screencapture-localhost-8000-dashboard-rtd-search-analytics-2019-09-18-01_05_52


ericholscher:
@dojutsu-user We should also remove the feature flag on the search queries now 👍

PR #6210, opened 2019-09-26
Re-sync Repository from VCS

saadmk11:
closes https://github.com/readthedocs/readthedocs.org/issues/6090


readthedocs/templates/projects/project_version_list.html
```diff
@@ -26,6 +26,15 @@
           <h1>{% trans "Active Versions" %}</h1>
         </div>
 
+        {% block resync_project %}
+          {% if request.user|is_admin:project %}
+            <form method="post" action="{% url 'project_repository_resync' project_slug=project.slug %}">
+              {% csrf_token %}
+              <input type="submit" value="{% trans "Re-sync Project" %}">
```
stsewd:
```suggestion
              <input type="submit" value="{% trans "Re-sync Versions" %}">
```

saadmk11:
@stsewd This looks better I think :)

Screenshot from 2019-09-27 00-14-19

readthedocs/projects/urls/private.py
```diff
@@ -31,6 +31,7 @@
     ProjectNoticationsDelete,
     ProjectRedirects,
     ProjectRedirectsDelete,
+    ProjectRepositoryReSync,
```
humitos: I'd rename this to `ProjectVersionsSync`.

readthedocs/projects/views/private.py
```diff
@@ -213,6 +214,39 @@ def form_valid(self, form):
...
+
+    def post(self, request, *args, **kwargs):
+        # pylint: disable=unused-argument
+        project = self.get_project()
+        sync = sync_versions(project)
+
+        if sync:
+            messages.success(
+                request,
+                _('Project repository re-sync triggered!'),
```
humitos: No need to use `!` in the messages.

readthedocs/projects/views/private.py
```diff
@@ -213,6 +214,39 @@ def form_valid(self, form):
...
+            )
+        else:
+            messages.error(
+                request,
+                _('Unable to re-sync project repository!'),
```
humitos:
```suggestion
                _('There was an error when syncing project's versions'),
```

readthedocs/templates/projects/project_version_list.html
```diff
@@ -22,6 +22,21 @@
...
       <div class="module-wrapper">
 
+        {% block resync_versions %}
+          {% if request.user|is_admin:project %}
+          <div class="button-bar">
+            <ul>
+              <li>
+                <form method="post" action="{% url 'project_repository_resync' project_slug=project.slug %}">
+                  {% csrf_token %}
+                  <input type="submit" value="{% trans 'Re-sync Versions' %}">
```
humitos: I'd use "Sync Versions" for the label of the button.

```suggestion
                  <input type="submit" value="{% trans 'Sync Versions' %}">
```

stsewd:
> The "Sync" does not affect "Active versions"

Actually it does, it activates a version if it follows semver and it's greater than the current stable. Also it updates latest and stable.

humitos:
OK. So, it makes more sense to be below the "Active versions" then :smile:

ericholscher:
I think this is a good interim UI, but I don't think it's quite ready. I think the better approach here is to make it an actual part of the page, with copy that explains what it does. Currently "Sync Versions" doesn't really mean anything to many users, and will likely be confusing. It seems like it would be more valuable with some copy that explains that it will update the versions from the VCS provider, and what that will do.

agjohnson:
I like to avoid putting copy in UI, it clutters the UI and is often just going to be skipped by the user. I think a refresh icon button communicates more that "resync versions" and doesn't require us to explain our UI.

readthedocs/projects/views/private.py
```diff
@@ -213,6 +214,33 @@ def form_valid(self, form):
...
+
+        if sync:
+            messages.success(
+                request,
+                _('Project version sync triggered'),
+            )
+        else:
+            messages.error(
+                request,
+                _("There was an error when syncing project's versions"),
```
agjohnson: Same feedback here, but also: there isn't much the user can do with this error. Do we have any additional information we can display in the `sync` variable? What error does a failure with the sync task point at otherwise? Perhaps there is more guidance that we can give the user.

As for the message, it's better to speak to the user -- something closer to:

```suggestion
                _("There was a problem updating the project versions"),
```

readthedocs/templates/projects/project_version_list.html
```diff
@@ -26,6 +26,21 @@
...
         </div>
 
+        {% block resync_versions %}
+          {% if request.user|is_admin:project %}
+            <div class="button-bar">
+              <ul>
+                <li>
+                  <form method="post" action="{% url 'project_version_sync' project_slug=project.slug %}">
+                    {% csrf_token %}
+                    <input type="submit" value="{% trans 'Sync Versions' %}">
```
agjohnson: I think a refresh icon button probably communicates what we need here and doesn't require us to add copy to our UI.

saadmk11: Okay :)

readthedocs/projects/views/private.py
saadmk11: @agjohnson Actually `sync_versions` function only calls a celery task `sync_repository_task` So, I don't know if we can show much more information from `sync` variable.

readthedocs/templates/projects/project_version_list.html
saadmk11: I have added this one:
Screenshot from 2019-11-29 01-33-27

Another option might be:
Screenshot from 2019-11-29 01-33-11

saadmk11:
I think this is still valid

ericholscher:
@agjohnson would be good for you to take a peek here. I think it's probably worth integrating into the new design.

PR #6215, opened 2019-09-27
Customize default build media storage for the FS

davidfischer:
Edit: closely mirror the logic to get the production media path:

https://github.com/readthedocs/readthedocs.org/blob/557c875fa4d74e1cb69146e005375595f80adf91/readthedocs/projects/models.py#L584-L598

### DIFF
--- a/readthedocs/builds/storage.py
+++ b/readthedocs/builds/storage.py
@@ -114,13 +114,11 @@ class BuildMediaFileSystemStorage(BuildMediaStorageMixin, FileSystemStorage):
         location = kwargs.pop('location', None)
 
         if not location:
-            # If no location was passed:
-            # - Use PRODUCTION_MEDIA_ARTIFACTS if it exists
-            # - Fallback to MEDIA_ROOT
-            if Path(settings.PRODUCTION_MEDIA_ARTIFACTS).exists():
-                location = settings.PRODUCTION_MEDIA_ARTIFACTS
-            else:
+            # Mirrors the logic of getting the production media path
+            if settings.DEFAULT_PRIVACY_LEVEL == 'public' or settings.DEBUG:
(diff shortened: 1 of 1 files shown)


davidfischer:
> do we actually want to pass a location when on the .org

In production, this class is not used on .org at all. We use an Azure storage class that mixes in `BuildMediaStorageMixin`. In development for .org, perhaps a better approach is to actually have `settings.PRODUCTION_MEDIA_ARTIFACTS` have meaning. That is, we could actually write build artifacts there as opposed to it being a meaningless, non-existent folder which it is now. In that case, we could just always make this class use `settings.PRODUCTION_MEDIA_ARTIFACTS`.

davidfischer:
The biggest drawback to what I outlined above as far as I can tell is that people will have to move their built artifacts to a different directory in dev.

agjohnson:
That might make sense. We can always work around having a secondary path, it's going to be much harder to tease apart the build output path, normal media path usage, and improper/old static file media path usage. Locally, we'd probably still run the local build storage on community development, we don't want to be writing to Azure from local dev.

We can come back to this case, but I'd agree that the production artifact path should be used in local development as well.

agjohnson:
I'll be applying this as hotfix to .com later, the change seems to look good

davidfischer:
> Locally, we'd probably still run the local build storage on community development, we don't want to be writing to Azure from local dev.

Just to be clear, I completely agree. I'm proposing basically having `prod_artifacts/media` exist in development.

ericholscher:
> I'm proposing basically having prod_artifacts/media exist in development.

I think this is probably the right approach. It also works well w/ the move to having all the build media served via the webs as well, instead of directly off the media server.

PR #6242, opened 2019-10-03
Allow only post requests for delete views

saadmk11:
closes: https://github.com/readthedocs/readthedocs.org/issues/6189

### DIFF
--- a/readthedocs/templates/projects/project_version_detail.html
+++ b/readthedocs/templates/projects/project_version_detail.html
@@ -19,11 +19,13 @@
 
     {% if request.user|is_admin:project %}
       {% if not version.active and version.built %}
+        <form name="version_delete_html" method="post" action="{% url "project_version_delete_html" project.slug version.slug %}">
+          {% csrf_token %}
+        </form>
         <p class="empty">
-          {% url "project_version_delete_html" project.slug version.slug as version_delete_url %}
           {% blocktrans trimmed %}
             This version is inactive but its documentation is still available online.
-            You can <a href="{{ version_delete_url }}">delete this version's documentation</a> if you want to remove it completely.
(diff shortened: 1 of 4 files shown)


readthedocs/templates/projects/project_version_detail.html
```diff
@@ -19,11 +19,13 @@ <h2> Editing {{ version.slug }} </h2>
...
       {% if not version.active and version.built %}
+        <form name="version_delete_html" method="post" action="{% url "project_version_delete_html" project.slug version.slug %}">
+          {% csrf_token %}
+        </form>
         <p class="empty">
-          {% url "project_version_delete_html" project.slug version.slug as version_delete_url %}
           {% blocktrans trimmed %}
             This version is inactive but its documentation is still available online.
-            You can <a href="{{ version_delete_url }}">delete this version's documentation</a> if you want to remove it completely.
+            You can <a href="#" onclick="document.forms['version_delete_html'].submit();">delete this version's documentation</a> if you want to remove it completely.
```
saadmk11: I'm using this here because using a button here doesn't look good also it breaks lot of CSS styles. Would love to get some Suggestions. :)

stsewd:
Thanks, can you attach some screenshots of the final result?

saadmk11:
@stsewd 

Screenshot from 2019-10-03 21-39-11
Screenshot from 2019-10-03 21-39-33

readthedocs/templates/projects/project_version_detail.html
stsewd: I don't have a suggestion for this, and also this is something that users shouldn't hit that much.

humitos: I think it makes sense for now.

I'd appreciate if you add this comment into the template itself (using `{# #}`) so we know why this done in this way in the future. Thanks!

saadmk11: @humitos Added comment.

Issue #6265, opened 2019-10-09
Message about number of inactive versions shown


humitos:
The new version listing page with the filter to activate versions is really cool. Although, it hides that only the first 100 versions are shown:

https://github.com/readthedocs/readthedocs.org/blob/558e143743576ada85d9a0a4abb5cc0693d3181d/readthedocs/projects/views/public.py#L370

This could confuse users, thinking that we are showing all the inactive versions. To solve this, I think we could add a small text saying that we only show the first 100 inactive versions.

davidfischer:
I'm on board with this. One pattern I've seen is to only show this message when there are 100 inactive versions.

PR #6276, opened 2019-10-10
Show message if version list truncated

davidfischer:
Show a message if we are showing the max inactive versions on the version list screen.

Technically, if the project has exactly 100 versions the message is shown even if nothing was truncated but I think that is acceptable in this instance.

Fixes https://github.com/readthedocs/readthedocs.org/issues/6265


readthedocs/templates/projects/project_version_list.html
```diff
@@ -134,8 +135,19 @@ <h1>{% trans "Activate a version" %}</h1>
           </div>
         {% endif %}
 
+
+
       </div>
     </div>
+
+    {% if inactive_versions|length >= max_inactive_versions %}
```
humitos: > Technically, if the project has exactly 100 versions the message is shown even if nothing was truncated but I think that is acceptable in this instance.

If you use `>` instead of `>=` don't you get the behavior you want?

saadmk11: Won't this `inactive_versions|length` always return 100 if there is more than 100 versions?
as we pass this to the context `inactive_versions[:max_inactive_versions]`

readthedocs/projects/views/public.py
```diff
@@ -367,7 +369,7 @@ def project_versions(request, project_slug):
     version_filter = request.GET.get('version_filter', '')
     if version_filter:
         inactive_versions = inactive_versions.filter(verbose_name__icontains=version_filter)
-    inactive_versions = inactive_versions[:100]
+    inactive_versions = inactive_versions[:max_inactive_versions]
```
ericholscher: Feels like the length check should be done here, before we truncate it. Otherwise won't the template never show additional versions because it's already cut off at `max_inactive_versions` total?

davidfischer: The message is shown because there will be 100 inactive versions and the check checks `>=`. It is an extra query to do a count but I don't think it'll be too costly.

readthedocs/templates/projects/project_version_list.html
```diff
@@ -134,8 +135,19 @@ <h1>{% trans "Activate a version" %}</h1>
...
+      <p>
+        {% blocktrans trimmed %}
+          Only the first {{ max_inactive_versions }} inactive versions were shown. Please <a href="#version-filter">filter the list</a>.
```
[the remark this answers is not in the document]
davidfischer: This will involve an extra query as I outlined https://github.com/readthedocs/readthedocs.org/pull/6276#discussion_r343881508 but I suppose the cost isn't too high. I can add it.

agjohnson:
I think a count would be more helpful here still. I'm not worried about the cost of the extra query. I'm fine waiting until this change is in.

davidfischer:
I just forgot about this one. I'm adding the extra query.

PR #6278, opened 2019-10-10
Store version media availability

davidfischer:
This PR allows the availability of particular download media to be saved on the version at the end of the build process. For example, does a particular version have a PDF? ePub? etc.

Currently, we rely on the presence of a file existing at a certain filesystem path to determine whether media/downloads are available. As we move toward storage being an abstraction on top of the filesystem where it is possibly remote (eg. Cloud storage) we don't want to have to query remote storage to know whether a version has a PDF.

This PR is the first step in changing that. The next is to write a job which calculates `has_pdf`/`has_epub`/`has_htmlzip` for the latest build of every version. Lastly, we can actually use the fields stored in the DB instead of [checking the filesystem](https://github.com/readthedocs/readthedocs.org/blob/044f2005739977600423d6925bc6a5a66299ed88/readthedocs/builds/models.py#L368).


readthedocs/api/v2/serializers.py
```diff
@@ -90,6 +90,9 @@ class Meta:
             'built',
             'downloads',
             'type',
+            'has_pdf',
+            'has_epub',
+            'has_htmlzip',
```
humitos: shouldn't we add these to fields to APIv3 Version detail endpoint as well?

I'm not sure, since the download link will appear only if it exists. So, maybe it's redundant.

readthedocs/rtd_tests/tests/test_api.py
```diff
@@ -2213,6 +2216,23 @@ def test_get_active_versions(self):
...
+        }
+        resp = self.client.patch(
+            reverse('version-detail', kwargs=data),
+            data=json.dumps({'built': False, 'has_pdf': True}),
+            content_type='application/json',
+            HTTP_AUTHORIZATION='Basic {}'.format(eric_auth),  # Eric is staff
+        )
+        self.assertEqual(resp.status_code, 200)
+        self.assertEqual(resp.data['built'], False)
+        self.assertEqual(resp.data['has_pdf'], True)
```
humitos:
```suggestion
        self.assertEqual(resp.data['has_pdf'], True)
        self.assertEqual(resp.data['has_epub'], False)
        self.assertEqual(resp.data['has_htmlzip'], False)
```

readthedocs/api/v2/serializers.py
agjohnson: These `has_*` fields feel like redundant fields for our API

davidfischer: I agree it is a little awkward. These fields are used for writing by the build servers but not reading by the general public.

PR #6286, opened 2019-10-14
Port additional features to proxito

### LINKED ISSUE #6331: Handle dirhtml builds with El Proxito
We might already have work around this, but raising here if not for tracking. Serving a dirhtml build doesn't work for paths that need the missing slash appended. There should be a redirect, and it sounds like we might need to make a second request to determine if the redirect is needed?

### LINKED ISSUE #6332: Handle serving README.html as default index file with Proxito
One of the rules we support in nginx is serving of README.html after index.html is tried. We need to be able to support this either in code or with nginx. Opening the issue here, but we can move if it is going to be addressed with nginx configuration.

ericholscher:
This moves over almost all of the doc serving logic from Django into Proxito. Most importantly:

* sitemap handling
* robots.txt handling
* 404.html handling for users

It also fixes some of the bugs that we saw during the initial deployment:

* Improperly serving `/en/latest/foo` where it should serve `foo/index.html`
* Not having support for README.html as an index file

This required additional nginx configuration changes, which I will put in another PR in ops. It would be cool to be able to test them against this, but not quite there yet. 

Closes #6331 
Closes #6332


readthedocs/proxito/views.py
```diff
@@ -222,26 +248,50 @@ def serve_docs(
...
-        root_path = storage.path('')
-        # Serve from Python
-        return serve(request, path, root_path)
-
-    # Serve via nginx
-    log.info('[Nginx serve] path=%s, project=%s', path, final_project)
-    return _serve_docs_nginx(
-        request, final_project=final_project, path=f'/proxito/{path}'
-    )
+        return _serve_docs_nginx(
```
humitos:
```suggestion
        return _serve_docs_python(
```

readthedocs/proxito/tests/test_full.py
```diff
@@ -185,6 +185,61 @@ def test_custom_robots_txt(self, storage_mock):
...
+            response.status_code, 200
+        )
+
+    @mock.patch('readthedocs.proxito.views.get_storage_class')
+    def test_404_storage_serves_404(self, storage_mock):
+        self.project.versions.update(active=True, built=True)
+
+        storage_mock()().exists.side_effect = [False, False, True]
+        response = self.client.get(
+            reverse('serve_error_404', kwargs={'proxito_path': '/en/fancy-version/index-exists'}),
```
humitos: The path is confusing here because the `index.html` does not exist in this case.

```suggestion
            reverse('serve_error_404', kwargs={'proxito_path': '/en/fancy-version/not-found'}),
```

readthedocs/proxito/tests/test_full.py
```diff
@@ -185,6 +185,61 @@ def test_custom_robots_txt(self, storage_mock):
...
+        storage_mock()().exists.assert_has_calls(
+            [
+                mock.call('html/project/fancy-version/index-exists/index.html'),
+                mock.call('html/project/fancy-version/index-exists/README.html'),
+                mock.call('html/project/fancy-version/404.html'),
+            ]
+        )
+        self.assertEqual(
+            response.status_code, 404
+        )
```
humitos: We may want to assert that we are receiving the content of the custom page here as well and not the default maze.

readthedocs/proxito/tests/test_full.py
```diff
@@ -185,6 +185,61 @@ def test_custom_robots_txt(self, storage_mock):
...
+
+    @mock.patch('readthedocs.proxito.views.get_storage_class')
+    def test_404_storage_paths_checked(self, storage_mock):
+        self.project.versions.update(active=True, built=True)
+        storage_mock()().exists.return_value = False
+        self.client.get(
+            reverse('serve_error_404', kwargs={'proxito_path': '/en/fancy-version/index-exists'}),
```
humitos: I'd change the path here as well, mentioning that it does not exist.

readthedocs/proxito/views.py
```diff
@@ -209,22 +187,70 @@ def serve_docs(
...
+        lang_slug=None,
+        version_slug=None,
+        filename='',
+):
+    """Take the incoming parsed URL's and figure out what file to serve."""
+
+    log.debug(
+        'project=%s, subproject=%s, lang_slug=%s, version_slug=%s, filename=%s',
+        project_slug, subproject_slug, lang_slug, version_slug, filename
+    )
```
humitos: I'd move this after `_get_project_data_from_request` and print the final ones (or both).

readthedocs/proxito/views.py
```diff
@@ -311,6 +336,90 @@ def _serve_docs_nginx(request, final_project, path):
...
+        for tryfile in ('404.html', '404/index.html'):
+            storage_root_path = final_project.get_storage_path(
+                type_='html',
+                version_slug=version_slug_404,
+                include_file=False,
+            )
+            storage_filename_path = os.path.join(storage_root_path, tryfile)
+            if storage.exists(storage_filename_path):
+                log.debug(
+                    'serving 404.html page current version: [project: %s] [version: %s]',
```
humitos:
```suggestion
                    'Serving custom 404.html page: [project: %s] [version: %s]',
```

humitos:
> I can do some QA once we have the PR for ops with the NGINX required changes.

robots.txt, sitemap.xml and custom 404 (on latest, on specific version and _not_ on latest) work properly with the NGINX fallback.

PR #6295, opened 2019-10-16
Fully working docker-compose file

humitos:
This PR configures all Read the Docs following the architecture that we use on production (multiple processes for different services) and allows developers to have an environment very close to what production looks like.

There were other attempts to support `docker-compose` in the past (#3245, #3003, etc) but none of them worked for different reasons. 6 months ago, we started working again on this idea but got blocked because the configuration to sync files between servers was too complicated, but also because the [_famous permission issue_](#2692) and we weren't able to make it work. Although, we did the [first step in that direction](#5556) by solving the issue for the permissions.

> I want to publicly thanks @raulcd for his help on the whole process.

Yesterday, I tried again now that we have "El Proxito" in front of everything and we use `django-storages` to upload the build's output to Cloud files, which reduces a lot the architecture needed for this. **It works!**

I'd like to have more people from @readthedocs/core team involve here and try to push this forward because _it will reduce the barrier to contributors a lot_ (no need of #4608) but also because it will make our life easier, avoiding problems on deploy that we can detect while developing (#5976). Besides, it adds the opportunity to test some operation changes before deploying and reducing the risk of failure.

### Services

This `docker-compose` starts:
- nginx: serving the main app at dev.readthedocs.io and proxito at *.dev.readthedocs.io
- proxito: serve docs from Cloud files (Azurite --Azure emulator)
- web: main application (dashboard)
- celery: celery application attending web,web01,reindex queues
- build: celery application without access to the DB attending builder,celery,default,build01 queues
- storage: Azurite to storage build's output and read them from proxito
- cache: redis 
- search: elasticsearch
- database: postgres

### Changes needed to Python code

- Docker shared volume used between build and the "docker in docker" container created by the build service itself. This avoid the permission issue.
- Serve from NGINX when using a specific setting (`RTD_DOCKER_COMPOSE`)
- Do not do `Domain` check when host is `web` (Docker service)
- Define `proxito` as valid host to serve docs
- ~~Fallback to normal `json` module if `orjson` is not installed (wasn't able to compile it on alpine linux)~~

### Testing this docker-compose file

1. pull this branch (`humitos/docker-compose`)
2. run `GITHUB_TOKEN=<your-github-token> docker-compose up`
3. access to http://docs.dev.readthedocs.io

> NOTE: if you don't have `readthedocs/build:latest` docker image already, you will need to pull it first running `docker pull readthedocs/build:latest` for a build to success

### Improvements / ToDo

- [ ] do not depend on readthedocs-ext (it's only used for custom AzureStorage classes but they can just be moved into .org)
- [x] find a way to automatically run `az` to create containers on first run
- [ ] pass arguments to `docker-compose up` to avoid running `collectstatic` and others each time
- [ ] properly set CORS on Azurite container
- [x] make custom 404 work properly (done at #6306)
- [ ] (optional) configure celery-beat
- [ ] (optional) setup cold storage for old builds


docker/Dockerfile
```diff
@@ -0,0 +1,41 @@
+FROM python:3.6-alpine
```
davidfischer: Alpine Linux definitely results in smaller docker images. However, since our production setup is Ubuntu, it's possible that Ubuntu is more appropriate. I don't have a very strong opinion about it though.

readthedocs/search/parse_json.py
```diff
@@ -1,7 +1,11 @@
 """Functions related to converting content into dict/JSON structures."""
 
 import logging
-import orjson as json
+try:
+    import orjson as json
+except ImportError:
+    # Fallback to standard json module of orjson does not load
```
davidfischer:
```suggestion
    # Fallback to standard json module if orjson does not load
```

docker-compose-search.yml
```diff
@@ -0,0 +1,35 @@
+# docker-compose-search.yml starts "search" service and override some
```
davidfischer: Why don't we just put elasticsearch into the same `docker-compose.yml` so that the same compose will setup search as well? Perhaps I'm missing something.

humitos:
> TypeError: __init__() got an unexpected keyword argument 'token_credential'

Hrm... I thought I fixed this. There is a small incompatibility between `django-storages` and `azure-storage-blob` versions. I will double check this.

> Overall, you might find it easier to use some sort of S3 compatible storage in a setup like this. There might be something more minimal than the Azure setup. I'm not 100% sure. Perhaps something like the S3 portion of [localstack](https://github.com/localstack/localstack)

I'm using [MinIO](https://hub.docker.com/r/minio/minio/) on corporate and it works great. I tried Azure here to be closer to production, though.

docker/Dockerfile
humitos: I agree. Ubuntu is probably better.

docker-compose-search.yml
humitos: The only reason to have it in another yaml file is because elasticsearch consumes a lot of RAM and most of the time is not needed for development. So, I'm doing it optional to start.

I wanted to have a better pattern, something like `docker-compose up --search` or similar but I didn't find a how to do it.

humitos:
> `TypeError: __init__() got an unexpected keyword argument 'token_credential'`

I added a commit that fixes this.

docker/Dockerfile
```diff
@@ -0,0 +1,43 @@
...
+
+RUN mkdir checkouts; cd checkouts
+
+COPY requirements readthedocs.org/requirements
+
+RUN pip install --no-cache-dir -r readthedocs.org/requirements/docker.txt
+
+# We clone the -ext here only to install the requirements at this point.
+# -ext is also mounted as volume from the upper directory
+RUN git clone --single-branch --branch humitos/upgrade-azure-storage https://${GITHUB_TOKEN}@github.com/readthedocs/readthedocs-ext
```
ericholscher: This broke when I tried to run it.

humitos: Ups. That branch was already merged in master and deleted. I fixed the Dockerfile --it should work now. Let me know.

docker-compose-search.yml
yakky: You can limit the ES memory consumption with `"ES_JAVA_OPTS=-Xms512m -Xmx512m"` environment variable, I don't think you can provide custom arguments or conditional container spin up to docker-compose

humitos: Thanks for the tip @yakky! I've done that at https://github.com/readthedocs/readthedocs.org/pull/6295/commits/6acd6f45c299d8a2c53646bbcb7148b16fd00511

humitos: Basically now you can run everything with search by calling:

```
docker-compose -f docker-compose.yml -f docker-compose-search.yml up
```

or without search

```
docker-compose up
```

docker/nginx/web.conf
```diff
@@ -0,0 +1,22 @@
...
+
+    location /favicon.ico {
+        proxy_pass http://storage:10000/devstoreaccount1/static/images/favicon.ico;
+        break;
+    }
+
+    # FIXME: this is a workaround because CORS on the storage.
+    # The command to enable it is failing without reason:
+    # pip install azure-cli
+    # AZURE_STORAGE_CONNECTION_STRING="UseDevelopmentStorage=true" az storage cors add --origins 'docs.dev.readthedocs.io' --services 'b' --methods GET OPTIONS HEAD --allowed-headers '*'
```
humitos: CORS it's not supported in Azurite yet: https://github.com/Azure/Azurite/issues/55#issuecomment-503380561

docker/entrypoints/proxito.sh
```diff
@@ -0,0 +1,5 @@
+#! /bin/sh
+
+../../docker/common.sh
+
+python3 manage.py runserver 0.0.0.0:8000
```
ericholscher: Shouldn't proxito run on a different port than webs?

docker/entrypoints/web.sh
```diff
@@ -0,0 +1,13 @@
...
+
+../../docker/common.sh
+
+python3 manage.py migrate
+cat ../../docker/scripts/createsuperuser.py | python3 manage.py shell
+
+# collect static in background
+python3 manage.py collectstatic --no-input &
+
+python3 manage.py loaddata test_data
```
ericholscher: Doesn't this file run every time we start the docker container? I think we only want it to run once, during initial setup?

readthedocs/core/views/serve.py
```diff
@@ -175,7 +175,7 @@ def _serve_file(request, filename, basepath):
     :raises: ``Http404`` on ``UnicodeEncodeError``
     """
     # Serve the file from the proper location
-    if settings.DEBUG or settings.PYTHON_MEDIA:
+    if (settings.DEBUG or settings.PYTHON_MEDIA) and not settings.RTD_DOCKER_COMPOSE:
```
ericholscher: Is this to enable DEBUG to be true?

readthedocs/doc_builder/environments.py
```diff
@@ -951,22 +975,8 @@ def get_container_host_config(self):
...
-            },
-        }
-
-        if settings.GLOBAL_PIP_CACHE and settings.DEBUG:
-            binds.update({
-                self.project.pip_cache_path: {
-                    'bind': self.project.pip_cache_path,
-                    'mode': 'rw',
-                },
-            })
```
ericholscher: We shouldn't be deleting code in this PR, unless it relates to getting this working I think?

readthedocs/doc_builder/environments.py
```diff
@@ -1039,9 +1049,11 @@ def create_container(self):
                 ),
                 name=self.container_id,
                 hostname=self.container_id,
+                volumes=self._get_binds(),
                 host_config=self.get_container_host_config(),
                 detach=True,
                 environment=self.environment,
+                user=settings.RTD_DOCKER_USER,
```
ericholscher: I worry a bit about this in prod, any reason this is needed?

readthedocs/settings/docker.py
```diff
@@ -0,0 +1,106 @@
+import os
```
ericholscher: This file should probably be called `docker_compose.py` or something. I think it's confusing otherwise.

docker/entrypoints/proxito.sh
humitos: There is no need for that since `proxito` is a different container than `web`, so there is no conflict there. One will be `http://proxito:8000/` and the other `http://web:8000/`.

If that's not what you are asking, I'm missing something.

docker/entrypoints/web.sh
humitos: Right. The same happens with the `migrate` and `collectstatic`. I don't have a good pattern yet. Do you have a suggestion?

readthedocs/doc_builder/environments.py
humitos: Yes. This avoid permission issues. There more context in this PR https://github.com/readthedocs/readthedocs.org/pull/5556

This won't affect production since we are defining `DOCKER_USER = 'docs:docs'` on default settings (https://github.com/readthedocs/readthedocs.org/pull/6295/files#diff-5973fe59a4d52278d55f884d145bd5d7R357) --which is the user and group used by default if this attribute is not specified.

readthedocs/doc_builder/environments.py
humitos: Ups. This was a mistake.

docker/entrypoints/web.sh
ericholscher: not really :/

docker/entrypoints/proxito.sh
ericholscher: I just find it a bit simpler to think about when they're on diff ports, but not an issue for now.

ericholscher:
This seems to make my CPU sit around 80% on one core when idle, so I think there's likely something we need to do to make it do less work. I'm not even running ES in that case, and docker makes my fans turn on :/

readthedocs/core/views/serve.py
humitos: Right. Depending on what you are working on, you will want `DEBUG=True` (e.g. something with the dashboard) or you will want `DEBUG=False` (e.g. working on custom 404s) but still, in both cases you want to serve files from NGINX.

docker/entrypoints/web.sh
humitos: My idea was to use a Python script,

- to make a query and if `test` user already exists don't call `loaddata`
- test if the manifest file is on the storage and don't call `collectstatic` in that case

Finally, these commands could be forced by using an environment variable and running docker compose like,

```
RUN_MIGRATE=t RUN_COLLECTSTATIC=t RUN_LOADDATA=t docker-compose up
```

humitos:
@ericholscher there is a way to limit CPU and memory with YAML file version 2.2 (https://docs.docker.com/compose/compose-file/compose-file-v2/#cpu-and-other-resources). Although, this is not going to be implemented on v3 --just for swarm mode: https://docs.docker.com/compose/compose-file/#resources

There is also a way to convert v3 into its v2 using `--compatibility` (https://github.com/docker/compose/pull/5684). It seems to stay in 2.x is the recommended way to use resource constraints with docker compose locally: https://github.com/docker/compose/issues/4513#issuecomment-281478365

ericholscher:
I don't think limiting it's resources is the answer, it's likely either a docker bug or something we need to adjust in our scripts. Limiting the CPU or similar will just make things slower, when it's already slower than running on my local machine, so definitely not an outcome we want :/

humitos:
@ericholscher when you have some time, please take another look at this PR (I added more commits after you looked last time) and let me know what you think. There may be some places where the code could be improved or decision can be made.

However, we can merge it as it is and continue improving it with smaller PRs.

docker/settings/web.py
```diff
@@ -0,0 +1,9 @@
+from .docker_compose import DockerBaseSettings
+
+class WebDevSettings(DockerBaseSettings):
+
+    # Needed to serve 404 pages properly
+    # NOTE: it may introduce some strange behavior
+    DEBUG = False
```
humitos: I think with the new implementation of El Proxito, this is not needed anymore.

requirements/docker.txt
```diff
@@ -0,0 +1,13 @@
+-r pip.txt
+# http://initd.org/psycopg/docs/install.html#binary-install-from-pypi
+psycopg2-binary==2.8.3
```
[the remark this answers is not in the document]
humitos: Thanks for the suggestion! I'm using -binary here just for simplicity and to avoid installing all the required packages into the docker image. We will be using this only for development at the moment.

humitos:
I added a small note on the YAML file and a also a small README in the docker folder. I'll merge the branch once tests pass.

PR #6306, opened 2019-10-18
Serve 404.html files using storage backend

humitos:
While reviewing El Proxito PR that increase the amount of docs served by it, I found that 404 and robots.txt didn't work (https://github.com/readthedocs/readthedocs-ops/pull/577#pullrequestreview-301251604). Although, when it was deployed, it did work and I thought that was wrong.

Today, I tested this locally using the docker compose branch (#6295) and I found that it works in production because we have the HTML duplicated (on webs and storage) and the files existence is done with `os.path.exists` and symlinks. Once we remove these HTML files from webs this will start failing.

This PR adapts this code to use the default storage backend instead to serve these special files.

### DIFF
--- a/readthedocs/core/views/__init__.py
+++ b/readthedocs/core/views/__init__.py
@@ -10,10 +10,11 @@ import logging
 from urllib.parse import urlparse
 
 from django.conf import settings
-from django.http import HttpResponseRedirect, Http404, JsonResponse
+from django.core.files.storage import get_storage_class
+from django.http import HttpResponseRedirect, Http404, JsonResponse, FileResponse
 from django.shortcuts import render, get_object_or_404, redirect
 from django.views.generic import TemplateView
-from django.views.static import serve as static_serve
+from django.http import FileResponse
 
 from readthedocs.builds.models import Version
 from readthedocs.core.utils.general import wipe_version_via_slugs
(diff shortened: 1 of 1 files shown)


humitos:
I just saw that `robots.txt` was moved to El Proxito in #6286 so maybe I need to revert that code. Although, the 404 handling is still needed since that's the fallback managed in the main django app instance.

humitos:
> I just saw that `robots.txt` was moved to El Proxito in #6286 so maybe I need to revert that code. Although, the 404 handling is still needed since that's the fallback managed in the main django app instance.

This is adapted in the latest commit.

Only 404 pages are being served with django storage in this PR.

ericholscher:
Believe this is already fixed and ported here: https://github.com/readthedocs/readthedocs.org/pull/6286

ericholscher:
We shouldn't be changing how the mainline code works, because it will blow up on the .com I believe. So we should be working in the proxito code and moving it there, like in the PR, unless I'm missing something?

humitos:
> Believe this is already fixed and ported here: #6286

I don't find where is the custom 404 page serve in that PR. Also, I don't think that serving custom 404 pages can be done in El Proxito. Otherwise, we will have to check for `storage.exists` on each request, right?

Hrm... Another idea is to fallback to El Proxito again on `location /proxito/` and use `storage.exists` in this case. That way, this check will be done only on 404 pages and there we can decide to serve the custom one or the RTD default ones. Makes sense?

> We shouldn't be changing how the mainline code works, because it will blow up on the .com I believe.

You are right here. I see three options:

- adding an if to know if we are in corporate or community
- block this PR until corporate uses S3 to upload html files
- test the fallback idea

humitos:
> Hrm... Another idea is to fallback to El Proxito again on `location /proxito/` and use `storage.exists` in this case. That way, this check will be done only on 404 pages and there we can decide to serve the custom one or the RTD default ones. Makes sense?

I did a small test for this and it works. It needs:

- the view modified in this PR, to be moved into `readthedocs.proxito.views`
- add a new `location @notfoundfallback` into nginx
- add `error_page 404 = @notfoundfallback` into `location /proxito/`
- define a `handler404` as the view modified in this PR

humitos:
This is my idea: https://github.com/readthedocs/readthedocs.org/pull/6322

ericholscher:
> I don't find where is the custom 404 page serve in that PR. Also, I don't think that serving custom 404 pages can be done in El Proxito. Otherwise, we will have to check for storage.exists on each request, right?

We will not have the HTML files on the disk after we ship proxito, so we don't have a choice. We are only checking for 404 pages on 404, so we don't have it on every request, only when we have a 404.

humitos:
Closing this for now. The changes needed for this will be probably added in #6286

PR #6307, opened 2019-10-18
Allow to use a different base for static documents

yakky:
This change allows to configure a different base for data storage than the project root

This might be useful mainly when using this project in a docker environment using shared volumes betweeb builder tasks, main application and webserver or for deployment which requires a different layout than the standard one

I cannot find a place where the documentation for this setting could be located or whether any sensible test is doable, but I'm more than willing to add both

### DIFF
--- a/readthedocs/core/symlink.py
+++ b/readthedocs/core/symlink.py
@@ -386,10 +386,11 @@ class Symlink:
 
 
 class PublicSymlinkBase(Symlink):
-    CNAME_ROOT = os.path.join(settings.SITE_ROOT, 'public_cname_root')
-    WEB_ROOT = os.path.join(settings.SITE_ROOT, 'public_web_root')
+    public_base = getattr(settings, 'PUBLIC_BASE', None) or settings.SITE_ROOT
+    CNAME_ROOT = os.path.join(public_base, 'public_cname_root')
+    WEB_ROOT = os.path.join(public_base, 'public_web_root')
     PROJECT_CNAME_ROOT = os.path.join(
-        settings.SITE_ROOT,
+        public_base,
         'public_cname_project',
     )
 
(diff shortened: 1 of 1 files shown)


humitos:
Hi @yakky! Thanks for your contribution. We are moving forward to not using symlinks anymore. They had give us a lot of problems. Instead, we are start using cloud storage with a small proxy as a first layer that will redirect the URL to the proper file.

All the code around symlinks will be completely removed in the next ~6 months, I'd say. Because of this reason, these changes are not useful for our current roadmap.

If you are still depending on symlinks, it may be worth to start thinking on follow the same path we are following to avoid conflicts in the future.

yakky:
> Hi @yakky! Thanks for your contribution. We are moving forward to not using symlinks anymore. They had give us a lot of problems. Instead, we are start using cloud storage with a small proxy as a first layer that will redirect the URL to the proper file.
> 
> All the code around symlinks will be completely removed in the next ~6 months, I'd say. Because of this reason, these changes are not useful for our current roadmap.
> 
> If you are still depending on symlinks, it may be worth to start thinking on follow the same path we are following to avoid conflicts in the future.

thanks for the reply: I'm looking forward to be able to use cloud storage in a local setup 
AFAIK it's still dependent on `readthedocsext` which is not open. Am I correct? Did I miss updates merged in the meantime which made it generally available?

humitos:
>  AFAIK it's still dependent on `readthedocsext` which is not open. Am I correct? Did I miss updates merged in the meantime which made it generally available?

Right. It depends on `readthedocsext` which is closed source for now. Although, that module is only used for the storage class, which only sets the names of the buckets/containers in the class --you should be able to reproduce it by yourself

yakky:
> > AFAIK it's still dependent on `readthedocsext` which is not open. Am I correct? Did I miss updates merged in the meantime which made it generally available?
> 
> Right. It depends on `readthedocsext` which is closed source for now. Although, that module is only used for the storage class, which only sets the names of the buckets/containers in the class --you should be able to reproduce it by yourself

Thanks, you mean that it just implements this pattern https://django-storages.readthedocs.io/en/latest/backends/azure.html#private-vs-public-access ?

humitos:
> Thanks, you mean that it just implements this pattern [django-storages.readthedocs.io/en/latest/backends/azure.html#private-vs-public-access](https://django-storages.readthedocs.io/en/latest/backends/azure.html#private-vs-public-access) ?

Right!

yakky:
> > Thanks, you mean that it just implements this pattern [django-storages.readthedocs.io/en/latest/backends/azure.html#private-vs-public-access](https://django-storages.readthedocs.io/en/latest/backends/azure.html#private-vs-public-access) ?
> 
> Right!

:bowing_man: 

Thanks for the hints. closing this

PR #6315, opened 2019-10-21
Revert "Adding RTD prefix for docker only in setting.py and all…

humitos:
Reverts readthedocs/readthedocs.org#6040

This is causing some troubles in production. See #6308


ericholscher:
Travis error seems like a random eslint error.

ericholscher:
Travis error seems like a random eslint error, and this doesn't change any JS files, so I think it's probably fine to merge.

PR #6321, opened 2019-10-22
Use django-cacheops to speed up serving docs and support DB going down

humitos:
`django-cacheops` allows us to cache querysets into a redis backend in a
granular way. We can define which models we want to cache and we can
specify which instances (el proxito, web, workers, etc) will use this
cache and which of them will invalidate it.

This initial approach configure El Proxito to make usage of this cache
when performing queries and configure Web/Celery/Build instances only
to invalidate the cache when a model changed.

This setup allows us to serve docs faster but also keep serving docs
during `CACHEOPS_TIMEOUT` seconds even when the DB goes down.

Because the way it's configured now (web is not using the cache), when
the DB goes down, the footer API call will fail and the flyout menu
won't be replaced properly. We could change this if we want by
performing Manual caching on that view or by enabling cacheops in Web
instance as well.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -144,6 +144,9 @@ services:
       - storage
     command: ["/usr/local/bin/create-containers.sh"]
 
+  cacheops:
+    image: redis:3.2.7
+
   cache:
     image: redis:3.2.7
 
--- a/docker/settings/proxito.py
+++ b/docker/settings/proxito.py
@@ -4,6 +4,30 @@ from .docker_compose import DockerBaseSettings
 class ProxitoDevSettings(DockerBaseSettings):
     ROOT_URLCONF = 'readthedocs.proxito.urls'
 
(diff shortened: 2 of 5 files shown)


humitos:
We haven't rolled out completely El Proxito yet and we are not sure we are going to need this. Although, the PR does work and the approach could be interesting.

humitos:
We haven't made a decision about what to do with this yet. El Proxito has been pretty stable so far and we may don't need this work at all. Although, I will leave it open for another month or so, and we can back when stale bot mentioned us again.

humitos:
This may not be needed. Our Proxito is pretty stable now and we are adding CDN in front of it. I'm closing it for now and we can revisit in the future if we think it's still a good idea.

humitos:
I may give it a second change and see if I can implement this PR quickly into our code under a feature flag to compare execution times with the regular implementation. Maybe we can make Proxito fast without too much effort.

PR #6322, opened 2019-10-22
Serve custom 404 pages from El Proxito and storage

humitos:
Continuation from #6306


ericholscher:
Yea, I thought about doing something like this. It feels pretty janky, but is basically the same solution that I was thinking about, and I can't think of a better solution.

docker/nginx/proxito.conf
```diff
@@ -31,11 +31,22 @@ server {
...
         proxy_intercept_errors on;
-        error_page 404 = @fallback;
+        error_page 404 = @notfoundfallback;
 
         add_header X-Served Nginx-Proxito-Sendfile always;
     }
 
+    # Serve 404 pages here
+    location @notfoundfallback {
+        proxy_pass http://proxito:8000/404$request_uri;
```
ericholscher: `proxy_pass http://proxito:8000/_proxito_404_$request_uri;`

PR #6326, opened 2019-10-24
Only copy HTML build output to web servers

humitos:
We do not need to copy localmedia, search, pdf and epub anymore because they are uploaded to Cloud storage and served from there directly via django-storages.

This PR removes this code completely because we don't really need anymore. Although, this may still needed for local development while we still use `readthedocs.builds.syncers.LocalSyncer` (won't be needed anymore if we decide to use #6295, though)

Another possibility here is to modify [`readthedocs.builds.syncers.RemotePuller`](https://github.com/readthedocs/readthedocs.org/blob/122a5e8aac8893c782fac1f6ab0e6dcc0c31d811/readthedocs/builds/syncers.py#L142) (used in production) to skip these paths. This will keep `LocalSyncer` to continue working on development.

Our corporate site is going to start using Cloud storage soon, so this PR also makes sense in that context as well. Although, that should happen in Corporate before merging this PR here.


agjohnson:
There are some test failures here to address before review maybe.

So far the change looks good! I won't sign off on it just yet as I think @davidfischer can provide the best feedback around storage and our asset syncing. I'm excited to get closer to not having web disks around though.

agjohnson:
Wouldn't that break file syncing on the commercial site, where we are not yet serving files from remote storage?

humitos:
> Wouldn't that break file syncing on the commercial site, where we are not yet serving files from remote storage?

I thought that we were serving files from remote storage if the `RTD_BUILD_MEDIA_STORAGE` was defined. Although, I just saw this conditional https://github.com/readthedocs/readthedocs.org/blob/f41a84db99b3bfc8ff4fde7b5ccec009ab434c2e/readthedocs/projects/views/public.py#L304

I think we can remove that condition since this URL (`storage.url(..)`) will contain the `AccessKeyId` with an expiration for private buckets:

https://github.com/readthedocs/readthedocs.org/blob/f41a84db99b3bfc8ff4fde7b5ccec009ab434c2e/readthedocs/projects/views/public.py#L311-L312

So, I think there is no need to keep serving files from disk in corporate.

humitos:
> There are more things that can be removed. Actually, all parameters that are not `html` can be removed from those functions.

We can't remove them all because they are used to call an API endpoint that set `has_*` fields. If you saw that I forget to remove some in particular, please point them out to me.

readthedocs/projects/tasks.py
```diff
@@ -1207,61 +1200,6 @@ def move_files(
         target = version.project.rtd_build_path(version.slug)
         Syncer.copy(from_path, target, host=hostname)
 
-    if search:
```
davidfischer: Overall, I'm fine removing this (large) block of code although I have no idea if there are any private installs using it. We probably want to make that clear in the changelog. 

Eventually however, it's likely that the entire `move_files` function can be removed as everything will be done with storage. I like this solution because it doesn't treat HTML as a special case separate from PDFs or ePubs which has been the source of some problems. We will also then be able to remove the syncers entirely.

humitos:
> PDFs and ePubs **are currently** served from disk but we could change that and that was my eventual plan.

Are you talking about corporate, community or both? My guess is that we are serving PDFs and ePubs in community from storage and in corporate from disk.

humitos:
Some good points were raised here:

- [ ] are we OK serving artifacts (pdf, epub, htmlzip) from S3 with `AccessKey` and `Expiration` parameters?
- [ ] we need to copy *all* the artifacts from media to S3 _before_ merging/deploying this PR into corporate
- [ ] make very explicit in the changelog the backward incompatibility that we are adding here
- [ ] development environment needs to adopt the change (`LocalSyncer` will stop working)

I'm :+1: going in this direction.

humitos:
Hrm... Another idea, maybe easier and that does not need code changes, could be to use the `NullSyncer` in community once El Proxito is deployed. That way, nothing will be copied to webs.

humitos:
> There are some test failures here to address before review maybe.

@agjohnson I updated the tests and all of them should be passing now.

ericholscher:
Of note, I'd like to hold off merging this until after the deploy when we start serving via proxito. I like having a good deprecation path, instead of turning off the old thing in the same deploy when we turn on the new thing.

Especially in this case, once this code is deployed, it's going to be almost impossible to revert, because we won't have the files on disk, and will have to try and copy them back down from cloud hosting.

humitos:
> This looks good to me once we have downloads serving properly from .com +1

This was deployed yesterday and it's working.

We can plan to merge this PR and deploy on next release.

humitos:
From our talk in chat:

> probably worth holding off, and doing it all in the same release as we remove the HTML syncing, and noting it in the changelog
> it will use the storage backends
> and sync properly
> it will only break when the builds & webs are different filesystems

agjohnson:
To clarify where we are with this PR: the worry here is about backwards compatibility and we will be timing this at the same time as removing HTML syncing, to make one tidy backwards incompatible release.

humitos:
We are probably closed to serve all the docs via El Proxito and remove all this code completely at this point. I'm not sure if it worth to keep this PR open or even merge it before #6535

ericholscher:
I think we can close this, and take a larger approach addressing a full removal of this code at some point.

PR #6344, opened 2019-11-02
Update install docs

agjohnson:
I walked through these docs again and hit more issues on new Ubuntu release
(SQLite and Python 3.7)

* Clean up copy on install docs
* Notes Python 3.6 is required, no 3.7+
* Convert literal blocks to code console blocks
* Add directions for using PostgreSQL, guide users away from SQLite
* Drop a block at the end that didn't sound like it belonged in the install
docs

Requires Sphinx 2+ and a yet-to-be-release sphinx-rtd-theme. This is mostly
to deal with the code console prefix, our theme will make these not part of
the selection.

Refs readthedocs/sphinx_rtd_theme#845
Requires a release of the sphinx theme

### DIFF
--- a/docs/development/install.rst
+++ b/docs/development/install.rst
@@ -2,11 +2,10 @@ Installation
 ============
 
 .. meta::
-   :description lang=en: Install a local instance of Read the Docs on your own servers with our step by step guide.
-
+   :description lang=en: Install a local development instance of Read the Docs with our step by step guide.
 
 Here is a step by step guide on how to install Read the Docs.
-It will get you to a point of having a local running instance.
+It will configure a local instance that can build and host documentation.
 
 Requirements
 ------------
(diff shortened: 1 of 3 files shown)


agjohnson:
This is not a core team priority, but noting these issues as I've hit them a few times, following this doc myself. I'm blocking this on sphinx theme release, but copy is reviewable for now. I'm not 100% I want docker instructions in these docs, but it's hidden by a tab to start and Postgres should definitely be our prescribed default database.

Code console looks like this on sphinx 2.2, and allows for prompt line and following STDOUT whenever we need it:
image

docs/development/install.rst
```diff
@@ -33,47 +31,143 @@ Additionally Read the Docs depends on:
...
+
+        You'll need to install the ``psycopg2`` Python package to allow Django to
+        use the database:
+
+        .. code:: console
+
+            $ python -m pip install psycopg2
+
+        This requires the PostgreSQL client libraries to be installed on your
+        system. If you do not have these, you can instead use:
```
humitos, line 161: I'd remove this and just go with the `-binary`

docs/development/install.rst
```diff
@@ -90,37 +184,51 @@ you need these libraries.
 Get and run Read the Docs
 -------------------------
 
-Clone the repository somewhere on your disk and enter to the repository::
+Clone the repository somewhere on your disk and enter to the repository:
+
+.. code:: console
```
humitos, line 189: Using ``.. prompt:: bash`` avoids adding `$` to the lines (adds it automatically) and make them un-selectable.

ericholscher:
@agjohnson  would be worth getting this merged or closed

docs/development/install.rst
agjohnson: The latest version of the theme should do this. `code:: prompt` did not give the same results when I tried.

agjohnson:
This is still low priority for me. No one reviewed this for several months, so the information is now stale and likely is outdated by docker install procedures.

PR #6365, opened 2019-11-09
rename docker settings

krptic07:
I have added the prefix RTD to the docker settings


krptic07:
My tests are failing, i just want to ask if the docker settings needed to be renamed?

PR #6370, opened 2019-11-12
Pin the node dependencies with a package-lock

davidfischer:
- Previously, we were ignoring this file leading to inconsistency
- This will ensure everyone uses the same version


humitos:
I was able to build the assets without any problem by pulling down this branch and running:

```
rm -rf node_modules
npm install
npm run build
```

I'm using `node` 8.11.2, just in case.

humitos:
We found that running this with a 6.6+ version of `npm` it modifies `package-lock.json`. See https://stackoverflow.com/questions/54411377/package-lock-json-file-package-with-optional-true

PR #6387, opened 2019-11-18
Redirect index files in proxito instead of serving

ericholscher:
This lets us just return a redirect if the file exists. This matches the
existing nginx behavior, except it actually appends the index filename
instead of just a slash.

I'm not sure which is better, but this is more explicit which feels good.


humitos:
We had a discussion on Slack. The first approach had a problem that users will see the `index.html` in the URL when they weren't hitting an URL with a trailing slash --which would be a problem for people using `dirhtml` builder.

So, we decided to change this behavior to redirect to the same URL with a trailing slash if the `index.html` file exists. This redirect will be caught again by El Proxito and return the `X-Accel` header that will be include the full path but the user won't see the `index.html` in the URL.

PR #6391, opened 2019-11-19
Release 3.10.0

ericholscher:


humitos:
I think the package lock file should not change. Last deploy, we realize that we need to use `npm ci` install of `npm install`.

ericholscher:
Yea, the `package-lock.json` file was updated when I ran `npm install npm`, because the version of NPM I had didn't have `npm ci`. The beauty of software.

PR #6394, opened 2019-11-19
Fix CSS and how we were handling html in automation rule UI

agjohnson:
Remove hard coded values in HTML for font awesome and override the
values with CSS instead.

The problem with how the underlying PR is written is that normally all of this is done with CSS, the values are not hard coded in the HTML. We can't override this on commercial styling as easy when we hardcode the value into HTML. This is a better example of how to do CSS, classes on the module and module elements, and how to add a font awesome icon to a button.

### DIFF
--- a/readthedocs/templates/builds/versionautomationrule_list.html
+++ b/readthedocs/templates/builds/versionautomationrule_list.html
@@ -1,8 +1,13 @@
 {% extends "projects/project_edit_base.html" %}
 
 {% load i18n %}
+{% load static %}
 
 {% block title %}{% trans "Automation Rules" %}{% endblock %}
+
+{% block extra_links %}
+  <link rel="stylesheet" type="text/css" href="{% static "projects/css/admin.css" %}" />
+{% endblock %}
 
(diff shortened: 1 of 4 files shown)


readthedocs/templates/builds/versionautomationrule_list.html
```diff
@@ -55,7 +62,9 @@
                 <li class="automation-rule">
                   <form method="post" action="{% url 'projects_automation_rule_move' project_slug=project.slug automation_rule_pk=rule.pk steps=1 %}">
                     {% csrf_token %}
-                    <input aria-label="{% trans 'Move down' %}" title="{% trans 'Move down' %}" type="submit" value="&#xf063">
+                    <button aria-label="{% trans 'Move down' %}" title="{% trans 'Move down' %}" class="automation-rule-down">
+                      <span>{% trans 'Move up' %}</span>
```
stsewd:
```suggestion
                      <span>{% trans 'Move down' %}</span>
```

agjohnson:
image

Updated the fixes for CSS, looks better now.

PR #6397, opened 2019-11-20
Allow to extend El Proxito views from commercial

humitos:
This is useful in commercial site to override `allowed_user` method to the required auth checks and decide if the file should be served to the user or not.

In the community site, this method should always return `True` since all projects/versions are public.

This is all the code changes required in the community site. Feel free to suggest a better pattern or better names for the methods.

This PR is based on #6396


humitos:
Yes. I'm not super happy with this design. Although, I didn't come up with a better one. As you said, the can find out a better solution later while we start testing/developing what we need in commercial. The change is not too big even if we want to change the direction later.

PR #6398, opened 2019-11-21
Perform redirects at DB level

humitos:
These queries a little more complex but they allow us to rely on the database to perform the filter and check for all the redirect in a single query than fetching all the instances of the Redirect objects and iterating over them.

This will perform only one query to the DB and the result will be the redirect we need to apply or no redirects at all.


readthedocs/redirects/utils.py
```diff
@@ -84,7 +84,7 @@ def get_redirect_response(request, full_path):
         language, version_slug, path = language_and_version_from_path(path)
 
     path, http_status = project.redirects.get_redirect_path_with_status(
-        path=path, language=language, version_slug=version_slug
+        path=path, full_path=full_path, language=language, version_slug=version_slug
```
humitos: I don't know why we were passing this `full_path` already. We are re-calculating it for one of the cases at `redirect_exact` method.

Passing it at this point, allows us to use it as a value to compare at db level instead from Python itself.

readthedocs/redirects/querysets.py
```diff
@@ -25,8 +25,55 @@ def api(self, user=None, detail=True):
...
+            path__startswith=F('from_url'),
+        )
+        page = Q(
+            redirect_type='page',
+            path__iexact=F('from_url'),
+        )
+        exact = (
+            Q(
+                redirect_type='exact',
+                from_url__endswith='$rest',  # Python implementation does "in"
```
ericholscher: This doesn't seem like enough of a query. It isn't actually testing the URL at all is it, just confirming it ends with `$rest`? It seems to also need a `path__startswith=F('from_url')` or something?

humitos: hrm... right. I'm not sure how to re-write this logic from Python into the query:

```
match = self.from_url.split('$rest')[0]
if full_path.startswith(match):
```

> It seems to also need a `path__startswith=F('from_url')` or something?

We would need to perform an operation over `from_url` field to split it first. I will try to find out this.

humitos: This function can help us https://docs.djangoproject.com/en/2.2/ref/models/database-functions/#replace but it's on Django 2.1

ericholscher: We should probably be able to combine a startswith and endswith, perhaps with an annotate that removes $rest or something. Seems complicated tho.

ericholscher: We can probably do it with another queryset tho, with a filter on `$rest` and then a `startswith`.

humitos: I pushed a possible solution using `Replace` from 2.2 --I'm not sure why it does not fully work yet.

humitos:
@ericholscher I update the PR and all tests are passing now. The main change was `output_field` to be `IntegerField` for the length. Also, I bugfix the typo for `htmldir` query.

humitos:
If we go in this direction, we will need to update the docs at https://docs.readthedocs.io/en/latest/user-defined-redirects.html where it mentions that redirects are only triggered on 404.

humitos:
> This looks good, but I do wonder about the performance of the queries. It should almost certainly be faster and cleaner than the current implementation tho :)

Even if they are faster and cleaner, one difference is that we are hitting this code for _all_ the files (css, js, html, etc) which could decrease the performance.

ericholscher:
> Even if they are faster and cleaner, one difference is that we are hitting this code for all the files (css, js, html, etc) which could decrease the performance.

Yea, that's a good point. We should probably filter those out and not do any of the redirect logic on them.

readthedocs/proxito/views/mixins.py
```diff
@@ -79,3 +84,47 @@ def _serve_401(self, request, project):
...
+    def get_redirect(self, project, lang_slug, version_slug, filename, full_path):
+        """
+        Check for a redirect for this project that matches ``full_path``.
+
+        :returns: the path to redirect the request and its status code
+        :rtype: tuple
+        """
+        redirect_path, http_status = project.redirects.get_redirect_path_with_status(
+            language=lang_slug,
+            version_slug=version_slug,
```
ericholscher: I'm not quite sure why we need the `lang_slug` and `version_slug` here. It feels a bit weird. I looked through the code and they're used a couple places, but they don't seem necessary. I wonder if we can just remove it and simplify the redirects code accordingly.

humitos: I didn't take a deep look at this on purpose. I didn't want to refactor all this without being sure that this is the direction we want to move forward but besides, because we are changing the whole logic of redirects and I felt it was too much. I'm opening an issue to track this refactor.

PR #6404, opened 2019-11-25
Serve non-html files from nginx (X-Accel-Redirect)

humitos:
Remove the "if" that checks if we are in community or commercial and always serve the files from storage (S3 on commercial). The URL generated will have an AccessKey and Expires arguments.

EDIT: this PR also returns a response with `X-Accel-Redirect` to serve the file from NGINX instead of redirecting the user to the container in BlobStorage (community). This allows us to use the `Content-Disposition` header and return the filename with the version slug on it.

Related to #6326 --we can probably merge both together.


humitos:
This change will bring a possible refactor of `Project.get_production_media_path` (probably remove it completely) and make more usage of `Project.get_storage_path` instead.

readthedocs/projects/views/public.py
```diff
@@ -291,46 +294,22 @@ def project_download_media(request, project_slug, type_, version_slug):
...
+
+    media_path = os.path.join(
+        settings.MEDIA_URL,
+        type_,
         project_slug,
         version_slug,
-        path.split('.')[-1],
+        '%s.%s' % (project_slug, type_.replace('htmlzip', 'zip')),
     )
-    response['Content-Disposition'] = 'filename=%s' % filename
```
ericholscher: I think this is actually the logic we want going forward. It lets us specify the actual filename of the download, instead of just linking out to a specific page. Linking back to the media server feels like going backwards, no?

humitos: I think we can't use a `Content-Disposition` header in a 302 redirect.

If we want to keep this header, I think we will need to proxy the request. On community, it seems we lost that header already when redirecting to Blob storage:

```
$ curl -IL https://readthedocs.org/projects/sphinx-notfound-page/downloads/epub/latest/
HTTP/2 302 
date: Mon, 25 Nov 2019 10:47:35 GMT
content-type: text/html; charset=utf-8
set-cookie: __cfduid=d8ba30f4746be12229d35f371d725f28d1574678854; expires=Wed, 25-Dec-19 10:47:34 GMT; path=/; domain=.readthedocs.org; HttpOnly
content-language: en
location: https://buildmedia.readthedocs.org/media/epub/sphinx-notfound-page/latest/sphinx-notfound-page.epub
vary: Accept-Language, Cookie
x-frame-options: DENY
x-content-type-options: nosniff
x-xss-protection: 1; mode=block
strict-transport-security: max-age=31536000;
x-deity: web01
cf-cache-status: DYNAMIC
expect-ct: max-age=604800, report-uri="https://report-uri.cloudflare.com/cdn-cgi/beacon/expect-ct"
server: cloudflare
cf-ray: 53b30b9a8aa4ee79-CDG

HTTP/1.1 200 OK
Content-Length: 44604
Content-Type: application/epub+zip
Last-Modified: Wed, 16 Oct 2019 15:25:34 GMT
ETag: 0x8D7524D16C2A75E
Vary: Origin
Server: nginx
x-ms-request-id: 2674898d-001e-0014-087d-a39e5f000000
x-ms-version: 2009-09-19
x-ms-lease-status: unlocked
x-ms-blob-type: BlockBlob
X-Deity: web03
Date: Mon, 25 Nov 2019 10:47:35 GMT

$ 
```

readthedocs/projects/views/public.py
```diff
@@ -266,71 +267,50 @@ def project_downloads(request, project_slug):
...
-            settings.MEDIA_URL,
-            type_,
-            project_slug,
-            version_slug,
-            '%s.%s' % (project_slug, type_.replace('htmlzip', 'zip')),
+        # URL without scheme and domain to perform an NGINX internal redirect
+        url = storage.url(storage_path)
+        url = urlparse(url)._replace(scheme='', netloc='').geturl()
+
+        return self._serve_media_nginx(
```
ericholscher: I think we probably want to just add a `media` flag to `_serve_docs`, and then it can handle the filename logic. I think we still want it to serve via Python when we have `PYTHON_MEDIA` on, so we should really be calling `_serve_docs` at the top-level, instead of only nginx.

humitos: I added a `download` argument to `_serve*` views. I needed to pass the `version_slug` all around these views to use them when creating the filename :(

readthedocs/proxito/views/serve.py
```diff
@@ -97,7 +97,7 @@ def get(self,
         if path[-1] == '/':
             path += 'index.html'
 
-        return self._serve_docs(request, final_project=final_project, path=path)
+        return self._serve_docs(request, final_project, version_slug, path=path)
```
ericholscher: I like using kwargs everywhere, if possible. I think it reduces issues downstream.

readthedocs/proxito/views/mixins.py
```diff
@@ -15,7 +16,7 @@ class ServeDocsMixin:
 
     """Class implementing all the logic to serve a document."""
 
-    def _serve_docs(self, request, final_project, path):
+    def _serve_docs(self, request, final_project, version_slug, path, download=False):
```
ericholscher: I wonder if we should default `version_slug` to None, since it isn't needed in the normal case to serve a file. That isn't a great solution either though, since it's a bit confusing to call it. Not really sure what is best.

ericholscher: The other option I guess is a top-level `_serve_media` that is similar to `_serve_docs`, which calls into the lower level functions. That way each function could keep only the kwargs it requires.

readthedocs/proxito/views/mixins.py
```diff
@@ -77,6 +98,11 @@ def _serve_docs_nginx(self, request, final_project, path):
         x_accel_redirect = iri_to_uri(path)
         response['X-Accel-Redirect'] = x_accel_redirect
 
+        if download:
+            filename_ext = urlparse(path).path.split('.')[-1]
```
davidfischer:
```suggestion
            filename_ext = urlparse(path).path.rsplit('.', 1)[-1]
```

This will probably never matter, but it's good to be defensive

PR #6408, opened 2019-11-26
Use ServeDocsPermissionsMixin to isolate this methods and be more clear

humitos:
Small refactor to make more clear where these method come from.


readthedocs/proxito/views/mixins.py
```diff
@@ -154,3 +154,38 @@ def get_redirect_response(self, request, redirect_path, http_status):
...
+        """
+        Return response for an unauthed hit at a documentation's version.
+
+        When a user does not have access to read documentation, we need to
+        return a response showing the proper message. This method produces that
+        reponse.
+
+        :rtype: django.http.HttpResponse
+        """
+        return self._serve_401(request, project)
```
ericholscher: I don't quite understand what this is meant to be doing. Are we using these additional methods, or are they just meant to be overwritten?

humitos: We are overriding these methods from corporate. They are meant to be overwritten.

This PR just move them outside the main class and put them into a mixin with better docstrings. Which is easier to understand where they come from. That's all. 

I could move the docstrings to the Base class, otherwise.

ericholscher: @humitos I think I ended up doing most of this PR in https://github.com/readthedocs/readthedocs.org/pull/6572 -- we should get this merged and combined with that one 👍

humitos: @ericholscher see https://github.com/readthedocs/readthedocs.org/pull/6572/files#r371770894

I'm fine with either. I think defaulting to `False` and force re-define it, it's safer, though.

ericholscher: Yea, it should default False 👍

readthedocs/proxito/views/mixins.py
```diff
@@ -175,3 +172,38 @@ def get_redirect_response(self, request, redirect_path, proxito_path, http_statu
...
```
ericholscher: This `_serve_401` method doesn't exist on this class. It shouldn't be called if we aren't implementing it.

PR #6416, opened 2019-11-27
Add AuthenticationMiddleware to El Proxito tests

humitos:
This middleware is required to run the tests in Corporate because the `allowed_user` method uses `request.user` to make the decision.

I'm not sure if there is a better way to do this by overriding one of this setting in the corporate site. We have been skipping these kind of tests, but I think El Proxito is important in corporate and it's better to run these tests as well.


humitos:
This is also failing because

```
django.urls.exceptions.NoReverseMatch: Reverse for 'features' not found. 'features' is not a valid view function or pattern name.
```

We may want to use a variable for the urlconf as we do in other test cases. I need to check this more in deep.

PR #6417, opened 2019-11-28
Add more supported field to APIv3 docs

### LINKED ISSUE #6385: Expand APIv3 docs showing all fields on POST/PATCH/PUT endpoints
There are some endpoints in our APIv3 documentation that do not show all the possible fields that we can send to the server. As a example, [Project update](https://docs.readthedocs.io/en/latest/api/v3.html#project-update) only shows 2 when there are a lot more that you can send.

humitos:
Only "Project Update" is updated. I didn't find more cases where we need to expand our docs to show all the available fields.

Closes #6385

### DIFF
--- a/docs/api/v3.rst
+++ b/docs/api/v3.rst
@@ -234,7 +234,18 @@ Project update
 
         {
             "name": "New name for the project",
-            "default_version": "v0.27.0"
+            "repository": {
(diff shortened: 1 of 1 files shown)


humitos:
No. APIv3 does not support that at creation time. Importing a project manually does not allow those fields either.

They may be useful for the APIv3, though. We may want to think a little more about this and consider to add them.

PR #6418, opened 2019-11-28
inv tasks to use when developing with docker

humitos:
This PR adds some `invoke` tasks to help developing with Docker and avoid writing a lot of `docker-compose` commands over and over again.

These tasks can be called like:

```bash
inv docker.up --no-search

inv docker.attach web

inv docker.restart proxito

inv docker.shell --container=proxito --running

inv docker.manage makemigrations

inv docker.manage shell_plus

inv docker.test --arguments='-e py36 -- -k "not automation_rule" --lf'
```

For example, `restart` helps us by restarting `nginx` automatically when it's needed.

Besides, this PR renames the `docker/` directory to `dockerfiles/` because we have installed a Python package named `docker` and it produces some conflicts. So, better to avoid this.


humitos:
I just added `--init` (default `False`) to the `inv docker.up` command so we can run the initial required commands the first time we create the all the containers/volumes (migrate, collectstatic, etc)

readthedocs/settings/docker_compose.py
```diff
@@ -24,7 +24,7 @@ class DockerBaseSettings(CommunityDevSettings):
     MULTIPLE_BUILD_SERVERS = ['build']
 
     # Enable auto syncing elasticsearch documents
-    ELASTICSEARCH_DSL_AUTOSYNC = True
+    ELASTICSEARCH_DSL_AUTOSYNC = True if 'SEARCH' in os.environ else False
```
ericholscher: This also is much nicer 👍

PR #6428, opened 2019-12-03
Use media availability instead of querying the filesystem

davidfischer:
This change allows us to query the database instead of the file system or remote storage to get whether a version has a download of a specific type.

Currently, we rely on the presence of a file existing at a certain filesystem or storage path to determine whether media/downloads are available. As we move toward storage being an abstraction on top of the filesystem where it is possibly remote (eg. Cloud storage) we don't want to have to query remote storage to know whether a version has a PDF.

Instead, we started caching whether a version has builds of a specific type (eg. a PDF, ePub) and storing it on the `Version` model. After we started storing that data, a script was run to store the results for all previously built versions.

This builds on work in https://github.com/readthedocs/readthedocs.org/pull/6278

## Future work

There's some additional work I didn't add into this PR to keep it minimal. However, we could add this cleanup if desired:

* Remove the `Project.get_downloads` method which gets project downloads for the default version. As far as I can tell, this is not used.
* Remove the `Project.has_pdf`, `Project.has_epub`, and `Project.has_htmlzip` methods. These do not appear to be used anymore. Alternatively, these could be changed to get the version object and query it instead of hitting the filesystem/storage. I think it really depends on whether we want to maintain this API.

### DIFF
--- a/readthedocs/rtd_tests/tests/test_footer.py
+++ b/readthedocs/rtd_tests/tests/test_footer.py
@@ -12,7 +12,6 @@ from readthedocs.builds.constants import BRANCH, LATEST, TAG
 from readthedocs.builds.models import Version
 from readthedocs.core.middleware import FooterNoSessionMiddleware
 from readthedocs.projects.models import Project
-from readthedocs.rtd_tests.mocks.paths import fake_paths_by_regex
 
 
 class Testmaker(APITestCase):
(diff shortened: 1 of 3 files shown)


readthedocs/rtd_tests/tests/test_footer.py
```diff
@@ -12,7 +12,6 @@
 from readthedocs.builds.models import Version
 from readthedocs.core.middleware import FooterNoSessionMiddleware
 from readthedocs.projects.models import Project
-from readthedocs.rtd_tests.mocks.paths import fake_paths_by_regex
```
humitos: :open_mouth:

PR #6434, opened 2019-12-04
Serve external version through El Proxito

humitos:
Allow El Proxito to serve `EXTERNAL` versions.

Before this PR, we were serving external version just by using an NGINX `proxy_pass` rule pointing to the Blob storage. This new behavior is cleaner but also required to serve these versions on corporate, since they require auth and El Proxito can handle it properly.

### Steps to test this PR

Import `readthedocs/test-builds` into your local instance:

```bash
python manage.py import_project_from_live test-builds
```
Then, open a Django shell and create an external version pointing to PR 13 in that repository:

```python
project = Project.objects.get(slug='test-builds')
version = Version.objects.create(project=project, type='external', active=True, verbose_name='13', slug='13', identifier='617672ce68a24c4397ae009a69b4d93850e4b8b6')

```

> You can also setup a OAuth GitHub application, connect your GH account, etc but that's too complicated just for testing accessing to an external version documentation.

Finally, trigger a build for that version:

```python
from readthedocs.core.utils import trigger_build
trigger_build(project, version)
```

At this point, you will be able to hit http://external-builds.community.dev.readthedocs.io/html/test-builds/13/index.html and see the documentation built from a PR properly.


readthedocs/proxito/urls.py
```diff
@@ -86,6 +87,18 @@
     #     name='docs_detail',
     # ),
 
+    # External builds
+    # (requires to be before single version)
+    url(
+        (
+            r'^html/(?P<project_slug>{project_slug})/'
```
humitos: If we want to have a nicer URL here, we need to change this code as well: https://github.com/readthedocs/readthedocs.org/blob/1876191b2b93c6e0d0965fa6503e06e39fe296c9/readthedocs/builds/models.py#L294-L303

ericholscher:
I feel like this needs additional changes to the nginx configs to work on .org, no?

readthedocs/builds/models.py
```diff
@@ -298,8 +299,15 @@ def get_absolute_url(self):
...
-            url = f'{settings.EXTERNAL_VERSION_URL}/html/' \
-                f'{self.project.slug}/{self.slug}/index.html'
+            scheme = 'https' if settings.PUBLIC_DOMAIN_USES_HTTPS else 'http'
+            path = self.project.get_storage_path(
+                type_='html',
+                version_slug=self.slug,
+                version_type=self.type,
+                include_file=False,
+            )
+            url = f'{scheme}://{self.project.subdomain()}/' \
```
ericholscher: I feel like calling functions in f-strings is a bit weird. Probably better to have it be a variable? I don't know though.

readthedocs/proxito/urls.py
```diff
@@ -86,6 +90,21 @@
     #     name='docs_detail',
     # ),
 
+    # External versions
+    # (requires to be before single version)
+    url(
+        (
+            r'^{DOC_PATH_PREFIX}external/html/(?P<project_slug>{project_slug})/'
```
ericholscher: I do wonder about having the `project_slug` here. I know users will be annoyed with how ugly and long these URL's are, and we should be able to get the slug from the Domain. If we do that though, we end up with the issues around subprojects & translations, so maybe it's better to just serve all subprojects & translations as their own project slug on the domain?

humitos: I don't have a good reply for this. I suppose we want to keep discussing around this before merging and going out with this approach.

humitos:
Originally, this PR was serving PR builds at a different domain than the docs domain:

> At this point, you will be able to hit http://external-builds.community.dev.readthedocs.io/html/test-builds/13/index.html and see the documentation built from a PR properly.

Then we decided that it could be good to serve it under the same domain as the docs:

> Also, the URL that it links to now is http://test-builds.community.dev.readthedocs.io/_/external/html/test-builds/13/index.html for me locally +1

Although, we realize that this is a potential security issue on projects with custom domains.

readthedocs/proxito/urls.py
```diff
@@ -86,6 +88,21 @@
     #     name='docs_detail',
     # ),
 
+    # External versions
+    # (RTD_EXTERNAL_VERSION_DOMAIN/html/<project-slug>/<version-slug>/<filename>)
+    # NOTE: requires to be before single version
+    url(
+        (
+            r'^external/html/(?P<project_slug>{project_slug})/'
```
humitos: Since we are serving in a complete different domain that already includes the `external` word, we may be able to remove the `/external/` part from this URL.

This will involve adapting [`Project.get_storage_path`](https://github.com/readthedocs/readthedocs.org/blob/0b41a1fcaf3c00ee0a89bbd3b48abd594f73aa4d/readthedocs/projects/models.py#L559-L562) and creating a new rule for NGINX that appends `/external` to the `proxy_pass` value when serving this domain.

We don't have to do this in PR, though.

humitos: I adapted the code to avoid the `external/` in the user-facing URL but keep it when serving it from El Proxito.

humitos:
> We'll need to also change the nginx configs in prod to support this, but that's good since we can easily deploy and test this.

If I understand correctly, we can safely deploy these changes and they won't have any effect in production until we change the NGINX config file to start using El Proxito. In fact, after deploying this, it should be working in our `test-builds` project which has El Proxito enabled.

So, I think we are good to merge this and safely deploy in the next deploy.

PR #6439, opened 2019-12-05
Better ES settings on docker-compose

humitos:
- Disable check for allocation which is not needed on development
- Set a better name to the node

This avoid the cluster going read-only mode because not "enough" space on disk,

```
search_1     | [2019-12-05T16:38:02,390][WARN ][o.e.c.r.a.DiskThresholdMonitor] [1qOjoH7] flood stage disk watermark [95%] exceeded on [1qOjoH7qRX2hXuMCo80xOg][1qOjoH7][/usr/share/elasticsearch/data/nodes/0] free: 4.2gb[4.7%], all indices on this node will be marked read-only
```

### DIFF
--- a/docker-compose-search.yml
+++ b/docker-compose-search.yml
@@ -31,6 +31,7 @@ services:
     image: docker.elastic.co/elasticsearch/elasticsearch:6.3.2
     environment:
       - discovery.type=single-node
+      - node.name=search
       - cluster.routing.allocation.disk.threshold_enabled=false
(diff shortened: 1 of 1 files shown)


humitos:
If you are already experimenting this problem, this command makes the index to be non read-only again (execute from `web` container):

```bash
# curl -H'Content-Type: application/json' -XPUT search:9200/_all/_settings?pretty -d'
{
    "index": {
        "blocks.read_only": false
    }
}'
```

PR #6453, opened 2019-12-12
Translations: fix plural forms

lexeii:
Hi there,

Note, first plural form in the Russian isn't only for 1, but also for 21, 31,
etc., so hardcoding "one" (or "1") in the first plural form isn't a good
decision, and it prevents to make correct translation on [transifex.com](https://transifex.com).
Please, see https://docs.readthedocs.io/en/latest/development/i18n.html#plurals
for general instructions.

Also, in Russian there are three forms of plural (I wonder where the four
forms came from?), therefore, the rules were corrected in accordance with
the source: http://docs.translatehouse.org/projects/localization-guide/en/latest/l10n/pluralforms.html

Finally, please update the translation files. I find it rather silly to waste
time translating on [transifex.com](https://transifex.com) the strings that are already gone.


lexeii:
I did not make changes to the code, so I can not imagine why the Travis CI tests failed. I apologize for this.

PR #6461, opened 2019-12-16
Add DOCKER_NORELOAD to compose settings

ericholscher:
This stops containers from auto-reloading.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -43,6 +43,7 @@ services:
       - database
       - cache
     environment:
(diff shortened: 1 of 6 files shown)


ericholscher:
I think the DOCKER is because it's for docker-compose. I think it's useful to prefix it, could also do `RTD_NO_RELOAD` but I think this is fine for now.

ericholscher:
I think this works for me for now, as it saves my CPU from burning up, we can make it cleaner as we go. 👍

PR #6466, opened 2019-12-17
Use django storage to build URL returned by El Proxito

humitos:
Isolate fixes from https://github.com/readthedocs/readthedocs.org/pull/6419/commits/6c2b187969bcf7837ed26f5f6511bfd0e7bab106#diff-fd3394188d8793947a3ccbb4924994a1R99 

With this change, we pass the full URL back to NGINX who will proxy it as is to the storage.

This fixes currently 404 in production for test-builds documentation for all versions.


humitos:
I brought some other commits from the other PR that fixes the tests in a good way by using a custom storage backend for tests.

PR #6488, opened 2019-12-23
Show debug toolbar when running docker compose

humitos:
We can't depend on DEBUG + INTERNAL_IPS (default check) because the IP of the docker container is dynamic.

### DIFF
--- a/readthedocs/settings/docker_compose.py
+++ b/readthedocs/settings/docker_compose.py
@@ -58,6 +58,10 @@ class DockerBaseSettings(CommunityDevSettings):
             }
         }
 
+    DEBUG_TOOLBAR_CONFIG = {
+        'SHOW_TOOLBAR_CALLBACK': lambda request: True,
+    }
+
     ACCOUNT_EMAIL_VERIFICATION = "none"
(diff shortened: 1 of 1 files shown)


davidfischer:
This is definitely an option. Another option I've seen in [Django Cookiecutter](https://github.com/pydanny/cookiecutter-django/blob/463c1f83e6896284b6c77b868b42797425a99c38/%7B%7Bcookiecutter.project_slug%7D%7D/config/settings/local.py#L64-L72) for example is to dynamically add the docker IP to `INTERNAL_IPS`:

```python
import socket

hostname, _, ips = socket.gethostbyname_ex(socket.gethostname())
INTERNAL_IPS += [ip[:-1] + "1" for ip in ips]
```

humitos:
Either way works for me. Since this setting it's only on docker compose file, I'm not too worry about exposing things.

On the other hand, the IP from the container is it the IP from where I'm accessing from the host?

humitos:
Merging this. Only affect development and both suggested solutions are good. I see the `return True` just easier to read.

PR #6517, opened 2020-01-13
Core team development standards guide

humitos:
Update the guide to mention docker-compose setup.

I'm interested in @ericholscher review to know if I'm missing some of the commands or workflows that we have been using so far.

On the other hand, I'd like @davidfischer to take a look at this to know if this guide is enough for now to jump into this environment setup.

### DIFF
--- a/docs/development/standards.rst
+++ b/docs/development/standards.rst
@@ -21,7 +21,7 @@ approximates our production environment, in order to spot bugs and logical
 inconsistencies before they make their way to production.
 
 Currently, core team members are migrating to a Docker Compose based
-solution that it's not fully recommended yet for contributing development.
+solution that is not yet recommended for contributing to development.
 
 This solution gives us many features that allows us to have an
 environment closer to production:
--- a/docs/development/standards.rst
+++ b/docs/development/standards.rst
@@ -153,8 +153,10 @@ save some work while typing docker compose commands. This section explains these
        you can run ``inv docker.attach web`` and jump into a pdb session
        (it also works with ipdb and pdb++)
 
-``inv docker.test {arguments}``
-    Runs the test suite inside the container.
+``inv docker.test``
+    Runs all the test suites inside the container.
+
+    * ``--arguments`` will pass arguments to Tox command (e.g. ``--arguments "-e py36 -- -k test_api"``)
 
 ``inv docker.pull``
     Downloads and tags all the Docker images required for builders.
--- a/docs/development/standards.rst
+++ b/docs/development/standards.rst
@@ -88,6 +88,10 @@ After cloning ``readthedocs.org`` repository, you need to
 
 #. build the Docker image for the servers:
 
+   .. warning::
+
+      This command could take a while to finish since it will download several Docker images.
+
    .. prompt:: bash
 
       inv docker.build
--- a/docs/development/standards.rst
+++ b/docs/development/standards.rst
@@ -8,6 +8,11 @@ Read the Docs, it might a be a good idea to follow these guidelines as well. The
 get started with a local installation of Read the Docs and can be used for local
 development by non-core team.
 
+.. warning::
+
+   Take into account that Core team does not offer support on this setup currently.
+
+
 Core team standards
 -------------------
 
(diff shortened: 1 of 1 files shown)


docs/development/standards.rst
```diff
@@ -35,70 +40,186 @@ Docker for builds
...
+
+   .. prompt:: bash
+
+      inv docker.pull
+
+#. start all the containers:
+
+   .. prompt:: bash
+
+      inv.docker up  --init
```
davidfischer:
```suggestion
      inv docker.up  --init  # --init is only needed the first time
```

davidfischer: There should be some explanation here about exactly what this starts and how to access it (eg. http://community.dev.readthedocs.io/).

humitos:
> This looks like a great start.

Cool! I'd be happy if you can follow this steps and let me know if any of them does not work or not make sense in the process.
 
> What is our plan with the [installation](https://docs.readthedocs.io/en/stable/development/install.html) and [search](https://docs.readthedocs.io/en/stable/development/search.html) pages? I think a lot of those could be removed or consolidated.

Yeah, we also have [build environments](https://docs.readthedocs.io/en/stable/development/buildenvironments.html) to update as well. Although, I think we are not ready to suggest users from the community to use this setup. We will need to update them once the Core team has already tested this implementation and we all agree to switch over what we support.

docs/development/standards.rst
```diff
@@ -8,21 +8,26 @@ Read the Docs, it might a be a good idea to follow these guidelines as well. The
...
 approximates our production environment, in order to spot bugs and logical
 inconsistencies before they make their way to production.
 
-Celery runs as a separate process
-    Celery needs to be run as a separate process, and core team will not use
-    in-process task execution via Celery's ``CELERY_ALWAYS_EAGER`` setting. The
-    ``CELERY_ALWAYS_EAGER`` setting doesn't work correctly, and masks bugs that
-    introduce task execution race conditions.
+Currently, core team members are migrating to a Docker Compose based
+solution that it's not fully recommend yet for Contributing development.
```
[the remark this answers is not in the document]
humitos: I'm happy to remove all the other setups that are not docker compose, but I think we agreed on doing this step by step. At the moment, only Eric and I have tested this I think.

docs/development/standards.rst
```diff
@@ -35,70 +40,186 @@ Docker for builds
...
+
+   ``readthedocs-ext`` is closed source for now and it's not available for non-core team members.
+
+Read the Docs extensions can be installed with the same pattern than `Adding a new Python dependency`_
+using the ``common.sh`` script. You can add the following line to the script:
+
+.. code-block:: bash
+
+   # common.sh
+   pip install -e ../readthedocs-ext/
```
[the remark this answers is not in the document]
humitos: That could be a good addition if we can do it in the `Dockerfile`. I would try to avoid running this install every time the containers speed up.

I think we could check if `GITHUB_TOKEN` was passed in the `docker.build` step and install -ext in the Dockerfile directly. It will need some work, though.

davidfischer:
> I'd be happy if you can follow this steps and let me know if any of them does not work or not make sense in the process.

I went through the process in its entirety and that's where my other comments came from. I'm comfortable merging this.

docs/development/standards.rst
```diff
@@ -8,21 +8,26 @@ Read the Docs, it might a be a good idea to follow these guidelines as well. The
...
 approximates our production environment, in order to spot bugs and logical
 inconsistencies before they make their way to production.
 
-Celery runs as a separate process
-    Celery needs to be run as a separate process, and core team will not use
-    in-process task execution via Celery's ``CELERY_ALWAYS_EAGER`` setting. The
-    ``CELERY_ALWAYS_EAGER`` setting doesn't work correctly, and masks bugs that
-    introduce task execution race conditions.
+Currently, core team members are migrating to a Docker Compose based
+solution that it's not fully recommended yet for contributing development.
```
ericholscher:
```suggestion
solution that is not yet recommended for contributing to development.
```

docs/development/standards.rst
```diff
@@ -35,70 +40,194 @@ Docker for builds
...
+After cloning ``readthedocs.org`` repository, you need to
+
+
+#. install `Docker <https://www.docker.com/>`_ following `their installation guide <https://docs.docker.com/install/>`_.
+
+#. install the requirements from ``common`` submodule:
+
+   .. prompt:: bash
+
+      pip install common/dockerfiles/requirements.txt
```
ericholscher:
```suggestion
      pip install -r common/dockerfiles/requirements.txt
```

PR #6535, opened 2020-01-18
Remove code replaced by El Proxito and stateless servers

humitos:
After working on #6326 and realizing that PR will introduce a breaking change, I started playing around by removing more code that we know is not going to be needed once El Proxito is fully rolled out. The idea is to make just one breaking change with all of these remotions together if that is possible. This PR removes these,

- Syncers: all the files generated by the builders are copied directly from there to the Django Storage backend configured
- Some web server tasks: there is no need to remove, sync or copy files anymore since these servers won't have any documentation or state on them
- Symlink code: files are not served from disk anymore. They are served from storage backend via El Proxito
- Subdomain and SingleVersion logic + urls: this is now handled by El Proxito

This PR will introduce a big breaking change and force core team and contributors to use our docker-compose based solution for their development environment.


ericholscher:
> Some web server tasks: there is no need to remove, sync or copy files anymore since these servers won't have any documentation or state on them

We still need to run these tasks, no? Just not on the web servers but in cloud storage?

humitos:
> We still need to run these tasks, no? Just not on the web servers but in cloud storage?

This is done inside the `UpdateDocsTaskStep` that runs in the builder now, at this method https://github.com/readthedocs/readthedocs.org/blob/0bdfeb61b4e277b0850a712fd4f2e27de2a31e30/readthedocs/projects/tasks.py#L784-L796

(there is a call to `storage.sync_directory` inside that method)

humitos:
Let's un-stale this PR for now. We are getting close to be able to remove all this code. We will be pinged back by stale bot in a month or so.

humitos:
I updated the PR with the latest changes. We are getting closer to be able to remove all this code.

ericholscher:
Yea, I'd like to get the symlink code removed this week if possible. It should make our tests a lot less spammy.

ericholscher:
@humitos looks like CI is unhappy with this, so still some work to do.

readthedocs/projects/views/private.py
```diff
@@ -458,14 +453,6 @@ def get_form(self, data=None, files=None, **kwargs):
         kwargs['user'] = self.request.user
         return super().get_form(data, files, **kwargs)
 
-    def form_valid(self, form):
-        broadcast(
-            type='app',
-            task=tasks.symlink_subproject,
```
ericholscher, line 464: I think a lot of the places we were calling symlink code we actually want to bust the cache for the project as well. Something to consider, though maybe we want to _add_ that logic in a new PR after this.

readthedocs/settings/base.py
```diff
@@ -190,8 +185,6 @@ def USE_PROMOS(self):  # noqa
         'django.contrib.auth.middleware.AuthenticationMiddleware',
         'django.contrib.messages.middleware.MessageMiddleware',
         'dj_pagination.middleware.PaginationMiddleware',
-        'readthedocs.core.middleware.SubdomainMiddleware',
-        'readthedocs.core.middleware.SingleVersionMiddleware',
```
ericholscher, line 194: I think we replace this in the proxito code, so probably need to update that as well.

readthedocs/settings/base.py
```diff
@@ -577,7 +561,6 @@ def USE_PROMOS(self):  # noqa
     GRAVATAR_DEFAULT_IMAGE = 'https://assets.readthedocs.org/static/images/silhouette.png'  # NOQA
     OAUTH_AVATAR_USER_DEFAULT_URL = GRAVATAR_DEFAULT_IMAGE
     OAUTH_AVATAR_ORG_DEFAULT_URL = GRAVATAR_DEFAULT_IMAGE
-    RESTRICTEDSESSIONS_AUTHED_ONLY = True
```
ericholscher, line 580: This seems unrelated?

readthedocs/settings/proxito/base.py
```diff
@@ -26,13 +26,9 @@ def MIDDLEWARE(self):  # noqa
         # Use our new middleware instead of the old one
         classes = super().MIDDLEWARE
         classes = list(classes)
-        index = classes.index(
-            'readthedocs.core.middleware.SubdomainMiddleware'
-        )
-        classes[index] = 'readthedocs.proxito.middleware.ProxitoMiddleware'
+        classes.append('readthedocs.proxito.middleware.ProxitoMiddleware')
```
ericholscher, line 29: 👍

readthedocs/projects/views/private.py
humitos: I agree. In _this_ particular case it was called when de-activating a version. The following line (which I didn't add) triggers a `clean_project_resources` which removes the version from the storage. Seems like someone already thought about this :smile:

humitos: It seems I already added some ones where I found the same case: https://github.com/readthedocs/readthedocs.org/pull/6535/files#diff-6e7074d6dee951393c702a7071bc0f1eR232

readthedocs/settings/base.py
humitos: Hrm... I'm not sure how this ended up here. I could re-add it, but we don't use it anymore, though. It's from https://github.com/mxsasha/django-restricted-sessions

humitos:
I worked on this to fix all the tests. I found these:

* some popen count changed because we are not running `ln -ns` anymore
* removed resolver tests using `/docs/` prefix --they were already migrated to `rtd_tests/tests/test_resolver.py` already
* removed some redirects started failing because they use old `/docs/` prefix --they were already migrated to `readthedocs/rtd_tests/tests/test_redirects.py` already

There are only 2 tests that are failing:

* is happening on `hg pull`, and I'm not sure why yet
* private doc serving under `/docs/django-kong/` --which I'm not sure if I should remove it directly

ericholscher:
> private doc serving under /docs/django-kong/ --which I'm not sure if I should remove it directly

Yea, we shouldn't support those URL's anymore, so we don't need tests for them :+1:

readthedocs/urls.py
```diff
@@ -46,8 +44,6 @@
     url(r'^accounts/gold/', include('readthedocs.gold.urls')),
     # For redirects
     url(r'^builds/', include('readthedocs.builds.urls')),
-    # For testing the 404's with DEBUG on.
-    url(r'^404/$', handler404),
```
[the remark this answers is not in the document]
humitos: I didn't remove 500. That is still working.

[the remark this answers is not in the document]
humitos: `/404/` will use the default Django handler and it should keep working normally as any other not found URL.

readthedocs/urls.py
```diff
@@ -22,7 +21,6 @@
 
 admin.autodiscover()
 
-handler404 = server_error_404
```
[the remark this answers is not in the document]
humitos: Regarding the 404, we are using the Django default on dashboard: https://docs.djangoproject.com/en/2.2/ref/views/#django.views.defaults.page_not_found --It keeps showing our custom `404.html` Maze Found.

Our custom handler is not more required because we are not doing any redirect now, just showing the Maze.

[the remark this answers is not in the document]
humitos: I just checked that running the code with DEBUG=False still shows the Maze Found page on any 404 page under the dashboard.

stsewd:
This can be removed too

https://github.com/readthedocs/readthedocs.org/blob/0b4eaddce9fa5ebbe243983c2ee8462df3cadacb/readthedocs/projects/models.py#L582-L582

humitos:
I'm going to merge this so we have 1 week testing it before deploying. Yay! :tada:

PR #6558, opened 2020-01-21
Fix proxito redirects breaking without a /

ericholscher:
This is an artifact of our old redirects code:

```
(Pdb) l
 59         match = re.match(
 60             r'^/(?P<language>%s)/(?P<version_slug>[^/]+)(?P<path>/.*)$' % LANGUAGES_REGEX,
 61             path,
 62         )
 63         import pdb; pdb.set_trace()
 64  ->     if match:
 65             language = match.groupdict()['language']
 66             version_slug = match.groupdict()['version_slug']
 67             path = match.groupdict()['path']
 68             return language, version_slug, path
 69         return None, None, path
(Pdb) match.groups()
('en', 'latest', '/foo.html')
(Pdb)
```

In particular, the regex was capturing the language as `everything up to /`, which then left the `/` always in the filename.

### DIFF
--- a/readthedocs/proxito/views/serve.py
+++ b/readthedocs/proxito/views/serve.py
@@ -183,15 +183,16 @@ class ServeError404Base(ServeRedirectMixin, View):
 
         # Always add a `/` to the filename to match our old logic:
         # https://github.com/readthedocs/readthedocs.org/blob/4b09c7a0ab45cd894c3373f7f07bad7161e4b223/readthedocs/redirects/utils.py#L60
+        redirect_file = filename
         if lang_slug and version_slug:
-            filename = '/' + filename
+            redirect_filename = '/' + filename
 
         # Check and perform redirects on 404 handler
         redirect_path, http_status = self.get_redirect(
             project=final_project,
             lang_slug=lang_slug,
             version_slug=version_slug,
-            filename=filename,
+            filename=redirect_filename,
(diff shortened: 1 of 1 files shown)


readthedocs/proxito/views/serve.py
```diff
@@ -181,13 +181,19 @@ def get(self, request, proxito_path, template_name='404.html'):
             filename=kwargs.get('filename', ''),
         )
 
+        # Always add a `/` to the filename to match our old logic:
+        # https://github.com/readthedocs/readthedocs.org/blob/4b09c7a0ab45cd894c3373f7f07bad7161e4b223/readthedocs/redirects/utils.py#L60
+        redirect_filename = filename
+        if lang_slug and version_slug:
+            redirect_filename = '/' + filename
```
humitos: @ericholscher If I understand correctly, our previous logic is wrong and our current one is correct. Although, since we have been supporting this, we want to keep supporting it --which means that we are making the "bug" to continue working.

If that's the case, wouldn't be better to update our data in the database to prepend a `/` to those "From URL" that are missing it and completely fix this issue instead of still supporting the bug?

ericholscher: all the redirects code works this way, so I think we need to be consistent in what we're passing in. Changing the DB feels really tricky and a larger refactor.

PR #6564, opened 2020-01-21
Fix /en/latest redirects

ericholscher:
This is caused by the URLConf not matching because of the missing last /.
So to fix it we:

* Add a custom URLconf that catches just the lang/version in a URL without a slash
* Pass a variable into the ServeDocs function so it doesn't try and serve it and 404's
* The 404 handlers will properly parse it using this new URLConf, and then properly find the index.html, appending a /


ericholscher:
David +0'd this in chat, so going to hotfix it to get things working in prod.

ericholscher:
This has been applied, and fixed the issue.

readthedocs/proxito/urls.py
```diff
@@ -108,6 +108,19 @@
...
+    # Hack /en/latest so it redirects properly
+    # We don't want to serve the docs here,
+    # because it's at a different level of serving so relative links break.
+    url(
+        (
+            r'^(?:projects/(?P<subproject_slug>{project_slug})/)?'
+            r'(?P<lang_slug>{lang_slug})/'
+            r'(?P<version_slug>{version_slug})$'.format(**pattern_opts)
+        ),
+        ServeDocs.as_view(redirect_root=True),
```
humitos: From the tests, I understand that the "root" is not the problem, but "directory indexing" is. I would suggest to adapt the name of the variable and comment related to match the logic.

readthedocs/proxito/urls.py
```diff
@@ -108,6 +108,19 @@
...
+        name='docs_detail',
```
humitos: Use a different name here,

```suggestion
        name='docs_detail_directory_indexing',
```

readthedocs/proxito/urls.py
```diff
@@ -108,6 +108,19 @@
         name='docs_detail',
     ),
 
+    # Hack /en/latest so it redirects properly
+    # We don't want to serve the docs here,
+    # because it's at a different level of serving so relative links break.
```
humitos: Why we don't make the previous URLConf with optional trailing slash instead of creating a new one?

readthedocs/proxito/urls.py
humitos: Also, if we are short-cutting the flow immediately in the `.get` method, there is no need to use `ServeDocs` here. We can just use `handler_404`/`fast_404` and return 404 directly.

readthedocs/proxito/urls.py
ericholscher: That is what the comment is explaining :)

readthedocs/proxito/urls.py
ericholscher: True 👍

dockerfiles/nginx/proxito.conf
```diff
@@ -46,18 +46,4 @@ server {
         add_header X-Served Proxito-404-Fallback always;
     }
 
-    # Main RTD site for non-doc serving needs
-    location @fallback {
```
humitos: I suppose we can remove this block from -ops as well.

ericholscher: Yea, I believe it should already be removed.

ericholscher: https://github.com/readthedocs/readthedocs-ops/pull/631/files#diff-e0738b03ea130e6f85f02d6e53914c3dL11

PR #6570, opened 2020-01-22
Run old redirect tests via El Proxito

humitos:
Most of the tests are copied from `readthedocs/rtd_tests/tests/test_redirects.py` into `readthedocs/proxito/tests/test_old_redirects.py` and adapted to work similarly to production using El Proxito as doc server.

Some notes:

* NGINX internal 404 redirection is emulated by re-defining the `handler404` with `ServeError404` and forcing `proxito_path` argument to be the full path.
* old crossdomain redirects are not passing on El Proxito (the path is not being parsed properly) and I'm leaving them marked as `xfail` for now.
* solve same issue as https://github.com/readthedocs/readthedocs.org/pull/6558/files#diff-fd3394188d8793947a3ccbb4924994a1R196 for redirects
* New storage backed is used to granularity control which file exists in the storage (together with a helper mixin for tests to update this list via a method)
* First check for dirindexing directories and then check for redirects (keep backward compatibility redirects)


readthedocs/proxito/views/serve.py
```diff
@@ -178,22 +180,18 @@ def get(self, request, proxito_path, template_name='404.html'):
...
-        # Always add a `/` to the filename to match our old logic:
+        # ``filepath`` is the path without ``/<lang>/<version>`` and without
+        # query, starting with a ``/``. This matches our old logic:
         # https://github.com/readthedocs/readthedocs.org/blob/4b09c7a0ab45cd894c3373f7f07bad7161e4b223/readthedocs/redirects/utils.py#L60
-        redirect_filename = filename
-        if lang_slug and version_slug:
-            redirect_filename = '/' + filename
+        # We parse ``filename`` to remove the query from it
+        schema, netloc, path, params, query, fragments = urlparse(filename)
+        filepath = path
```
humitos: I change this name here to make it more clear. I would expect that `filename=configuration.html`, but this is `filepath=/section/configuration.html`.

I think is a little more clear now, but I'm open for suggestions.

readthedocs/proxito/views/serve.py
```diff
@@ -178,22 +180,18 @@ def get(self, request, proxito_path, template_name='404.html'):
...
-            filename=redirect_filename,
-            full_path=proxito_path,
-        )
-        if redirect_path and http_status:
-            return self.get_redirect_response(request, redirect_path, http_status)
+        # we can't check for lang and version here to decide if we need to add
+        # the ``/`` or not because ``/install.html`` is a valid path to use as
+        # redirect and does not include lang and version on it. It should be
+        # fine always adding the ``/`` to the beginning.
+        filepath = '/' + filepath.lstrip('/')
```
humitos: I added a small protection for `filepath` that already starts with `/` here. This is the case of crossdomain redirects (that I couldn't make them work yet, BTW)

readthedocs/proxito/views/serve.py
```diff
@@ -164,10 +164,12 @@ def get(self, request, proxito_path, template_name='404.html'):
         the Docs default page (Maze Found) is rendered by Django and served.
         """
         # pylint: disable=too-many-locals
+        log.info('Executing handler404. proxito_path=%s', proxito_path)
```
ericholscher: This isn't configured as a handler404. We should just say `404 handler`, since that has a specific name that isn't correct in this case.

readthedocs/proxito/tests/test_old_redirects.py
```diff
@@ -0,0 +1,686 @@
...
+
+    def test_redirect_exact_with_rest(self):
+        """
+        Exact redirects can have a ``$rest`` in the ``from_url``.
+
+        Use case: we want to deprecate version ``2.0`` and replace it by
+        ``3.0``. We write an exact redirect from ``/en/2.0/$rest`` to
+        ``/en/3.0/``.
+        """
+        self._storage_exists([
```
ericholscher: I don't understand why this file needs to exist -- does the redirect only get called when the file _doesn't_ exist?

readthedocs/proxito/views/serve.py
ericholscher: we use `filename` for this everywhere, so we should keep it consistent.

readthedocs/proxito/views/serve.py
ericholscher: Wouldn't we want this protection before we parse it with urlparse?

readthedocs/proxito/views/serve.py
humitos: He... it's configured as a `handler404` when running tests. But yeah, in production it's different. I'll update the string.

readthedocs/proxito/tests/test_old_redirects.py
humitos: It seems it's wrong and not needed. I should pasted this by mistake while refactoring.

readthedocs/proxito/views/serve.py
humitos: I'm not sure I have a good answer for that. What would be the case that we are covering?

humitos:
I made some final changes, but this is ready to go to me.

PR #6572, opened 2020-01-23
Use our standard auth mixin for proxito downloads

ericholscher:
This is the same logic as in the Proxito views,
it just extends it to the downloads.

### DIFF
--- a/readthedocs/projects/views/public.py
+++ b/readthedocs/projects/views/public.py
@@ -300,24 +300,30 @@ class ProjectDownloadMediaBase(ServeDocsMixin, View):
                      not the actual Project permissions.
         """
         if self.same_domain_url:
-            resp = self._version_same_domain_url(
+            # It uses the request to get the ``project``. The rest of arguments come
+            # from the URL.
+            final_project, lang_slug, version_slug, filename = _get_project_data_from_request(  # noqa
                 request,
-                type_,
-                lang_slug,
-                version_slug,
-                subproject_slug,
+                project_slug=None,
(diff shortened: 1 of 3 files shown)


readthedocs/projects/views/public.py
```diff
@@ -363,10 +364,11 @@ def _version_same_domain_url(
...
         )
-        version = get_object_or_404(
-            final_project.versions.public(user=request.user),
-            slug=version_slug,
-        )
+
+        if not self.allowed_user(request, final_project, version_slug):
+            return self.get_unauthed_response(request, final_project)
+
+        version = final_project.versions.public(user=request.user).filter(slug=version_slug).first()
```
davidfischer: Don't we need to raise a 404 here if `version == None`?

ericholscher: True -- I will re-add that logic.

readthedocs/proxito/views/mixins.py
```diff
@@ -118,6 +118,9 @@ def _serve_401(self, request, project):
         log.debug('Unauthorized access to %s documentation', project.slug)
         return res
 
+    def allowed_user(self, *args, **kwargs):
+        return True
```
humitos: My idea was to default this to `False` to avoid mistakes if it's not overwritten in the class that inherit from this. Forcing the class that inherit from this to re-define it.

ericholscher: Yea, this worried me a bit, we should fix it in another PR.

PR #6581, opened 2020-01-24
Mount proper shared docker volume

humitos:
I introduce this bug in #6539.


humitos:
Merging to avoid people breaking their local instances.

PR #6590, opened 2020-01-27
Support Ethical Ads server for development

humitos:
These are the steps to make local RTD instance to hit the Ads Server

1. export `HOSTIP` before running the RTD instance
1. create a User in Ethical Ads server
1. create a Token for this User
1. put the token in the `ADSERVER_API_KEY` setting

It would be good if we could load some testing data in the Ads Server (test user with publisher, campaigns, flights, ads, etc)


readthedocs/settings/docker_compose.py
```diff
@@ -24,6 +24,14 @@ class DockerBaseSettings(CommunityDevSettings):
     MULTIPLE_APP_SERVERS = ['web']
     MULTIPLE_BUILD_SERVERS = ['build']
 
+    # https://docs.docker.com/engine/reference/commandline/run/#add-entries-to-container-hosts-file---add-host
+    # export HOSTIP=`ip -4 addr show scope global dev wlp4s0 | grep inet | awk '{print \$2}' | cut -d / -f 1`
+    HOSTIP = os.environ.get('HOSTIP')
```
ericholscher: I'd like to use `host.docker.internal` if we're on a mac. That way there's no configuration needed.

humitos: I pushed a change for this.

PR #6592, opened 2020-01-27
Remove webs from broadcast list

humitos:
Use an empty list so webs are not called to execute any of the following tasks:

- syncing files
- remove directories
- update metadata

This is a different approach we could use for #6326 and/or #6535 while we test how it works.

### DIFF
--- a/readthedocs/settings/docker_compose.py
+++ b/readthedocs/settings/docker_compose.py
@@ -21,7 +21,12 @@ class DockerBaseSettings(CommunityDevSettings):
     SLUMBER_API_HOST = 'http://web:8000'
     RTD_EXTERNAL_VERSION_DOMAIN = 'external-builds.community.dev.readthedocs.io'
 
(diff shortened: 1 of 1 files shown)


ericholscher:
This setting is overridden in prod. We can also just set the FILE_SYNCER setting to the `NullSyncer` for a similar effect in prod.

humitos:
> This setting is overridden in prod. 

Right. This is just a way to share what's the change proposal.

> We can also just set the FILE_SYNCER setting to the `NullSyncer` for a similar effect in prod.

Yes. This is what we are already using in Docker compose setup. Although, tasks are still triggered and symlink code (+ other) is being executed in the web servers.

ericholscher:
I worry this might have other effects, given that we're doing a broadcast to nothing, but we depend on the broadcast to actually complete to run the tasks afterwards, so this feels riskier than just null routing the normal broadcasts.

PR #6596, opened 2020-01-28
Revert "Update celery requirements to its latest version"

agjohnson:
This reverts commit 5939bca415b5fac3ec12ff6fb9557b38b3be84bd.

We noticed a lot of strange issues, including Celery dying without an
error, using the upgraded packages and even 4.4.0. Reverting back to
something that works for now.

### DIFF
--- a/requirements/pip.txt
+++ b/requirements/pip.txt
@@ -30,9 +30,15 @@ pyyaml==5.1.2
 Pygments==2.4.2
 
 # Basic tools
-redis==3.3.11
-kombu==4.6.7
-celery==4.3.0
+# Redis 3.x has an incompatible change and fails
+# https://stackoverflow.com/questions/53331405/django-compress-error-invalid-input-of-type-cachekey
+# https://github.com/sebleier/django-redis-cache/pull/162
+redis==2.10.6  # pyup: ignore
+# Kombu >4.3 requires redis>=3.2
+kombu==4.3.0  # pyup: ignore
+# Celery 4.2 is incompatible with our code
+# when ALWAYS_EAGER = True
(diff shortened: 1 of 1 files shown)


agjohnson:
Hotfixed on relcorp only

ericholscher:
Seemed ambitious that all the issues would be fixed given how many caveats we had in the requirements file :/ Not sure how to move forward with this upgrade though.

benjaoming:
Are there issues tracking this now?

https://github.com/readthedocs/readthedocs.org/pull/6448 and https://github.com/readthedocs/readthedocs.org/issues/4643 are closed.

humitos:
No. We only have this note in the code: https://github.com/readthedocs/readthedocs.org/blob/97124cfd804ab0dc26d2630c718e3efd9c79130e/requirements/pip.txt#L40-L42

We rely a lot on Celery to make Read the Docs to work. Since it's hard to test locally all the possible scenarios we usually try to upgrade Celery only when needed or try to make a deploy that only involves upgrading Celery to be able to monitor it from close.

IIRC, when we deployed the latest version, workers where randomly disconnected or they stopped consuming tasks and start doing nothing. That's why we decided to revert this change.

I'm curious, why you are interested in upgrading Celery, @benjaoming ?

benjaoming:
@humitos I'm running Read the Docs on Python 3.7 (didn't realize it was intended for 3.6). After a lot of setup, I realized that Celery was broken (the classical `async` syntax keyword issue) and now I'm running it with a local patch that upgrades Celery.

I'll let you know if I experience the same hang-ups :)

PR #6629, opened 2020-02-05
Initial attempt to serve PR builds at `readthedocs.build`

ericholscher:
This updates the logic so that we can serve docs from a URL that looks like this:

http://test-builds--13.dev.readthedocs.build/en/13/

We will take the Project & Version from the domain, and then the rest of the URL will be parsed normally.


readthedocs/proxito/views/serve.py
```diff
@@ -45,9 +45,16 @@ def get(self,
     ):  # noqa
         """Take the incoming parsed URL's and figure out what file to serve."""
 
+        # Handle external domain
+        if request.external_domain:
+            self.version_type = EXTERNAL
+            log.warning('Replacing version slug from URL old=%s new=%s',
+                        version_slug, request.host_version_slug)
+            version_slug = request.host_version_slug
```
ericholscher: This logic should go into some kind of helper, just haven't set it up yet.

readthedocs/projects/tasks.py
```diff
@@ -810,6 +810,23 @@ def store_build_artifacts(
         :param pdf: whether to save PDF output
         :param epub: whether to save ePub output
         """
+        # TODO: Remove this logic from `update_app_instances`
+        # It's in both places to make sure it runs.
```
humitos: Good catch!

readthedocs/proxito/middleware.py
```diff
@@ -59,6 +62,15 @@ def map_host_to_project_slug(request):
             request, 'core/dns-404.html', context={'host': host}, status=400
         )
 
+    if external_domain in host:
+        # Serve custom versions on external-host-domain
+        if external_domain_parts == host_parts[1:]:
+            project_slug, version_slug = host_parts[0].split('--', 1)
```
humitos: This will probably fail with projects with `--` on their slug. It's probably and edge case, that we don't want to consider, though. Maybe 3 dashes is better?

humitos: How our code will handle this case?

readthedocs/proxito/views/serve.py
humitos: Yes. This could probably be a method of `ServeDocsMixin`.

readthedocs/proxito/middleware.py
ericholscher: We strip double `--`'s in slugs, so we should be fine.

humitos:
> http://test-builds--13.dev.readthedocs.build/en/13/

we are going to use a different domain for corporate, correct?

readthedocs/proxito/middleware.py
davidfischer: If we really want to, we could use another `.` in the domain. We just have to make sure we get the right wildcard cert (eg. a cert for `*.org.readthedocs.build` is not good for `*.*.org.readthedocs.build`).

readthedocs/proxito/views/serve.py
```diff
@@ -148,7 +148,7 @@ class ServeDocs(SettingsOverrideObject):
     _default_class = ServeDocsBase
 
 
-class ServeError404Base(ServeRedirectMixin, View):
+class ServeError404Base(ServeRedirectMixin, ServeDocsMixin, View):
```
humitos: We are inheriting a lot of functionality (`allow_user`, `_serve_docs`, etc) that it's not required in this view. We are rendering the template from django instead serving a doc.

I think I'd prefer if we can have a small mixin with the method `get_version_from_host` instead, which will make this easier to read.

readthedocs/proxito/views/utils.py
```diff
@@ -49,6 +41,7 @@ def _get_project_data_from_request(
     if current_project.single_version:
         if lang_slug and version_slug:
             filename = os.path.join(lang_slug, version_slug, filename)
+            log.warning('URL looks like versioned, fixing, url=%s', filename)
```
humitos:
```suggestion
            log.warning('URL looks like versioned on a single version project. Making them None url=%s', filename)
```

readthedocs/settings/docker_compose.py
```diff
@@ -19,7 +19,7 @@ class DockerBaseSettings(CommunityDevSettings):
     PUBLIC_API_URL = 'http://community.dev.readthedocs.io'
     RTD_PROXIED_API_URL = PUBLIC_API_URL
     SLUMBER_API_HOST = 'http://web:8000'
-    RTD_EXTERNAL_VERSION_DOMAIN = 'external-builds.community.dev.readthedocs.io'
+    RTD_EXTERNAL_VERSION_DOMAIN = 'dev.readthedocs.build'
```
humitos:
```suggestion
    RTD_EXTERNAL_VERSION_DOMAIN = 'community.dev.readthedocs.build'
```

humitos:
I understand that the URL for production will be like `http://test-builds--13.org.readthedocs.build/en/13/`.

Since we have the canonical project's slug in the host, having a lang in the path is also irrelevant --we won't have a URL like `http://test-builds--13.org.readthedocs.build/ja/13/`. In case of Japanese, it will have another project slug, like `http://test-builds-ja--13.org.readthedocs.build/ja/13/` and we end up with duplicated lang _and_ version in this case.

humitos:
I pulled down the PR and it works properly locally. Although, I'm not sure why it's not required to add a `server_name *.dev.readthedocs.build` in our proxito nginx conf.

readthedocs/core/resolver.py
```diff
@@ -226,21 +230,31 @@ def _get_canonical_project(self, project, projects=None):
             return self._get_canonical_project(next_project, projects)
         return project
 
+    def _get_external_subdomain(self, project, version_slug):
+        """Determine domain for an external version."""
+        project = self._get_canonical_project(project)
```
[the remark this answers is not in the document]
ericholscher: I just copied the existing logic, I think it's to make sure it's the canonical project, if it's called from elsewhere.

readthedocs/projects/models.py
```diff
@@ -1304,6 +1305,7 @@ def get_absolute_url(self):
             project=self.project,
             version_slug=self.version.slug,
             filename=self.path,
+            external=False,
```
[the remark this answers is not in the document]
ericholscher: It defaults to `None`, not False.

readthedocs/proxito/middleware.py
```diff
@@ -59,6 +62,21 @@ def map_host_to_project_slug(request):
             request, 'core/dns-404.html', context={'host': host}, status=400
         )
 
+    if external_domain in host:
+        # Serve custom versions on external-host-domain
+        if external_domain_parts == host_parts[1:]:
+            try:
+                project_slug, version_slug = host_parts[0].split('--', 1)
```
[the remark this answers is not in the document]
ericholscher: Perhaps, but I think it's OK for now.

readthedocs/proxito/middleware.py
```diff
@@ -59,6 +62,21 @@ def map_host_to_project_slug(request):
...
+                request.external_domain = True
+                request.host_version_slug = version_slug
+                log.debug('Proxito External Domain: host=%s', host)
```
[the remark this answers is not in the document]
ericholscher: We don't call it `external versions` other places, I don't quite understand why we should here.

readthedocs/proxito/tests/test_full.py
```diff
@@ -105,15 +105,15 @@ def test_single_version_external_serving(self):
             active=True,
             project=self.project,
         )
-        url = '/html/project/10/awesome.html'
-        host = 'external-builds.dev.readthedocs.io'
+        url = '/awesome.html'
+        host = 'project--10.dev.readthedocs.build'
```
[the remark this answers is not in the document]
ericholscher: Yep.

readthedocs/proxito/views/serve.py
```diff
@@ -64,28 +73,20 @@ def get(self,
...
             'Serving docs: project=%s, subproject=%s, lang_slug=%s, version_slug=%s, filename=%s',
             final_project.slug, subproject_slug, lang_slug, version_slug, filename
         )
 
         # Handle a / redirect when we aren't a single version
         if all([
                 lang_slug is None,
-                version_slug is None,
+                # External domains will always have a version
+                version_slug is None or hasattr(request, 'external_domain'),
```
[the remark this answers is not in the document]
ericholscher: Yea, I think it's just confusing in general. It's a weird edge case that we want to serve a specific version of the HTML but with a URL config that is a `single_version`.

ericholscher:
> I pulled down the PR and it works properly locally. Although, I'm not sure why it's not required to add a server_name *.dev.readthedocs.build in our proxito nginx conf.

Because nginx always has a default server, so it's just handling it properly when it gets hit.

readthedocs/proxito/views/serve.py
humitos: Are you doing this small refactor?

readthedocs/proxito/middleware.py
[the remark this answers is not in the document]
humitos: Because they are external versions. We have `Version.type = 'external'` and and `Version.objects.external`, etc. So, basically all the versions that come from Pull Request we are naming them "External Versions". 

That's my point.

readthedocs/proxito/views/serve.py
ericholscher: I'm not sure that having even more mixins is really a good solution either. I feel that having too many mixins makes the code hard to read, so I'm not convinced that it's better.

ericholscher: In particular, all of the doc-serving code needs `get_version_from_host`, so it would require adding a new mixin to *all* the classes in this function, and I don't see a lot of value in that.

readthedocs/core/resolver.py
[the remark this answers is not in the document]
ericholscher: Good point. I looked and the other call wasn't necessary either, so I've removed both. I just wanted to make sure they were implemented the same way, but it looks like both can be made faster :)

PR #6646, opened 2020-02-11
Skip tasks for project that are marked to be skipped

humitos:
We need a way to not execute the build/sync tasks even after the tasks were triggered and the project marked as `skip=True`.

This code needs to be improved, but it shows where we need to the checks and stop the task.

### DIFF
--- a/readthedocs/projects/tasks.py
+++ b/readthedocs/projects/tasks.py
@@ -250,6 +250,13 @@ class SyncRepositoryTaskStep(SyncRepositoryMixin):
             self.version = self.get_version(version_pk)
             self.project = self.version.project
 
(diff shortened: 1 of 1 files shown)


readthedocs/projects/tasks.py
```diff
@@ -250,6 +250,13 @@ def run(self, version_pk):  # pylint: disable=arguments-differ
             self.version = self.get_version(version_pk)
             self.project = self.version.project
 
+            if not Project.objects.is_active(self.project):
```
humitos: Also, this does not work because we don't have access to the DB here. We probably need to adding this to the API response.

ericholscher:
We solved this in another PR.

PR #6651, opened 2020-02-12
Cache tags -> commits for performance

davidfischer:
This optimization only really improves things when there are 1000s of tags.


davidfischer:
I ran this locally with the [rtd-git-stresstest](https://github.com/readthedocs/rtd-git-stresstest) repository (~35k tags, ~35k commits) and the build time was ~45s. It was timing out before. The build time is actually longer than this amount but some amount of cloning which took a few minutes wasn't included in the timing.

PR #6653, opened 2020-02-13
Default python version per Docker image

humitos:
Instead of always picking the latest minor version when the user just put '2' or '3' in their YAML file, we use a default Python version per Docker image.

This allows us to upgrade our Docker images (`latest` or `stable`) introducing new versions of Python (like `3.8`) without forcing users without specifying a Python version to use the new Python version.

We can easily change the `default_version` in our `latest` or `stable` image when we the new Python version gets more stable and we know we won't affect too many users with the change.

> We can release a `6.0` (plus `testing`) Docker image from current `master` after this gets merged and deployed. We need to create another PR to add this value into the Docker setting, though. See #6654 

Related #6324

### DIFF
--- a/readthedocs/config/config.py
+++ b/readthedocs/config/config.py
@@ -295,7 +295,7 @@ class BuildConfigBase:
             )
         return settings.DOCKER_IMAGE_SETTINGS[build_image]['python']['supported_versions']
 
-    def get_default_python_version_for_image(self, build_image, python):
+    def get_default_python_version_for_image(self, build_image, python_version):
         """
         Return the default Python version for Docker image and Py2 or Py3.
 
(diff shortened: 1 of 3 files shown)


readthedocs/config/config.py
```diff
@@ -296,6 +295,28 @@ def get_valid_python_versions_for_image(self, build_image):
...
 
+    def get_default_python_version_for_image(self, build_image, python):
+        """
+        Return the default Python version for Docker image and Py2 or Py3.
+
+        :param build_image: the Docker image complete name, already validated
+            (``readthedocs/build:4.0``, not just ``4.0``)
+        :type build_image: str
+
+        :param python: major Python version (``2`` or ``3``) to get its default
```
ericholscher: This should probably be `python_version`, otherwise below its confusing when used as `[build_image]['python']['default_version'][python]`

PR #6654, opened 2020-02-13
New docker release (6.0 and testing)

humitos:
- Update our Docker settings to support `6.0` and `testing` images in our YAML file.
- Tag 5.0 as stable
- Tag 6.0 as latest

Edit: this PR includes changes from #6653 so be careful when merging.

### DIFF
--- a/readthedocs/settings/base.py
+++ b/readthedocs/settings/base.py
@@ -424,6 +424,15 @@ class CommunityBaseSettings(Settings):
                 },
             },
         },
+        'readthedocs/build:testing': {
+            'python': {
+                'supported_versions': [2, 2.7, 3, 3.5, 3.6, 3.7, 3.8, 'pypy3.5'],
+                'default_version': {
+                    2: 2.7,
+                    3: 3.7,
+                },
+            },
+        },
     }
 
     # Alias tagged via ``docker tag`` on the build servers
--- a/readthedocs/settings/base.py
+++ b/readthedocs/settings/base.py
@@ -415,7 +415,7 @@ class CommunityBaseSettings(Settings):
                 },
             },
         },
-        'readthedocs/build:6.0rc1': {
+        'readthedocs/build:6.0': {
             'python': {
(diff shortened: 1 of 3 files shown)


readthedocs/settings/base.py
```diff
@@ -369,29 +369,74 @@ def USE_PROMOS(self):  # noqa
...
+                'supported_versions': [2, 2.7, 3, 3.5, 3.6, 3.7, 3.8, 'pypy3.5'],
+                'default_version': {
+                    2: 2.7,
+                    3: 3.7,
+                },
+            },
         },
-        'readthedocs/build:6.0rc1': {
-            'python': {'supported_versions': [2, 2.7, 3, 3.5, 3.6, 3.7, 3.8, 'pypy3.5']},
+        'readthedocs/build:testing': {
```
ericholscher: I think this should be labeled as 7.0, and `testing` used as an alias, like latest and stable. I don't think we want a specific `testing` image, otherwise it's done differently than all our other images.

readthedocs/settings/base.py
```diff
@@ -369,29 +369,74 @@ def USE_PROMOS(self):  # noqa
...
+                },
+            },
         },
     }
 
     # Alias tagged via ``docker tag`` on the build servers
     DOCKER_IMAGE_SETTINGS.update({
-        'readthedocs/build:stable': DOCKER_IMAGE_SETTINGS.get('readthedocs/build:4.0'),
-        'readthedocs/build:latest': DOCKER_IMAGE_SETTINGS.get('readthedocs/build:5.0'),
+        'readthedocs/build:stable': DOCKER_IMAGE_SETTINGS.get('readthedocs/build:5.0'),
```
davidfischer: Wouldn't this change everybody who was pointed at `stable` in their config file to use `readthedocs/build:5.0` when previously they were using `readthedocs/build:4.0`? I'm not super familiar with the backwards compatibility of our docker images but is that what we want?

PR #6669, opened 2020-02-17
Add azurite --loose option

agjohnson:
I was getting 5xx errors from azurite every other load of a static asset
file. I found this issue and added a fix that was mentioned:
https://github.com/Azure/Azurite/issues/337

I can not reccomend this fix and I did not try to spend any time
understanding Azurite.

### DIFF
--- a/docker-compose.override.yml
+++ b/docker-compose.override.yml
@@ -76,7 +76,7 @@ services:
       - "10000:10000"
     networks:
       readthedocs:
-    command: ["azurite-blob", "--silent", "--blobPort", "10000", "--blobHost", "0.0.0.0", "--location", "/data"]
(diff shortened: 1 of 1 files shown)


agjohnson:
Also, to clarify, the `--loose` option did resolve the problem I was seeing. I'm able to consistently reload the assets now. I'd be fine pinning the Azurite image or adding the `--loose` option. I don't think it makes a whole lot of sense to spend time figuring out Azurite as we don't use it anywhere outside local development.

humitos:
What's the hash of your azurite image?

humitos:
I re-downloaded the azurite docker image and it works now :man_shrugging:

agjohnson:
My azurite is:

```
mcr.microsoft.com/azure-storage/azurite         latest              741b0142b740        10 days ago         292MB
```

PR #6673, opened 2020-02-18
Use PUBLIC_DOMAIN_USES_HTTPS for resolver tests

humitos:
Parity with production but also allow to run these tests in corporate.


humitos:
Hrm... it seems the resolver does not respect the setting?

PR #6676, opened 2020-02-19
Allow user to set `build.image: testing` in the config file

humitos:
We have a new `testing` image now that we can recommend to users when they need something extra that it's now in our current image, but will be available in the following image.

This allow us to not have to do work manually (like modifying `container_image` in the db), clearly communicate to the user that they are using something that it's not released yet and also allow them to opt-in/out when they want.

### DIFF
--- a/readthedocs/config/config.py
+++ b/readthedocs/config/config.py
@@ -268,9 +268,9 @@ class BuildConfigBase:
 
         The user can use any of this values in the YAML file. These values are
         the keys of ``DOCKER_IMAGE_SETTINGS`` Django setting (without the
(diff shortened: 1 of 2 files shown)


ericholscher:
Believe @agjohnson had some reservations about letting users set this "forever". I think I'm +0 on it, so that users can opt into a new image without having to pester us to setup a feature flag -- I think it will lead to less support over time than folks who have broken builds because of testing, but not sure.

agjohnson:
Yeah, a bit. I worry users are opting in to an image that solves a problem they are having, where the testing image is much more of a beta test area for us. I'd much rather we increase release cycles and tell the user "use X image when it's released, which won't be long", not "become a beta tester for our docker images".

PR #6680, opened 2020-02-19
Add ability to sort dashboard by modified date

ericholscher:
This only implemented the backend, so you can pass a GET arg to it to have
it sort.

The goal will be to have a proper UI for this after the UI refactor, but I
want this now :)


ericholscher:
This is on the dashboard page, not the version list.

PR #6683, opened 2020-02-20
Disable more loggings in development environment

humitos:


readthedocs/settings/docker_compose.py
```diff
@@ -57,6 +57,15 @@ def LOGGING(self):
                 'handlers': ['null'],
                 'propagate': False,
             },
+            'azure.storage.common._auth': {
```
ericholscher: I have this patched locally, and just did 
```suggestion
            'azure.storage.common': {
```

I think that will solve this in the future.

PR #6689, opened 2020-02-21
Add feature flag to just completely skip sync and symlink operations

agjohnson:
We can control this on a per-project basis. Once we remove symlinking
and Syncers, this can go away entirely.

### DIFF
--- a/readthedocs/projects/models.py
+++ b/readthedocs/projects/models.py
@@ -1505,6 +1505,7 @@ class Feature(models.Model):
     UPDATE_CONDA_STARTUP = 'update_conda_startup'
     CONDA_APPEND_CORE_REQUIREMENTS = 'conda_append_core_requirements'
     ALL_VERSIONS_IN_HTML_CONTEXT = 'all_versions_in_html_context'
+    SKIP_SYNC = 'skip_sync'
 
     FEATURES = (
         (USE_SPHINX_LATEST, _('Use latest version of Sphinx')),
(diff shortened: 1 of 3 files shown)


agjohnson:
I'm not confident I protected all of the sync/symlink calls. I didn't put the same logic in the actual tasks as we shouldn't fire off thousands of tasks that are effectively noop tasks. This did save the webs from being hammered with thousands of symlink tasks for the project however.

davidfischer:
This worked for me in my testing to stop symlinking versions for specific projects. This didn't protect the `symlink_subproject` function but those subprojects would technically be different projects and could be feature flagged themselves. :+1:

readthedocs/projects/models.py
```diff
@@ -1505,6 +1505,7 @@ def add_features(sender, **kwargs):
     UPDATE_CONDA_STARTUP = 'update_conda_startup'
     CONDA_APPEND_CORE_REQUIREMENTS = 'conda_append_core_requirements'
     ALL_VERSIONS_IN_HTML_CONTEXT = 'all_versions_in_html_context'
+    SKIP_SYNC = 'skip_sync'
```
humitos: nit: I would call it `SKYP_WEB_SERVERS_SYNC` to make it explicit.

humitos:
I don't think it worth the effort, but we could enable this Feature flag automatically on `Project.save` when we detect that `skip = True` and remove it otherwise.

PR #6695, opened 2020-02-24
Skip promoting new stable if current stable is not `machine=True`

humitos:
When the "stable" is managed by the user (because they have a
`origin/stable` branch, for example) we do not need to calculate which
would be the version to be promoted as stable.

So, we return immediately (without doing any computation) if the
current stable is not managed by Read the Docs.


readthedocs/projects/models.py
```diff
@@ -1022,15 +1022,21 @@ def update_stable_version(self):
...
+
         versions = self.versions(manager=INTERNAL).all()
         new_stable = determine_stable_version(versions)
         if new_stable:
-            current_stable = self.get_stable_version()
             if current_stable:
                 identifier_updated = (
                     new_stable.identifier != current_stable.identifier
                 )
-                if identifier_updated and current_stable.machine:
```
ericholscher: Won't this break users who are setting `stable` manually?

ericholscher: Oh, I see how it works now. Ignore me :)

humitos:
This is valid. I just need to write some tests for it.

ericholscher:
@humitos would be useful to get the tests fixed up here, so we can ship it 👍

PR #6708, opened 2020-02-26
Workaround SameSite cookies

### LINKED ISSUE #6705: Gold users seeing ads after Django2 upgrade
After the Django 2 upgrade today, gold users are currently seeing ads. This is due to setting `SameSite=Lax` on our session cookie. The cookie is not sent with the ads API request.

While we could revert that part of this change, in the next couple weeks [Chrome is going to start not sending cookies for cross-domain requests](https://www.chromium.org/updates/same-site) even if `SameSite` is unset. Instead, I think we need a workaround or some alternative here. We may do something along the lines of what [Auth0 is doing](https://auth0.com/blog/browser-behavior-changes-what-developers-need-to-know/#SameSite.None) by setting 2 cookies: one with `SameSite` and one without.

This does not affect projects that have been designated as ad-free. Those are still ad-free.

davidfischer:
In Django 2.2, the default SameSite value for cookies became `SameSite=Lax`. This causes cookies to not be sent with cross-domain XHRs (eg. from `*.readthedocs.io` to `readthedocs.org`). We could not set this attribute at all by setting `SESSION_COOKIE_SAMESITE = None` but starting soon Chrome is going to treat unset SameSite the same as `SameSite=Lax`. Chrome wants you to explicitly set `SameSite=None` to get the old cross-site behavior.

There are two problems with this. Firstly, [some browsers completely reject cookies with `SameSite=None`](https://www.chromium.org/updates/same-site/incompatible-clients) and secondly, Django<3.1 doesn't support setting `SameSite=None` out of the box (`SESSION_COOKIE_SAMESITE = None` means `SameSite` is unset rather than set to `None`).

The workaround is to set 2 cookies -- the first one is set to `SameSite=None` and the 2nd with SameSite unset and to have a custom middleware that forces `SameSite=None`. The first cookie may be rejected completely. We also set a 2nd cookie without the SameSite attribute at all. If the first cookie is present, always use it. This is the solution for new browsers that understand `SameSite=None`. If the first cookie is missing but the 2nd is present, use that as the session cookie.

See: https://web.dev/samesite-cookie-recipes/
See: https://auth0.com/blog/browser-behavior-changes-what-developers-need-to-know/#SameSite.None

Much of the added code is copied directly from Django 2.2's SessionMiddleware.

Fixes  #6705

### DIFF
--- a/readthedocs/core/middleware.py
+++ b/readthedocs/core/middleware.py
@@ -1,12 +1,17 @@
 import logging
+import time
 
 from django.conf import settings
 from django.contrib.sessions.backends.base import SessionBase
+from django.contrib.sessions.backends.base import UpdateError
 from django.contrib.sessions.middleware import SessionMiddleware
+from django.core.exceptions import SuspiciousOperation
 from django.core.exceptions import MultipleObjectsReturned, ObjectDoesNotExist
 from django.http import Http404, HttpResponseBadRequest
 from django.urls.base import set_urlconf
+from django.utils.cache import patch_vary_headers
 from django.utils.deprecation import MiddlewareMixin
+from django.utils.http import http_date
 from django.utils.translation import ugettext_lazy as _
 from django.shortcuts import render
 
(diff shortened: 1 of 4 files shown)


readthedocs/core/middleware.py
```diff
@@ -188,34 +193,131 @@ def process_request(self, request):
...
             if (
                 request.path_info.startswith(url) and
-                settings.SESSION_COOKIE_NAME not in request.COOKIES
+                settings.SESSION_COOKIE_NAME not in request.COOKIES and
+                self.cookie_name_fallback not in request.COOKIES
             ):
                 return response
-        return super().process_response(request, response)
+
+        try:
```
ericholscher: Did we heavily modify the code below, or just small edits to support both cookie names? Would be useful to better understand what exactly we've changed.

davidfischer: Mostly it's just small changes. I can add a comment trying to explain it although the places with changes are commented.

davidfischer: I added comments about what changed. It isn't just Auth0 making this recommendation. This is recommended directly from the [Google web team](https://web.dev/samesite-cookie-recipes/).

PR #6709, opened 2020-02-26
Docker settings improvements

davidfischer:
These changes were just small tweaks I made to the settings while working on #6708. These changes make it easier to override values like `PRODUCTION_DOMAIN` when necessary.

### DIFF
--- a/readthedocs/settings/docker_compose.py
+++ b/readthedocs/settings/docker_compose.py
@@ -12,14 +12,15 @@ class DockerBaseSettings(CommunityDevSettings):
     RTD_DOCKER_USER = f'{os.geteuid()}:{os.getegid()}'
     DOCKER_LIMITS = {'memory': '1g', 'time': 900}
     USE_SUBDOMAIN = True
-    STATIC_URL = 'http://community.dev.readthedocs.io/devstoreaccount1/static/'
 
     PRODUCTION_DOMAIN = 'community.dev.readthedocs.io'
     PUBLIC_DOMAIN = 'community.dev.readthedocs.io'
-    PUBLIC_API_URL = 'http://community.dev.readthedocs.io'
+    PUBLIC_API_URL = f'http://{PRODUCTION_DOMAIN}'
     RTD_PROXIED_API_URL = PUBLIC_API_URL
     SLUMBER_API_HOST = 'http://web:8000'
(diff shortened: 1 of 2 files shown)


davidfischer:
One other change I considered but didn't make in this PR because it requires discussion is to make `STATIC_URL` not actually point to storage at all and use storage only for media. Currently if we were doing significant tweaks to the theme or static files, those would involve running `collectstatic` after each change which could be tedious.

Instead, perhaps we should just serve static files directly from the web. Media files would still be served from storage. This could be accomplished with something like:

```python
## settings.py
STATIC_URL = "/static/"

## urls.py
if settings.DEBUG:
    urlpatterns += static(settings.STATIC_URL, document_root=settings.STATIC_ROOT)
```

readthedocs/settings/docker_compose.py
```diff
@@ -110,14 +111,14 @@ def DATABASES(self):  # noqa
     AZURE_ACCOUNT_KEY = 'Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw=='
     AZURE_CONTAINER = 'static'
     AZURE_STATIC_STORAGE_CONTAINER = AZURE_CONTAINER
-    AZURE_MEDIA_STORAGE_HOSTNAME = 'community.dev.readthedocs.io'
+    AZURE_MEDIA_STORAGE_HOSTNAME = PRODUCTION_DOMAIN
```
humitos: In an ideal world, media and static would be different domains but Azurite had problems with CORS headers when I implemented this.

humitos:
> One other change I considered but didn't make in this PR because it requires discussion is to make `STATIC_URL` not actually point to storage at all and use storage only for media. Currently if we were doing significant tweaks to the theme or static files, those would involve running `collectstatic` after each change which could be tedious.

It would be good to have this easily configurable maybe, for these cases you mention here. I like using the storage for static files because it helps us to find issues with it --as it did already with manifest on S3.

readthedocs/settings/docker_compose.py
davidfischer: It does look like it has since added support for CORS but I didn't spend the time to figure out how to add it. It isn't clear if it is allowed to set globally as opposed to per object either. If it's per object, that probably isn't useful to us.

PR #6717, opened 2020-02-27
Only output debug logging from RTD app

ericholscher:
In normal settings only log INFO and above,
but in dev we should log DEBUG to console.
This should only output RTD debug messages though,
because others are only logging INFO and above via the logger.

### DIFF
--- a/readthedocs/settings/base.py
+++ b/readthedocs/settings/base.py
@@ -603,7 +603,7 @@ class CommunityBaseSettings(Settings):
         },
         'handlers': {
             'console': {
-                'level': 'INFO',
+                'level': 'DEBUG',
                 'class': 'logging.StreamHandler',
                 'formatter': 'default'
             },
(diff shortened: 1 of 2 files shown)


readthedocs/settings/base.py
```diff
@@ -621,7 +621,7 @@ def USE_PROMOS(self):  # noqa
             '': {  # root logger
                 'handlers': ['debug', 'console'],
                 # Always send from the root, handlers can filter levels
-                'level': 'DEBUG',
+                'level': 'INFO',
```
humitos: If the root logger filters by INFO, can we set DEBUG level in the handler later? With this change, is DEBUG logging under `debug.log` file still working?

ericholscher: We want to be able to log DEBUG messages, but only from our application. So we want DEBUG to be output, but only sent via the RTD app, not all libraries.

PR #6726, opened 2020-02-28
Improve Proxito 404 handler to render user-facing Maze when needed

### LINKED ISSUE #6721: Project domains starting with a number are broken
## Details
Every project domain starting with a number (eg. 54street) doesn't show the documentation but instead `SORRY This page does not exist yet`.

It even happens for projects that do not exist. For example, look at the different behaviour of this two urls where both projects are not existent:

https://street54.readthedocs.io/en/latest/ (returns `Not Found`)
https://54street.readthedocs.io/en/latest/ (returns `SORRY This page does not exist yet`)

## Expected Result
The documentation should be visible at the project's url.

## Actual Result
The error `SORRY This page does not exist yet.` is always returned, even when build is successfull.

humitos:
Use a specific 404 handler in El Proxito to decide if it's called to
be executed as a NGINX internal redirect (`fast_404` view) or if was
called to render the 404 user-facing page (`ServeError404` view).

This way, we keep the fast 404 view for all our internal handling but
we use the standard Django flow to render 404 pages for user.

Closes #6721


readthedocs/proxito/views/utils.py
```diff
@@ -19,6 +19,24 @@ def fast_404(request, *args, **kwargs):
...
+    Decide what 404 page return depending if it's an internal NGINX redirect.
+
+    We want to return fast when the 404 is used as an internal NGINX redirect to
+    reach our ``ServeError404`` view. However, if the 404 exception was risen
+    inside ``ServeError404`` view, we want to render the default Read the Docs
+    Maze page.
+    """
+
+    if request.resolver_match.url_name != 'proxito_404_handler':
+        return fast_404(request, exception, template_name)
```
ericholscher: When would this ever be True? It's only configured on our proxito views, so it seems like we could just use the normal RTD 404 handler here, no?

humitos: If you hit a URL of a non existing project (*) there are going to be two 404 happening there:

1. internal NGINX redirect, produced at decorator project mapping: https://github.com/readthedocs/readthedocs.org/blob/804887591a0882b0f6d3603a0034b5cc12e79755/readthedocs/proxito/views/decorators.py#L76
2. user-facing 404, in our ServeError404 view itself: https://github.com/readthedocs/readthedocs.org/blob/804887591a0882b0f6d3603a0034b5cc12e79755/readthedocs/proxito/views/serve.py#L280

The 1) is going to execute `fast_404` and 2) is going to execute the user-facing one.

We could use always the user-facing one, but we "loose the speed" of the `fast_404` in those cases. I'm fine with that, since it's simpler, though.

(*) any 404 happening inside the `ServeDocs` should fastly respond 404 so it's cached by our own `ServeError404` handler. In the end, the last 404 which will be shown to the user must be the one that calls `render(template)`, all the others could be the fast one.

humitos:
I fixed two tests, but there is one that seems unrelated to my changes. It needs to be fixed, though since it's the one that checks for an infinite redirect --but looks that some change needs to be done in the code, not just in the test.

PR #6729, opened 2020-03-02
Add feature flag for branch & tag syncing to API.

ericholscher:
This helps us lower load on specific projects that have a lot of these and aren’t using them.


ericholscher:
I've tested this locally, and I've confirmed it *deletes* the existing Versions associated with the tags and branches. This should help performance a lot 👍

agjohnson:
This conflicts a little with the hotfix feature flag PR to skip symlinking and sync operations. I've merged that PR so this can be resolved.

PR #6739, opened 2020-03-04
Add TOS PDF output

agjohnson:
Unfortunately, RTD doesn't support this output because it's a secondary
file. You have to generate this file locally, and you have to also
disable `-W` option, as there are warnings that fail the build,
unrelated to this change.


humitos:
> You have to generate this file locally, and you have to also disable `-W` option, as there are warnings that fail the build, unrelated to this change.

We can add a comment about this above the latex documents config, or we can even create a new Makefile entry to generate only this file that already include the `-W` argument.

ericholscher:
We could certainly add support for building/syncing these files, but linking them directly from our dashboard is a bit tricky, but I'd be 👍 with building in basic support for it at least.

PR #6751, opened 2020-03-05
Use `storage.exists` on HEAD method

humitos:
When request method is HEAD we can't use ``storage.url`` because the signature calculated will consider GET as method. `django-storages` does not support passing the HTTP method into ``storage.url`` yet. Because of this, the response won't be exactly the same between GET and HEAD since we are not passing the headers returned by the storage itself.

We just use `storage.exists` and return 200 with an empty body or raise 404.

### DIFF
--- a/readthedocs/proxito/views/serve.py
+++ b/readthedocs/proxito/views/serve.py
@@ -133,6 +133,17 @@ class ServeDocsBase(ServeRedirectMixin, ServeDocsMixin, View):
             # Signature and Expire time is calculated per file.
             path += 'index.html'
 
+        if request.method == 'HEAD':
+            # When request method is HEAD we can't use ``storage.url`` because
+            # the signature calculated will consider GET as method.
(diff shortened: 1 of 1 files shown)


jschneier:
Merged the corresponding PR

ericholscher:
Thanks @jschneier!

ericholscher:
@humitos should we merge this, or should we add a stoages dependency to the commit that fixes it?

humitos:
@ericholscher the immediate fix I'd say is our PR. I'm not sure what else we are introducing if we upgrade `django-storages` right now. We are using 1.7.2 and the latest version is 1.9.1

I'm not really concern about incompatibility changes in the django-storage package itself, but on azure (mainly) and boto3 dependencies.

My plan would be to merge this PR and deploy tomorrow and then django-storage + azure/boto3 and test it for at least one week locally.

PR #6763, opened 2020-03-09
Pull/Push cached environment using storage

humitos:
Before starting clonig the repository we check of a cached environment in the storage. If there is one, we pull it and extract it under `project.doc_path` (using the same structure that we have been using) and use those cached files.

Once the build has finished and it was successful, we compress all the environment files (pip cache, checkout, conda and virtualenv environments) and push it to the storage.

The cached is cleaned up when the version is wiped.

This feature is under a Feature flag (`CACHED_ENVIRONMENT`), so we can select specific projects to start testing out this cached environment. It can help on projects with big/huge repositories or that install a lot of dependencies.

### DIFF
--- a/readthedocs/projects/tasks.py
+++ b/readthedocs/projects/tasks.py
@@ -124,8 +124,8 @@ class CachedEnvironmentMixin:
             with open(tmp_filename, mode='wb') as local_fd:
                     local_fd.write(remote_fd.read())
 
-            tar = tarfile.TarFile(tmp_filename)
-            tar.extractall(self.version.project.doc_path)
+            with tarfile.open(tmp_filename) as tar:
+                tar.extractall(self.version.project.doc_path)
 
             # Cleanup the temporary file
             if os.path.exists(tmp_filename):
--- a/readthedocs/projects/tasks.py
+++ b/readthedocs/projects/tasks.py
@@ -136,25 +136,28 @@ class CachedEnvironmentMixin:
             return
 
         project_path = self.project.doc_path
-        paths = [
-            os.path.join(project_path, 'checkouts', self.version.slug),
-            os.path.join(project_path, 'envs', self.version.slug),
-            os.path.join(project_path, 'conda', self.version.slug),
-            os.path.join(project_path, '.cache'),
+        directories = [
+            'checkouts',
+            'envs',
+            'conda',
         ]
 
         _, tmp_filename = tempfile.mkstemp(suffix='.tar')
         # open just with 'w', to not compress and waste CPU cycles
         with tarfile.open(tmp_filename, 'w') as tar:
-            for path in paths:
+            for directory in directories:
+                path = os.path.join(
+                    project_path,
+                    directory,
+                    self.version.slug,
+                )
+                arcname = os.path.join(directory, self.version.slug)
                 if os.path.exists(path):
-                    tar.add(
-                        path,
-                        arcname=os.path.join(
-                            os.path.basename(os.path.dirname(path)),
-                            self.version.slug,
-                        )
-                    )
+                    tar.add(path, arcname=arcname)
+
+            # Special handling for .cache directory because it's per-project
+            path = os.path.join(project_path, '.cache')
+            tar.add(path, arcname='.cache')
 
         storage = get_storage_class(settings.RTD_BUILD_ENVIRONMENT_STORAGE)()
         with open(tmp_filename, 'rb') as fd:
--- a/readthedocs/projects/tasks.py
+++ b/readthedocs/projects/tasks.py
@@ -119,7 +119,7 @@ class CachedEnvironmentMixin:
                     'msg': msg,
                 }
             )
-            tmp_filename = tempfile.mktemp(suffix='.tar')
+            tmp_filename = tempfile.mkstemp(suffix='.tar')
             remote_fd = storage.open(filename, mode='rb')
             with open(tmp_filename, mode='wb') as local_fd:
                     local_fd.write(remote_fd.read())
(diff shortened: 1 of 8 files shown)


humitos:
Some notes to keep in mind while reviewing this PR:

- I tried to put these method under `before_vcs` and `build_complete` signals, but the code wasn't really clean. Besides it needed some other changes to the signal it self since we are not passing the required information (e.g. `build` object is a dictionary instead of a Build instance)
- The cached environment is pushed back to the storage _only_ on success builds to avoid corrupted data
- In case of corrupted cached environment, the user can just wipe the version to start from a fresh environment
- If we want to start testing this out in production, we need to define `RTD_BUILD_ENVIRONMENT_STORAGE`
- This solution should work either on Amazon S3 and Microsoft Blob Storage (it uses django-storage behind the scenes)
- I haven't testing this with a huge repository yet
- I'm not using compression at all to avoid using CPU cycles for this. We want to go fast here

readthedocs/projects/tasks.py
```diff
@@ -87,6 +89,87 @@
...
+
+        project_path = self.project.doc_path
+        paths = [
+            os.path.join(project_path, 'checkouts', self.version.slug),
+            os.path.join(project_path, 'envs', self.version.slug),
+            os.path.join(project_path, 'conda', self.version.slug),
+            os.path.join(project_path, '.cache'),
+        ]
+
+        tmp_filename = tempfile.mktemp(suffix='.tar')
```
agjohnson: `mktemp` has been deprecated for a while. It would be better to use tempfile creation in a context manager to handle file cleanup on exceptions/etc

readthedocs/projects/tasks.py
```diff
@@ -87,6 +89,87 @@
...
+        # open just with 'w', to not compress and waste CPU cycles
+        with tarfile.open(tmp_filename, 'w') as tar:
+            for path in paths:
+                if os.path.exists(path):
+                    tar.add(
```
agjohnson: Noted in chat, we will want to be defensive on tar operations, as we want to preserve and extended attributes and special file types (symlinks especially) on tar operations. A good place to start with tar for full system backups is `tar --xattrs -czpvf ...` -- so maybe the equivalent of this call? If you research full system backups using tar, you'll come across some recommendations i'm sure.

readthedocs/projects/tasks.py
humitos: I will change this code to not use `mktemp` then. However, using a context manager with a temporary file will automatically delete the file when we are out of the context. We need to write on the temporary file, close it, and then read it to upload to the storage.

I'm not too worries about cleaning up the file since it's inside a container that will remove all the data created there once it's closed (although, using LocalEnvironment could be a problem if not deleted)

I will replace `mktemp` with `mkstemp` which is the recommended way to get a tmp filename.

readthedocs/projects/tasks.py
humitos: I didn't find anything particularly useful. The recommended attributes to create a full system backup are the defaults in Python. So, I think we are good here.

readthedocs/projects/tasks.py
```diff
@@ -87,6 +89,94 @@
...
+                    'msg': msg,
+                }
+            )
+            tmp_filename = tempfile.mkstemp(suffix='.tar')
+            remote_fd = storage.open(filename, mode='rb')
+            with open(tmp_filename, mode='wb') as local_fd:
+                local_fd.write(remote_fd.read())
+
+            with tarfile.open(tmp_filename) as tar:
+                tar.extractall(self.version.project.doc_path)
```
ericholscher: I wonder if it's worth abstracting the archive format here. It looks like `shutil` has an abstraction over it. I imagine we'll want to play with `tar` and `gzip` at least, because I imagine bandwidth and CPU will be a tradeoff here that we want to be able to tweak.

https://docs.python.org/3/library/shutil.html#archiving-operations

readthedocs/projects/tasks.py
```diff
@@ -271,6 +361,7 @@ def run(self, version_pk):  # pylint: disable=arguments-differ
             with environment:
                 before_vcs.send(sender=self.version, environment=environment)
                 with self.project.repo_nonblockinglock(version=self.version):
+                    self.pull_cached_environment()
```
ericholscher: Do we not `push` it back up during `sync`?

humitos: I did this on purpose. Since we have per-version cached environment and sync operation is not tied to a version --well, it's tied but we always pass the default one.

The other problem could be a race condition. Even if we would pass different versions to sync, it may conflict with the build itself trying to both push the cache to the storage at the same time or even sync after build (loosing the real cache)

> I could add this explanation as a comment in the code, probably

readthedocs/projects/tasks.py
humitos: Hrm... Interesting! I didn't know about `shutil.make_archive`. I will take a deeper look.

humitos: Having an abstraction is a good idea to be able to play here. Although, it seems that `shutil.make_archive` does not gives us too much ability for this since it's not too customizable: we can't change lower level attributes.

Besides, we would need Python 3.8 because since that version they are using modern PAX (``--xattrs`` equivalent):

> Changed in version 3.8: The modern pax (POSIX.1-2001) format is now used instead of the legacy GNU format for archives created with format="tar".

On the other hand, `tarfile` allows us to play with `gzip`, `bz2` and `lzma` and with compression level as well.

readthedocs/projects/tasks.py
agjohnson: Oh I see, that makes sense! `mkstemp` is a minor upgrade then

humitos:
@agjohnson there is a small PR that just defines the required settings on docker compose environment and other in -ops

I don't think there is any other differences compared with community.

PR #6771, opened 2020-03-11
Upgrade django-storages to support URLs with more http methods

humitos:
This PR upgrades django-storages to a not yet released version to support HEAD method when generating storage URLs that are private (include signed tokens).

I checked the changelog and there is nothing breaking as far as I can tell. All the tests that I did locally using Azurite worked properly.

Note that I'm removing `azure-storage-blob` and `azure-storage-common` from our pinned versions because those are correctly specified on django-storage dependencies as `azure-storage-blob >=1.3.1,<12.0.0`.

I'm interesting on merging this sooner than later, so we can have a week or similar of testing locally before deploying it. Best way to test this is to pull down the PR, re-generate the docker image completely (without cache: `inv docker.build`) and do some storage related operations (build, serve docs, wipe, collectstatic, etc)

### DIFF
--- a/readthedocs/builds/storage.py
+++ b/readthedocs/builds/storage.py
@@ -186,3 +186,16 @@ class BuildMediaFileSystemStorage(BuildMediaStorageMixin, FileSystemStorage):
         if not self.exists(path):
             return [], []
         return super().listdir(path)
+
+    def url(self, name, *args, **kwargs):  # noqa
+        """
+        Override to accept extra arguments and ignore them all.
+
+        This method helps us to bring compatibility between Azure Blob Storage
+        (which does not use the HTTP method) and Amazon S3 (who requires HTTP
+        method to build the signed URL).
+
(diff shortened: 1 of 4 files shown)


ericholscher:
Seems tests aren't installing the proper version.

humitos:
It's installing the proper version. However, I didn't realize that this is not going to work as is because I added the support for `http_method` only in S3 backend and we are using Azure here.

Actually, it does not make sense in Azure because [it seems the token is not based on HTTP method](https://docs.microsoft.com/en-us/python/api/azure-storage-blob/azure.storage.blob.baseblobservice.baseblobservice?view=azure-python-previous#generate-blob-shared-access-signature-container-name--blob-name--permission-none--expiry-none--start-none--id-none--ip-none--protocol-none--cache-control-none--content-disposition-none--content-encoding-none--content-language-none--content-type-none-) :confused:

PR #6779, opened 2020-03-14
Perform redirects at db level

humitos:
This is another attempt to perform redirects at DB level and be able to put them at Proxito level, allowing to redirect to be executed not only on 404s.

It continues the work we did at #6398 but that we reverted at #6430 because we found an issue with Exact Redirects with `from_url` smaller than 5 chars long. This PR workarounds that case, giving us the opportunity to keep exploring this idea of performing redirects at DB level.

### DIFF
--- a/readthedocs/redirects/querysets.py
+++ b/readthedocs/redirects/querysets.py
@@ -1,6 +1,8 @@
 """Queryset for the redirects app."""
 
 from django.db import models
+from django.db.models import Value, CharField, IntegerField, Q, F, ExpressionWrapper
+from django.db.models.functions import Substr, Length
 
 from readthedocs.core.utils.extend import SettingsOverrideObject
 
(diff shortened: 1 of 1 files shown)


humitos:
I checked how many of these redirects we have:

- .org: 51
- .com: 0

Most of them are `/` and `''` which do not work anyways. The rest may be worth to double check also, because they would only work if the project is single version since there is not enough chars to create a valid redirect with less than 5 chars in a multi-version project (`/en/` is 4 chars already)

```
In [26]: Counter(list(Redirect.objects.annotate(from_url_length=ExpressionWrapper(Length('from_url'), output_field=IntegerField())).filter(from_url_length__lt=5).filter(redirect_type='exact'
    ...: ).values_list('from_url', flat=True)))
Out[26]: 
Counter({'': 7,
         '/': 25,
         '/ko': 1,
         '/en': 1,
         '/who': 1,
         '/why': 1,
         '/faq': 2,
         '/bar': 1,
         '/foo': 1,
         '/rus': 1,
         '/soc': 1,
         '*': 2,
         '?q': 1,
         '/en/': 3,
         '3.1': 1,
         '/ko/': 2})
```

I'd say that it worth to have a talk around this and discuss if it's doable, but it seems we are pretty close to this, IMO.

humitos:
Closing in favor of #6780

PR #6782, opened 2020-03-14
Properly proxy the Proxito headers via nginx/sendfile

ericholscher:
This is a hack, but it works.

### DIFF
--- a/dockerfiles/nginx/proxito.conf
+++ b/dockerfiles/nginx/proxito.conf
@@ -42,6 +42,8 @@ server {
 
         # Nginx is silly, so we have to *set* these then use them.
         # Using them directly gets the values from the storage upstream.
+        # I think the `set` must get read before doing the proxy_pass
+        # but the `add_header` is read after
         set $rtd_project $upstream_http_x_rtd_project;
         add_header X-RTD-Project $rtd_project always;
         set $rtd_version $upstream_http_x_rtd_version;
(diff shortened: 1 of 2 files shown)


dockerfiles/nginx/proxito.conf
```diff
@@ -20,8 +20,12 @@ server {
...
         proxy_pass http://proxito:8000;
         proxy_set_header Host $host;
-        proxy_intercept_errors on;
-        error_page 404 = @notfoundfallback;
+        proxy_set_header X-Real-IP $remote_addr;
+        proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
+        proxy_set_header X-Forwarded-Host $host;
+
+        # Don't handle 404's here because they are from the application
+        # We only want to catch 404's from the storage backend.
```
humitos: Hrm... if our application does `get_object_or_404` we want to trigger our handler, right?

ericholscher: Hmm, perhaps, but it seems like in those cases we want the default 404, not a users 404?

humitos: Right. That's how it's working right now. By removing this nginx config, the user will get the `fast_404` response instead the default render.

We are returning the correct 404 in the handler: https://github.com/readthedocs/readthedocs.org/blob/2091b2dd74ed90827759860816823f18c96885c6/readthedocs/proxito/views/utils.py#L22-L37

So, basically if it does not come from `_proxito_404_` URL we will return just `Not found.` as response if we remove this.

humitos: Removing this config, if you set `DEBUG = False` in proxito settings and access to http://notfound.community.dev.readthedocs.io/en/latest/ you get just `Not Found.`

ericholscher: Reverted

PR #6784, opened 2020-03-16
Add CORS headers to Azurite

### LINKED ISSUE #6724: Add CORS support on Azurite for docker compose
Since https://github.com/readthedocs/readthedocs.org/pull/6709 we are pinning an Azurite image with seems it has support for CORS headers.

I've tried this already but CORS was not supported in Azurite at that time:

https://github.com/readthedocs/readthedocs.org/blob/572d3747fba765f17ed40d0678004215f4c3a0ef/dockerfiles/nginx/web.conf#L12-L16

To fix this issue, we need

* run a similar command on startup (maybe using the `azure-cli` container)
* remove that nginx block
* update Django settings to re-point static domain
* maybe others

Reference: https://github.com/readthedocs/readthedocs.org/pull/6709#discussion_r384641247

humitos:
I may be missing something because the URLs generated by Django are not considering `STATIC_URL` and pointing them to `localhost:10000` but they are still pointing to `community.dev.readthedocs.io` for some reason.

Besides that, I think the rest of the changes are all the required ones.

Closes #6724


ericholscher:
Is this still needed?

humitos:
It's better to have it. Currently we are hitting NGINX and proxy the request to the storage instead of hitting the storage directly --which is the correct thing.

readthedocs/settings/docker_compose.py
```diff
@@ -22,7 +22,9 @@ class DockerBaseSettings(CommunityDevSettings):
     SLUMBER_API_HOST = 'http://web:8000'
     RTD_EXTERNAL_VERSION_DOMAIN = 'org.dev.readthedocs.build'
 
-    STATIC_URL = f'http://{PRODUCTION_DOMAIN}/devstoreaccount1/static/'
+    # storage:10000 is running the Azure Blob Storage and it's exposed to the
+    # host on the same port that user's can access via localhost
+    STATIC_URL = f'http://localhost:10000/devstoreaccount1/static/'
```
ericholscher, line 27: Should it continue to use `PRODUCTION_DOMAIN`, but on port 10000? That seems a bit cleaner, and should still resolve to localhost.

humitos: Yeah. I think you are right. I may messed things up here.

I will update this PR with latest master and see if it still works :)

humitos: So, yes. Although, the way to specify this is with our `AZURE_STATIC_STORAGE_HOSTNAME` that it's used to override the domain returned by `storage.url` in our Mixin.

This STATIC_URL just need to be `/devstorageaccount1/static/`

BTW, I'm using `assets.community.dev.readthedocs.io` so it's a different domain and it follows the same idea than production.

PR #6793, opened 2020-03-18
Use `tar` to extract cache, write by chunks and new build state

humitos:
We have experimented memory leak issues when extracting a file using the Python API. We are using regular `tar` for this operation now since the test done in the server with 2.5Gb has shown better results regarding memory usage and extracting time as well.

Besides, instead of reading the whole file into memory (`fd.read()`), we are iterating over its chunks and saving these small chunks into disk.

Finally, we are adding a new build status "Pulling cache" for now to communicate that the build is not in "Triggered" state, but it's actually doing something.

Related to #6763


readthedocs/projects/tasks.py
```diff
@@ -122,10 +131,20 @@ def pull_cached_environment(self):
...
-                local_fd.write(remote_fd.read())
-
-            with tarfile.open(tmp_filename) as tar:
-                tar.extractall(self.version.project.doc_path)
+                for chunk in remote_fd.chunks():
+                    local_fd.write(chunk)
+
+            safe_makedirs(self.version.project.doc_path)
+            # Using ``tar`` command because ``tarfile`` has a memory leak when
+            # extracting big files (2.5Gb)
```
humitos: I'm not 100% convinced that this is true. I isolated the problem but I wasn't able to find that `.extractall` consumes more memory than `tar` command when it's _outside_ read the docs environment yet.

humitos: This was not the problem in the end. The problem was IO and copying the big file multiple times.

humitos:
Closing in favor of #6799 and #6800

PR #6799, opened 2020-03-20
Use storage.open API correctly for tar files (build cached envs)

humitos:
Instead of downloading the file into a temporary file (`storage.open`
and then `.read`) to save the file into another temporary file (our
loop reading by chunks) we just use the `storage.open` response as the
input of the `tarfile.open` function and extract it from there.

Related to #6793


humitos:
Merging. Failing tests are unrelated to this PR (master seems broken as well)

PR #6803, opened 2020-03-23
Use `conda env update --prune` to re-use the conda environment

humitos:
This is similar to what we are doing with pip already: instead of creating the environment from scratch on each build, we can re-use the one that was created before.

The option `--prune` will remove all the packages that were removed from the environment.yml file.

> Under a feature flag so we can start testing it without affecting all the projects

https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/manage-environments.html#updating-an-environment

### DIFF
--- a/readthedocs/doc_builder/python_environments.py
+++ b/readthedocs/doc_builder/python_environments.py
@@ -479,18 +479,37 @@ class Conda(PythonEnvironment):
             self._append_core_requirements()
             self._show_environment_yaml()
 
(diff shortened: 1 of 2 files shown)


humitos:
`--quiet` is not quiet when using `update` because of a bug: https://github.com/conda/conda/issues/9108

humitos:
This is a better solution https://github.com/readthedocs/readthedocs.org/pull/6815

PR #6806, opened 2020-03-23
Add feature flag letting us ignore `/projects/` on subproject URLs

ericholscher:
This is in production on .com for a customer,
and now this will let us turn it on for users in general.

### DIFF
--- a/readthedocs/proxito/views/serve.py
+++ b/readthedocs/proxito/views/serve.py
@@ -18,6 +18,7 @@ from readthedocs.builds.constants import LATEST, STABLE, EXTERNAL
 from readthedocs.builds.models import Version
 from readthedocs.core.utils.extend import SettingsOverrideObject
 from readthedocs.projects import constants
+from readthedocs.projects.models import Feature, ProjectRelationship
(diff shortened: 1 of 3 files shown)


readthedocs/proxito/views/serve.py
```diff
@@ -98,6 +99,18 @@ def get(self,
                 'Invalid URL for project with versions. url=%s, project=%s',
                 filename, final_project.slug
             )
+            if final_project.has_feature(Feature.PROXITO_SUBPROJECT_PATH):
+                new_subproject_slug, rest = filename.split('/', 1)
+                sub = ProjectRelationship.objects.filter(
```
humitos: nit: I prefer if we don't chunk names in variables. Here `sub` and in the tests `feat`.

humitos:
> Although, it does not implement 100% what we already have deployed for this customer.

I updated our current solution at NGINX level to keep the exact same behavior without moving this feature to El Proxito for now. See https://github.com/readthedocs/readthedocs-corporate-ops/pull/396

ericholscher:
CLosing this for now. I think there's better approach in #6868

PR #6812, opened 2020-03-24
Use django autoreload to restart celery workers

humitos:
Use Django internals (`file_changed` signal) to detect that a file was changed and Celery `pool_restart` to communicate with workers so they gracefully reload the Python modules for all the tasks they have registered.

This should help with CPU consumption and removes the needing of using `watchmedo` 4 times to inpect the same files (from web, from proxito, from celery and from builder). Now, we just need web and proxito.

Uses https://github.com/readthedocs/common/pull/57

### DIFF
--- a/readthedocs/builds/signals.py
+++ b/readthedocs/builds/signals.py
@@ -2,7 +2,28 @@
 
 """Build signals."""
 
-import django.dispatch
+import logging
+import os
+
+from django.conf import settings
+from django.dispatch import receiver, Signal
+from django.utils.autoreload import file_changed
+
+from readthedocs.worker import app
+
+build_complete = Signal(providing_args=['build'])
 
 
-build_complete = django.dispatch.Signal(providing_args=['build'])
+log = logging.getLogger(__file__)
+
+
+if all([
+        # restart workers in local docker compose only
+        settings.RTD_DOCKER_COMPOSE,
+        # restart workers only from web instance
+        os.environ['DJANGO_SETTINGS_MODULE'].endswith('web_docker'),
+]):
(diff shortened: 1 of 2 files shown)


readthedocs/builds/signals.py
```diff
@@ -2,7 +2,28 @@
...
+
+
+log = logging.getLogger(__file__)
+
+
+if all([
+        # restart workers in local docker compose only
+        settings.RTD_DOCKER_COMPOSE,
+        # restart workers only from web instance
+        os.environ['DJANGO_SETTINGS_MODULE'].endswith('web_docker'),
```
ericholscher, line 24: Why are we only restarting the webs?

ericholscher:
FWIW I tried this locally, and it still destroys my CPU :(

readthedocs/builds/signals.py
humitos: We are restarting all the workers.

This check is here because proxito and web will be checking for file changes in disk and this signal will be triggered from both, and we want to restart the workers only once. I just picked `web` but I could have picked `proxito`.

humitos:
> FWIW I tried this locally, and it still destroys my CPU :(

:cry:, the last thing that I have in mind is to run `proxito` with `runserver --no-reload` and find a way to restart it from `web` who is running with reload enabled. That way, we will be watching files only in one process.

ericholscher:
Yea, I don't think this work has much value. Docker needs to get fixed on macOS to solve the problem, all these workaround aren't going to do anything.

humitos:
I'll close this PR. We can come back to it in the future if we found it could be useful.

PR #6813, opened 2020-03-24
Use watchman when calling `runserver` in local development

humitos:
watchman is the replacement of pyinotify and it's the recommended way on Linux and MacOS.

https://docs.djangoproject.com/en/2.2/ref/django-admin/#runserver

Django will automatically detect that this is installed and use watchman instead of pyinotify, showing this log in the terminal:

```
django.utils.autoreload:597[20]: INFO Watching for file changes with WatchmanReloader
```

This change could help to reduce the CPU usage when running docker-compose locally.

This is a continuation of the work done in #6812.


humitos:
I'm merging. Let's see how goes. At least is seems better in my laptop. We can easily revert if needed.

PR #6833, opened 2020-03-30
Upgrade celery and its dependencies

humitos:
This is required to gracefully stop celery workers when using scaling groups.

### DIFF
--- a/requirements/pip.txt
+++ b/requirements/pip.txt
@@ -30,15 +30,9 @@ pyyaml==5.1.2
 Pygments==2.4.2
 
 # Basic tools
-# Redis 3.x has an incompatible change and fails
-# https://stackoverflow.com/questions/53331405/django-compress-error-invalid-input-of-type-cachekey
-# https://github.com/sebleier/django-redis-cache/pull/162
-redis==2.10.6  # pyup: ignore
-# Kombu >4.3 requires redis>=3.2
-kombu==4.3.0  # pyup: ignore
-# Celery 4.2 is incompatible with our code
-# when ALWAYS_EAGER = True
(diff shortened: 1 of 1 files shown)


agjohnson:
I'm hesitant to do this given the number of issues we've had upgrading celery. It's not noted in comments what the underlying problem with the upgrade was, but I'd want to confirm that the celery upgrade is not a new problem being introduced.

It's still not clear what needs to happen with supervisord, as the default time to wait before moving from a SIGTERM to SIGKILL is something like 10s. Was this tested with a long running task and supervisord on shutdown?

requirements/pip.txt
```diff
@@ -30,15 +30,11 @@ pyyaml==5.1.2
...
-# Redis 3.x has an incompatible change and fails
-# https://stackoverflow.com/questions/53331405/django-compress-error-invalid-input-of-type-cachekey
-# https://github.com/sebleier/django-redis-cache/pull/162
-redis==2.10.6  # pyup: ignore
-# Kombu >4.3 requires redis>=3.2
-kombu==4.3.0  # pyup: ignore
-# Celery 4.2 is incompatible with our code
-# when ALWAYS_EAGER = True
-celery==4.1.1  # pyup: ignore
+# redis==3.4.1 makes the celery workers to randomly get killed
```
ericholscher, line 33: On what redis server version? I assume it's a mismatch between client & server.

humitos: Engine Version Compatibility:
Version compatibility of the engine that will be run on your nodes
2.8.6

humitos:
This is not needed anymore. We should go in the direction of #6960 instead.

PR #6847, opened 2020-04-01
Limit concurrent builds

humitos:
Add a Feature flag to be able to limit concurrent builds for specific
projects. This will allow us to protect ourselves for projects
triggering a lot of builds via the API or for a few very long builds
of the same project blocking all the rest of users.

If the limit is reached, the Build is marked as "Queue" and an error
message is shown in the build's detail page saying that the concurrent
limit was reached and the build is re-triggered to be executed in 5
minutes again.

Screenshot_2020-04-01_22-48-07


readthedocs/projects/tasks.py
```diff
@@ -510,6 +512,26 @@ def run(
             self.commit = commit
             self.config = None
 
+            if self.project.has_feature(Feature.LIMIT_CONCURRENT_BUILDS):
+                response = api_v2.build.running.get(project__slug=self.project.slug)
```
humitos, line 516: We could use APIv3 here (https://docs.readthedocs.io/en/stable/api/v3.html#builds-listing) if we add a new permission class to allow build user to list builds. As this will require extra work and we are only using APIv2 from builders, I didn't want to mix them here.

readthedocs/doc_builder/exceptions.py
```diff
@@ -53,6 +53,10 @@ class BuildTimeoutError(BuildEnvironmentError):
     message = ugettext_noop('Build exited due to time out')
 
 
+class BuildMaxConcurrencyError(BuildEnvironmentError):
+    message = ugettext_noop('Concurrent limit reached ({limit}), retrying in 5 minutes.')
```
ericholscher, line 57:
```suggestion
    message = ugettext_noop('Concurrency limit reached ({limit}), retrying in 5 minutes.')
```

readthedocs/projects/tasks.py
```diff
@@ -510,6 +512,26 @@ def run(
...
+
+                    # This is done automatically on the environment context, but
+                    # we are executing this code before creating one
+                    api_v2.build(self.build['id']).patch({
+                        'error': BuildMaxConcurrencyError.message.format(
+                            limit=settings.RTD_MAX_CONCURRENT_BUILDS,
+                        ),
+                        'state': BUILD_STATE_QUEUED,
+                    })
+                    self.task.retry(exc=BuildMaxConcurrencyError, throw=False)
```
ericholscher, line 532: Do we want to explicitly set a time here? We also likely want to set a large max number of retries.

humitos: We can set this explicitly here. Currently, it's taking these numbers from the `@app.task` decorator at https://github.com/readthedocs/readthedocs.org/blob/759683771a2ccea2c439ee6327ac6754adf99cdc/readthedocs/projects/tasks.py#L403-L404

humitos: What numbers do you think that work better for this case?

ericholscher: Not sure, but we likely want almost infinite retries for concurrency I think? But that might have other issues if applied at the task level.

humitos: I'm fine increasing the `max_retries` to something like 10 or 15. With 15 it will be Queued for almost 2hs (105m). So, the user will need to hit the concurrency for that time; which I think it's not common.

We are starting at 4, and it's per-project, not per-organization.

humitos: Hrm... What I said is not 100% true, because the task will be re-queued and put it at the end of it. So, builders would need to be at full capacity (not taking this task) for that time.

readthedocs/api/v2/views/model_views.py
```diff
@@ -276,6 +282,20 @@ class BuildViewSetBase(UserSelectViewSet):
...
+        detail=False,
+        permission_classes=[permissions.IsAdminUser],
+        methods=['get'],
+    )
+    def running(self, request, **kwargs):
+        project_slug = request.GET.get('project__slug')
+        queryset = (
+            self.get_queryset()
+            .filter(project__slug=project_slug)
+            .exclude(state__in=[BUILD_STATE_FINISHED, BUILD_STATE_QUEUED])
```
agjohnson, line 295: This case is where I had issues with overloading the build state to communicate build step to the user in the UI. There are likely other queries/methods we need to update whenever we add build states like this -- for instance, anything checking just for `BUILD_STATE_TRIGGERED` and should be also checking for our new state, `BUILD_STATE_QUEUED`?

This code is correct, but it's hard to know what fallout could be without some thorough testing. Queries like these should probably be QuerySet methods so that we're not continually reproducing logic that we need to update.

humitos: I don't have a strong opinion here, but a slight preference to add a new state. I just wanted to mention why I thought it could be useful.

The difference in UI is that going to the Build list page, you can have a big picture about what's happening: Triggered, Queued, Building, Passed, Failed. Where Triggered means that you are waiting for your turn to build, while Queued means that you have reached the concurrency limit: "those builds are waiting "because of you" (in some way), not us.

Having too many builds in Triggered could cause the user to contact us because they think we are not processing their builds: "the waiting time is too high", where in fact, they have reached the limit.

On the other hand, we use [`Build.is_stale`](https://github.com/readthedocs/readthedocs.org/blob/ee403c589ff0131a87dff57a4639bfa96bc86ce2/readthedocs/builds/models.py#L873) to show a small "Warning" icon in the build list if the build has been in Triggered state for more than 5 minutes. I'd not show this icon for Queued state.

Finally, I did a quick grep trying to find filters via `BUILD_STATE_TRIGGERED` and I didn't find anything important. Although, for `BUILD_STATE_FINISHED` I found this, https://github.com/readthedocs/readthedocs.org/blob/ee403c589ff0131a87dff57a4639bfa96bc86ce2/readthedocs/api/v3/filters.py#L41-L45 but I'm not sure it will be affected, because it considered Triggered as running anyways.

I agree we could move them to a QuerySet and avoid repetition, though.

ericholscher: I say lets keep it Triggered for now, and we can adjust as we go.

There are ways to change the UX for the user without using the state, if needed, or audit all the queries. We need to do additional work before we ship it for users anyway.

agjohnson: Good points! Some of this felt like core team opinions, so I got feedback on the UI changes. What it came back to was that builds should be in "Queued" state if the user has to wait for the build -- regardless of whether we introduce a delay due to concurrency limits, or the build is queued because of build backup. "Triggered" doesn't communicate the same thing as "Queued".

So, I think my opinion is that adding "Queued" is not a problem, but all builds should become "Queued" state when they are put into the build queue. This might leave no room for "Triggered" state if so, and so probably back to my original point, should be dropped in favor of "Queued".

For now, I'd agree we can leave it "Triggered" and move more deliberately to clearer language for the users. We can add more UI later and guide our UI decisions based on our technical implementation.

Here are the feedback notes:

* "Triggered" vs "Queued" does communicate the difference between build state. "Triggered" implies there is something RTD needs to do still, while "Queued" implies the build will be grabbed eventually.
* "I don't need to worry about builds in a queued state, and would have more patience with these builds vs triggered state"
* "Queued" state, when builds go over concurrency, makes sense and is mostly obvious. "I'd have more patience with builds in queued state"
* Expectation is that builds should be picked up and move quickly from a "Triggered" state into "Building" state
* Builds stuck in "queued" for a long time would eventually be a problem
* A build stuck in "Triggered" state **does not imply that we will ever start the build**. If the queue is temporarily backed up, it was expected that the build would enter a "Queued" state, not stay in "Triggered" state. "Triggered" state is more worrying.

humitos: :+1: --I've updated this PR and removed the QUEUED state for now. We can come back to this discussion later if needed.

readthedocs/projects/tasks.py
humitos: I updated the PR with `max_retries=25`

ericholscher:
@humitos I'm 👍 to merge once the tests are fixed.

ericholscher:
Test failure looks unrelated.

PR #6850, opened 2020-04-02
Avoid failing on empty Webhook URL

humitos:
Skip webhook URL if it's empty.

Since we send first the webhook notifications, if the user has an
empty webhook configured, none of the email notification were sent.


humitos:
This is wrong. We are catching the exception below. We should just make this exception a WARNING instead of ERROR.

PR #6853, opened 2020-04-02
Change name of celery default queue

humitos:


readthedocs/settings/base.py
```diff
@@ -334,7 +334,7 @@ def USE_PROMOS(self):  # noqa
     CELERYD_PREFETCH_MULTIPLIER = 1
     CELERY_CREATE_MISSING_QUEUES = True
 
-    CELERY_DEFAULT_QUEUE = 'celery'
+    CELERY_DEFAULT_QUEUE = 'build:default'
```
ericholscher, line 337: I don't think we want this in our base settings, do we? Nothing in a normal RTD install will consume from this queue right now -- this should stay the same, and change in our ops code.

humitos: I've updated this in common as well, just to keep it similar: https://github.com/readthedocs/common/pull/58

I don't have a strong preference, though.

ericholscher: I think we shouldn't change this in a backwards incompat way without a good reason, since it really is an ops change not a change to the codebase.

ericholscher:
This got fixed in our ops repo.

PR #6861, opened 2020-04-06
Make development docs a bit easier to find

ericholscher:


docs/index.rst
```diff
@@ -161,8 +161,8 @@ of Read the Docs and the larger software documentation ecosystem.
 
 * **Getting involved with Read the Docs**:
   :doc:`Contributing <contribute>` |
+  :doc:`Docker development setup </development/standards>` |
   :doc:`roadmap` |
-  :doc:`gsoc` |
```
davidfischer, line 165: Was this purposefully removed? This will make that doc not listed anywhere really.

docs/index.rst
```diff
@@ -161,8 +161,8 @@ of Read the Docs and the larger software documentation ecosystem.
 
 * **Getting involved with Read the Docs**:
   :doc:`Contributing <contribute>` |
+  :doc:`Docker development setup </development/standards>` |
```
davidfischer, line 164:
```suggestion
  :doc:`Development setup </development/standards>` |
```

I don't think there's any need to reference "docker" here. This is just the preferred development setup for working on RTD.

docs/index.rst
ericholscher: Yea -- It should still be in the sidebar. Doesn't seem worth calling out since we're not in it this year?

PR #6876, opened 2020-04-07
Add 12 hour caching to our robots.txt serving

ericholscher:
This should help reduce load on our servers when serving robots.txt.
This file should change very slowly,
and when it does it's not super important for it to change quickly.

This also makes sitemaps the same 12 hours,
since 3 days seemed a bit aggressive.


ericholscher:
This needs the tests to be updated. I'm guessing they are now being cached :)

readthedocs/proxito/views/serve.py
```diff
@@ -355,7 +356,7 @@ class ServeRobotsTXT(SettingsOverrideObject):
 class ServeSitemapXMLBase(View):
 
     @method_decorator(map_project_slug)
-    @method_decorator(cache_page(60 * 60 * 24 * 3))  # 3 days
+    @method_decorator(cache_page(60 * 60 * 12))  # 12 hours
```
humitos, line 359: This makes sense. We are communicating in the sitemap for latest to "daily".

ericholscher:
> Maybe using @override_settings with CACHES={} on these tests?

I tried that and it didn't seem to work, I am explicitly clearing the cache now which worked locally 👍

PR #6877, opened 2020-04-07
Make the status name in CI configurable via setting

ericholscher:
We had the same one on .org/.com and it was overwriting each other.


ericholscher:
Ah yea, I had that locally but didn't push it up 👍

PR #6879, opened 2020-04-08
Allocate docker limits based on server size.

ericholscher:
This assumes 1-builder per server.
It is still able to be overwritten by settings,
so to have this take effect we need to set:

```
DOCKER_LIMITS = None
```

### DIFF
--- a/readthedocs/doc_builder/constants.py
+++ b/readthedocs/doc_builder/constants.py
@@ -35,7 +35,8 @@ DOCKER_LIMITS = {
 # This assumes 1-builder per server
 try:
     total_memory = int(subprocess.check_output("free -m | awk '/^Mem:/{print $2}'", shell=True))
-except Exception:
+except ValueError:
+    # On systems without a `free` command it will return a string to int and raise a ValueError
     log.exception('Failed to get memory size. Using defaults docker limits')
     total_memory = 0
 
--- a/readthedocs/doc_builder/constants.py
+++ b/readthedocs/doc_builder/constants.py
@@ -60,7 +60,8 @@ elif total_memory > 2000:
         'time': 600,
     })
 
-DOCKER_LIMITS.update(settings.DOCKER_LIMITS)
+if hasattr(settings, 'DOCKER_LIMITS'):
+    DOCKER_LIMITS.update(settings.DOCKER_LIMITS)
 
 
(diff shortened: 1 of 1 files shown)


readthedocs/doc_builder/constants.py
```diff
@@ -25,9 +26,43 @@
...
-DOCKER_LIMITS = {'memory': '200m', 'time': 600}
+DOCKER_LIMITS = {
+    'memory': '200m',
+    'time': 600,
+}
+
+# Set docker limits dynamically based on system memory
+# This assumes 1-builder per server
+try:
+    total_memory = int(subprocess.check_output("free -m | awk '/^Mem:/{print $2}'", shell=True))
```
ericholscher, line 37: We could use psutil here: https://pypi.org/project/psutil/ -- but it would add a dependency. Not sure the tradeoff.

humitos:
> `DOCKER_LIMITS = None`

Will need to be set to `{}` because we are doing and .update finally at this line,

https://github.com/readthedocs/readthedocs.org/pull/6879/files#diff-3fbfb50cb375355069e83003f500d2c8R63

ericholscher:
> Wouldn't be better to have these dynamic limits in our -ops repo?

Why? It seems useful for anyone running RTD.

humitos:
> Why? It seems useful for anyone running RTD.

Just asking because we are changing the default from 200m to be ~75% of the total memory in the computer. Only that.

readthedocs/doc_builder/constants.py
agjohnson: I'd worry a little about this call failing for some reason, then falling through to the except because it's a generic Exception, and then we could be in a place where we are setting the default amount on servers without noticing.

`psutil` might give us more confidence in this check, or the except should be much tighter. The failure case is likely MacOS doing something different, or MacOS free/awk doing something different than Linux free/gnu awk, so we could also put this default set operation in a `if DEBUG` inside a `@property` setting. We likely never want to settle on the defaults in prod

ericholscher:
> This logic might fit better inside a settings property method inside global level in constants

Yea, I just wanted to allow users to override it still with a setting if needed. We could have two settings for this, but that just seems confusing?

ericholscher:
I will merge this now, and see how it goes in prod. We can always adjust, but I'd like to test building one deploy builder image today.

PR #6882, opened 2020-04-09
Redirect HTTP -> HTTPS for custom domains

### LINKED ISSUE #4641: Support HTTPS redirect

davidfischer:
- Only redirect if the appropriate flag is set on the Domain (default is not to redirect)
- Use a 302 redirect for now. Ideally this is a 301 redirect, but redirecting other people's domains with a 301 is always a bit risky. If things go smoothly, it's an easy change to switch from a 301 to a 302.

Fixes #4641

### DIFF
--- a/readthedocs/proxito/middleware.py
+++ b/readthedocs/proxito/middleware.py
@@ -87,7 +87,9 @@ def map_host_to_project_slug(request):  # pylint: disable=too-many-return-statem
         if domain.https and not request.is_secure():
             # Redirect HTTP -> HTTPS (302) for this custom domain
             log.debug('Proxito CNAME HTTPS Redirect: host=%s', host)
-            return redirect('https://{}{}'.format(host, request.get_full_path()))
+            resp = redirect('https://{}{}'.format(host, request.get_full_path()))
+            resp['X-RTD-Redirect'] = 'https'
+            return resp
 
         return project_slug
 
--- a/readthedocs/proxito/middleware.py
+++ b/readthedocs/proxito/middleware.py
@@ -8,7 +8,7 @@ Additional processing is done to get the project from the URL in the ``views.py`
 import logging
 
 from django.conf import settings
-from django.shortcuts import render
+from django.shortcuts import render, redirect
 from django.utils.deprecation import MiddlewareMixin
 
 from readthedocs.projects.models import Domain, Project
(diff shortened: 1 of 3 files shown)


readthedocs/proxito/middleware.py
```diff
@@ -83,6 +83,12 @@ def map_host_to_project_slug(request):  # pylint: disable=too-many-return-statem
         project_slug = domain.project.slug
         request.cname = True
         log.debug('Proxito CNAME: host=%s', host)
+
+        if domain.https and not request.is_secure():
+            # Redirect HTTP -> HTTPS (302) for this custom domain
+            log.debug('Proxito CNAME HTTPS Redirect: host=%s', host)
+            return redirect('https://{}{}'.format(host, request.get_full_path()))
```
ericholscher, line 90: If possible, I'd like to see a custom RTD header on here so we know where this request comes from similar to https://github.com/readthedocs/readthedocs.org/blob/d030d67b4ba83974b2d5df16b5bd71254487ebaf/readthedocs/proxito/views/mixins.py#L179

davidfischer: I think I'll do `X-RTD-Redirect` and the value will be `https`. I might use the same header with the value `canonicalize` when doing the domain canonicalization.

davidfischer:
Hold off on merging this. I'm considering a refactor here that might make it easier to canonicalize the domain as well. I might move the actual redirection to `ServeRedirectMixin.system_redirect` and merely set an attribute on the request object in the Middleware (or rather `map_host_to_project_slug`)

davidfischer:
I put an alternative implementation (based on this one) in #6905. That one uses the regular resolver code to actually perform the redirect and I think I like that solution a bit better overall. I'm open to feedback on either/both though.

If we want to roll out the HTTP -> HTTPS change separately from canonicalizing the domain, we can merge this, roll this out, and merge that one in the future.

davidfischer:
I'm going to close this in favor of #6905 which seems closer to rolling out.

PR #6887, opened 2020-04-09
Add CloudFlare Cache tags support

ericholscher:
This lets us purge the CDN more easily.
This requires a version in the files_changed signal.

### DIFF
--- a/readthedocs/proxito/views/mixins.py
+++ b/readthedocs/proxito/views/mixins.py
@@ -126,6 +126,7 @@ class ServeDocsMixin:
         response['X-RTD-Version'] = version_slug
         # Needed to strip any GET args, etc.
         response['X-RTD-Path'] = urlparse(path).path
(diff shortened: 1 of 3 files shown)


readthedocs/proxito/views/mixins.py
```diff
@@ -126,6 +126,8 @@ def _serve_docs_nginx(self, request, final_project, version_slug, path, download
         response['X-RTD-Version'] = version_slug
         # Needed to strip any GET args, etc.
         response['X-RTD-Path'] = urlparse(path).path
+        # Include the project & project-version so we can do larger purges if needed
+        response['Cache-Tags'] = f'{final_project.slug}-{version_slug},{final_project.slug}'
```
humitos, line 130: With this, we can decide to purge "project & version" or the whole "project" (all versions)? That's why we are separating this by a comma?

davidfischer: That's correct. In the docs I linked above, cache tags are comma separated.

PR #6888, opened 2020-04-10
Fix URLs like `/projects/subproject` from 404ing when they don't end with a slash

ericholscher:
This was hit by users.
You can test it here: https://docs.celeryproject.org/projects/kombu


ericholscher:
Looks good locally:

```
-> curl -IL http://test-builds.community.dev.readthedocs.io/projects/subproject
HTTP/1.1 302 Found
Server: nginx/1.17.5
Date: Fri, 10 Apr 2020 21:09:30 GMT
Content-Type: text/html; charset=utf-8
Content-Length: 0
Connection: keep-alive
Location: http://test-builds.community.dev.readthedocs.io/projects/subproject/en/latest/
X-RTD-System-Redirect: True
X-Frame-Options: DENY
X-Content-Type-Options: nosniff
X-XSS-Protection: 1; mode=block
Vary: Accept-Language, Cookie
Content-Language: en
X-Served: Django-Proxito
```

ericholscher:
We also have an issue with single-single subprojects, eg `http://test-builds.community.dev.readthedocs.io/projects/single-version`
where they 404 currently, and with that change, it serves the files but all media breaks (because it needs the `/`)

readthedocs/proxito/views/serve.py
```diff
@@ -87,6 +93,14 @@ def get(self,
         ]):
             return self.system_redirect(request, final_project, lang_slug, version_slug, filename)
 
+        # Handle `/projects/subproject` URL redirection
+        if all([
+                final_project.single_version,
```
humitos, line 98: I'm not sure to understand why we are only checking this case when its parent project is Single Version. Why not when it's not?

humitos:
> Test are not passing for some reason as well.

Ah, we need to update the matching of the URLs because there is a new attribute now.

readthedocs/proxito/views/serve.py
ericholscher: We only check this case when it's at the end of a URL, which isn't true for any other URL pattern.

humitos: Ah! OK. It wasn't working previously because we were forcing a trailing `/` in the URL to match the regex!

humitos:
Weird... our own build on RTD failed because of timeout. Is our project already _that_ big?

PR #6895, opened 2020-04-13
On Azure .exists blob timeout, log the exception and return False

humitos:
We have been experimenting some issues with Azure Blob storage. Our current timeout is the django-storages default (20 secs). We want to reduce it to ~2s, but without breaking the code and start returning 500. So, we log the error and return False.

### DIFF
--- a/readthedocs/storage/azure_storage.py
+++ b/readthedocs/storage/azure_storage.py
@@ -34,7 +34,7 @@ class AzureBuildMediaStorage(BuildMediaStorageMixin, OverrideHostnameMixin, Azur
         """
         return super().url(name, expire)
 
-    def exists(self, name, timeout=None):
+    def exists(self, name):
         """Override to catch timeout exception and return False."""
         try:
             return super().exists(name)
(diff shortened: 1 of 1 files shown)


humitos:
I'm not sure that this really works. I wasn't able to reproduce it locally with lower azure timeout, or even stopping the storage container.

readthedocs/storage/azure_storage.py
```diff
@@ -30,6 +34,14 @@ def url(self, name, expire=None, http_method=None):  # noqa
         """
         return super().url(name, expire)
 
+    def exists(self, name, timeout=None):
```
ericholscher, line 37: We don't actually use the `timeout` here, so probably shouldn't accept it.

humitos: Meh! I was working in the other idea about passing a timeout here and forgot to remove it :/

humitos:
Yeah. Feel the same. I don't have another idea about how to test it, but I don't understand why it does not work when stopping the storage container either. The custom `.exists` function is executed (I put a pdb there and it stopped), but the timeout seems to not be respected.

I think _the code_ won't fail, but it may just does not do what we expect it to do :confused:

ericholscher:
@humitos I'm happy to merge it once we fix the tests, and we can see what happens in prod.

PR #6897, opened 2020-04-13
Load YAML files safely

humitos:
Use a custom YAML loader to ignore all the unknown tags. This allow us to
support YAML files that have custom tags that are irrelevant for Read the Docs)
but are required for MkDocs to build the documentation properly.

Close: #6889

### DIFF
--- a/readthedocs/config/parser.py
+++ b/readthedocs/config/parser.py
@@ -4,8 +4,6 @@
 
 import yaml
 
-from .utils import yaml_load_safely
-
 __all__ = ('parse', 'ParseError')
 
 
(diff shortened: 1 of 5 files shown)


ericholscher:
Shouldn't this add a test that it actually ignores unknown tags?

humitos:
@ericholscher this is the test https://github.com/readthedocs/readthedocs.org/pull/6897/files#diff-c5e29e92356c212b708de92ebb71be8cR15

ericholscher:
I'm still not convinced this is a great UX for users. Their YAML files will now load but be missing data that they would expect to be in it. I understand the use-case for parsing a users mkdocs.yml, but we shouldn't ever support this for our own config file, right?

humitos:
That's true. The correct approach would be to ignore these tags when reading the mkdocs.yml but keep using the regular `yaml.safe_load` for our own `.readthedocs.yaml` file.

PR #6898, opened 2020-04-14
Allow CORS for Embed API in our PR builds domain

humitos:
This PR allows us to use the Embed API in our PR builds.

I'm not sure if this could be a security issue, but I think it's not. It would be good to have some opinions on this. I'm assuming that if you were able to create a PR, you have access to the whole docs repository already.

### DIFF
--- a/readthedocs/core/signals.py
+++ b/readthedocs/core/signals.py
@@ -54,6 +54,14 @@ def decide_if_cors(sender, request, **kwargs):  # pylint: disable=unused-argumen
         if domain.exists():
             return True
 
+    if all([
+            request.path_info.startswith('/api/v2/embed/'),
+            request.method in SAFE_METHODS,
+            host == settings.RTD_EXTERNAL_VERSION_DOMAIN,
+    ]):
+        # Allow CORS for embed API in our PR builds
+        return True
+
     valid_url = False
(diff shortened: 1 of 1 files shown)


ericholscher:
I'd prefer to have this be a general solution. Do the other endpoints eg. footer break on the PR builds? We should make it so that the PR build domain is accepted for a project, not special-case only the embed API.

humitos:
`/api/v2/footer_html/` works, but it seems we are managing that CORS somewhere else, not in this function. `/api/v2/docsearch/` fails because CORS.

So, if we want to have a general solution, we could,

1. add `r'^https?://(.+)\.readthedocs\.build$'` to `CORS_ORIGIN_REGEX_WHITELIST` instead: https://github.com/readthedocs/readthedocs.org/blob/master/readthedocs/settings/base.py#L473-L476

2. add `/api/v2/` in the `WHITELIST_URLS` variable in our own function, and checking it could be our domain as well at https://github.com/readthedocs/readthedocs.org/blob/master/readthedocs/core/signals.py#L75-L80

ericholscher:
@humitos #1 seems like the best option 👍

humitos:
OK. I saw that you opened the PR in -ops. I'm closing this one since it's not required anymore.

PR #6900, opened 2020-04-14
Fix single version infinite redirect

ericholscher:


readthedocs/proxito/views/serve.py
```diff
@@ -97,7 +97,7 @@ def get(self,
         if all([
                 final_project.single_version,
                 filename == '',
-                not subproject_slash,
+                subproject_slug and not subproject_slash,
```
humitos, line 100:
```suggestion
                subproject_slug,
                not subproject_slash,
```

humitos: hrm... what would be the value of `subproject_slug` in this case?

PR #6907, opened 2020-04-14
Split X-RTD-Version-Method header into two HTTP headers.

humitos:
* Version-Method reflects how the version slug was obtained:
  * external_domain: slug from external versions' domain (split after --)
  * url: slug was taken directly from the URL
* Project-Method indicates how the project slug was obtained
  * rtdheader: X-RTD-SLUG header
  * subdomain: slug is the first part in the host before readthedocs.io
  * cname: slug is from a custom domain


ericholscher:
This seems technically correct, but I don't know if it's useful. Having one header for all of these seems like the best UX, I don't want to be returning a million headers, so I'd just rename it `X-Project-Method` if you care about it.

readthedocs/proxito/views/mixins.py
```diff
@@ -129,13 +129,15 @@ def _serve_docs_nginx(self, request, final_project, version_slug, path, download
...
-            response['X-RTD-Version-Method'] = 'subdomain'
+            response['X-RTD-Project-Method'] = 'subdomain'
+        if hasattr(request, 'cname'):
+            response['X-RTD-Project-Method'] = 'cname'
         if hasattr(request, 'external_domain'):
             response['X-RTD-Version-Method'] = 'external_domain'
-        if hasattr(request, 'cname'):
-            response['X-RTD-Version-Method'] = 'cname'
+        else:
+            response['X-RTD-Version-Method'] = 'url'
```
ericholscher, line 140: Let's remove this logic entirely and only send it on `external_domain`, that removes a lot of my issue with it.

readthedocs/proxito/views/mixins.py
```diff
@@ -129,13 +129,15 @@ def _serve_docs_nginx(self, request, final_project, version_slug, path, download
         # Include the project & project-version so we can do larger purges if needed
         response['Cache-Tags'] = f'{final_project.slug}-{version_slug},{final_project.slug}'
         if hasattr(request, 'rtdheader'):
-            response['X-RTD-Version-Method'] = 'rtdheader'
+            response['X-RTD-Project-Method'] = 'rtdheader'
         if hasattr(request, 'subdomain'):
-            response['X-RTD-Version-Method'] = 'subdomain'
+            response['X-RTD-Project-Method'] = 'subdomain'
+        if hasattr(request, 'cname'):
```
ericholscher, line 135: These should be `elif`'s

readthedocs/proxito/views/mixins.py
humitos: If we don't send this header from Proxito, how would be the NGINX config? Do we need to add an `if` there to avoid checking for this variable?

ericholscher: I'm not sure. I guess this is fine for now if we care, but I think for PR's, we know where the version came from, so I don't know if this is super useful. 🤷

PR #6911, opened 2020-04-15
Add hoverxref to our docs

ericholscher:


tox.ini
```diff
@@ -31,6 +31,10 @@ commands =
 [testenv:docs]
 description = build readthedocs documentation
 changedir = {toxinidir}/docs
+setenv =
+    {[testenv]setenv}
+    READTHEDOCS_PROJECT=docs
+    READTHEDOCS_VERSION=latest
```
davidfischer, line 37: These envvars need to be set in `docs/Makefile` or otherwise set in `docs/conf.py`. Otherwise the normal way of building locally `cd docs && make html` will fail with:

```
Warning, treated as error:
hoverxref extension is not fully configured. Set hoverxref_project and hoverxref_version in your conf.py file.
make: *** [html] Error 2
```

humitos: In fact, these two configs is something that have been bothering me and other users as well. See https://github.com/readthedocs/sphinx-hoverxref/issues/59

I want to communicate this problem to users to avoid unexpected behaviors and I think the correct way to do that is using `sphinx.utils.logging.warning`. Although, many users use `-W` in their builds and they break.

I'd like if you want to jump to that issue and give your opinion there. I feel that I should just change them to be INFO instead to avoid breaking builds randomly.

ericholscher: Yea, I'd prefer not to have to set it, but I wanted the docs to properly build. 👍 on moving it to an `INFO`.

ericholscher: I've removed the env changes, will wait on a PR to the extension for merging.

humitos: PR at https://github.com/readthedocs/sphinx-hoverxref/pull/64

humitos:
Now that the PR where merged in master, this build should just work.

ericholscher:
Rebased & pushed i t up.

ericholscher:
🎉

PR #6921, opened 2020-04-17
Add basic support for lower priority PR builds

ericholscher:
To test this I added this to our tasks:

```
@app.task(ignore_result=True)
def priority_test(priority):
    import time
    time.sleep(1)
    print(priority)
```

Then ran this script in a `manage.py shell_plus`:

```
from readthedocs.projects.tasks import priority_test

for i in range(50):
    priority_test.apply_async((i%10,), queue="default", priority=i%10)

from redis import Redis
r = Redis.from_url(settings.BROKER_URL)
for k in r.keys('default*'):
    print(k.decode())
    print(r.llen(k))
    #print(r.delete(k))
```


readthedocs/core/utils/__init__.py
```diff
@@ -148,6 +148,11 @@ def prepare_build(
         # Send Webhook notification for build triggered.
         send_notifications.delay(version.pk, build_pk=build.pk, email=False)
 
+    if version.type == EXTERNAL:
+        # External builds should be lower priority.
+        # We only support [1,2] currently, so set it to 2 (higher number being higher priority)
```
humitos, line 153:
```suggestion
        # We only support [1,2] currently, so set it to 1 (higher number being higher priority)
```

readthedocs/settings/base.py
```diff
@@ -336,6 +336,14 @@ def USE_PROMOS(self):  # noqa
     CELERYD_PREFETCH_MULTIPLIER = 1
     CELERY_CREATE_MISSING_QUEUES = True
 
+    CELERY_DEFAULT_PRIORITY = 1
```
humitos, line 339: Shouldn't this be 2, so all the tasks go with high priority but only PR with low priority?

humitos: Or here 1, but 0 on the PR option.

humitos:
The example shows it works locally. It puts 20 tasks on each queue (it shows 19 because I suppose 2 are consumed immediately by the workers and they are gone when it checks redis). After that, all the tasks with Priority 1 are consumed first and then all the tasks with priority 0.

ericholscher:
@humitos I cleaned up the logic around it now that we have it working.

readthedocs/projects/constants.py
```diff
@@ -374,3 +374,9 @@
 # Git provider names
 GITHUB_BRAND = 'GitHub'
 GITLAB_BRAND = 'GitLab'
+
+# Set 3 priorities, [low, medium, high] -- default is medium
+# Leave some space on each side of the set to expand if needed
+CELERY_LOW = 3
+CELERY_MEDIUM = 5
+CELERY_HIGH = 7
```
humitos, line 382: nit: maybe these fit better in `builds.contants`?

readthedocs/core/utils/__init__.py
```diff
@@ -150,6 +151,11 @@ def prepare_build(
         # Send Webhook notification for build triggered.
         send_notifications.delay(version.pk, build_pk=build.pk, email=False)
 
+    options['priority'] = CELERY_HIGH
+    if version.type == EXTERNAL:
```
humitos, line 155: Want we check for being a translation here before checking if external and assigning `CELERY_MEDIUM`?

readthedocs/rtd_tests/tests/test_core_utils.py
```diff
@@ -169,6 +171,7 @@ def test_trigger_build_rounded_time_limit(self, update_docs):
             'queue': mock.ANY,
             'time_limit': 3,
             'soft_time_limit': 3,
+            'priority': CELERY_HIGH,
```
humitos, line 174: We should modify/add a test that triggers a PR and check it's CELERY_LOW.

ericholscher:
Addressed review feedback, going to ship this and QA this week 👍

PR #6923, opened 2020-04-17
Docs on supporting root domains

davidfischer:
I tested this process end-to-end and was able to add an ALIAS record on namecheap's DNS, Cloudflare issued the SSL cert, and RTD served it without issue.

### DIFF
--- a/docs/custom_domains.rst
+++ b/docs/custom_domains.rst
@@ -27,8 +27,8 @@ You can also host your documentation from your own domain.
 
 .. note::
 
-   We don't currently support pointing subdomains or naked domains to a project using ``A`` records.
-   It's best to point a subdomain, ``docs.example.com`` for example, using a CNAME record.
+   We don't currently support pointing subdomains or root domains to a project using A records.
+   DNS A records require a static IP address and our IPs may change without notice.
 
 .. tabs::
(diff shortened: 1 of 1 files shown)


docs/custom_domains.rst
```diff
@@ -27,16 +27,17 @@ You can also host your documentation from your own domain.
...
 
 .. tabs::
 
    .. tab:: Read the Docs Community
       
       In order to setup your custom domain, follow these steps:
 
-      #. Add a CNAME record in your DNS that points the domain to ``readthedocs.io``
+      #. For a subdomain like ``docs.example.com``, add a CNAME record in your DNS that points the domain to ``readthedocs.io``.
+         For a root domain like ``example.com`` use an ANAME or ALIAS record pointing to ``readthedocs.io``.
```
humitos, line 40: Is this also valid for commercial?

davidfischer: I thought I responded to this. I did not test this fully on .com. I don't see any reason why it wouldn't work but I think we should end-to-end test it before blessing it for commercial. Currently, it is just in the .org tab.

ericholscher:
Merging this for deploy this week 👍

PR #6925, opened 2020-04-17
CAA records are not only for users of Cloudflare DNS

davidfischer:
This change moves the CAA record note out of the Cloudflare note. Users of any DNS provider can configure CAA records which prevent us/Cloudflare from issuing SSL certificates.

I also noted that if somebody has a CAA record for the root domain that it also applies to the subdomain.

### DIFF
--- a/docs/custom_domains.rst
+++ b/docs/custom_domains.rst
@@ -63,22 +63,26 @@ You can also host your documentation from your own domain.
          they do not yet allow us to acquire SSL certificates for those domains.
          Follow the new setup to have a SSL certificate.
 
-      .. warning:: Notes for Cloudflare users
+      .. admonition:: Certificate Authority Authorization (CAA)
 
-         - If your domain has configured CAA records, please do not forget to include
-           Cloudflare CAA entries, see their `Certification Authority Authorization (CAA)
-           FAQ <https://support.cloudflare.com/hc/en-us/articles/115000310832-Certification-Authority-Authorization-CAA-FAQ>`__.
+         If your custom domain — either the subdomain you're using or the root domain — has configured CAA records,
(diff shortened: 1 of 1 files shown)


ericholscher:
I wonder if we should move some of these into a troubleshooting section, instead of inside the main .org docs for this?

davidfischer:
> I wonder if we should move some of these into a troubleshooting section, instead of inside the main .org docs for this?

CAA also applies on .com although customers will need to add Amazon to the CAA records.

ericholscher:
> CAA also applies on .com although customers will need to add Amazon to the CAA records.

Right, but currently its only shown in the .org tab of the docs. We should probably add a troubleshooting section that mentions this, and has tabs for the proper config for .com/.org as we do for custom domains. 

https://docs--6925.org.readthedocs.build/en/6925/custom_domains.html#custom-domain-support

davidfischer:
I added a separate CAA admonition for Amazon in the RTD for Business tab.

docs/custom_domains.rst
```diff
@@ -63,22 +63,26 @@ You can also host your documentation from your own domain.
          they do not yet allow us to acquire SSL certificates for those domains.
          Follow the new setup to have a SSL certificate.
 
-      .. warning:: Notes for Cloudflare users
+      .. admonition:: Certificate Authority Authorization (CAA)
```
ericholscher, line 66: We should explicitly mention what record users need to add. The Cloudflare docs on this aren't super useful, having just walked a user through it. I believe we need the digicert records.

Should we consider migrating to Lets Encrypt? Everyone I've seen that has misconfigured this, has set it to LE, so it might solve a lot of the issues.

ericholscher: I pushed a small patch to add this.

PR #6926, opened 2020-04-17
Index date and ID together on builds

davidfischer:
Currently the production Django admin for Builds times out. This is a fix for that.

- The Django admin [automatically adds the ID to the ordering](https://docs.djangoproject.com/en/2.2/ref/contrib/admin/#django.contrib.admin.ModelAdmin.ordering) in order to get a deterministic sort order.
- In our setup, this *greatly* slows down the query since date and ID are not indexed together.
- ~~This index may be very slow to add on .org where the Build table has 6.8M rows.~~ In testing, it was added in less than a minute.

Below is the query plan in production. Virtually all the time is spent on sorting.
```
Limit  (cost=1301869.22..1301869.72 rows=200 width=805) (actual time=42943.729..42943.815 rows=200 loops=1)
   ->  Sort  (cost=1301869.22..1317972.23 rows=6441205 width=805) (actual time=42943.723..42943.797 rows=200 loops=1)
         Sort Key: builds_build.date DESC, builds_build.id DESC
         Sort Method: top-N heapsort  Memory: 266kB
         ->  Hash Join  (cost=39213.29..1023484.97 rows=6441205 width=805) (actual time=947.943..35212.336 rows=6785440 loops=1)
               Hash Cond: (builds_build.project_id = projects_project.id)
               ->  Merge Right Join  (cost=8.83..489575.94 rows=6441205 width=223) (actual time=0.118..12322.444 rows=6785440 loops=1)
                     Merge Cond: (builds_version.id = builds_build.version_id)
                     ->  Index Scan using builds_version_pkey on builds_version  (cost=0.43..81188.92 rows=1584185 width=85) (actual time=0.006..1396.431 rows=1590883 loops=1)
                     ->  Index Scan using builds_build_version_id_cd1a89966ebc2f3 on builds_build  (cost=0.56..324291.61 rows=6441205 width=138) (actual time=0.005..7627.731 rows=6785440 loops=1)
               ->  Hash  (cost=20343.87..20343.87 rows=217487 width=582) (actual time=928.183..928.183 rows=217648 loops=1)
                     Buckets: 8192  Batches: 64  Memory Usage: 1513kB
                     ->  Seq Scan on projects_project  (cost=0.00..20343.87 rows=217487 width=582) (actual time=0.011..184.260 rows=217648 loops=1)
 Planning time: 0.699 ms
 Execution time: 42944.572 ms
```

~~I'm not 100% sure this will fix the problem but~~ it should be pretty easy to verify before we roll it out:

```sql
CREATE INDEX CONCURRENTLY deterministic_build_sort ON builds_build ("date", "id");
-- Ensure that the Builds admin loads in the Django admin
DROP INDEX CONCURRENTLY deterministic_build_sort;
```


davidfischer:
> However, won't this fail if we click on another field that it's not date in that page, like state?

Probably. However, to filter by state we probably want to use the filters not the sorts. Those already work in production.

> Another different approach could be to override ModelAdmin.get_ordering

I can give this a try.

ericholscher:
Though I guess its a DB migration, so not easy to deploy to a single server :)

davidfischer:
> Another different approach could be to override ModelAdmin.get_ordering

I tested this and the deterministic build ordering overrides this method. Specifically, even with a method like:

```python
    def get_ordering(self, request):
        return ['-date']
```

...the final query generated by the admin ends with `ORDER BY "builds_build"."date" DESC, "builds_build"."id" DESC`. It looks like there is a [different solution](https://stackoverflow.com/questions/32419190/django-admin-incorrectly-adds-order-by-into-query) but it is not as clean.

davidfischer:
> I think this is fine. However, won't this fail if we click on another field that it's not `date` in that page, like `state`? I suppose that Django will still add `id` and we will end up ordering by `state` _and_ `id` which is again, not indexed.

Actually, the final query when you sort by state ends with `ORDER BY "builds_build"."state" ASC, "builds_build"."date" DESC, "builds_build"."id" DESC`. It's possible that a (date, ID) index might still help there!

davidfischer:
I tested adding the index in production. It does allow the build admin to load. Filtering works like a charm. However, sorting by other columns can cause the build admin to still time out. Also searching can also trigger timeouts. I still think having this index is better than not having it as this gives an easy look into what builds are in which states across our servers from the admin.

PR #6944, opened 2020-04-22
Add an unresolver similar to our resolver

ericholscher:
Turn a URL into the component parts that our views would use to process them.

Look at the tests for an example.

### DIFF
--- a/readthedocs/core/unresolver.py
+++ b/readthedocs/core/unresolver.py
@@ -36,7 +36,7 @@ class UnresolverBase:
 
         # Handle returning a response
         if hasattr(project_slug, 'status_code'):
-            return UnresolvedObject(None, None, None, None, None)
+            return None
 
         _, __, kwargs = url_resolve(
             path,
--- a/readthedocs/core/unresolver.py
+++ b/readthedocs/core/unresolver.py
@@ -1,0 +1,68 @@
+import logging
+from urllib.parse import urlparse
+from collections import namedtuple
+
+from django.urls import resolve as url_resolve
+from django.test.client import RequestFactory
+
+from readthedocs.core.utils.extend import SettingsOverrideObject
+from readthedocs.proxito.middleware import map_host_to_project_slug
+from readthedocs.proxito.views.utils import _get_project_data_from_request
+from readthedocs.proxito.views.mixins import ServeDocsMixin
+
+log = logging.getLogger(__name__)
+
+UnresolvedObject = namedtuple(
+    'Unresolved', 'project, language_slug, version_slug, filename, fragment')
+
+
+class UnresolverBase:
+
+    def unresolve(self, uri):
+        """
+        Turn a URL into the component parts that our views would use to process them.
+
+        This is useful for lots of places,
+        like where we want to figure out exactly what file a URL maps to.
+        """
+        parsed = urlparse(uri)
+        domain = parsed.netloc.split(':', 1)[0]
+        path = parsed.path
+
+        # TODO: Make this not depend on the request object,
+        # but instead move all this logic here working on strings.
+        request = RequestFactory().get(path=path, HTTP_HOST=domain)
+        project_slug = request.host_project_slug = map_host_to_project_slug(request)
+
+        # Handle returning a response
+        if hasattr(project_slug, 'status_code'):
+            return UnresolvedObject(None, None, None, None, None)
+
+        _, __, kwargs = url_resolve(
+            path,
+            urlconf='readthedocs.proxito.urls',
+        )
+
+        mixin = ServeDocsMixin()
+        version_slug = mixin.get_version_from_host(request, kwargs.get('version_slug'))
+
+        final_project, lang_slug, version_slug, filename = _get_project_data_from_request(  # noqa
+            request,
+            project_slug=project_slug,
+            subproject_slug=kwargs.get('subproject_slug'),
+            lang_slug=kwargs.get('lang_slug'),
+            version_slug=version_slug,
+            filename=kwargs.get('filename', ''),
+        )
+
+        log.info('Unresolved: %s', locals())
+        return UnresolvedObject(final_project, lang_slug, version_slug, filename, parsed.fragment)
+
+
+class Unresolver(SettingsOverrideObject):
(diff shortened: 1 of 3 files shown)


readthedocs/core/unresolver.py
```diff
@@ -0,0 +1,41 @@
...
+
+    final_project, lang_slug, version_slug, filename = _get_project_data_from_request(  # noqa
+        request,
+        project_slug=project_slug,
+        subproject_slug=kwargs.get('subproject_slug'),
+        lang_slug=kwargs.get('lang_slug'),
+        version_slug=kwargs.get('version_slug'),
+        filename=kwargs.get('filename', ''),
+    )
+    return (final_project, lang_slug, version_slug, filename)
```
humitos, line 41: nit: maybe is better to return an object, so we can access it via attributes: `url.filename`,etc. It could be a class itself or a namedtuple.

readthedocs/core/unresolver.py
```diff
@@ -0,0 +1,41 @@
+from urllib.parse import urlparse
+
+from django.urls import resolve as url_resolve
+from django.test.client import RequestFactory
+
+from readthedocs.proxito.middleware import map_host_to_project_slug
+from readthedocs.proxito.views.utils import _get_project_data_from_request
+
+
+def unresolve(uri):
```
humitos, line 10: nit: making this a class will help us to extend it later if needed (similar to the resolver class)

readthedocs/core/unresolver.py
ericholscher: Yea, this is just an example.

readthedocs/core/unresolver.py
```diff
@@ -0,0 +1,68 @@
+import logging
+from urllib.parse import urlparse
+from collections import namedtuple
+
+from django.urls import resolve as url_resolve
+from django.test.client import RequestFactory
+
+from readthedocs.core.utils.extend import SettingsOverrideObject
+from readthedocs.proxito.middleware import map_host_to_project_slug
+from readthedocs.proxito.views.utils import _get_project_data_from_request
```
davidfischer, line 10: If we're using it outside of proxito, it probably shouldn't be private (underscore prefixed).

readthedocs/core/unresolver.py
```diff
@@ -0,0 +1,68 @@
...
+        path = parsed.path
+
+        # TODO: Make this not depend on the request object,
+        # but instead move all this logic here working on strings.
+        request = RequestFactory().get(path=path, HTTP_HOST=domain)
+        project_slug = request.host_project_slug = map_host_to_project_slug(request)
+
+        # Handle returning a response
+        if hasattr(project_slug, 'status_code'):
+            return UnresolvedObject(None, None, None, None, None)
```
humitos, line 39: Why not just return `None` instead?

readthedocs/rtd_tests/tests/test_unresolver.py
```diff
@@ -0,0 +1,85 @@
...
+import pytest
+
+from readthedocs.rtd_tests.tests.test_resolver import ResolverBase
+from readthedocs.core.unresolver import unresolve
+from readthedocs.projects.models import Domain
+
+
+@override_settings(
+    PUBLIC_DOMAIN='readthedocs.io',
+    RTD_EXTERNAL_VERSION_DOMAIN='dev.readthedocs.build',
```
humitos, line 12: nit: I think it's better if we use the real production value. It's easier to visualize and compare.

readthedocs/rtd_tests/tests/test_unresolver.py
```diff
@@ -0,0 +1,85 @@
...
+)
+@pytest.mark.proxito
```
humitos, line 14: I don't think proxito settings are required for this. We are just testing an object.

readthedocs/rtd_tests/tests/test_unresolver.py
```diff
@@ -0,0 +1,85 @@
...
+class UnResolverTests(ResolverBase):
+
+    def test_unresolver(self):
+        parts = unresolve('http://pip.readthedocs.io/en/latest/foo.html#fragment')
+        self.assertEqual(parts.project.slug, 'pip')
+        self.assertEqual(parts.language_slug, 'en')
+        self.assertEqual(parts.version_slug, 'latest')
+        self.assertEqual(parts.filename, 'foo.html')
+        self.assertEqual(parts.fragment, 'fragment')
```
humitos, line 23: I love these tests! Super easy to follow! :100:  More "unit test" that our current tests that needs to call Django internals

readthedocs/rtd_tests/tests/test_unresolver.py
```diff
@@ -0,0 +1,85 @@
...
+            canonical=True,
+        )
+        parts = unresolve('http://docs.foobar.com/en/latest/')
+        self.assertEqual(parts.project.slug, 'pip')
+        self.assertEqual(parts.language_slug, 'en')
+        self.assertEqual(parts.version_slug, 'latest')
+        self.assertEqual(parts.filename, '')
+
+    def test_unresolver_single_version(self):
+        self.pip.single_version = True
```
humitos, line 53: Isn't it required to `.save` the project here? :thinking:

readthedocs/core/unresolver.py
humitos: It will be easier to perform checks by checking `if parts` than `if parts.<somefield>` --looks more intuitive to me, since `<somefield>` could be `None` and be a valid object as well (`fragment`, for example)

ericholscher: Makes sense 👍

readthedocs/rtd_tests/tests/test_unresolver.py
ericholscher: This is the dev value, which I think makes sense for testing.

readthedocs/rtd_tests/tests/test_unresolver.py
ericholscher: They import the proxito code tho, at least currently

PR #6949, opened 2020-04-22
Refactor how we handle GitHub webhook events

humitos:
I found this code very hard to follow when trying to understand if we were triggering a sync version twice. I tried to refactor it trying to return immediately we know that we have all the data required to make a decision about "what type of GitHub event we received" trying to minimize the context in needed in the head.

The only special case is the PUSH event with `created`/`deleted` attribute that it's sent together with a CREATE/DELETE event.

> It's easier to review this code in a side-by-side diff.

### DIFF
--- a/readthedocs/api/v2/views/integrations.py
+++ b/readthedocs/api/v2/views/integrations.py
@@ -365,6 +365,34 @@ class GitHubWebhookView(WebhookMixin, APIView):
         return digest.hexdigest()
 
     def handle_webhook(self):
+        """
+        Handle GitHub webhook events.
+
+        It checks for all the events we support currently:
+
+        - PUSH: Triggered on a push to a repository branch. Branch pushes and repository tag pushes
+          also trigger webhook push events.
+
+          .. note::
+
+            ``created`` and ``deleted`` indicate if the push was a branch/tag created or deleted.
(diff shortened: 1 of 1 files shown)


ericholscher:
@humitos what do we need to do to QA this?

humitos:
I'd check the normal flows. There is no changes in the logic.

- opening a PR and triggers a build
- closing a PR removes the version
- pushing to a branch triggers a build
- pushing a new tag that becomes stable triggers a build

We can check the HTTPExchange objects in the RTD admin and/or the Webhook event on GitHub to validate the responses ([PRs responses](https://github.com/readthedocs/readthedocs.org/blob/f90cf35d100c8164fdc7662f2cc749e9f8e88c29/readthedocs/api/v2/views/integrations.py#L221-L225) and [pushing responses](https://github.com/readthedocs/readthedocs.org/blob/f90cf35d100c8164fdc7662f2cc749e9f8e88c29/readthedocs/api/v2/views/integrations.py#L208-L213))

> All of them should do a sync versions (indicated in the response) except for projects with new webhooks (`CREATE/DELETE` event setup) on `PUSH+create=True`, that should be skipped since it will be managed by the `CREATE/DELETE` webhook.

PR #6954, opened 2020-04-23
Add canonical to the Domain listview in the admin

davidfischer:


davidfischer:
> Might be worth making some of these list_editable for easier changes.

Since they affect user/customer domains, I don't change them frequently. Do you? I have actually never edited multiple domains.

ericholscher:
That's a good point -- I guess we'll only edit them once, so doesnt' need to be fast. 👍

PR #6960, opened 2020-04-24
Stop builders gracefully on SIGTERM

humitos:
When SIGTERM is received by the process that is building a build, we only log a
message instead of passing it to the child processes (`pip` running inside the
Docker container, for example).

Otherwise, the `pip` process also receives the SIGTERM and it kills itself,
producing a `exit_status != 0`, registering this as a BuildCommand and making
the whole build to fail. After that, celery "stops gracefully" with a failed build.

This is useful combined with supervisor to make our builders to stop gracefully
in this scenario:

1. supervisorctl stop build
2. celery receives the SIGTERM and logs the warning message
3. supervisor waits for `stopwaitsecs` before sending SIGKILL

### DIFF
--- a/readthedocs/projects/tasks.py
+++ b/readthedocs/projects/tasks.py
@@ -11,6 +11,7 @@ import json
 import logging
 import os
 import shutil
+import signal
 import socket
 import tarfile
 import tempfile
@@ -404,7 +405,6 @@ class SyncRepositoryTaskStep(SyncRepositoryMixin, CachedEnvironmentMixin):
     default_retry_delay=7 * 60,
 )
 def update_docs_task(self, version_pk, *args, **kwargs):
-    import signal
 
     def sigterm_received(*args, **kwargs):
         log.warning('SIGTERM received. Waiting for build to stop gracefully after it finishes.')
--- a/readthedocs/projects/tasks.py
+++ b/readthedocs/projects/tasks.py
@@ -404,6 +404,15 @@ class SyncRepositoryTaskStep(SyncRepositoryMixin, CachedEnvironmentMixin):
     default_retry_delay=7 * 60,
 )
 def update_docs_task(self, version_pk, *args, **kwargs):
+    import signal
+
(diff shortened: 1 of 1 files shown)


readthedocs/projects/tasks.py
```diff
@@ -404,6 +404,15 @@ def run(self, version_pk):  # pylint: disable=arguments-differ
...
 )
 def update_docs_task(self, version_pk, *args, **kwargs):
+    import signal
+
+    def sigterm_received(*args, **kwargs):
+        log.warning('SIGTERM received. Waiting for build to stop gracefully after it finishes.')
+
+    # Do not send the SIGTERM signal to childs
+    # (pip is automatically killed when receives SIGTERM and make the build to fail one command and stop build)
+    signal.signal(signal.SIGTERM, sigterm_received)
```
ericholscher, line 414: I wonder where this code should live really, in this task seems resonable, but I wonder if there's a celery-level function that would make more sense.

humitos: Yes. I had the same concern. I was thinking on `readthedocs.worker` but that's executed on normal runs as well.

humitos: It seems there is a "correct" way of doing this via a Celery plugin. I just found https://github.com/MnogoByte/celery-graceful-stop that replace the SIGTERM handler for the workers as https://github.com/MnogoByte/celery-graceful-stop/blob/master/celery_graceful_stop/bootsteps.py#L35

I will keep reading to see if we can do something similar in a nicer way.

readthedocs/projects/tasks.py
```diff
@@ -404,6 +404,15 @@ def run(self, version_pk):  # pylint: disable=arguments-differ
     default_retry_delay=7 * 60,
 )
 def update_docs_task(self, version_pk, *args, **kwargs):
+    import signal
```
ericholscher, line 407: We should move this import.

Issue #6868, opened 2020-04-07
Allow users to customize the URL Pattern for their projects


ericholscher:
Historically, we've offered a subset of URL patterns for a project. The default is `/$lang/$version/$filename`. We also offer multiple projects on a single domain via subproject & translations: `/projects/$subproject/$lang/$version/$filename`.  But we have a lot of users who want different things from these URLs:

* no `/projects/` or a different string there
* No language but still have a version

## Issue with the current approach

One big issue is we end up with conflicting regexs that can match. For example someone who wanted `/$version/$filename`, with a version and no language, could have a version of `en`. Our regexs wouldn't know how to actually parse this, and we'd have to guess.

## Proposed solution

It would be much cleaner if we could parse a specific project's URLs using a known regex. If we let users configure this, we would then be able to only have 1 regex that we're testing, and always have proper matches. For example in our config:

```
# Only version
readthedocs:
  url:
    format: '/$en/$filename'

# Custom subproject
readthedocs:
  url:
    format: '/components/$subproject/$en/$filename'
```

## Implementation

I think we could add this to the config file. We'd then sync it to the Database on build, and cache it in proxito for serving.

The actual URL resolution could use either the Django URL resolver or just the regex model. I'm imagining something like this:

https://github.com/readthedocs/readthedocs.org/blob/e166c1ccc327bcfd2e41850939ca59b101820f35/readthedocs/proxito/views/serve.py#L173-L178

agjohnson:
I'm having trouble picturing how the actual resolution would work in this case, so i don't have much input there. Either seem reasonable, however I haven't dived deep into the proxito changes yet.

I'm -1 on using the config file for project level settings. Project level settings should be in the UI and database, controlling project level settings in version level config files is confusing UX as it's not easy to see which version is the canonical source. We can skip UI for now, I think the functionality is more important than allowing configuration. Adding a form field can be handled later

ericholscher:
@agjohnson there's a PR that already implements this as a DB field: https://github.com/readthedocs/readthedocs.org/pull/6963

ericholscher:
This is now merged and we're able to support it via an email to our support. I'm going to close this, since it's now implemented. It will be promoted more in the next release of our product, but the code is there 👍

PR #6963, opened 2020-04-25
Add ability for users to set their own URLConf

ericholscher:
This lets users configure how they want their URLs to look.
This could include not having a version or language,
or having a different subpath for subprojects.

Refences https://github.com/readthedocs/readthedocs.org/issues/6868

# Example

This serves docs from `http://kong.community.dev.readthedocs.io/latest/index.html` by settings `project.urlconf = '$version/$filename`. 

Screenshot 2020-04-25 14 31 36

From the logs:

```
[25/Apr/2020 21:31:19] readthedocs.proxito.middleware:134[586]: INFO Setting URLConf: project=Kong, urlconf=$version/$filename, real_urlconf=(?P<version_slug>(?:[a-z0-9A-Z][-._a-z0-9A-Z]*?))/(?P<filename>(?:.*))
```

# TODO

There's still a bunch of work we need to do to make this work:

- [x] Make the resolver respect the users URLs
- [x] Change a lot of our other logic in places that depends on existing URL structure
- [x] Lots and lots of QA testing


readthedocs/proxito/middleware.py
```diff
@@ -113,4 +117,29 @@ def process_request(self, request):  # noqa
...
+        project = Project.objects.get(slug=request.host_project_slug)
+
+        # This is hacky because Django wants a module for the URLConf,
+        # instead of also accepting string
+        if project.has_feature(Feature.PROJECT_URL_ROUTES) and project.urlconf:
+            class fakeurlconf:
+                urlpatterns = [
+                    re_path(project.real_urlconf, ServeDocs.as_view()),
+                    re_path('^' + project.real_urlconf, ServeDocs.as_view()),
+                    re_path('^/' + project.real_urlconf, ServeDocs.as_view()),
```
ericholscher, line 129: Just for debugging

readthedocs/proxito/middleware.py
```diff
@@ -113,4 +117,29 @@ def process_request(self, request):  # noqa
...
+                ]
+
+            log.info(
+                'Setting URLConf: project=%s, urlconf=%s, real_urlconf=%s',
+                project, project.urlconf, project.real_urlconf
+            )
+
+            # Stop Django from caching URLs
+            ns = time.mktime(time.gmtime())
```
ericholscher, line 138: This would be the model modified date probably

readthedocs/proxito/middleware.py
agjohnson: I had to bring in the 404, 500, and all of the proxito urls here as well.

ericholscher: @agjohnson I didn't need the 404/500 yet, but I might be missing something.

ericholscher: I've gone ahead and added them.

readthedocs/projects/models.py
```diff
@@ -541,13 +555,117 @@ def get_production_media_url(self, type_, version_slug):
...
+
+        class ProxitoURLConf:
+
+            """A URLConf dynamically inserted by Proxito."""
+
+            proxied_urls = [
+                re_path(
+                    r'{proxied_api_url}api/v2/'.format(proxied_api_url=self.proxied_api_url),
+                    include('readthedocs.api.v2.proxied_urls'),
+                    name='fake_proxied_api'
```
humitos, line 643: Why do we need to proxy these URLs with `/hello/_/` prefix (following my previous comment)? Why not just use `_/` as we are currently doing?

readthedocs/proxito/middleware.py
```diff
@@ -166,6 +167,24 @@ def process_request(self, request):  # noqa
...
 
+        project = Project.objects.get(slug=request.host_project_slug)
+
+        # This is hacky because Django wants a module for the URLConf,
+        # instead of also accepting string
+        if project.urlconf:
+
+            # Stop Django from caching URLs
+            project_timestamp = project.modified_date.strftime("%Y%m%d.%H%M%S")
+            url_key = f'rtd.urls.fake.{project.slug}.{project_timestamp}'
```
humitos, line 178:
```suggestion
            url_key = f'readthedocs.urls.fake.{project.slug}.{project_timestamp}'
```

Just for consistency.

readthedocs/proxito/middleware.py
```diff
@@ -166,6 +167,24 @@ def process_request(self, request):  # noqa
...
+
+            log.info(
+                'Setting URLConf: project=%s, url_key=%s, urlconf=%s',
```
humitos, line 181:
```suggestion
                'Setting URLConf: project=%s url_key=%s urlconf=%s',
```

readthedocs/projects/models.py
```diff
@@ -541,13 +555,117 @@ def get_production_media_url(self, type_, version_slug):
...
+                ),
+                re_path(
+                    r'{proxied_api_url}downloads/'
+                    r'(?P<lang_slug>{lang_slug})/'
+                    r'(?P<version_slug>{version_slug})/'
+                    r'(?P<type_>[-\w]+)/$'.format(
+                        proxied_api_url=self.proxied_api_url,
+                        **pattern_opts),
+                    ProjectDownloadMedia.as_view(same_domain_url=True),
+                    name='fake_proxied_downloads'
```
humitos, line 653:
```suggestion
                    name='urlconf_proxied_downloads'
```

Using `user_urlconf_` as prefix for these URL names will make it easier to follow.

readthedocs/projects/models.py
```diff
@@ -201,6 +204,17 @@ class Project(models.Model):
...
         ),
     )
+    urlconf = models.CharField(
+        _('Documentation URL Configuration'),
+        max_length=255,
+        default=None,
+        null=True,
+        help_text=_(
+            'Supports the following keys: $language, $version, $subproject, $filename. '
+            'An example `$language/$version/$filename`.'
```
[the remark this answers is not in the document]
ericholscher: Not sure why `Default:` is better here? There isn't a default for this field.

readthedocs/projects/models.py
```diff
@@ -541,13 +555,117 @@ def get_production_media_url(self, type_, version_slug):
...
+        Used for the proxied_api_host in javascript.
+
+        This needs to start with a slash at the root of the domain,
+        and end without a slash
+        """
+        if self.urlconf:
+            # Add our proxied api host at the first place we have a $variable
+            # This supports both subpaths & normal root hosting
+            url_prefix = self.urlconf.split('$', 1)[0]
+            return '/' + url_prefix.strip('/') + '/_'
```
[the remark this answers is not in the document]
ericholscher: What part don't you understand? It's passed as `proxied_api_host` in JS, which is where we hit the backend proxied API's. 

It's a path in this case, but it can also be a fully formed URL, so we're using the existing name.

readthedocs/projects/models.py
```diff
@@ -541,13 +555,117 @@ def get_production_media_url(self, type_, version_slug):
...
+    def proxied_api_url(self):
+        """
+        Like the api_host but for use as a URL prefix.
+
+        It can't start with a /, but has to end with one.
+        """
+        return self.proxied_api_host.strip('/') + '/'
+
+    @property
+    def real_urlconf(self):
```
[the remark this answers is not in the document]
ericholscher: The point here is that it's converted into a regex, instead of being the user-facing simplified version. I don't think `user_urlconf` conveys that at all.

readthedocs/projects/models.py
```diff
@@ -541,13 +555,117 @@ def get_production_media_url(self, type_, version_slug):
...
+        Convert User's URLConf into a proper django URLConf.
+
+        This replaces the user-facing syntax with the regex syntax.
+        """
+        to_convert = self.urlconf
+
+        # We should standardize these names so we can loop over them easier
+        to_convert = to_convert.replace(
+            '$version',
+            '(?P<version_slug>%s)' % pattern_opts['version_slug']
```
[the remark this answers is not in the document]
ericholscher: This is how our URLConf's already work.

readthedocs/projects/models.py
ericholscher: Because we need to be able to support proxied API's at non-root URLs for a users who want to use a proxied subpath `docs.company.com/rtd-hosted-project/_/`

readthedocs/projects/models.py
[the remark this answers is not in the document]
humitos: Because that URL is what Read the Docs uses by default: `/en/latest/filename.html`. It's not the default of the field, but that's the help text presented to the user.

readthedocs/projects/models.py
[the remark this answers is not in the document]
humitos: I'm reading `user_urlconf` as the urlconf given by the user (`/$language/$version/`) and `django_urlconf` as an already user urlconf parsed and converted to what django expects.

readthedocs/projects/models.py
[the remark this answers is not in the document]
ericholscher: Hrm, maybe not.

readthedocs/projects/models.py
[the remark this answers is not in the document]
ericholscher: I think `regex_urlconf` works reasonably well.

PR #6969, opened 2020-04-27
Limit concurrency in translations

humitos:
It happens that projects using the same repository for all their translations
will trigger multiple builds on each commit. Adding this concurrency limitation
here helps us to protect ourselves when this happens.


readthedocs/api/v2/views/model_views.py
```diff
@@ -289,9 +290,26 @@ class BuildViewSetBase(UserSelectViewSet):
     )
     def running(self, request, **kwargs):
         project_slug = request.GET.get('project__slug')
+        query = Q(project__slug=project_slug)
+
+        project = get_object_or_404(Project, slug=project_slug)
```
ericholscher, line 295: Do we actually want to 404 here? Does the client expect this?

readthedocs/api/v2/views/model_views.py
```diff
@@ -289,9 +290,26 @@ class BuildViewSetBase(UserSelectViewSet):
...
+                Q(project__slug=project_slug)
+            )
+
         queryset = (
             self.get_queryset()
-            .filter(project__slug=project_slug)
+            .filter(query)
             .exclude(state__in=[BUILD_STATE_TRIGGERED, BUILD_STATE_FINISHED])
         )
         return Response({'count': queryset.count()})
```
ericholscher, line 315: I think we want to return a `project_count` and a `translation_count`, because we likely don't want to apply the same logic in both places.

humitos: What would be those cases where we want that difference in logic?

Actually, I was thinking about simplifying this by just moving all the logic to a QuerySet and return just `True` or `False` depending if the limit is reached or not, instead of comparing all of this here and in the `trigger_build` function as well.

So, just `Project.objects.reached_max_concurrency(project)` -> `True` or `False` and we use this in the endpoint and in the `trigger_build` function. This also helps us to override it from corporate to have a different behavior. Similar to what we do with `Project.objects.is_active(project)`.

readthedocs/api/v2/views/model_views.py
humitos: I suppose we want a 404 if the project does not exist, yes. However, we are not currently handling any exception in the client yet.

readthedocs/api/v2/views/model_views.py
ericholscher: I mentioned above, but 50 builds across all translations limited at 2 concurrency is a terrible experience. We likely want to be able to change the logic of how we handle translations vs. project builds.

humitos:
@ericholscher I put the logic all in one place (`BuildQuerySet`) and modify the endpoint to return more information. 

I think the logic for the translations is correct and will make more sense once we have the priority queues, but we can adapt it as we wish there if needed. I currently counts "main + translations" as a whole.

humitos:
Even if we don't want the logic around limiting translations builds (which is easy to adapt/remove), the refactor around the code to put all the login in one place is still valid/good.

readthedocs/projects/tasks.py
```diff
@@ -500,19 +500,18 @@ def run(
...
-                if builds_running >= max_concurrent_builds:
+                try:
+                    response = api_v2.build.concurrent_limit.get(project__slug=self.project.slug)
+                    concurrency_limit_reached = response.get('limit_reached', False)
+                    max_concurrent_builds = response.get(
+                        'max_concurrent',
+                        settings.RTD_MAX_CONCURRENT_BUILDS,
+                    )
+                except Exception:
+                    concurrency_limit_reached = False
```
ericholscher, line 511: We should log this.

PR #6987, opened 2020-04-29
Log sync_repository_task when we run it

ericholscher:
This should make it more obvious what our builders are doing

### DIFF
--- a/readthedocs/projects/tasks.py
+++ b/readthedocs/projects/tasks.py
@@ -359,6 +359,11 @@ class SyncRepositoryTaskStep(SyncRepositoryMixin, CachedEnvironmentMixin):
                 update_on_success=False,
                 environment=self.get_rtd_env_vars(),
             )
+            log.info(
+                'Running sync_repository_task: project=%s, version=%s',
+                self.project.slug,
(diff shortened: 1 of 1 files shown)


readthedocs/projects/tasks.py
```diff
@@ -359,6 +359,11 @@ def run(self, version_pk):  # pylint: disable=arguments-differ
                 update_on_success=False,
                 environment=self.get_rtd_env_vars(),
             )
+            log.info(
+                'Running sync_repository_task: project=%s, version=%s',
```
humitos, line 363: nit:
```suggestion
                'Running sync_repository_task: project=%s version=%s',
```

(just to keep compatibility with others and make parsing logs easier)

ericholscher: @humitos we should probably switch to structlog, which does this for us :)

humitos: I don't know structlog, but I will read about it. In fact, I wanted to suggest to review all our logs for a while, but I don't think we had the time and energy to do that yet. I'd vote +1 to whatever thing that we use that makes our logs better.

Issue #6989, opened 2020-04-29
Request to enable CONDA_APPEND_CORE_REQUIREMENTS feature flag for project cocotb


eric-wieser:
Filing an issue as requested [here](https://docs.readthedocs.io/en/stable/guides/feature-flags.html#feature-flags).

Brought on by https://github.com/cocotb/cocotb/pull/1706, and a stop gap until #6965 is merged and released.

humitos:
Hey @eric-wieser! I just enabled this in your project. Let me know if that worked as you expected.

eric-wieser:
Thanks @humitos.

Is there any way to see the output of `conda env create`? Unfortunately the `--quiet` flag hides all of the information about which package versions were installed.

humitos:
Hrm... I don't think we currently have a way to change that :disappointed:

eric-wieser:
Is it done that way because the logs are huge and usually useless?

humitos:
I'd say yes. In particular because of the progress bar that makes all the output to be broken anyways.

```
-q, --quiet           Do not display progress bar.
```

It seems there is no way to show the useful information without showing the progress bar.

eric-wieser:
Would it be possible to run `conda list` as an extra command then?

humitos:
I suggested that already at https://github.com/readthedocs/readthedocs.org/issues/5006#issuecomment-453824161 but didn't succeed :smile:

PR #6992, opened 2020-04-29
Show a list of packages installed on environment

humitos:
Use `pip list` and `conda list` depending on the Python environment used.

I went ahead and propose an implementation for this since it was requested again at https://github.com/readthedocs/readthedocs.org/issues/6989#issuecomment-621177598 . So, worth re-consider to add it.

Edit: this behavior is under a feature flag.

### Virtualenv

Screenshot_2020-04-29_15-06-32

### Conda

Screenshot_2020-04-29_15-07-10


humitos:
@eric-wieser would this be enough for your use case?

eric-wieser:
Looks ideal, thanks!

agjohnson:
My largest problem with adding debug commands like this is that they don't mean much to users most of time and will make the command output page harder to visually scan for relevant information. It's certainly helpful when you need to debug build issues however.

A few options:

- Use a `debug: true` readthedocs.yaml option. This would be harder for the user to know about and would make debug commands even less used. It is explicit, which is nice though.
- My +1 would probably be behind adding an attribute to BuildCommand for `debug=True` or something similar. We could filter or de-emphasize these commands on the display -- a filter might be as simple as `[ ] Show debug output`.
- Separate but related, we absolutely need to get rid of the silly path prefix that is sprinkled through our command output. It is wholly unimportant to the user and makes the page incredibly hard to scan. This change alone might alter my opinion on whether we need to special case debug output commands in the build output page.

This change is reasonable and seems helpful to users. The discussion above is a more general one about adding more of these types of commands and planning on how to proceed without further degrading the build detail page experience.

eric-wieser:
> and will make the command output page harder to visually scan for relevant information

The output is already grouped and collapsed by command, so users can simply choose not to expand this command.

agjohnson:
> The output is already grouped and collapsed by command, so users can simply choose not to expand this command.

There is still a lot of visual pollution on the output that makes it hard to scan these commands, and the commands we care about are the ones that users interact with the most: sphinx building and pip install.

eric-wieser:
> and the commands we care about are the ones that users interact with the most: sphinx building and pip install.

This is exactly the reason I'm asking for this command in the first place - upon switching from pip to conda (with some embedded pip dependencies), I no longer get any install output from either pip or conda. See https://readthedocs.org/projects/cocotb/builds/10935365/ for an example of my dilemma.

humitos:
> * Use a `debug: true` readthedocs.yaml option.

I don't see too much value on adding this as an option in the yaml and enabling/disabling it per version. I think it's more useful if it's enabled by default and you can interactively hide/show it (as you propose in the following points. You usually wants to debug something that already failed.
 
> * My +1 would probably be behind adding an attribute to BuildCommand for `debug=True` or something similar. We could filter or de-emphasize these commands on the display -- a filter might be as simple as `[ ] Show debug output`.

I'd also vote :+1: for this implementation. I really like this idea.

> * Separate but related, we absolutely need to get rid of the silly path prefix that is sprinkled through our command output.

:+1: on this as well.

@agjohnson All the points mentioned are really good improvements towards a general solution and I like that --In fact, we should create 2 separate issues to track them. However, I don't think we should block this PR on on design decision at this point (or waiting on them to be implemented).

This PR is already using a feature flag to enable this package listing command in a per-project way. So, it should be harmless for most of users and useful for the ones who needs this behavior currently.

agjohnson:
Yup, agreed. I think this PR can continue. The next step is already handled on the template project, removal of the path from the command line. This makes it far easier to scan the output lines for what you're looking for. Whether we take on the engineering time of a `debug=true` command, I'm not 100% sold yet and I think removal of the paths from the command line resolve my issues with the view pollution.

humitos:
Cool! I created two different issues to track those points. I'm removing design decision from this PR so we can continue with the review.

humitos:
@eric-wieser Hi! I enabled this flag in your project. Let me know if it does work as you expected.

eric-wieser:
Thanks, it shows everything I needed to see:

```
# packages in environment at /home/docs/checkouts/readthedocs.org/user_builds/cocotb/conda/latest:
#
# Name                    Version                   Build  Channel
_libgcc_mutex             0.1                        main  
alabaster                 0.7.12                     py_0  
asn1crypto                1.3.0                    py38_0  
babel                     2.8.0                      py_0  
backcall                  0.1.0                    pypi_0    pypi
breathe                   4.18.0                   pypi_0    pypi
ca-certificates           2020.1.1                      0  
cairocffi                 1.1.0                    pypi_0    pypi
cairosvg                  2.4.2                    pypi_0    pypi
certifi                   2020.4.5.1               py38_0  
cffi                      1.14.0           py38h2e261b9_0  
chardet                   3.0.4                 py38_1003  
click                     7.1.2                    pypi_0    pypi
cocotb                    1.4.0.dev0               pypi_0    pypi
commonmark                0.9.1                    pypi_0    pypi
cryptography              2.8              py38h1ba5d50_0  
cssselect2                0.3.0                    pypi_0    pypi
decorator                 4.4.2                    pypi_0    pypi
defusedxml                0.6.0                    pypi_0    pypi
docutils                  0.16                     py38_0  
freetype                  2.9.1                h8a8886c_1  
idna                      2.9                        py_1  
imagesize                 1.2.0                      py_0  
incremental               17.5.0                   pypi_0    pypi
ipython                   7.14.0                   pypi_0    pypi
ipython-genutils          0.2.0                    pypi_0    pypi
jedi                      0.17.0                   pypi_0    pypi
jinja2                    2.11.2                     py_0  
jpeg                      9b                   h024ee3a_2  
ld_impl_linux-64          2.33.1               h53a641e_7  
libedit                   3.1.20181209         hc058e9b_0  
libffi                    3.2.1                hd88cf55_4  
libgcc-ng                 9.1.0                hdf63c60_0  
libpng                    1.6.37               hbc83047_0  
libstdcxx-ng              9.1.0                hdf63c60_0  
libtiff                   4.1.0                h2733197_0  
markupsafe                1.1.1            py38h7b6447c_0  
mock                      4.0.2                      py_0  
ncurses                   6.2                  he6710b0_1  
olefile                   0.46                       py_0  
openssl                   1.1.1g               h7b6447c_0  
packaging                 20.3                       py_0  
parso                     0.7.0                    pypi_0    pypi
pexpect                   4.8.0                    pypi_0    pypi
pickleshare               0.7.5                    pypi_0    pypi
pillow                    7.1.2            py38hb39fc2d_0  
pip                       20.0.2                   py38_1  
prompt-toolkit            3.0.5                    pypi_0    pypi
ptyprocess                0.6.0                    pypi_0    pypi
pycparser                 2.20                       py_0  
pyenchant                 3.0.1                    pypi_0    pypi
pygments                  2.6.1                      py_0  
pyopenssl                 19.1.0                   py38_0  
pyparsing                 2.4.7                      py_0  
pysocks                   1.7.1                    py38_0  
python                    3.8.2                hcf32534_0  
pytz                      2020.1                     py_0  
readline                  8.0                  h7b6447c_0  
readthedocs-sphinx-ext    1.0.3                    pypi_0    pypi
recommonmark              0.6.0                    pypi_0    pypi
requests                  2.23.0                   py38_0  
setuptools                46.1.3                   py38_0  
six                       1.14.0                   py38_0  
snowballstemmer           2.0.0                      py_0  
sphinx                    3.0.3                      py_0  
sphinx-argparse           0.2.5                    pypi_0    pypi
sphinx-issues             1.2.0                    pypi_0    pypi
sphinx_rtd_theme          0.4.3                      py_0  
sphinxcontrib-applehelp   1.0.2                      py_0  
sphinxcontrib-devhelp     1.0.2                      py_0  
sphinxcontrib-domaintools 0.3                      pypi_0    pypi
sphinxcontrib-htmlhelp    1.0.3                      py_0  
sphinxcontrib-jsmath      1.0.1                      py_0  
sphinxcontrib-makedomain  0.1.1                    pypi_0    pypi
sphinxcontrib-qthelp      1.0.3                      py_0  
sphinxcontrib-serializinghtml 1.1.4                      py_0  
sphinxcontrib-spelling    5.0.0                    pypi_0    pypi
sqlite                    3.31.1               h62c20be_1  
tinycss2                  1.0.2                    pypi_0    pypi
tk                        8.6.8                hbc83047_0  
toml                      0.10.0                   pypi_0    pypi
towncrier                 19.2.0                   pypi_0    pypi
traitlets                 4.3.3                    pypi_0    pypi
urllib3                   1.25.8                   py38_0  
wcwidth                   0.1.9                    pypi_0    pypi
webencodings              0.5.1                    pypi_0    pypi
wheel                     0.34.2                   py38_0  
xz                        5.2.5                h7b6447c_0  
zlib                      1.2.11               h7b6447c_3  
zstd                      1.3.7                h0b5b093_0
```

humitos:
Awesome! Thanks for reporting back

PR #6993, opened 2020-04-29
Mention concurrent builds limitation in "Build Process"

humitos:
This is the smallest amount of work to mention this.

However, I think that this whole page needs some work. I'll see if I can start doing that in another PR, but I think it worth to mention this somewhere in the meanwhile.

### DIFF
--- a/docs/builds.rst
+++ b/docs/builds.rst
@@ -5,7 +5,7 @@ Every documentation build has limited resources.
 Our current build limits are:
 
 * 15 minutes of CPU
-* 1GB of RAM memory
+* 3GB of RAM memory
 * 2 concurrent builds
 
 We can increase build limits on a per-project basis,
--- a/docs/builds.rst
+++ b/docs/builds.rst
@@ -11,6 +11,7 @@ Our current build limits are:
 
 * 15 minutes of CPU
 * 1GB of RAM memory
+* 2 concurrent builds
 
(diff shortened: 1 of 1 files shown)


docs/builds.rst
```diff
@@ -1,16 +1,12 @@
...
-Files: `tasks.py`_ - `doc_builder/`_
-
-.. _tasks.py: https://github.com/readthedocs/readthedocs.org/blob/master/readthedocs/projects/tasks.py
-.. _doc_builder/: https://github.com/rtfd/readthedocs.org/tree/master/readthedocs/doc_builder
-
 Every documentation build has limited resources.
 Our current build limits are:
 
 * 15 minutes of CPU
 * 1GB of RAM memory
```
ericholscher, line 8: These changed as well, no?

humitos: Yes. I think the default is 15m and 3GB now. I'll update.

docs/builds.rst
```diff
@@ -130,14 +126,14 @@ by default the image used is ``readthedocs/build:latest``,
 but you can change that using a :doc:`config-file/index`.
 
 .. note::
-   
+
```
ericholscher, line 129: Does this not break?

ericholscher: Guess not, huh.

PR #6994, opened 2020-04-29
Exclusively handle proxito HSTS from the backend

davidfischer:
- This will allow removing the HSTS config for proxito from nginx
- This assumes we want HSTS for public domain subdomains if `settings.PUBLIC_DOMAIN_USES_HTTPS`

I didn't make HSTS for the public domain customizable (setting max-age, subdomains, etc.). However, we could add additional values for that. Both community and RTD for Business will use the same HSTS values.

### DIFF
--- a/readthedocs/proxito/middleware.py
+++ b/readthedocs/proxito/middleware.py
@@ -132,10 +132,30 @@ class ProxitoMiddleware(MiddlewareMixin):
         return None
 
     def process_response(self, request, response):  # noqa
-        """Set the Strict-Transport-Security (HSTS) header for a custom domain if max-age>0."""
-        if hasattr(request, 'domain'):
+        """
+        Set the Strict-Transport-Security (HSTS) header for docs sites.
+
+        * For the public domain, set the HSTS header if settings.PUBLIC_DOMAIN_USES_HTTPS
+        * For custom domains, check the HSTS values on the Domain object.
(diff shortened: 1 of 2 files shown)


readthedocs/proxito/middleware.py
```diff
@@ -132,10 +132,30 @@ def process_request(self, request):  # noqa
...
+          The domain object should be saved already in request.domain.
+        """
+        host = request.get_host().lower().split(':')[0]
+        public_domain = settings.PUBLIC_DOMAIN.lower().split(':')[0]
+
+        hsts_header_values = []
+
+        if not request.is_secure():
+            # Only set the HSTS header if the request is over HTTPS
+            return response
```
ericholscher, line 149: 👍 I ran into this earlier :)

readthedocs/proxito/middleware.py
```diff
@@ -132,10 +132,30 @@ def process_request(self, request):  # noqa
...
+
+        if settings.PUBLIC_DOMAIN_USES_HTTPS and (public_domain in host or host == 'proxito'):
```
ericholscher, line 151: I don't think we want `proxito` here, since thats used locally and we don't have SSL there. 

```suggestion
        if settings.PUBLIC_DOMAIN_USES_HTTPS and public_domain in host:
```

PR #6995, opened 2020-04-29
Add ability to rebuild a specific build

ericholscher:
This is most useful for PR builds,
but I've enabled it for all builds currently.
It might make sense to only enable it for PR builds,
or other builds where we care about rebuilding a specific commit.

It definitely isn't the final UX, but it works reasonably well for now:

Screen Shot 2020-04-29 at 12 20 51 PM

Fixes https://github.com/readthedocs/readthedocs.org/issues/6524


readthedocs/builds/views.py
```diff
@@ -54,14 +54,32 @@ def post(self, request, project_slug):
...
+        commit = request.POST.get('commit')
+        build_pk = request.POST.get('build_pk')
+
         version = get_object_or_404(
             self._get_versions(project),
             slug=version_slug,
         )
 
+        # Set either the build or None
+        build = Build.objects.filter(pk=build_pk).first()
```
saadmk11, line 66: If a `build_pk` from a different project is passed through the form maybe we should check it before we start any tasks.

```suggestion
        build = version.builds.filter(pk=build_pk).first()
```

readthedocs/core/utils/__init__.py
```diff
@@ -207,17 +214,19 @@ def trigger_build(project, version=None, commit=None, record=True, force=False):
     :rtype: tuple
     """
     log.info(
-        'Triggering build. project=%s version=%s commit=%s',
+        'Triggering build. project=%s version=%s commit=%s build=%s',
         project.slug,
         version.slug if version else None,
         commit,
+        build.pk
```
saadmk11, line 221: In the build list page `build_pk` is not passed in the form. So I think trying to get `build.pk` for logging will cause an error.
```suggestion
        build.pk if build else None
```

agjohnson:
:+1:  on the feature, but could we either move the button or replace it with a link? This feels a bit unpolished if we it's going to be live for all projects.

image

ericholscher:
I hit this pretty frequently when something weird happens with the PR builder. I think it's a useful enough thing to support, since other ways of supporting this are weird hacks.

readthedocs/core/utils/__init__.py
```diff
@@ -112,7 +111,21 @@ def prepare_build(
...
+        build.commit = commit
+        build.commands.all().delete()
+        build.save()
+        kwargs['build_pk'] = build.pk
+
+        # Show a step in the UI that the build is being rebuilt
+        BuildCommandResult.objects.create(
+            build=build, command='Rebuilding build', exit_code=0,
+            start_time=datetime.utcnow(), end_time=datetime.utcnow(),
+        )
```
agjohnson, line 126: I'd probably remove this entirely, I don't think it's worth introducing UI confusion in the command output. The best place to solve this would be either popping up a confirmation box for the user on click, or simply revert the state back to show the UI as "loading". Either would communicate to the user that the build is restarting

ericholscher: I'm more worried about other folks looking at it and being confused about why it was running again after it had already built. I think we don't do a great job of showing history on builds -- we hit this issue with failures & retries as well, so we need some kind of pattern for it.

ericholscher: I've updated this to use the new build status, which is a better solution 👍

ericholscher:
This should be ready for a final review with the new status change 👍

humitos:
I took a look at this PR today trying to recover the feature and finish it.

@ericholscher is there a strong reason why we are re-triggering _the same build object_ instead of just trigger a new build with a specific commit? IMO, reading the code just makes it more complex passing the `build` object around different methods, adding a new build status, etc.

I'm thinking that we can just call `trigger_build(commit=)` with the commit received from the dashboard (via `?commit=` or via `?build_pk=` if the build is associated with a particular commit). With this approach, it would create a new `Build` object and we would keep all the build outputs, giving users more context about why the first time it fails and maybe the subsequent re-triggered ones passed.

I'll try opening a PR with this idea.

humitos:
I opened a PR at https://github.com/readthedocs/readthedocs.org/pull/8227 to show my idea but I'm still not convinced this is a good feature and I think it will create more problems than benefits.

While it may be a good feature to "re-build the latest failed build from a PR", it will for sure create problems for all the other use cases that involve an old commit:

- re-build an _old_ commit from this external version
- re-build an _old_ commit from `latest`
- re-build an _old_ commit from `stable`
- re-build an _old_ commit from any other version

In all these cases, the user will be building **and publishing** an old version of their documentation, and they may not even notice that they just break their docs. They will end up with that version of their (now out-of-date) docs pointing to an old commit.

Basically, there is going to be a discrepancy between what you have in your repository and what you have published, breaking one of the main benefits of RTD. Read the Docs' case is different from a regular CI because it does not only do checks but also generates and publishes assets.

Considering the best use case for this feature the one I mentioned in the first paragraph, there are currently at least some workarounds that are not ideal but way safer from a user's perspective:

- add a new commit to the PR, or
- close and re-open the PR

If any, it seems we should only allow to re-build on that particular case. Still, I don't think it adds a lot of value.

astrojuanlu:
I agree with the concerns raised by @humitos. I still think this is a good feature though, provided that we narrow down the scope.

- Reopening and closing the PR also triggers _all_ other CI checks users have. This might be undesirable.
- The most likely reason to do this is that some temporary network failure ruined the build. Therefore, I don't think users will want to build ancient commits. And if they want, they already have other methods.

Therefore, I think we could limit the feature as follows:

- Only rebuilds of PR builds are allowed. No danger in rebuilding an old published version, that should be done with the existing UI.
- Does not allow rebuilding _ancient_ commits. Therefore, we can push a reasonable timeout, say 24 hours.

What do you think? Pinging @pllim , who probably has interest in this (I don't remember if we are discussing the feature in some other issue).

stsewd:
I think the idea was to allow to re-build a PR or version from the latest commit.

pllim:
24 hours might not work for some. A commit can still be a latest commit but weeks old before a reviewer can get to it and realized a rebuild is needed, though I guess in that case, reviewer can open/close as usual and cancel the other CI (not pretty but it is a workaround).

@astrojuanlu , I do vaguely recall talking about this though what I wanted more is the ability to skip building RTD entirely via `[skip ci]`, `[ci skip]`, `[skip rtd]`, and `[rtd skip]`.

astrojuanlu:
Ah yes, that's #7660 :)

ericholscher:
Agreed with @astrojuanlu -- I think we probably only want to allow it for the last build on a PR branch. It should be simple enough to figure this out (`pk == Build.objects.filter(version=pr_version).first()`). I don't feel strongly about reusing a build object or creating a new one. I initially reused it so that the URL would be the same. I don't think either is obviously better in terms of UX, so whatever is simpler is probably best.

pllim:
I cannot wait to put this into good use. Thanks! 😸

dimpase:
> I'm not sure to understand the use case for this. What are the scenarios where you want to re-build the same commit?

In our case we have to rely on our package supplied via a wheel (it depends on an external C++ library, and the rtd builder
cannot handle building the package). Now, our wheels are built on GitHub, using https://github.com/pypa/cibuildwheel
- which is triggered by push/PR.
If the wheel is broken, it need not lead to an error in rtd builder; it often passes, but the docs are broken.
So we cannot rely on the rebuild button only available for errored out builds.

Ideally,  we would like to be able to install wheels on PyPI and only then trigger the rtd rebuild, but we'd be OK with rebuild
button being available for passed builds (after we install the wheels to PyPI, we can just press it...)

astrojuanlu:
@dimpase sorry it took so long to respond. Would an incoming webhook suit your use case? You can trigger builds externally https://docs.readthedocs.io/en/stable/integrations.html#using-the-generic-api-integration

PR #7001, opened 2020-04-29
Start storing JSON data for PR builds

ericholscher:
This will let us do a couple things:

* Make the embed API work in PR previews
* Allow us to start indexing search for PRs once we want that feature-set


ericholscher:
@stsewd this doesn't index search in PR's yet. We need a way to make it operationally less overhead. I'm thinking a totally different search index for PR's, so that they shouldn't effect normal prod search.

PR #7006, opened 2020-04-30
Add Priority 0 to Celery

ericholscher:
We use 0 here because some things still put a task in the queue with no priority
I don't fully understand why, but this seems to solve it.


ericholscher:
Merging this so we can test it a bit more locally before we ship it next week.

PR #7007, opened 2020-05-01
Add proxito headers to redirect responses

ericholscher:
This lets us set the cache and debug headers on all our proxito responses.
This is a step towards letting us cache pages a longer amount of time.

This includes and fixes #6907 as well 👍


davidfischer:
This looks like a cleaner solution!

readthedocs/proxito/middleware.py
```diff
@@ -144,6 +145,8 @@ def process_response(self, request, response):  # noqa
 
         hsts_header_values = []
 
+        self.add_proxito_headers(request, response)
```
davidfischer, line 148: Rather than adding a single-use mixin in a separate file, couldn't this method just be a method on this class?

ericholscher: Yea, I thought we'd have to use it more places, but we haven't yet, so that seems fine. 👍

PR #7019, opened 2020-05-05
Cleanup unresolver logging

ericholscher:


ericholscher:
Seems travis is busted on sending status, it passed.

Issue #3984, opened 2018-04-20
Use Celery Task event methods instead of updating at context manager level


agjohnson:
Raising this because of #3864 and some issues we hit today.

Currently we're using context managers to try to grab and report all the exceptions that we can. We could actually be using `Task.on_failure`[0], `Task.on_success`[1], and `Task.on_retry`[2] though. Instead of grabbing and reporting exceptions to users through the context manager, instead just bubble these up to Celery and update the build state at task failure/sucess.

This would be tried in steps:

* First, break out the update build code from the context manager. This will live in a utility function so it can be shared. Input is build_pk and the build state
* Next, make an exception subclass that can bubble up to Celery. That is, one that is not reported to users, but raises up through the context managers
* Make VCS locking a bubble up exception subclass, use `autoretry_on` to retry the task. (Do we mark as build failed here? Possible, but maybe we get away from that pattern)
* Get rid of generic handling of errors outside buildenvs and use `on_failure` for this

At this point, we should be catching, reporting, and notifiying (we're not, see #3864) through the use of celery.

Next stage, if this is going smoothly, is to reduce the reporting logic on the context managers and see if celery can report these more consistently.

Overall, this will reduce the duplication in logic that we're using to report/update builds, and exceptions will be reported in one spot.

0: http://docs.celeryproject.org/en/latest/userguide/tasks.html#on_failure
1: http://docs.celeryproject.org/en/latest/userguide/tasks.html#on_success
2: http://docs.celeryproject.org/en/latest/userguide/tasks.html#on_retry

PR #7031, opened 2020-05-06
Example about how to kill a running task

humitos:
:warning: **DO NOT MERGE** :warning: 

This PR is an example of how to cancel an already triggered build.

From what I've seen, it would be good to implement https://github.com/readthedocs/readthedocs.org/issues/3984 before working on a solution to cancel builds. Or, implement _only_ the part needed for now to handle the `BuildKilled` exception.

I think implementing that issue first is better because otherwise, we need to handle the exception `BuildKilled` in many different places inside the code of the task itself:

- environment context manager
- _all the methods_ that are ran inside the task itself

The current implementation "works", but it's not handling the exception in all the places needed. So, depending _when_ you click on "Cancel build" you will get a gracefully killed build (proper set of all the attributes --`success=False`, etc) or a hard killed build.

### DIFF
--- a/readthedocs/builds/models.py
+++ b/readthedocs/builds/models.py
@@ -687,6 +687,12 @@ class Build(models.Model):
         _('Cold Storage'),
         help_text='Build steps stored outside the database.',
     )
+    task_id = models.CharField(
+        _('Celery task id'),
+        max_length=36,
+        null=True,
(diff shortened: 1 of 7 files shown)


humitos:
We are de-duplicating builds for now, only. Not cancelling already running builds.

PR #7037, opened 2020-05-06
Use theme release 0.5.0rc1 for docs

agjohnson:
Also upgrade Sphinx to 3+


agjohnson:
The issue was a duplicate definition in `requirements/pip.txt` for Sphinx. I upgraded there but have no clue if we can use the version in our code.

If this works and we can't use sphinx 3.0.3 in code, then i'll make a mess of our requirements files to install Sphinx 3 after the docs requirements file.

ericholscher:
I'm getting a flash of a different font when I first load the docs, not sure whats going on there.

agjohnson:
Opened up https://github.com/readthedocs/sphinx_rtd_theme/issues/913 to address this. The behavior seems to be prod docs produce invisible text while loading and 0.5.0rc1 uses the old behavior for unstyled text.

agjohnson:
I've rebased and i'm testing changes that separate docs dependencies from application dependencies. This is to work around the inability to upgrade docs Sphinx independent of the application Sphinx dependency.

ericholscher:
Believe we should ship this now right?

tox.ini
```diff
@@ -48,7 +48,7 @@ commands =
 
 [testenv:docs-lint]
 description = run linter (rstcheck) to ensure there aren't errors on our docs
-deps = -r{toxinidir}/requirements/docs-lint.txt
+deps = -r{toxinidir}/requirements/docs.txt
```
ericholscher, line 51: Don't we still need rstcheck?

ericholscher: Looks like it's already in `docs.txt` -- so we should remove the docs-lint file.

ericholscher: (removed in a commit)

ericholscher:
@agjohnson looks like we have conflicting version issues, so I think that's why we had `local-docs-build`. I can't figure out a way to make it work :/

ericholscher:
Actually -- I figured it out by hacking the install into the test command 👍

ericholscher:
Shipping!

requirements/testing.txt
```diff
@@ -1,5 +1,5 @@
 -r pip.txt
--r local-docs-build.txt
+-r docs.txt
```
agjohnson: Why do we need docs requirements here at all?

agjohnson:
>  looks like we have conflicting version issues, so I think that's why we had local-docs-build

Yeah, I separated docs from pip requirements because of this. We could add back this line eventually:
https://github.com/readthedocs/readthedocs.org/pull/7037/files#diff-c7e6b68f10ee36406c7a76f4c4d988d6L1

But i'm not convinced that's the best pattern. I think your patch makes sense.

I believe we can also do:

```
deps = 
    -r{toxinidir}/requirements/pip.txt
    -r{toxinidir}/requirements/docs.txt
```

ericholscher:
Yea, I tried that but it runs them both at the same time and conflicts, which is why I had to hack it.

PR #7038, opened 2020-05-06
Fix translation error

luciopileggi:


agjohnson:
Hi @luciopileggi! We managed all of our translation changes using Transifex. Could you migrate this change there so the team there can help?

ericholscher:
Closing this since they should go to transifex, thanks for the contribution!

PR #4585, opened 2018-08-29
Respect user language when caching homepage

agjohnson:
Previously it was using the first language that hit the page to cache.

Fixes #4582

PR #7049, opened 2020-05-08
Cache the HomepageView for an hour

ericholscher:
This should reduce our DB load a good bit,
since the results of this query almost never change.

I thought we were already caching this page,
but it doesn't look like it.


davidfischer:
I think we're just caching the featured projects block. There were some issues around caching respecting language (#4585) but that's about it.

ericholscher:
Hrm ok -- I will have to look through the templates and figure out what is causing all our queries then.

PR #7083, opened 2020-05-14
Document HSTS support

### LINKED ISSUE #7062: Document HSTS support
We now have the ability to configure HSTS for a hosted domain. It's currently an admin-only option, but we should document that it's available via an email to support.

Refs https://github.com/readthedocs/readthedocs.org/issues/6984

davidfischer:
Fixes #7062

### DIFF
--- a/docs/custom_domains.rst
+++ b/docs/custom_domains.rst
@@ -99,6 +99,19 @@ You can also host your documentation from your own domain.
 
          .. _Amazon CAA guide: https://docs.aws.amazon.com/acm/latest/userguide/setup-caa.html
 
+Strict Transport Security
++++++++++++++++++++++++++
+
+By default, we do not return a `Strict Transport Security header`_ (HSTS) for user custom domains.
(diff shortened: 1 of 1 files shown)


docs/custom_domains.rst
```diff
@@ -99,6 +99,19 @@ You can also host your documentation from your own domain.
 
          .. _Amazon CAA guide: https://docs.aws.amazon.com/acm/latest/userguide/setup-caa.html
 
+Strict Transport Security
++++++++++++++++++++++++++
+
+By default, we do not return a `Strict Transport Security header`_ (HSTS) for user custom domains.
+This is a conscious decision as it can be misconfigured in a not easily reversible way.
+For both |org_brand| and |com_brand|, HSTS for custom domains can be set upon request.
```
humitos, line 107: Maybe is worth to mention how to make that request, with the link to the email?

ericholscher: We could mention contacting support, but I think we don't need to include a link here, since it would be different for each site?

PR #7000, opened 2020-04-29
Add an guide about embedding content

ericholscher:
This is a start of talking about our Embed API to users.
It likely needs some cleaning up,
and we need to document the API itself on the API v2 page.

PR #7089, opened 2020-05-18
Guide for Embed API

humitos:
Document two different use cases for the Embed API.

My goals with this PR are:

* communicate that Embed API exists
* show two different ways of usage with real examples
* let the user know that they can build something new on top of it
* promote sphinx-hoverxref
* have clickable working examples

I'm not mentioned https://github.com/agjohnson/readthedocs-client-js because I think it needs more work or at least a review and testing that it's still working. I think it makes sense to just mention and AJAX call for now and then we can update this page when we have polished our `readthedocs-client-js`.

Rendered version https://docs--7089.org.readthedocs.build/en/7089/guides/embedding-content.html


ericholscher:
How does this compare to https://github.com/readthedocs/readthedocs.org/pull/7000?

humitos:
Haha, I totally forgot about that PR :smile:.  However, I'd avoid mentioning our JS client for now (as #7000 does) since we haven't touched or polished yet. Besides, I think this guide is more complete, IMHO.

Please, take a look at this and let me know which you prefer.

ericholscher:
I've pushed some updates here. I moved the hoverxref section up since it's the most functionalky for users. I found the screenshot I found really confusing. I'd crop it more. I reduced the width, since I got confused when reading the page and thought it was the actual page content.

humitos:
I changed the image using one from our own documentation and centered it. I think it looks more differentiated from the page's content itself and better because the tooltip is nicer :)

PR #7095, opened 2020-05-19
Document Embed APIv2 endpoint

humitos:
I added the documentation under https://docs.readthedocs.io/en/stable/api/v2.html. However, we have a note at the top saying that it's deprecated and you should use APIv3, but APIv3 does not have this endpoint yet.

It's probably fine for now to have it here, but we will need to migrate it to APIv3 at some point.

https://docs--7095.org.readthedocs.build/en/7095/api/v2.html

### DIFF
--- a/docs/api/v2.rst
+++ b/docs/api/v2.rst
@@ -343,7 +343,7 @@ Embed
 
         {
             "content": [
-                "..."
+                "<div class=\"section\" id=\"read-the-docs-features\">\n<h1>Read the Docs..."
             ],
             "headers": [
                 {
--- a/docs/api/v2.rst
+++ b/docs/api/v2.rst
@@ -317,6 +317,80 @@ Build detail
 
     Some fields primarily used for UI elements in Read the Docs are omitted.
 
+
+Embed
+~~~~~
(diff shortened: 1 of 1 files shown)


docs/api/v2.rst
```diff
@@ -317,6 +317,84 @@ Build detail
...
+            }
+        }
+
+    :>json string content: HTML content of the section.
+    :>json object headers: section's headers in the document.
+    :>json string url: URL of the document.
+    :>json object meta: meta data of the requested section.
+
+    :query string project: Read the Docs project's slug.
+    :query string version: Read the Docs version's slug.
```
ericholscher, line 386: We should mention these are not required and have a default value (`latest`).

docs/api/v2.rst
```diff
@@ -317,6 +317,84 @@ Build detail
...
+    :query string doc: document to fetch content from.
+    :query string section: section within the document to fetch.
+    :query string path: full path to the document including extension.
```
ericholscher, line 389: These should all mention if they are optional or required, and what the defaults are.

PR #7097, opened 2020-05-19
Add a setting for storing pageviews

ericholscher:
This is so we can disable it if needed based on DB load.


agjohnson:
Would this make more sense behind a feature flag and done on a per-project basis?

ericholscher:
It certainly could, at least if we wanted to test the load.

PR #7099, opened 2020-05-19
Add a tip in EmbedAPI to use Sphinx reference in section

humitos:
Mention that can be used a Sphinx reference to avoid changes in title breaking
the EmbedAPI content referenced.

### DIFF
--- a/docs/guides/embedding-content.rst
+++ b/docs/guides/embedding-content.rst
@@ -84,7 +84,7 @@ and show a modal when the user clicks in a "Help" link.
 
        .. in your .rst document file
 
-       .. _unbreakable-section-reference
+       .. _unbreakable-section-reference:
 
        Creating an automation rule
        ---------------------------
--- a/docs/guides/embedding-content.rst
+++ b/docs/guides/embedding-content.rst
@@ -73,6 +73,13 @@ from our own docs and will populate the content of it into the ``#help-container
 You can modify this example to subscribe to ``.onclick`` Javascript event,
 and show a modal when the user clicks in a "Help" link.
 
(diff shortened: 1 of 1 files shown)


docs/guides/embedding-content.rst
```diff
@@ -73,6 +73,13 @@ from our own docs and will populate the content of it into the ``#help-container
 You can modify this example to subscribe to ``.onclick`` Javascript event,
 and show a modal when the user clicks in a "Help" link.
 
+.. tip::
+
+   You can use a `Sphinx reference`_ as ``section`` argument instead of the slug of the title (i.e. ``creating-an-automation-rule``)
```
ericholscher, line 78: I think an example is better here, this `i.e.` section doesn't make sense to me?

humitos: Yeah. I don't know what I was thinking when I wrote this. I'm reading it now and I don't understand it :stuck_out_tongue_closed_eyes:

docs/guides/embedding-content.rst
```diff
@@ -75,10 +75,29 @@ and show a modal when the user clicks in a "Help" link.
...
+    Take into account that if the title changes, your ``section`` argument will break.
+    To avoid that, you can manually define Sphinx references above the sections you don't want to break.
+    For example,
+
+    .. code-block:: rst
+       :emphasize-lines: 3
+
+       .. in your .rst document file
+
+       .. _unbreakable-section-reference
```
ericholscher, line 87:
```suggestion
       .. _unbreakable-section-reference:
```

PR #7106, opened 2020-05-20
Store pageviews via signals, not tasks

davidfischer:
Sending a task per pageview quickly overwhelmed our tasks queues. Instead, this uses a signal we are already sending. Currently, this will result in no more queries than normal unless the project has the Feature flag `STORE_PAGEVIEWS`. However, if the project has that set, there will be 2 additional queries.

### DIFF
--- a/readthedocs/analytics/signals.py
+++ b/readthedocs/analytics/signals.py
@@ -1,0 +1,32 @@
+"""Django signals for the analytics app."""
+from django.db.models import F
+from django.dispatch import receiver
+from django.utils import timezone
+
+from readthedocs.api.v2.signals import footer_response
+from readthedocs.projects.models import Feature
+
+from .models import PageView
+
+
+@receiver(footer_response)
+def increase_page_view_count(sender, **kwargs):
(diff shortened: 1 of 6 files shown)


readthedocs/analytics/signals.py
```diff
@@ -0,0 +1,32 @@
...
+@receiver(footer_response)
+def increase_page_view_count(sender, **kwargs):
+    """Increase the page view count for the given project."""
+    del sender  # unused
+    context = kwargs['context']
+
+    project = context['project']
+    version = context['version']
+    # footer_response sends an empty path for the index
+    path = context['path'] or '/'
```
ericholscher, line 21: I imagine we might want to catch `/index.html` and `/` as being the same at some point, but seems overcomplicated for now.

davidfischer: I actually believe this code will handle it correctly.

davidfischer: It doesn't. This change now stores the exact path requested.

PR #7123, opened 2020-05-25
De-duplicate builds

humitos:
When a build is triggered, it could be marked as to be skipped by builders if:

* there is already a build running/queued for the same commit

  Multiple builds of the same commit will lead to the same results. So, we skip
  it if there is one already running/queued.

* there is already a build queued for the same version

  When building a version without specifying the commit, the last commit is
  built. In this case, we can only skip the new triggered build if there is one
  build queued because both builds will just pick the same commit.

A different approach of #7031


readthedocs/core/utils/__init__.py
```diff
@@ -165,8 +165,42 @@ def prepare_build(
...
+        # position in the queue and go back to the end of it --penalization
+        log.warning(
+            'Marking build to be skipped by builder. project=%s version=%s build=%s commit=%s',
+            project.slug,
+            version.slug,
+            build.pk,
+            commit,
+        )
+        build.error = DuplicatedBuildError.message
+        build.success = False
```
ericholscher, line 197: I wonder if we should make this nullable and set it to `None` here? It didn't succeed or fail, so it kinda makes sense.

readthedocs/core/utils/__init__.py
```diff
@@ -165,8 +165,42 @@ def prepare_build(
...
+        )
+    else:
+        skip_build = Build.objects.filter(
+            project=project,
+            version=version,
+            state=BUILD_STATE_TRIGGERED,
+        ).count() > 1
+    if skip_build:
+        # TODO: we could mark the old build as duplicated, however we reset our
+        # position in the queue and go back to the end of it --penalization
```
ericholscher, line 188: I don't see where we're changing the queue position?

agjohnson:
I'd like to see build state eventually decoupled from build status (or whatever you want to call this concept). How we do this is mostly a matter of modeling, but I see state and status as two competing concepts -- at least as it will apply to our new UX.

I'd describe build state as *where* in the build process the build is. This allows us to show progression to the user in the form of a progress bar or in the build listing. Our steps now are fairly broad:

* Queued
* Cloning
* Installing
* Building
* Uploading
* Finished

I would describe status as *why* the build is in a particular state. This is helpful for communicating more details about state to the user, but it doesn't help describe progression. Queued is an important state description, but "Queued" vs "Queue due to build concurrency" is not helpful in describing how far along the build is. Build status would include why a build is queued, why a build was rejected, and error states like why a build failed. Currently, these are all weak concepts in our modeling/UI.

I've long wanted something like a status field on Build with an identifier that maps to a list of build status/exception classes. We would store metadata around the status/exception on builds -- for example save `build.status_code = Build.STATUS_TIMEOUT` -- not just store the error as a character model field or use some obtuse template logic to display copy that exists in templates only.

I'd propose we stop adding to the list of build states -- that is, describing build flow progression -- and move all reporting of *why* a build is in a particular state to more flexible modeling.

So, for this particular case, I'd suggest:

* `build.state = 'finished'`
* `build.status_code = Build.STATUS_SKIPPED_DUPLICATE.status_code`

From there, we have many more options with class attributes/etc to help describe the build in templates.

readthedocs/core/utils/__init__.py
humitos: We are queuing the new build at the end of the queue, but we already had a build closer to be executed. If we "cancel" the build that was closer to be executed because the new build is going to replace it, we loose that position in the queue.

readthedocs/core/utils/__init__.py
humitos: It makes sense to me. However, I don't think we can change it now without changing the UI (knockout code) that renders in the frontend; which I'm not sure if we want to touch at this point.

humitos:
@agjohnson super interesting. I'm aligned with what you are saying here.

> `build.status_code = Build.STATUS_SKIPPED_DUPLICATE.status_code`

We don't have a `status_code` currently. We have a `exit_code`, but I don't think it means the same. Are you suggesting to add a new field? If that's the case, should we change the UI with this change as well, or you are just thinking to use this new field in the redesign?

ericholscher:
I think this makes sense generally.

I also think we need to think more about "success" in this case as well. Here, the build doesn't fail or succeed, but is mostly just a no-op. I think we have weird edge cases here already with badges, where we show failed badges more often than we should because of this issue.

agjohnson:
> Are you suggesting to add a new field?

Sorry, yeah my example was adding a new field for this, which would be separate from exit_code and would have special meaning internally. I'm not sure if we have a good middle ground though, and don't want to hold up this PR on a larger overhaul of error/status reporting in the UI. However, I did remove a lot of the complex template logic used to display build various state blocks from the build detail UI in anticipation of having stronger error/status reporting.

If the the simple, temporary solution is re-using the error field, then I think that is fine. We can prioritize adding some nicer error/status reporting before I get to a more mature release with the templates. I don't think we're talking about a lot of work though, but I'd probably only support better error/status modeling on the new templates

> I also think we need to think more about "success" in this case as well

This is a great point as well. We might want to turn `build.success` from a boolean into a select field of `failed`, `success`, `skipped`, and maybe even `unknown`. I'd like to see skipped builds as a different state than builds that succeeded or failed.

humitos:
I made some improvements here and did some more testing locally. It's ready for a re-review.

> it's easier to test with multiple single process builders (see https://github.com/readthedocs/common/pull/61#discussion_r431684874)

Build output's page looks like this:

Screenshot_2020-06-03_17-50-17

If we want to change how the error is displayed [this is the code generates the Error section](https://github.com/readthedocs/readthedocs.org/blob/6ce7617993c6447828b244ec2ff142d9917a0403/readthedocs/templates/builds/build_detail.html#L192-L199).

* "Error" title is hardcoded
* `Build.error` is what it's shown to the user

The Build list's page shows "Passed" and "Failed" for each build. If we want to change that [this is the code that generates that](https://github.com/readthedocs/readthedocs.org/blob/6ce7617993c6447828b244ec2ff142d9917a0403/readthedocs/templates/core/build_list_detailed.html#L7)

* it's based on `state='finished'` and `success`
* we could use `status_code` instead

In both cases, we could inject these Exception classes the template context to share the string/error/message to display.

I'm still not sure if we want to do these template changes in this PR or not (cc @agjohnson). We could keep everything the same and make it behaves different if `build.status_code != None` to keep backward compatibility.

readthedocs/core/utils/__init__.py
ericholscher: Yea, I just worry this will break the badge display.

readthedocs/doc_builder/exceptions.py
```diff
@@ -59,6 +59,8 @@ class BuildMaxConcurrencyError(BuildEnvironmentError):
 
 class DuplicatedBuildError(BuildEnvironmentError):
     message = ugettext_noop('Duplicated build.')
+    exit_code = 1
+    status_code = 0
```
ericholscher, line 63: This should be a constant probably, and something other than 0?

humitos: I made an `Enum` for this. I don't think it's a problem if it's 0. Why did you think it's better to have it different than 0? and what value?

humitos:
Hrm... I'm not sure what happened with Travis :shrug:

readthedocs/doc_builder/exceptions.py
ericholscher: Because 0 to me means "normal" status -- we want a status that represents the issue of "duplicated build", which should be assigned a specific number.

PR #7124, opened 2020-05-25
Use 2 builders with celery concurrency at 1

humitos:
This gives us more parity with production where 2 builds can't be ran at the
same time in one builder.

This is useful to avoid `VersionLocked` error when testing locally,
de-duplication of builds (#7123), usage of `build:default` and `build:large` for the router, etc.

Requires https://github.com/readthedocs/common/pull/61/


humitos:
I'm closing this one because we decide to follow another direction: use `--scale` in the docker compose file.

PR #7166, opened 2020-06-08
Use `CharField.choices` for `Build.status_code`

humitos:
We want to be able to communicate more than True/False status in this field.
This will help us to render different error messages to the user depending on
the `Build.status_code`.

#7123 was approved and merged, but I missed that usage of `status_code` there (I don't know _why_ I used a boolean :shrug:) but that is incorrect.

Also from https://github.com/readthedocs/readthedocs.org/pull/7123#discussion_r434726641, I think it's better to use a `CharField.choices` that will communicate better the meaning, instead of a plain number.


readthedocs/builds/models.py
```diff
@@ -615,7 +616,14 @@ class Build(models.Model):
         choices=BUILD_STATE,
         default='finished',
     )
-    status_code = models.BooleanField(_('Status code'), null=True, default=None, blank=True)
+    status_code = models.CharField(
```
ericholscher, line 619: I don't think `code` is really the right word for this, since that usually means a number (like `exit_code`). I think `status` is probably fine.

I worry this is likely going to confuse things a bit with our `state` field. We should explain a bit more in a comment at least the vision behind how this field is different than `state`.

humitos: OK. I will use `state` and copy what Anthony wrote in the other PR about their differences.

humitos:
@ericholscher please, take a look at this again. We do need this to be merged before the release.

Issue #2759, opened 2017-03-28
Alter the use of RemoteRepository


agjohnson:
Thinking about this more, `RemoteRepository` should not be a mapping of user access to remote repository objects, but instead should be just a list of accumulated remote repository objects. There should be a secondary table that provides a many to many mapping between remote users and remote repository objects. This lets us associate a project to a GitHub repo, for example, and multiple users would get access to the project on our side.

- [ ] Add a model for remote user <-> remote repository
- [ ] Add a migration that moves this data around
- [ ] Make sync operations update both tables

aswinshenoy:
I wish to work this issue @agjohnson, is this still open?
Regards

humitos:
This was done in originally https://github.com/readthedocs/readthedocs.org/pull/7536, and the work continued in the branch https://github.com/readthedocs/readthedocs.org/tree/remote-repository-normalization

I think we can close this issue. I'm trying to target the start of data sync for next deploy.

PR #7169, opened 2020-06-09
Design document for RemoteRepository DB normalization

humitos:
Initial document to discuss the path forward to implement

Reference: https://github.com/readthedocs/readthedocs.org/issues/2759


ericholscher:
This makes sense to me. I wonder if it's worth considering only storing the data we actually care about, and creating model fields for it, instead of the entire JSON blob? I guess we've hit times where we wanted extra data that we didn't store previously, but we could always re-query the API to get it, if needed. That would also cut down a good bit on the amount of data storage required.

humitos:
That's a good question. One thing I'd mention is that having the `json` field as is helped me to understand the previous modelling, how it was used and the differences between the "duplicated" `RemoteRepository` objects.

I think it's a trade off between adding more work from our side (add these fields, populate them and maintain this code) vs. db storage.

There is a lot of API URLs in the response that most of them are useless for our case. The rest of the data looks useful, though. Maybe keep using a JSON field but removing these URLs before storing it?

I have a slightly preference to keep the whole json as is and rely on Postgres JSON internals, because seems simpler and safer.

humitos:
Also, something to note is that the `json` field is independent from the Social Network, so it will definitely have different data.

humitos:
> Yea, I think at least trimming known useless data would be good. I mostly just want to get the data size reduced.

I replied too quickly last time. We are already [parsing the relevant data and storing it as `RemoteRepository` fields *outside* the `json` field](https://github.com/readthedocs/readthedocs.org/blob/d13c934c3acbffd6148ed125c4dd9c511f9e9108/readthedocs/oauth/services/github.py#L118-L133).

So, we may not require the `json` field at all. Instead, we may need to add some extra field in case we are missing something important.

> The other issue is, what is the API response format changes? It seems like that could wreck havoc on our side if we aren't parsing the incoming content to get the values we want.

If the API response change, even if we are parsing the relevant fields, it may break. We do not have have a good way to protect ourselves here. To avoid this, we should pin to an API version on each service (e.g. https://developer.github.com/v3/media/#request-specific-version)

docs/development/design/refactor-remote-repository.rst
```diff
@@ -0,0 +1,101 @@
...
+This document describes the current usage of ``RemoteRepository`` objects and proposes a new normalized modeling.
+
+
+Goals
+=====
+
+* De-duplicate data stored in our database.
+* Save only one ``RemoteRepository`` per GitHub repository.
+* Use an intermediate table between ``RemoteRepository`` and ``User`` to store associated remote data for the specific user.
+* Make this model usable from our SSO implementation.
```
humitos, line 15: This implies adding a new field `remote_id=` to make queries faster and do not depend on `full_name`.

docs/development/design/refactor-remote-repository.rst
```diff
@@ -0,0 +1,101 @@
...
+Use this modeling for SSO
+-------------------------
+
+We can get the list of ``Project`` where a user as access:
+
+.. code-block:: python
+
+   admin_remote_repositories = RemoteRepository.objects.filter(
+       users__contains=request.user,
+       users__remoterelation__json__admin=True,  # False for read-only access
```
humitos, line 99: We already have `admin=` as a field in `RemoteRepository` (outside the JSON field). We should keep it as a field in the `RemoteRelation` to allow quicker access to it.

docs/development/design/refactor-remote-repository.rst
```diff
@@ -0,0 +1,101 @@
...
+* Use Post ``JSONField`` to store associated ``json`` remote data.
+* Do not disconnect ``Project`` and ``RemoteRepository`` when a user delete/disconnects their account.
+
+
+Non-Goals
+=========
+
+* Keep ``RemoteRepository`` in sync with GitHub repositories.
+* Delete ``RemoteRepository`` objects deleted from GitHub.
+* Listen to GitHub events to detect ``full_name`` changes and update our objects.
```
humitos, line 25: If we add a `remote_id=` field, this sync is not even needed.

docs/development/design/refactor-remote-repository.rst
```diff
@@ -0,0 +1,101 @@
...
+
+
+Goals
+=====
+
+* De-duplicate data stored in our database.
+* Save only one ``RemoteRepository`` per GitHub repository.
+* Use an intermediate table between ``RemoteRepository`` and ``User`` to store associated remote data for the specific user.
+* Make this model usable from our SSO implementation.
+* Use Post ``JSONField`` to store associated ``json`` remote data.
```
humitos, line 16: This could be considered a "Nice to have" in case we don't have the time to make all the changes required to depend on PostgreSQL in tests. If we keep the `admin=` field in the relationship, filtering over the JSON is not needed.

docs/development/design/refactor-remote-repository.rst
```diff
@@ -0,0 +1,101 @@
...
+  * basic information (provider, last login, etc)
+  * provider's specific data saved in a JSON under ``extra_data``
+* ``allauthsocialaccount.models.SocialToken``
+  * token to hit the API on behalf the user
+
+
+We *don't create* any ``RemoteRepository`` at this point.
+They are created when the user jumps into "Import Project" page and hit the circled arrows.
+It triggers `sync_remote_repostories`_ task in background that updates or creates ``RemoteRepositories``,
+but **it does not delete** them.
```
humitos, line 48: This changed for GitHub when implemented https://github.com/readthedocs/readthedocs.org/pull/7183

docs/development/design/refactor-remote-repository.rst
```diff
@@ -0,0 +1,101 @@
...
+When a user connect their account to a social account, we create a
+
+* ``allauth.socialaccount.models.SocialAccount``
+  * basic information (provider, last login, etc)
+  * provider's specific data saved in a JSON under ``extra_data``
+* ``allauthsocialaccount.models.SocialToken``
+  * token to hit the API on behalf the user
+
+
+We *don't create* any ``RemoteRepository`` at this point.
```
humitos, line 45: On corporate we are syncing in background at login (and/or social account connected) for SSO --may be useful to bring that pattern into community as well.

humitos:
@ericholscher I updated the document with the latest changes in context. This is ready to merge IMO, and we can workaround some implementation details in the PR that implements this. I'm tempt to say that we can remove the `json` field completely if we are not going to use it anyways --we have all the data we need in DB model fields.

humitos:
This is _required_. It's ready for review and merge, IMO. The implementation will take some time, tho.

docs/development/design/refactor-remote-repository.rst
humitos: I think we can probably wait here until we migrate to Django 3.x that implements this natively and avoid making this migration twice. We don't really need  json field right now.

humitos:
I'm merging this since we are already implementing this and this document reflects almost exactly what we are doing. I just added a "Rollout plan" section with more info about this.

Issue #2412, opened 2016-09-13
Deleted repository not detected


beruic:
## Details

I have deleted the repository at https://github.com/beruic/bootstrap-tagsinput, but it does not disappear when i refresh in the repository importer at https://readthedocs.org/dashboard/import/.

humitos:
I confirm that I'm able to reproduce this. I created a repository called `1` in GitHub and I imported without problem to RTD. Then I deleted the project from RTD and also from GitHub. After that I hit re-sync in RTD and I still seeing my `1` repository to be able to import.

beruic:
I just checked BTW. The deleted repository still appears in my list.

humitos:
After #7183, GitHub integration in cleaning up old `RemoteRepository` from that user _only if it's not connected to a Project_. This will have impact in the "Import Project" page that won't show stale repositories that do not exist anymore.

The same work/code needs to be ported to GitLab and Bitbucket to be able to close this issue.

PR #7183, opened 2020-06-11
Modifications required to implement GitHub SSO in commercial

humitos:
Changes on this PR:

- user is part of an organization if there is at least one `Project` imported under that `Organization` that it's linked to a `RemoteRepository` the user owns and the organization has `SSOIntegration` enabled

- `Organization.members` and `Organization.users` methods were moved to `AdminPermission.members`

   > Note that `.members` and `.users` return exactly the same users

- delete `RemoteRepository` instances when syncing GitHub (only those instances that are not tied to a `Project`) (Fixes, partially, #2412)

### DIFF
--- a/readthedocs/core/permissions.py
+++ b/readthedocs/core/permissions.py
@@ -8,14 +8,38 @@ from readthedocs.core.utils.extend import SettingsOverrideObject
 class AdminPermissionBase:
 
     @classmethod
-    def is_admin(cls, user, project):
+    def admins(cls, obj):
+        from readthedocs.projects.models import Project
+        from readthedocs.organizations.models import Organization
+
+        if isinstance(obj, Project):
+            return project.users.all()
+
+        if isinstance(obj, Organization):
+            return obj.owners.all()
+
+    @classmethod
+    def members(cls, obj):
+        from readthedocs.projects.models import Project
+        from readthedocs.organizations.models import Organization
+
+        if isinstance(obj, Project):
+            return project.users.all()
+
+        if isinstance(obj, Organization):
+            return User.objects.filter(
+                Q(teams__organization=obj) | Q(owner_organizations=obj),
+            ).distinct()
+
+    @classmethod
+    def is_admin(cls, user, obj):
         # This explicitly uses "user in project.users.all" so that
         # users on projects can be cached using prefetch_related or prefetch_related_objects
-        return user in project.users.all() or user.is_superuser
+        return user in cls.admins(obj) or user.is_superuser
 
     @classmethod
     def is_member(cls, user, obj):
-        return user in obj.users.all() or user.is_superuser
+        return user in cls.members(obj) or user.is_superuser
 
 
 class AdminPermission(SettingsOverrideObject):
--- a/readthedocs/core/permissions.py
+++ b/readthedocs/core/permissions.py
@@ -1,6 +1,8 @@
 # -*- coding: utf-8 -*-
 
 """Objects for User permission checks."""
+from django.contrib.auth.models import User
+from django.db.models import Q
 
 from readthedocs.core.utils.extend import SettingsOverrideObject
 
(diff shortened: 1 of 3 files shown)


readthedocs/organizations/models.py
```diff
@@ -93,7 +94,23 @@ def get_absolute_url(self):
 
     @property
     def users(self):
-        return (self.members.all() | self.owners.all().distinct()).distinct()
+        sso_users = User.objects.none()
+        # TODO: find another way to check if we are under corporate site
```
ericholscher, line 98: Can't we just override this on the .com?

humitos: I think we don't have a pattern to override a method from a Model.

The only way I'm thinking now (since this is a queryset method) could be move this method into the `OrganizationQuerySet` and use `SettingsOverrideObject` in that class. Makes sense?

readthedocs/organizations/querysets.py
```diff
@@ -15,14 +15,24 @@ class BaseOrganizationQuerySet(models.QuerySet):
...
+        )
+
+        # Add organizations where the user has GitHub SSO projects
+        # TODO: change this Subquery once we have normalized our db
+        # https://github.com/readthedocs/readthedocs.org/pull/7169
+        orgs |= self.filter(
+            projects__remote_repository__full_name__in=Subquery(user.oauth_repositories.values('full_name')),
+            ssointegration__isnull=False,
+        )
+        return orgs.distinct()
```
humitos, line 30: This can definitely moved to .com using our existent pattern.

readthedocs/organizations/models.py
humitos: It will be weird, though. It will work like `Organization.objects.users(organization=slug)` but it will return `User` instances instead of `Organization` what is what I'd expect.

readthedocs/organizations/querysets.py
humitos: I moved this already.

readthedocs/organizations/models.py
ericholscher: What is this logic even used for? I really don't like having a `readthedocsinc` import in this codebase, it feels really janky. We should probably just change the logic where this is used to call code that isn't in this codebase.

humitos: It's used in,

- `AdminPermissionBase.is_member` (`obj` could be an Organization)
- `CorporateAdminPermission.is_member`
- old migrations (oauth.0003_move_github)
- corporate profile details

I'm tempted to remove this `users` property together with `is_member` (we talked about the latter already)

readthedocs/organizations/models.py
```diff
@@ -93,7 +94,25 @@ def get_absolute_url(self):
...
+            # Members
+            self.members.all() |
+            # Owners
+            self.owners.all().distinct() |
+            # SSO Users
+            sso_users
+        ).distinct()
 
     @property
     def members(self):
```
humitos, line 118: This method needs to be adapted as well.

It does not make sense with the new usage of SSO.

readthedocs/organizations/models.py
humitos: OK. I moved this code to `CorporateAdminPermission.is_member`, where it seems it's the correct place for this code. Then, when we need to check for membership, we should use that classmethod.

I removed these `users` and `member` properties to avoid confusions (they won't return SSO members). We could re-add them later if we need them.

humitos: > I removed these `users` and `member` properties to avoid confusions (they won't return SSO members)

I finally moved them into the `AdminPermission` class, so we can override them from corporate. Corporate requires the `Organization.members` method.

readthedocs/core/permissions.py
```diff
@@ -1,21 +1,47 @@
...
+        from readthedocs.projects.models import Project
+        from readthedocs.organizations.models import Organization
+
+        if isinstance(obj, Project):
+            return obj.users.all()
+
+        if isinstance(obj, Organization):
+            return User.objects.filter(
+                Q(teams__organization=obj) | Q(owner_organizations=obj),
+            ).distinct()
```
ericholscher, line 34: I like seeing this logic backported into the .org 👍

PR #7187, opened 2020-06-11
Document the fact that PR builds are now enabled on .org

ericholscher:


ericholscher:
Yea, that'll come once it's rolled out fully on .com as well 👍

PR #7198, opened 2020-06-16
Add a feature flag to use latest RTD Sphinx ext

davidfischer:
This is useful to test the latest versions of the extension without pushing it live for everyone.


readthedocs/doc_builder/python_environments.py
```diff
@@ -361,7 +361,11 @@ def install_core_requirements(self):
                     negative='sphinx<2',
```
ericholscher, line 361: I know this isn't inherently related, but I think we should drop the `<2` on the positive here too, and likely consider a newer default.

davidfischer: Seems easy enough. I'll make another PR for that.

davidfischer: https://github.com/readthedocs/readthedocs.org/pull/7201

PR #7201, opened 2020-06-16
Feature flag for using latest Sphinx

davidfischer:
The latest Sphinx flag will now mean actually using the latest Sphinx.

### DIFF
--- a/readthedocs/doc_builder/python_environments.py
+++ b/readthedocs/doc_builder/python_environments.py
@@ -352,12 +352,10 @@ class Virtualenv(PythonEnvironment):
                 ),
             )
         else:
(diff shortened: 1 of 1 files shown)


davidfischer:
Actually we should be careful here. I checked prod and this feature flag is on for a very large number of projects >10k. This could change how a lot of projects are built. Thoughts?

humitos:
I'd say that we should remove the feature flag to those projects, and start over with our testings. They are not using it currently since it's installing the same version.

davidfischer:
> I'd say that we should remove the feature flag to those projects, and start over with our testings. They are not using it currently since it's installing the same version.

That sounds reasonable. I blanked out that feature flag and we can now use it!

PR #7209, opened 2020-06-19
Update CAA docs

davidfischer:
- Note that CAA records are usually at the root
- They don't *have* to be at the root but you can't have a CNAME and CAA record for the same subdomain

### DIFF
--- a/docs/custom_domains.rst
+++ b/docs/custom_domains.rst
@@ -51,10 +51,11 @@ You can also host your documentation from your own domain.
 
       .. admonition:: Certificate Authority Authorization (CAA)
 
-         If your custom domain — either the subdomain you're using or the root domain — has configured CAA records,
+         If your custom domain has configured CAA records,
          please do not forget to include Cloudflare CAA entries to allow them to issue a certificate for your custom domain.
+         CAA records are typically configured at the root domain.
          See the `Cloudflare CAA FAQ`_ for details.
-         We need a record that looks like this: ``0 issue "digicert.com"`` in response to ``dig +short CAA <domain>``
+         We need a record that looks like this: ``0 issue "digicert.com"`` in response to ``dig +short CAA <rootdomain>``
 
          .. _Cloudflare CAA FAQ: https://support.cloudflare.com/hc/en-us/articles/115000310832-Certification-Authority-Authorization-CAA-FAQ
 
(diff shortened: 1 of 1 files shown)


docs/custom_domains.rst
```diff
@@ -51,10 +51,11 @@ You can also host your documentation from your own domain.
 
       .. admonition:: Certificate Authority Authorization (CAA)
 
-         If your custom domain — either the subdomain you're using or the root domain — has configured CAA records,
+         If your custom domain has configured CAA records,
          please do not forget to include Cloudflare CAA entries to allow them to issue a certificate for your custom domain.
+         CAA records are typically configured at the root domain.
          See the `Cloudflare CAA FAQ`_ for details.
-         We need a record that looks like this: ``0 issue "digicert.com"`` in response to ``dig +short CAA <domain>``
+         We need a record that looks like this: ``0 issue "digicert.com"`` in response to ``dig +short CAA <rootdomain>``
```
humitos, line 58: Does it worth to put an example here? Like, if your domain is `docs.example.org`, your root domain is `example.org`.

davidfischer: I reworked this section and added an example.

davidfischer:
I updated the docs to add a troubleshooting section where I moved the CAA notes, the Cloudflare-to-Cloudflare notes, and discuss delays, retrying, and avoiding downtime.

docs/custom_domains.rst
```diff
@@ -112,6 +81,68 @@ for our own domains including ``*.readthedocs.io``, ``*.readthedocs-hosted.com``
...
+    0 issue "comodoca.com"
+    0 issue "letsencrypt.org"
+
+If there are CAA records for your domain that do not allow the certificate authorities that Read the Docs uses,
+you may see an error message like ``pending_validation: caa_error: YOURDOMAIN.TLD``
+in the Read the Docs dashboard for your domain.
+You will need to update your CAA records to allow us to issue the certificate.
+
+* For |org_brand|, we use Cloudflare which uses Digicert as a CA. See the `Cloudflare CAA FAQ`_ for details.
+* For |com_brand|, we use AWS Certificate Manager as a CA. See the `Amazon CAA guide`_ for details.
```
ericholscher, line 125: I do think an example of what a "proper" record looks like for confirmation would be useful, but maybe a bit much.

davidfischer: There's a community example a few lines up (palletsprojects.com). Since this is outside the tab, it's not as nice to provide an example .com and a .org example.

PR #7212, opened 2020-06-22
Documentation for Sigle Sign-On feature on commercial

humitos:
Initial documentation for SSO on Read the Docs for Business.

My goals here are:

- mention different types of SSO:
  - authentication & authorization (GitHub)
  - authentication (verified email address)
- make it clear where permissions are managed
  - social accounts (GitHub, etc)
  - normal Read the Docs auth (Admins teams)
- mention that this is still on Beta and enabled by request only
- avoid confusions mentioning differences with "regular organizations": users only needs to know how SSO works

Rendered version: https://docs--7212.org.readthedocs.build/en/7212/commercial/single-sign-on.html


docs/commercial/single-sign-on.rst
```diff
@@ -0,0 +1,74 @@
...
+   This feature only exists on `Read the Docs for Business <https://readthedocs.com/>`__.
+
+
+Single Sign-On is supported on |com_brand| for Pro and Enterprise plans.
+:abbr:`SSO (Single Sign-On)` will allow you to manage all the users' permissions directly on GitHub.
+
+.. note::
+
+   Currently, we only supports GitHub as Identity Provider.
+   We plan to support GitLab and Bitbucket soon as well.
```
ericholscher, line 15: I don't think either of these need to be notes, or we should try and keep only 1 of them, since a bunch of notes is sloppy.

humitos:
@eric I did another iteration on this document. I tried to focus on the distinction of the two different types of SSO we currently support: Social Account and `@company.com` email address. Besides, I made the normal actions (e.g. grant admin access) as sections so it's easier to see and find them.

I think it's too much better than the previous one and makes things clearer. I'd like to have "better making names" for the two different types, but at least, I think it communicates clearly what the differences are. I suppose we can grow from here...

humitos:
Note with this we will have 3 different types of Auth for organizations:

1. Read the Docs Auth:
   - Owners
     - admin of the organization's settings
     - admin of _all_ projects
     - can import projects
   - Admins
     - admins of projects under "Admin Team"
     - can import projects
     - every user must have to be invited to the team
   - Members
     - read access of projects under "Read Team"
     - every user must have to be invited to the organization
2. Verified email:
   - Owners
     - admin of the organization's settings
     - admin of _all_ the projects
   - Admins
     - admins of projects under "Admin Team"
     - can import projects
     - every user must have to be invited to the team
   - Members
     - any user with `@company.com` verified email
     - read access to _all_ the projects on the organization
     - there is no need to invite/add this users manually to any Team
2. GitHub:
   - Owners
     - admin of the organization's settings
     - _no_ admin of _any_ projects under the organization
   - There is no concept of teams at all
   - Admin access on Projects is managed under GitHub
   - Read access on Projects is managed under GitHub

ericholscher:
@humitos I think #1 and #2 should be exactly the same, except for the membership with email.

humitos:
@ericholscher what do you mean exactly with "membership with email"? What access a user would have after signup with a `@company.com` email if the organization already have 3 projects imported?

> NOTE, I added "can import project" on Verified Email -> Admins, that was missing

humitos:
The main important difference that make this feature useful (between #1 and #2) is that "users with `@company.com` email will have _read_ access to all the projects of the organization" --the rest is exactly the same.

docs/commercial/single-sign-on.rst
```diff
@@ -0,0 +1,90 @@
...
+
+You can add a user under an "Admin Team" to grant admin permissions to all the projects under that Team.
+This can be done under :guilabel:`Teams` > :guilabel:`Admins` > :guilabel:`Invite Member`.
+
+
+Grant access to import a project
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Making the user member of any "Admin Team" under your organization (as mentioned in the previous section),
+they will be granted access to import a project.
```
ericholscher, line 87: They can only import projects into that team, right?

docs/commercial/single-sign-on.rst
```diff
@@ -0,0 +1,90 @@
...
+
+Note that to be able to import a project, that user must have at least **read** permissions in the GitHub, Bitbucket or GitLab repository associated,
```
ericholscher, line 89: This seems wrong, no? We can't set webhooks on these repos, and I think it isn't how it works currently right?

docs/commercial/single-sign-on.rst
```diff
@@ -0,0 +1,90 @@
...
+
+SSO with GitHub, Bitbucket or GitLab
+------------------------------------
+
+Using an Identity Provider that supports authentication and authorization allows you to manage
+"who have access to what projects on Read the Docs" directly from the provider itself.
+In case you want a user to have access to your documentation project under Read the Docs,
+that user just needs to be granted permissions in the GitHub, Bitbucket or GitLab repository associated with it.
+
+Note the users created under Read the Docs must have their GitHub, Bitbucket or GitLab
```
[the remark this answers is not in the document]
humitos: It's fine using "VCS provider" as a general concept here I think. We use "VCS account" already in other documents, https://docs.readthedocs.io/en/stable/connected-accounts.html. Probably better to stick with that term.

I think it makes sense to mention them explicitly at the beginning maybe and use the general term in the rest of the document. It's good to avoid confusions with projects imported from outside these three providers. However, they have to be imported _manually_ which is not discoverable in .com

docs/commercial/single-sign-on.rst
humitos: Yes. In fact, if you are an admin, you can _only_ import projects under the Admins teams you belong to. You _cannot_ import projects on "Read" teams, even if you belong to --which is a bug, I'd say. See https://github.com/readthedocs/readthedocs-corporate/issues/929

I'm not sure if this has to be clarified here, since it's not related to SSO. It probably fits better on https://docs.readthedocs.io/en/stable/commercial/organizations.html

docs/commercial/single-sign-on.rst
humitos: Right. This is a mistake.

humitos: Wait... This is confusing.

Currently, _without SSO_, if the project is public and you have **read** permissions on it, you can imported it, see https://github.com/readthedocs/readthedocs.org/blob/eb34b08ee05acc9f3c4df0a4dd1fed2422a998df/readthedocs/projects/static-src/projects/js/import.js#L123-L128 . In that case, you will be able to import the project, but the webhook won't be connected, ssh key not setup and we will show these notifications:

Screenshot_2020-07-07_19-32-25

I changed that behavior to only make `is_locked = !self.admin()` only when SSO is enabled because it doesn't make sense to import a project that you are not able to modify later (GitHub SSO won't give you access to Admin tab if you don't have **admin** permissions in the GH repo).

Summarizing,

* current behavior with RTD Auth, allows you to import public projects with only **read** permissions
* GitHub SSO, allows you to import projects _only_ if you **admin** on the GH repository
* Verified Email, same as RTD Auth

I think, for corporate, it makes sense to _always_ force to be admin on the GH repository to be able to import a project; because the reasons that you mentioned. After imported, you will end up with a half configured project. This would change the current behavior, tho.

humitos: I changed this behavior to allow import projects only with `admin` permissions in the repository for Email SSO and VCS SSO.

I think, we should eventually do the same for RTD Auth and make all behave the same under corporate: "admin permissions on RemoteRepository are required to import a project".

docs/commercial/single-sign-on.rst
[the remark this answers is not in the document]
humitos: I made some small changes for this. I left it explicit where I thought it makes sense to make it explicit and used "VCS social provider" or "VCS provider" as well.

docs/commercial/organizations.rst
```diff
@@ -25,7 +25,7 @@ The best way to think about this relationship is:
 .. warning::
 
    Owners, Members and Teams behave differently if you are using
-   :ref:`SSO with GitHub, Bitbucket or GitLab <SSO with GitHub, Bitbucket or GitLab>`.
+   :ref:`SSO with VCS social provider (GitHub, Bitbucket or GitLab) <commercial/single-sign-on:SSO with VCS social provider (GitHub, Bitbucket or GitLab)>`
```
ericholscher, line 28: `VCS social provider` feels even more verbose and isn't a term I've ever heard before, and we're still including the full set of 3. I'm not convinced this is better :)

humitos:
OK. I finally used "VCS provider" when the document refers to GitHub/GitLab/Bitbucket and "VCS repository" when it refers to a specific repository associated to a Read the Docs' project. I'm not 100% convinced, but seems a little better at least :)

ericholscher:
Looks good 👍

ericholscher:
Excited to have this public 🎉

Issue #7235, opened 2020-06-25
Remote repository URL validation fails on Django admin


agjohnson:
### Reproduce

* Add RemoteRepository with SSH URL that looks like a standard GitHub URL: `git@github.com:readthedocs/test-builds.git`
* Open admin, try to make a change to RemoteRepository instance
* You'll get a validation error on SSH URL field.

I wasn't able to even edit this to a valid URL, even HTTPS URL wasn't allowed:
![image](https://user-images.githubusercontent.com/1140183/85767556-5e069380-b6d5-11ea-96e4-27b8ee52a3af.png)

Seems we need to alter the validation on the URL fields. This admin view can't be used otherwise

agjohnson:
This doesn't throw an exception but the field isn't updated in the admin either. I don't know if that is actually a problem or not though.

PR #7243, opened 2020-06-29
Update RemoteRepository ssh_url validation

saadmk11:
closes https://github.com/readthedocs/readthedocs.org/issues/7235


readthedocs/oauth/models.py
```diff
@@ -134,7 +134,7 @@ class RemoteRepository(models.Model):
         _('SSH URL'),
         max_length=512,
         blank=True,
-        validators=[URLValidator(schemes=['ssh'])],
+        validators=[URLValidator(schemes=['https', 'ssh', 'git'])],
```
humitos, line 137: Why the SSH URL would allow `https` as a scheme on it?

The URL that allows HTTPS is the `clone_url` instead.

saadmk11: Opps! I miss read the issue. I'll fix it. Thanks :+1:

humitos: Did you test that `git@github.com:...` is a valid URL for the URLValidator?

We should, at least, check that URLs from `clone_url` and `ssh_url` (https://developer.github.com/v3/repos/#response) are valid.

We use these at https://github.com/readthedocs/readthedocs.org/blob/041edb30e9fb83e12d15b7238d90cbf9f6b29d8d/readthedocs/oauth/services/github.py#L123-L129.


We should probably check the same for GitLab and Bitbucket.

humitos: > Opps! I miss read the issue. I'll fix it. Thanks +1

I think the original issue is confusing because it shows a HTTPS URL on the `ssh_url` field, which I thin it's wrong :)

saadmk11: > Did you test that `git@github.com:...` is a valid URL for the URLValidator?

Yes :)

GitHub and GitLab uses `git` url for `ssh_url` and Bitbucket uses `ssh` url

GitHub: https://developer.github.com/v3/repos/#get-a-repository
GitLab: https://docs.gitlab.com/ce/api/projects.html
Bitbucket: https://bitbucket.org/api/2.0/repositories/

humitos: It does not work for me:

```
In [15]: from django.core.validators import URLValidator                                                                                                                                                                                        

In [16]: URLValidator(schemes=['ssh', 'git'])('git@github.com:humitos/readthedocs.org')                                                                                                                                                         
...
ValidationError: ['Enter a valid URL.']

In [17]:
```

saadmk11: @humitos I'm sorry I used this scheme to test `git://github.com/user/project-name.git` 
from Django URLValidator docs [(valid URI schemes):](https://docs.djangoproject.com/en/3.0/ref/validators/#django.core.validators.URLValidator.schemes) https://www.iana.org/assignments/uri-schemes/prov/git
Django by default only supports URL's with `://` in the `URLField`

https://github.com/django/django/blob/779e615e362108862f1681f965ee9e4f1d0ae6d2/django/core/validators.py#L105-L107

```python
scheme = value.split('://')[0].lower()
if scheme not in self.schemes:
    raise ValidationError(self.message, code=self.code)
```

humitos: Right, but that's not the SSH URL that GitHub returns in its API: 

```json
"ssh_url": "git@github.com:octocat/Hello-World.git",
```

> from https://developer.github.com/v3/repos/#list-repositories-for-the-authenticated-user

So, after those object are created, if you go to the admin and try to modify another field, it will fail the validation and won't let you update the other field.

saadmk11: Yes, that's right `GitHub` and `GitLab` sends `git@....` URL as `ssh_url`. 
To fix this I think we have to create a Custom URLValidator and override the `__call__` method and copy paste code as we can't call `super()` but not sure if thats a good idea.

agjohnson:
Yeah, correct, it's the GitHub URL returned from the API that is the problem. We'll need a custom validator

saadmk11:
I have added a Regex validator and changed it to ``CharField``

**Reason for Using CharField**

``URLField`` contains an attribute ``default_validators`` which already has ``URLValidator()`` and it can only be removed by creating a custom field and overriding the attribute, otherwise it will always run `URLValidator()`` which we do not want.
With that, we'll need to create a new Form ``URLField`` and override `default_validators``  to not use ``URLValidator()`` for the new ``GitSSHURLField``.

I was thinking if that's overkill for this change. Would love to know your thoughts on this. If we follow the overriding path then it may look like this, 

https://gist.github.com/saadmk11/172962c0f97b55d1e559aebbbed0f24f

CC: @humitos @agjohnson

humitos:
> `URLField` contains an attribute `default_validators` which already has `URLValidator()` and it can only be removed by creating a custom field and overriding the attribute, otherwise it will always run `URLValidator()` which we do not want.

Inheriting `URLField` would bring us other benefits that `CharField` would not?

BTW, the code with the CharField option looks cleaner and simpler. However, if we need to use this in different places maybe having our own custom field is better. I don't think it's the case, tho, right?

saadmk11:
> Inheriting `URLField` would bring us other benefits that `CharField` would not?

URLField Provides the ``URLValidator()`` on top of ``CharField`` which validates the regular URL stuctures.

> BTW, the code with the CharField option looks cleaner and simpler. However, if we need to use this in different places maybe having our own custom field is better. I don't think it's the case, tho, right?

We are using this time of URL (``git@...``) only in this model.
I think Creating a new field for just one usecase is overkill.

If we create a new field that inherits from URLField then we have to create a new form URLField and overide the ``default_validations`` or use the form CharFIeld for the new model URLFIeld.

So, I'm not sure if that would be the best route. We can go on any route, @humitos would like to know your thoughts.

Issue #7216, opened 2020-06-22
Don't allow Domain objects matching production domain to be created


ericholscher:
Currently we still allow users to accidently create Domain's for their actual docs sites. So you can create a Domain that is like `foo.readthedocs.io`. We should change the form to check for [PUBLIC_DOMAIN](https://docs.readthedocs.io/en/latest/development/settings.html?highlight=public_domain#public-domain) and likely the [PRODUCTION_DOMAIN](https://docs.readthedocs.io/en/latest/development/settings.html?highlight=public_domain#production-domain) as well, so that this raises a validation error.

PR #7244, opened 2020-06-29
Don't allow Domain name matching production domain to be created

saadmk11:
closes https://github.com/readthedocs/readthedocs.org/issues/7216

### DIFF
--- a/readthedocs/projects/forms.py
+++ b/readthedocs/projects/forms.py
@@ -634,6 +634,19 @@ class DomainBaseForm(forms.ModelForm):
             domain_string = parsed.netloc
         else:
             domain_string = parsed.path
+
+        if (
+            settings.PRODUCTION_DOMAIN and
+            settings.PRODUCTION_DOMAIN in domain_string
+        ):
+            raise forms.ValidationError('Enter a valid domain name.')
+
+        if (
+            settings.PUBLIC_DOMAIN and
+            settings.PUBLIC_DOMAIN in domain_string
+        ):
+            raise forms.ValidationError('Enter a valid domain name.')
+
         return domain_string
 
     def clean_canonical(self):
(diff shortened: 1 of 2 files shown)


readthedocs/projects/forms.py
```diff
@@ -634,6 +634,19 @@ def clean_domain(self):
             domain_string = parsed.netloc
         else:
             domain_string = parsed.path
+
+        if (
+            settings.PRODUCTION_DOMAIN and
+            settings.PRODUCTION_DOMAIN in domain_string
```
ericholscher, line 640: I think we probably want a `endswith` in both of these, to reduce false positives.

readthedocs/projects/forms.py
```diff
@@ -634,6 +634,19 @@ def clean_domain(self):
...
+        ):
+            raise forms.ValidationError('Enter a valid domain name.')
```
ericholscher, line 642: We should mention exactly why this is invalid (`{PRODUCTION_DOMAIN} is not a valid domain`

readthedocs/projects/forms.py
```diff
@@ -634,6 +634,23 @@ def clean_domain(self):
...
+                f'{settings.PRODUCTION_DOMAIN} is not a valid domain.'
+            )
+
+        if (
+            settings.PUBLIC_DOMAIN and
+            domain_string.endswith(settings.PUBLIC_DOMAIN)
+        ):
+            raise forms.ValidationError(
+                f'{settings.PUBLIC_DOMAIN} is not a valid domain.'
+            )
```
humitos, line 652: This can be done in a for and de-duplicate this code, right?

saadmk11: Okay :+1:

humitos:
Thanks @saadmk11, you rock!

PR #7259, opened 2020-07-03
[WIP] Add support for "requires" key in Python install config and allow extra keys beginning with "x-"

NiklasRosenstein:
* Adds support for a "requires" install step in the Python install config. This allows you to specify requirements directly in the readthedocs config.
* Allow extra keys in the v2 configuration file if the key name starts with `x-`.

My motivation for this change is to reduce the number of files needed if 

a) the requirements are needed only for readthedocs, and thus it makes sense to put them in the configuration directly, and
b) add the ability to enrich the readthedocs config with non-standard data for custom workflows

### DIFF
--- a/docs/config-file/v2.rst
+++ b/docs/config-file/v2.rst
@@ -175,6 +175,27 @@ Example:
   manage the build, this setting will not have any effect. Instead
   add the extra requirements to the ``environment`` file of Conda.
 
+Requirements List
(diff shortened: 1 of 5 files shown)


NiklasRosenstein:
Hey @stsewd ! Thanks for your comments. I've submitted two separate PRs per your suggestion: #7288 #7289

PR #7285, opened 2020-07-13
Allow users to set hidden on versions via APIv3

humitos:


humitos:
The docs are already updated: https://github.com/readthedocs/readthedocs.org/blob/97124cfd804ab0dc26d2630c718e3efd9c79130e/docs/api/v3.rst#L515-L522

It seems you added it to the docs when added `hidden` field, but you didn't added to the serializer.

PR #7288, opened 2020-07-13
allow "x-" options in readthedocs.yml configuration

NiklasRosenstein:
This Pull Request will allow that options starting with `x-` can be added to the `.readthedocs.yml` configuration. This is useful for custom workflows where it makes sense to embed additional options in the Read the Docs configuration. The `x-` convention is also used by Docker-Compose, for example.

My personal use case for this is to embed the configuration for [`readthedocs-custom-steps`](https://pypi.org/project/readthedocs-custom-steps/) in the configuration directly, instead of a having to use a separate `.readthedocs-custom-steps.yml` file.

### DIFF
--- a/readthedocs/config/config.py
+++ b/readthedocs/config/config.py
@@ -1085,10 +1085,13 @@ class BuildConfigV2(BuildConfigBase):
         }
 
         Will return `['key', 'name']`.
+
+        Ignores keys that start with ``x-``.
         """
-        if isinstance(value, dict) and value:
-            key_name = next(iter(value))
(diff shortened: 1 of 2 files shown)


agjohnson:
I would vote no. I don't really want 2 separate tools using the same configuration file. I'd sooner support native custom build steps, I'm not sure it needs to be a separate extension.

humitos:
I'm closing this PR. We are moving in a different direction to support custom commands. There are some discussions about this at #8215 #8190

PR #7295, opened 2020-07-15
Enable SessionAuthentication on APIv3 endpoints

humitos:
This is required to be able to use the APIv3 from the dashboard in the new templates we are creating.

@agjohnson is this enough for the work you need at the moment?

> I'm not documenting this yet so we start using it internally, but after that we should change the documentation a little.

### DIFF
--- a/readthedocs/api/v3/views.py
+++ b/readthedocs/api/v3/views.py
@@ -2,7 +2,7 @@ import django_filters.rest_framework as filters
 from django.utils.safestring import mark_safe
 from rest_flex_fields.views import FlexFieldsMixin
 from rest_framework import status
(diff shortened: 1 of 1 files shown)


agjohnson:
Yeah, i assumed this was the only piece we needed. You can test by simply making a call from the debug console to the API v3 endpoint. It should give you a valid response. That is all I would need to make a client request to APIv3 for dashboard interactions

humitos:
I tested this with the Browseable API by being logged in. However, I followed your example as well and it worked.

Not logged in,

Screenshot_2020-07-21_13-02-19

After logging in,

Screenshot_2020-07-21_13-02-47

PR #7310, opened 2020-07-22
Sync RemoteRepository and RemoteOrganization in all VCS providers

### LINKED ISSUE #2412: Deleted repository not detected

humitos:
- Refactor `Service.sync` to be unique for all the VCS providers
- Make `sync_repositories` and `sync_organizations` to return `repositories`
  and `(organizations, repositories)` to allow deleting old instances where the
  user does not have more access
- rename `BitBucketService.sync_teams` to `BitBucketService.sync_organizations`
  to keep our internal naming

With this commit, all VCS providers (GitHub, GitLab and BitBucket) will start
deleting old RemoteRepository instances (not only GitHub as it's currently
working) but also deleting old RemoteOrganization instances as well.

This is a step previous to #2759 and it's required for GitLab/BitBucket VCS SSO 

Closes #2412


readthedocs/oauth/services/base.py
```diff
@@ -181,11 +182,30 @@ def paginate(self, url, **kwargs):
...
+        Sync repositories (RemoteRepository) and organizations (RemoteOrganization).
+
+        - creates a new RemoteRepository/Organization per new repository
+        - updates fields for existing RemoteRepository/Organization
+        - deletes old RemoteRepository/Organization that are not present for this user
+        """
+        repos = self.sync_repositories()
+        organizations, organization_repos = self.sync_organizations()
+
+        # Delete RemoteRepository where the user doesn't have access anymore
```
ericholscher, line 199: I wonder if we should run these delete tasks across all users as a one-time CLI command, to clean up our DB. There's likely a ton of these sitting around.

humitos: This one-time command would require to sync _all VCS_ account connected and delete all the old objects. That may be too much. We may want to limit the amount of accounts re-synced and run it several times by chunks. It will remove a lot of stale data.

However, as this cleanup is in the core of the sync action, all active accounts will be cleanup automatically when they login next time or, when they click the arrow from Import Project page.

humitos:
I'm going to merge this since it's useful as is it. We can keep the conversation about re-sync all the accounts and make a final decision, tho.

PR #7315, opened 2020-07-23
added info on custom hosting with GitHub Pages

maltfield:
added info on custom hosting with GitHub Pages

### DIFF
--- a/docs/custom_installs/github_pages.rst
+++ b/docs/custom_installs/github_pages.rst
@@ -1,0 +1,27 @@
+Hosting on GitHub Pages
+=======================
+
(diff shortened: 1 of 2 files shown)


ericholscher:
Thanks for your contribution! However, these are docs about using Read the Docs, not using our theme on other hosting providers. I'd recommend putting these docs somewhere else where they are more appropriate. Perhaps as a doc in our theme repository: https://github.com/readthedocs/sphinx_rtd_theme/

maltfield:
...but you already have a section on custom installs.

It seems appropriate that this section should at least provide *links* to other documentation on "appropriate" sites, which is what I've done with this commit.

ericholscher:
That is documentation on installing our code base, which this is documenting. I think it's more fitting for the theme itself, since it related to that project, not anything to the Read the Docs hosting service.

Issue #7313, opened 2020-07-23
orjson is outdated (Python 3.8 compat)


benjaoming:
## Details

Hi friends, working on a development setup where I was incidentally on Python 3.8. I think this could easily be fixed, as orjson doesn't seem to have any backwards-incompatible changes, apart from [this](https://github.com/ijl/orjson/releases/tag/3.0.0):

> orjson.dumps() serializes dataclasses.dataclass and uuid.UUID instances by default. The options OPT_SERIALIZE_DATACLASS and OPT_SERIALIZE_UUID can still be specified but have no effect.

* Read the Docs project URL: n/a
* Build URL (if applicable): n/a
* Read the Docs username (if applicable): n/a

## Expected Result

I am not able to build ReadTheDocs w/ Python 3.8 on a system w/o Rust. I assume it's because the old Wheel doesn't come with pre-compiled binaries for Py3.8. Which is fair and expectable.

This works up until Python 3.7.

By using a newer version, it's possible again to install.

It's recommended to use `orjson>=3,<4`, see: https://github.com/ijl/orjson + several bug fixes: https://github.com/ijl/orjson/releases

## Actual Result

```
(readthedocs) ➜  readthedocs.org git:(4f25c132e) ~ pip install orjson==2.0.7 -i https://pypi.org/simple
Collecting orjson==2.0.7
  Downloading orjson-2.0.7.tar.gz (480 kB)
     |████████████████████████████████| 480 kB 5.2 MB/s 
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
    Preparing wheel metadata ... error
    ERROR: Command errored out with exit status 1:
     command: /home/user/.virtualenvs/readthedocs/bin/python3 /home/user/.virtualenvs/readthedocs/lib/python3.8/site-packages/pip/_vendor/pep517/_in_process.py prepare_metadata_for_build_wheel /tmp/tmp0il1_f4d
         cwd: /tmp/pip-install-yhst16n9/orjson
    Complete output (6 lines):
    
    Cargo, the Rust package manager, is not installed or is not on PATH.
    This package requires Rust and Cargo to compile extensions. Install it through
    the system's package manager or via https://rustup.rs/
    
    Checking for Rust toolchain....
    ----------------------------------------
[...]

PR #7332, opened 2020-07-28
Update to latest release of orjson

benjaoming:
Researching if an upgrade is possible, re: https://github.com/readthedocs/readthedocs.org/issues/7313


benjaoming:
Thanks @humitos - I need to have a look at the `inv` tool, will get back with a result :+1:

humitos:
@benjaoming it's basically our tool to start the development instance. Just want to double check that it will work after merging. You can read more at https://docs.readthedocs.io/en/stable/development/standards.html

humitos:
I think this PR won't be required anymore after #7421 gets merged.

PR #7336, opened 2020-07-29
Re-sync GitHub (RemoteRepository and RemoteOrganization) on webhook

humitos:
Add `member` event when creating the Read the Docs' webhook on each repository
imported under Read the Docs.

This webhook will communicate back to us when a collaborator was
added/removed/changed from that repository in particular. This allow us to
trigger the re-sync method for the GitHub service on that user and
create/delete/change `RemoteRepository` objects for that user.

Examples,

* user is removed from having access to a repository -> remove RemoteRepository
* user is added access to a repository -> create RemoteRepository
* user is removed admin access but keep read access -> `.admin=False`

Related to #7185 

Documentation reference: https://docs.github.com/en/developers/webhooks-and-events/webhook-events-and-payloads#member

### DIFF
--- a/readthedocs/api/v2/views/integrations.py
+++ b/readthedocs/api/v2/views/integrations.py
@@ -6,6 +6,9 @@ import json
 import logging
 import re
 
+from allauth.socialaccount.providers.github.provider import GitHubProvider
+from allauth.socialaccount.models import SocialAccount
+
 from django.shortcuts import get_object_or_404
 from rest_framework import permissions, status
 from rest_framework.exceptions import NotFound, ParseError
@@ -27,6 +30,7 @@ from readthedocs.core.views.hooks import (
     build_external_version,
 )
 from readthedocs.integrations.models import HttpExchange, Integration
+from readthedocs.oauth.services.github import GitHubService
 from readthedocs.projects.models import Project, Feature
 
 
@@ -34,6 +38,7 @@ log = logging.getLogger(__name__)
 
 GITHUB_EVENT_HEADER = 'HTTP_X_GITHUB_EVENT'
 GITHUB_SIGNATURE_HEADER = 'HTTP_X_HUB_SIGNATURE'
+GITHUB_MEMBER = 'member'
 GITHUB_PUSH = 'push'
 GITHUB_PULL_REQUEST = 'pull_request'
 GITHUB_PULL_REQUEST_OPENED = 'opened'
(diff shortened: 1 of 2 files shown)


readthedocs/api/v2/views/integrations.py
```diff
@@ -459,6 +464,16 @@ def handle_webhook(self):
...
 
+        # Re-sync repositories for the user if any permission has changed
+        if event == GITHUB_MEMBER:
+            uid = self.data.get('member').get('id')
+            socialaccount = SocialAccount.objects.get(
+                provider=GitHubProvider.id,
+                uid=uid,
+            )
+            service = GitHubService(user=socialaccount.user, account=socialaccount)
+            service.sync()
```
humitos, line 475: We may want to do this async via a Celery task. We already have [`sync_remote_repositories`](https://github.com/readthedocs/readthedocs.org/blob/2565e275ff1c3d45af1d3b9e387a7bc5591d8c41/readthedocs/oauth/tasks.py#L27) but we should add an argument to accept which service to sync instead of doing it on all of them --if we want to be precise here.

humitos:
This PR is useful. I haven't had the time to update it with the newer code, but we will eventually need it.

humitos:
> This webhook will communicate back to us when a collaborator was added/removed/changed from that repository in particular

Take into account that this only works for repository collaborators. There is another webhook called `membership` that works for users added/removed from an organization's team: https://docs.github.com/en/developers/webhooks-and-events/webhook-events-and-payloads#membership . However, it's only available for organization webhooks and we are not making usage of them.

----

Even if we merge this PR, all the Read the Docs projects have to re-sync their webhooks (:sad:) to make usage of the new `member` event. New projects will use it automatically.

readthedocs/api/v2/views/integrations.py
agjohnson: Yeah, :+1: on moving this to a celery task. Lately, we have definitely been feeling the effects of executing these sorts of tasks in-process on the webs. `sync()` is perfectly fine otherwise :+1:

agjohnson (changes requested):
I'd agree that celery task is the most correct here. We've had issues lately with blocking web processes and should avoid that pattern going forward.

humitos:
We are moving away from OAuth Application to GitHub Application: #8445 

If anything, we will implement the re-sync via the events we can subscribe there.

ericholscher:
I'd love to get this working via Celery. I think it might help us keep SSO in sync.

humitos:
@ericholscher 

> I don't fully understand the issues with it, but I do think we want some kind of way to do real-time updates on permissions.

The implementation of this webhook may make you feel that it solves the problem, where it does not. That's what I'm trying to say. It will keep permissions over repositories updated, but it won't for permissions granted via organizations and teams --which is the main problem we want to solve here.

That said, it may help in some cases and add more confusion in others.

The bad UX because we update permissions once a day, will still be present even with this webhook.

PR #7346, opened 2020-07-31
Add a middleware for referrer policy

davidfischer:
This sets a [referrer policy](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Referrer-Policy) with the value of `strict-origin-when-cross-origin` for the RTD dashboard. Proxito (which serves documentation sites) will use `no-referrer-when-downgrade` so that API calls from docs sites to readthedocs.org will report the referrer.

Setting the setting `SECURE_REFERRER_POLICY=None` will turn off the middleware.

I didn't use the existing [django-referrer-policy](https://github.com/ubernostrum/django-referrer-policy) module because this new middleware is setup so that it can just be removed when we upgrade to Django 3. Referrer policy is a built-in feature of Django's SecurityMiddleware (which we use) in Django 3. The django-referrer-policy uses a different setting name than what Django uses in Django3.

### DIFF
--- a/dockerfiles/nginx/proxito.conf
+++ b/dockerfiles/nginx/proxito.conf
@@ -64,6 +64,8 @@ server {
         add_header X-RTD-Redirect $rtd_redirect always;
         set $cache_tag $upstream_http_cache_tag;
         add_header Cache-Tag $cache_tag always;
+        set $referrer_policy $upstream_http_referrer_policy;
+        add_header Referrer-Policy $referrer_policy always;
     }
 
     # Serve 404 pages here
--- a/readthedocs/settings/proxito/base.py
+++ b/readthedocs/settings/proxito/base.py
@@ -11,6 +11,7 @@ class CommunityProxitoSettingsMixin:
 
     ROOT_URLCONF = 'readthedocs.proxito.urls'
     USE_SUBDOMAIN = True
+    SECURE_REFERRER_POLICY = "no-referrer-when-downgrade"
 
     @property
     def DATABASES(self):
--- a/readthedocs/core/middleware.py
+++ b/readthedocs/core/middleware.py
@@ -188,7 +188,7 @@ class ReferrerPolicyMiddleware:
     def __init__(self, get_response):
         self.get_response = get_response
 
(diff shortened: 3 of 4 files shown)


davidfischer:
Chrome is going to make `strict-origin-when-cross-origin` the default [in the next few months](https://developers.google.com/web/updates/2020/07/referrer-policy-new-chrome-default). Maybe we should use the same for Proxito/API calls and live with the slightly reduced data. It's probably not needed.

dockerfiles/nginx/proxito.conf
```diff
@@ -64,6 +64,8 @@ server {
         add_header X-RTD-Redirect $rtd_redirect always;
         set $cache_tag $upstream_http_cache_tag;
         add_header Cache-Tag $cache_tag always;
+        set $referrer_policy $upstream_http_referrer_policy;
+        add_header Referrer-Policy $referrer_policy always;
```
ericholscher, line 68: This needs a corresponding `ops` change as well.

readthedocs/settings/proxito/base.py
```diff
@@ -11,6 +11,7 @@ class CommunityProxitoSettingsMixin:
 
     ROOT_URLCONF = 'readthedocs.proxito.urls'
     USE_SUBDOMAIN = True
+    SECURE_REFERRER_POLICY = "no-referrer-when-downgrade"
```
ericholscher, line 14: This makes sense 👍

dockerfiles/nginx/proxito.conf
davidfischer: Yes it does although the critical piece is getting the stricter policy for the dashboard.

davidfischer: I added a PR for the corresponding ops change.

readthedocs/settings/proxito/base.py
davidfischer: To be honest, I think `strict-origin-when-cross-origin` would be fine for docs sites as we don't really use the additional data as far as I'm aware but at least we have an easy upgrade path this way.

PR #7443, opened 2020-09-03
Update links to advertising

davidfischer:
- Don't rely on the redirects if possible
- Add UTM codes
- Highlight EthicalAds.io in ethical ads docs


ericholscher (approved):
Looks good. I don't know if we need full utm codes, I was thinking something closer to what codefund did with a single GET arg `?ref=readthedocs_footer` and `?ref=readthedocs_docs`, since we aren't actually capturing this data into a marketing tool. Though once we have server logs we could build some basic data around it, but I'm not sure having 3 pieces of data is more useful than 1.

davidfischer:
> I don't know if we need full utm codes

Works for me.

PR #7449, opened 2020-09-03
Truncate the beggining of the commands' output

humitos:
Instead trunctaing the last part of the command, we truncate the beggining of
it. This allows to understand what's happening if the command fails (e.g. LaTeX
step, https://readthedocs.org/projects/platformio/builds/11763095/) with more context to be able to solve it.


ericholscher (approved):
Yea, that's a bad bug :(

PR #7458, opened 2020-09-05
correct invalid YAML

AvdN:
In YAML, scalars cannot start with a `*` as this character is an indicator character ( https://yaml.org/spec/1.2/spec.html#id2772075 , rule [14]) used for denoting aliases ( https://yaml.org/spec/1.2/spec.html#id2786196 ). Therefore this key needs to be quoted.


stsewd (approved):
Thanks!

PR #7459, opened 2020-09-05
correct inconsistent indentation of YAML

AvdN:
The YAML examples sometimes use two, sometimes use three spaces to indent block mapping values that are collections
(see the examples for https://docs.readthedocs.io/en/stable/config-file/v2.html#build resp.  https://docs.readthedocs.io/en/stable/config-file/v2.html#requirements-file )

For sequences that are mapping values (and need as a miminum, an indent of two with an offset of zero, i.e. left aligned with the key) things are even more diverse:
- the first YAML example has an offset of two in an indent of four.
- the second example for **formats** ( https://docs.readthedocs.io/en/stable/config-file/v2.html#formats ), has an offset of three in an indent of five
- the value for `extra_requirements for **python** ( https://docs.readthedocs.io/en/stable/config-file/v2.html#python ) has an offset of four in an indent of six

This PR consistently formats mappings with an indent of two and sequences with an indent of four and offset of two.

I can make another PR with other indent schemes, this update was programmatically done using:

```
import sys
from pathlib import Path
import ruamel.yaml

yaml = ruamel.yaml.YAML()
yaml.indent(mapping=2, sequence=4, offset=2)

in_path = Path('v2.rst.org')
out_path = Path('v2.rst')
code_indent = 3  # YAML relative to '.. code-block:: yaml'
pattern = '.. code-block:: yaml'
pattern_indent = None
yaml_str = ""

class ReAddIndent:
    def __init__(self, indent):
        self.indent = ' ' * indent

    def __call__(self, s):
        # indent all non-empty lines
        return '\n' + '\n'.join([(self.indent + line if line else line) for line in s.splitlines()]) + '\n'


with out_path.open('w') as out:
    for line_nl in in_path.open():
        line = line_nl.rstrip('\n')
        if pattern_indent is not None:
            leading = line[0:pattern_indent + code_indent].strip()
            #if not yaml_str and not leading: # leading empty line(s)
            #    print(line, file=out)
            #    continue
            if leading != '':  # end of YAML block
                yaml.dump(yaml.load(yaml_str), out, transform=ReAddIndent(pattern_indent+code_indent))
                pattern_indent = None
                print(line, file=out)
                continue
            yaml_str += line[pattern_indent+code_indent:] + '\n'
            continue
        elif (ind := line.find(pattern)) >= 0:
            yaml_str = ""
            pattern_indent = ind
        print(line, file=out)
```


stsewd (approved):
Thanks! Looks like you need to resolve a conflict, so we can merge this PR

PR #7489, opened 2020-09-17
Limit concurrency per-organization

humitos:
Allow us to limit concurrency per-organization besides per-project. Both limits are compatible and can be used together, giving more priority to the limits per-project. This means we can limit a whole organization to 10 concurrent builds but a particular project from the organization could be limited to only 2.

> NOTE: I think this is the first time we put logic code for `Organization` into the community site. I think it's fine now since we have all the models here. We could override the `BuildQuerySetBase.concurrent` method from corporate, but I think it's just more work without real benfits


readthedocs/builds/querysets.py
```diff
@@ -149,12 +154,22 @@ def concurrent(self, project):
...
         concurrent = (
             self.filter(query)
             .exclude(state__in=[BUILD_STATE_TRIGGERED, BUILD_STATE_FINISHED])
-        ).count()
+        ).distinct().count()
 
-        max_concurrent = project.max_concurrent_builds or settings.RTD_MAX_CONCURRENT_BUILDS
+        max_concurrent = (
+            project.max_concurrent_builds or
+            (organization.max_concurrent_builds if organization else 0) or
```
ericholscher, line 170: Do we want `else 0` here? Will the 0 cause the `or` to move to the default? It seems a bit confusing.

readthedocs/organizations/models.py
```diff
@@ -69,6 +69,11 @@ class Organization(models.Model):
         help_text='Docs and builds are disabled for this organization',
         default=False,
     )
+    max_concurrent_builds = models.IntegerField(
```
ericholscher, line 72: I think we want this to come from the organizations plan, not from the organization itself. We can have overrides on the org, but in general it should come from the Plan. We need to think about how to architect this, so it's easy to override on the .com.

readthedocs/builds/querysets.py
humitos: Yes, this may be a little tricky. Basically the `0` here works as `False` and makes the `or` to continue to the next value:

```
In [27]: None or 0 or 5
Out[27]: 5
```

readthedocs/organizations/models.py
humitos: Yes, I agree. The max per organization should come from the organization's plan. We could:

1. update the `organization.max_concurrent_builds` via a signal when the organization's plan changes (listening to `Organization.post_save` or similar) to keep this value updated
1. write `QuerySet`/`Manager` (similar to [`CorporateProjectQuerySet.is_active`](https://github.com/readthedocs/readthedocs-corporate/blob/8a8dc755631e5e4513f830836378fa02865fdbc5/readthedocsinc/projects/querysets.py#L80-L101)) method that's overridden in .com to return `org.max_concurrent_builds or plan.max_concurrent_builds` 

1) sounds more like a hack and may be hard to keep them in sync if something fails (or if we change the max per org at a Plan level. 2) looks more similar to the pattern we have been following, so it's probably better.

ericholscher: Yea, I'd prefer #2. No reason to create a cache to have issues with if you can query the value directly.

humitos: Great! I updated this code to use a QuerySet and created a PR in commercial to override this and consider the plan limits. Let me know!

ericholscher (approved):
This looks good 👍

PR #7493, opened 2020-09-18
Add proper div names on our ad placements

ericholscher:
This will be deprecated here soon,
but for now this will make that data more clean.


humitos (commented):
I'm not sure if I have the context about this change. I guess we just want human-readable names here but don't really know how/where they are used. We may want to keep the postfix random number as it was, tho: `rtd-explicit-2918`.

ericholscher:
Why do we want the random number?

humitos:
I don't have the answer to that, but I suppose that it was added for some reason :sweat_smile:, obfuscate or avoid id collisions, perhaps?

ericholscher:
I talked with David, and these were originally done for ad blocking, but it didn't work :) So I think they are safe to remove, tho this code will be removed in a couple weeks anyway, so I'm fine if this doesn't ship.

humitos (approved):
OK!

PR #7506, opened 2020-09-23
No longer proxy RTD ads through RTD servers

davidfischer:
- Instead of proxying ads through `readthedocs.org/api/...`, hit EthicalAds directly. This has a lot of performance advantages from a server scalability perspective since there won't be any process on RTD servers waiting on ad responses.
- This relies on additional data (whether ad free, community ads only, keywords) being sent in ~~the Footer API response or in having~~ a new API endpoint.
- Rely on viewport detection in the ad client (https://github.com/readthedocs/ethical-ad-client/pull/29) which will be present in the beta version
- Removes the footer ad type and the fixed footer ad types entirely. Instead, we will place a regular ad in the footer if the sidebar would push the ad off the screen.
- From a timing perspective, when the DOM is ready, we immediately query the new RTD user data endpoint and concurrently load the ethical ad client. When the RTD user data endpoint returns, that's when we request an ad. The performance of this shouldn't be significantly worse than our current code.

## Open Questions/To-Do

* [X] We might bring back the sustainability API but instead of proxying the connection to the ad server, it just returns the data necessary to display an ad. (Edit: we did this)
  - To get the functionality on par with our existing system, we also want to send keywords related to the project to the ad server. This is used for content targeting.
  - We have a few projects on RTD where revenue is shared. These have specific publisher IDs which we will also need to sent to the ad server.
* [X] The ad client only works for image or text ads. Any other custom types might need a modification. The [ad client appends](https://github.com/readthedocs/ethical-ad-client/blob/cba5bbe33c86e179ae855e81c514d3e0aeb53d8e/index.js#L69) `v1` to either the image or text ad type. (Edit: added in https://github.com/readthedocs/ethical-ad-client/pull/31)
* [X] There's some additional functionality I stripped out in the initial implementation to pick other ad types (footer ads, mostly). This will need to come back with some modifications to support the ad client. (Edit: these features are back)

## Screenshot

Screen Shot 2020-09-23 at 3 41 23 PM


readthedocs/core/static-src/core/js/doc-embed/footer.js
```diff
@@ -78,6 +79,10 @@ function init() {
             }
             injectFooter(data);
             setupBookmarkCSRFToken();
+
+            if (!data.ad_free_user && !data.ad_free_project) {
+                sponsorship.init();
```
ericholscher, line 84: I imagine this is going to add a decent bit of latency to ad viewing, I wonder if we should do the ads request always, but vary the display based on this data?

ericholscher (commented):
This seems like a good start. I see the issues arounds special casing "publishers within a publisher" and some of the other integration data we were doing on the server side. I don't have a great solution tho, but I think it'll be useful to figure out if we can.

readthedocs/core/static-src/core/js/doc-embed/sponsorship.js
```diff
@@ -3,224 +3,53 @@
...
-        }
-
-        return {
-            'div_id': element_id,
-            'display_type': display_type,
-            'priority': priority,
-        };
+        container = $('<div />').attr('style', 'text-align:center').appendTo(selector);
+        $('<div />')
+            .attr('data-ea-publisher', "readthedocs")
```
ericholscher, line 45: We could include the publisher group in the footer response to adjust this for our revshare folks, but that does require blocking this on the server response.

readthedocs/api/v2/views/footer_views.py
```diff
@@ -216,6 +216,20 @@ def _get_context(self):
         }
         return context
 
+    def is_ad_free_user(self, user):
+        if not settings.USE_PROMOS:
+            return True
+        if user.is_authenticated and hasattr(user, 'gold') and hasattr(user, 'goldonce') and (user.gold.exists() or user.goldonce.exists()):
```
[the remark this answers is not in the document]
davidfischer: I believe the logic is correct. It is an OR except first checking that the attributes exist. Since this API can be called even when the gold app or the donate app are not installed, this extra logic is necessary here.

readthedocs/api/v2/views/footer_views.py
```diff
@@ -216,6 +216,20 @@ def _get_context(self):
...
+            return True
+        return False
+
+    def is_ad_free_project(self, project):
+        if not settings.USE_PROMOS:
+            return True
+        if project and hasattr(project, 'gold_owners') and (project.gold_owners.exists() or project.ad_free):
```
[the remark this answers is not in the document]
davidfischer: The check `hasattr(project, 'gold_owners')` just checks whether that object is present, not whether there actually is a gold owner. This is needed because the gold app may not be installed.

readthedocs/core/static-src/core/js/doc-embed/footer.js
davidfischer: It will add some latency, but I'm not sure how you can actually do what you're saying. Either you add the `ethicalads.js` meaning you want an ad or you don't. To determine that, you need to know some additional data about the user or project. Having a stripped down API that just gets the project/user data would likely be faster although there would be an additional concurrent API call.

readthedocs/core/static-src/core/js/doc-embed/sponsorship.js
davidfischer: I don't think you can initialize ads without first getting the project/user data. All the decisions you want to make require that data.

ericholscher: One idea: 99% of our pageviews aren't revshare or gold users, so I still think loading the ad JS makes sense at pageload, but we can hide the ad display until we get the footer data back. That will keep ads at the same latency as now, and we just need to re-request them on revshare projects. We could even dump some metadata into the built HTML for revshare projects to be able to read that data prior to the initial request. 

This seems like it'll get us almost every ad view in the same latency as the previous approach, if not faster.

readthedocs/api/v2/views/footer_views.py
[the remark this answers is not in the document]
ericholscher: Gotcha -- we should probably just check for that explicitly (`if 'donate' in settings.INSTALLED_APPS`), or at least add a comment. It isn't clear that's the logic from the code.

readthedocs/core/static-src/core/js/doc-embed/footer.js
ericholscher: I think we do the ad request, but we can hide it with CSS until we confirm we want to show it. That way as soon as the footer responds, we can display the ad, instead of doing another full round trip.

davidfischer: I don't think we should be hiding ads with CSS and showing them based on API responses. I think this will lead to complication. Instead, perhaps there's a way to add some integration points to the ad client so we can control when requests for an ad are made.

ericholscher: I think we can start with this approach for now. But I do think trying to reduce latency on the ad display is important.

readthedocs/api/v2/views/footer_views.py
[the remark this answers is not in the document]
humitos: I think Eric is right.

In the case that `hasattr(user, 'gold') -> True` and `user.gold.exists() -> True`, but `hasattr(user, 'goldonce') -> False` it won't enter the IF block but it seems that it's currently ad free.

However, I'm not sure if it's possible that `gold` exists but not `goldonce`, tho.

Maybe this change?

```python
if user.is_authenticated and ((hasattr(user, 'gold') and user.gold.exists()) or (hasattr(user, 'goldonce') and user.goldonce.exists())):
``

File: readthedocs/api/v2/views/footer_views.py (hunk 4 above)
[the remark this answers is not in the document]
Comment (humitos) [reply readthedocs/api/v2/views/footer_views.py:229]: When this if's gets complicated on the `or` things, I try to split per each case to avoid confusions, like:

```python
if not settings.USE_PROMOS:
  # commercial site
  return True

if project.ad_free:
  # marked manually as ad free
  return True

if project and hasattr(project, 'gold_owners') and project.gold_owners.exists():
  # project's owner is gold member
  return True

return False
```

You end up with multiple exits where the output is the same, `True`, but it's clearer to read, IMO.

readthedocs/core/static-src/core/js/doc-embed/footer.js
humitos: > I don't think we should be hiding ads with CSS and showing them based on API responses

Is it possible to do the request for the ad, save all the data needed (view URL, click URL, image URL, text, etc) without creating the HTML element to display it yet, until we receive the response from the footer and _there_ create the HTML element and show it?

I assume the ad-client is not thought to work like that at this point, but I think we could have the best of both ideas with lower latency.

ericholscher:
We talked about this on a call, and the proposed path forward:

* Have revshare users add the EA `div` to their docs directly, so we know the publisher
* Update the ad client so it can fetch ads, but not display them
* Add a lightweight ads api that just returns gold user status
* Request the ad & footer at pageload, then only display the ad once we confirm it isn't a gold user.

readthedocs/core/static-src/core/js/doc-embed/footer.js
ericholscher: Yea, this is what we settled on 👍

readthedocs/api/v2/views/footer_views.py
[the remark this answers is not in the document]
davidfischer: After discussing, we're going to check `settings.INSTALLED_APPS`.

davidfischer:
This has been updated and I updated the description. It is ready for a review.

readthedocs/core/static-src/core/js/doc-embed/sponsorship.js
```diff
@@ -3,51 +3,43 @@
...
-
-        return {
-            'div_id': element_id,
-            'display_type': display_type,
-        };
-    }
-    return null;
+function inject_ads_client() {
+    var script = document.createElement("script");
+    script.src = "https://media.ethicalads.io/media/client/beta/ethicalads.min.js";
```
ericholscher, line 19: We should rebase this to get the latest updates.

ericholscher (approved):
Looks good 🎉

readthedocs/core/static-src/core/js/doc-embed/sponsorship.js
davidfischer: You mean regenerate this beta client?

ericholscher: Yea, so it has the latest other client features.

davidfischer: I built the latest beta client.

davidfischer:
This is testing out really well. Here's how you can test it:

* Set `USE_PROMOS = True` in `readthedocs/settings/docker_compose.py`
* Build the latest static assets on RTD `npm run build`
* Load ads on a built project. Verify that the view (`https://server.ethicalads.io/proxy/view/...`) is triggered only if the ad is in the viewport. The ad should be loaded after the gold user/project is checked.

ericholscher:
Tested this locally and it looks 💯

Issue #7509, opened 2020-09-25
Add RemoteRepository API v3 endpoint


agjohnson:
In order to enable search as you type project creation, an API v3 endpoint is needed for the RemoteRepository model. It should allow filtering by `contains` for the repository name, at very least.

saadmk11:
It would be great to have a search option in the import page, It would make a smoother import experience. :100:

PR #7510, opened 2020-09-26
Add List API Endpoint for `RemoteRepository` and `RemoteOrganization`

saadmk11:
closes https://github.com/readthedocs/readthedocs.org/issues/7509

### DIFF
--- a/docs/api/v3.rst
+++ b/docs/api/v3.rst
@@ -1681,7 +1681,7 @@ Remote Organization listing
 
     The ``results`` in response is an array of remote organizations data.
 
-    :query string name__contains: return remote organizations with containing the name
+    :query string name: return remote organizations with containing the name
     :query string vcs_provider: return remote organizations for specific vcs provider (``github``, ``gitlab`` or ``bitbucket``)
 
     :requestheader Authorization: token to authenticate.
@@ -1805,7 +1805,7 @@ Remote Repository listing
 
     The ``results`` in response is an array of remote repositories data.
 
-    :query string name__contains: return remote repositories containing the name
+    :query string name: return remote repositories containing the name
     :query string vcs_provider: return remote repositories for specific vcs provider (``github``, ``gitlab`` or ``bitbucket``)
     :query string organization: return remote repositories for specific remote organization (using remote organization ``slug``)
     :query string expand: allows to add/expand some extra fields in the response.
--- a/docs/api/v3.rst
+++ b/docs/api/v3.rst
@@ -1807,7 +1807,7 @@ Remote Repository listing
 
     :query string name__contains: return remote repositories containing the name
     :query string vcs_provider: return remote repositories for specific vcs provider (``github``, ``gitlab`` or ``bitbucket``)
-    :query string organization: return remote repositories for specific remote organization (Using remote organization ``slug``)
+    :query string organization: return remote repositories for specific remote organization (using remote organization ``slug``)
     :query string expand: allows to add/expand some extra fields in the response.
                           Allowed values are ``project`` and ``organization``.
                           Multiple fields can be passed separated by commas.
--- a/docs/api/v3.rst
+++ b/docs/api/v3.rst
@@ -1624,7 +1624,6 @@ Organization projects list
        }
 
 
-
 Remote Repositories
 ~~~~~~~~~~~~~~~~~~~
 
--- a/docs/api/v3.rst
+++ b/docs/api/v3.rst
@@ -1694,7 +1694,7 @@ Remote Repositories are the importable repositories connected via
 ``GitHub``, ``GitLab`` and ``BitBucket``.
 
 
-Remote Repositories listing
+Remote Repository listing
 +++++++++++++++++++++++++++
 
 
@@ -1729,6 +1729,16 @@ Remote Repositories listing
             "previous": null,
             "results": [
                 {
+                    "organization": {
+                        "avatar_url": "https://avatars.githubusercontent.com/u/12345?v=4",
+                        "created": "2019-04-29T10:00:00Z",
+                        "modified": "2019-04-29T12:00:00Z",
+                        "name": "Organization Name",
+                        "pk": 1,
+                        "slug": "organization",
+                        "url": "https://github.com/organization",
+                        "vcs_provider": "github"
+                    },
                     "avatar_url": "https://avatars3.githubusercontent.com/u/test-rtd?v=4",
                     "clone_url": "https://github.com/rtd/project.git",
                     "created": "2019-04-29T10:00:00Z",
--- a/docs/api/v3.rst
+++ b/docs/api/v3.rst
@@ -1624,6 +1624,69 @@ Organization projects list
        }
 
 
+Remote Organizations
+~~~~~~~~~~~~~~~~~~~
+
+Remote Organizations are the importable organizations connected via
+``GitHub``, ``GitLab`` and ``BitBucket``.
+
+
+Remote Organization listing
++++++++++++++++++++++++++++
+
+
+.. http:get:: /api/v3/remote/organizations/
+
+    Retrieve a list of all Remote Organizations for the authenticated user.
+
+    **Example request**:
+
+    .. tabs::
+
+        .. code-tab:: bash
+
+            $ curl -H "Authorization: Token <token>" https://readthedocs.org/api/v3/remote/organizations/
+
+        .. code-tab:: python
+
+            import requests
+            URL = 'https://readthedocs.org/api/v3/remote/organizations/'
+            TOKEN = '<token>'
+            HEADERS = {'Authorization': f'token {TOKEN}'}
+            response = requests.get(URL, headers=HEADERS)
+            print(response.json())
+
+    **Example response**:
+
+    .. sourcecode:: json
+
+        {
+            "count": 20,
+            "next": "api/v3/remote/organizations/?limit=10&offset=10",
+            "previous": null,
+            "results": [
+                {
+                    "avatar_url": "https://avatars.githubusercontent.com/u/12345?v=4",
+                    "created": "2019-04-29T10:00:00Z",
+                    "modified": "2019-04-29T12:00:00Z",
+                    "name": "Organization Name",
+                    "pk": 1,
+                    "slug": "organization",
+                    "url": "https://github.com/organization",
+                    "vcs_provider": "github"
+                }
+            ]
+        }
+
+
+    The ``results`` in response is an array of remote repositories data.
+
+    :query string name: return remote organizations with matching name
+    :query string vcs_provider: return remote organizations for specific vcs provider (``github``, ``gitlab`` or ``bitbucket``)
+
+    :requestheader Authorization: token to authenticate.
+
+
 Remote Repositories
 ~~~~~~~~~~~~~~~~~~~
 
--- a/docs/api/v3.rst
+++ b/docs/api/v3.rst
@@ -1676,7 +1676,9 @@ Remote Repositories listing
                     "name": "project",
                     "pk": 1,
                     "ssh_url": "git@github.com:rtd/project.git",
-                    "vcs": "git"
+                    "vcs": "git",
+                    "vcs_provider": "github",
+                    "admin": true
                 }
             ]
         }
--- a/docs/api/v3.rst
+++ b/docs/api/v3.rst
@@ -1635,7 +1635,7 @@ Remote Repositories listing
 +++++++++++++++++++++++++++
 
 
-.. http:get:: /api/v3/remoterepositories/
+.. http:get:: /api/v3/remote/repositories/
 
     Retrieve a list of all Remote Repositories for the authenticated user.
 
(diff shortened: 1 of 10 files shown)


humitos (commented):
This is a good start. However, I'm not sure if we have to implement this endpoint before https://github.com/readthedocs/readthedocs.org/issues/2759 (and https://github.com/readthedocs/readthedocs.org/issues/7169) because we are changing how `RemoteRepository` will behave completely.

Maybe the API response won't change, but would be good to double check that, tho. Want to take a look at this?

Also, we may need to be able to filter by user account (github, gitlab, etc) as we are currently doing on API v2:

- https://readthedocs.org/api/v2/remote/repo/?org=162925 (filter by organization)
- https://readthedocs.org/api/v2/remote/repo/?own=gitlab (filter by account)

(you can see your own API URLs with the `org=` and `own=` values at https://readthedocs.org/dashboard/import/)

The idea of using `/remote` as a namespace may be good as well. I'd call it `/remotes/repositories/` in the new APIv3.

In APIv2 we also have `/remote/account` and `/remote/org` that I think they will be required here as well (could be done in a different PR, but should be considered in the whole implementation).

If you have the time and energy to work on these things, I'm happy to help you with the process, chat and/or talk about it.

saadmk11:
> This is a good start. However, I'm not sure if we have to implement this endpoint before #2759 (and #7169) because we are changing how `RemoteRepository` will behave completely.
> 
> Maybe the API response won't change, but would be good to double check that, tho. Want to take a look at this?

Sure! 👍  I did not see the changes being done with remote repositories. I'll look into the changes and see if there is anything we need to change about this API.

> Also, we may need to be able to filter by user account (github, gitlab, etc) as we are currently doing on API v2:

I was thinking about that, but did not know what are the requirements for API V3 are. Thanks for letting me know 💯 


> The idea of using `/remote` as a namespace may be good as well. I'd call it `/remotes/repositories/` in the new APIv3.

This sound like a good idea. 👍 

> In APIv2 we also have `/remote/account` and `/remote/org` that I think they will be required here as well (could be done in a different PR, but should be considered in the whole implementation).

Okay, so we need all the API endpoints that are required for the import process.

> If you have the time and energy to work on these things, I'm happy to help you with the process, chat and/or talk about it.
Sure, I would love to work on this, recently working with different API's and liking it a lot, and RTD has one of the best designed API 💯 

It would be great if you are able to help me with the process, Let me know when you are free to talk or chat about it, maybe on slack?

humitos:
> It would be great if you are able to help me with the process, Let me know when you are free to talk or chat about it, maybe on slack?

I'm on Slack on Europe time. Feel free to ping me there when you have time.

saadmk11:
Marking this as blocked as me and @humitos talked, this should be implemented after #2759 (and #7169)

humitos:
@saadmk11 this could be a good pull request to cleanup and finish now we have the new structure deployed.

docs/api/v3.rst
```diff
@@ -1624,6 +1624,149 @@ Organization projects list
...
+                    "url": "https://github.com/organization",
+                    "vcs_provider": "github"
+                }
+            ]
+        }
+
+
+    The ``results`` in response is an array of remote organizations data.
+
+    :query string name: return remote organizations with matching name
```
humitos, line 1684: I don't thinks this query string is required. I suppose that we will end up with an endpoint `/api/v3/remote/organizations/<slug>/` instead.

However, this `name=` attribute could be used as the `__contains` filter.

docs/api/v3.rst
```diff
@@ -1624,6 +1624,149 @@ Organization projects list
        }
 
 
+Remote Organizations
+~~~~~~~~~~~~~~~~~~~~
+
+Remote Organizations are the importable organizations connected via
+``GitHub``, ``GitLab`` and ``BitBucket``.
```
humitos, line 1631: These organizations are not "importable" :smile:

docs/api/v3.rst
```diff
@@ -1624,6 +1624,149 @@ Organization projects list
...
+
+    .. sourcecode:: json
+
+        {
+            "count": 20,
+            "next": "api/v3/remote/repositories/?limit=10&offset=10",
+            "previous": null,
+            "results": [
+                {
+                    "organization": {
```
humitos, line 1732: I think that by default we shouldn't return the `organization` object in the response. We can make this attribute expandable (`?expand=organization`).

docs/api/v3.rst
```diff
@@ -1624,6 +1624,149 @@ Organization projects list
...
+                    "admin": true
+                }
+            ]
+        }
+
+
+    The ``results`` in response is an array of remote repositories data.
+
+    :query string name: return remote repositories with matching name
+    :query string vcs: return remote repositories for specific vcs (``git``, ``svn``, ``hg`` or ``bzr``)
```
humitos, line 1763: I think we can remove this one. Is there a good use case?

docs/api/v3.rst
```diff
@@ -1624,6 +1624,149 @@ Organization projects list
...
+    :query string vcs_provider: return remote repositories for specific vcs provider (``github``, ``gitlab`` or ``bitbucket``)
+    :query string organization: return remote repositories for specific remote organization (Using remote organization ``pk``)
```
humitos, line 1765: I think we should use the remote organization `slug`, instead. Most of the API is based/using `slug` now.

docs/api/v3.rst
```diff
@@ -1624,6 +1624,149 @@ Organization projects list
...
+                    "vcs_provider": "github",
+                    "admin": true
+                }
+            ]
+        }
+
+
+    The ``results`` in response is an array of remote repositories data.
+
+    :query string name: return remote repositories with matching name
```
humitos, line 1762: Same here. I think we should use this as `__contains`. If this is the use case, we should explain it better.

We could also use `name__contains=` as attribute name as well. How are we calling these attributes in other APIv3 endpoints? We should follow the same pattern.

readthedocs/api/v3/serializers.py
```diff
@@ -891,3 +892,56 @@ class Meta:
...
+        read_only_fields = fields
+
+
+class RemoteRepositorySerializer(serializers.ModelSerializer):
+    admin = serializers.SerializerMethodField('is_admin')
+    organization = RemoteOrganizationSerializer()
+
+    class Meta:
+        model = RemoteRepository
+        fields = [
```
humitos, line 920: I think we can make `project` field expandable as well (`?expand=project`).

docs/api/v3.rst
```diff
@@ -1624,6 +1624,149 @@ Organization projects list
...
+                        "created": "2019-04-29T10:00:00Z",
+                        "modified": "2019-04-29T12:00:00Z",
+                        "name": "Organization Name",
+                        "pk": 1,
+                        "slug": "organization",
+                        "url": "https://github.com/organization",
+                        "vcs_provider": "github"
+                    },
+                    "avatar_url": "https://avatars3.githubusercontent.com/u/test-rtd?v=4",
+                    "clone_url": "https://github.com/rtd/project.git",
```
humitos, line 1743:
```suggestion
                    "clone_url": "https://github.com/organization/project.git",
```

Make sure the example matches the organization. There are some URLs pointing to `rtd` but the organization object says `organization`.

readthedocs/api/v3/serializers.py
```diff
@@ -891,3 +892,56 @@ class Meta:
...
+            'admin',
+            'avatar_url',
+            'ssh_url',
+            'clone_url',
+            'html_url',
+            'vcs',
+            'vcs_provider',
+            'created',
+            'modified',
+            'organization',
```
humitos, line 934: I think we are missing some fields here:

* `private`
* `default_branch`

readthedocs/api/v3/tests/test_remoteorganizations.py
```diff
@@ -0,0 +1,48 @@
...
+        )
+        social_account = fixture.get(SocialAccount, user=self.me, provider=GITHUB)
+        fixture.get(
+            RemoteOrganizationRelation,
+            remote_organization=self.remote_organization,
+            user=self.me,
+            account=social_account
+        )
+
+    def test_remote_organization_list(self):
```
humitos, line 38: Would be good to have a test that uses the `name=` filter which is the one we will use the most.

readthedocs/api/v3/tests/test_remoterepositories.py
```diff
@@ -0,0 +1,74 @@
...
+            admin=True
+        )
+        fixture.get(
+            RemoteOrganizationRelation,
+            remote_organization=self.remote_organization,
+            user=self.me,
+            account=social_account
+        )
+
+    def test_remote_repository_list(self):
```
humitos, line 64: Same here. Would be good to have a test that uses `?name=` filter.

readthedocs/api/v3/views.py
```diff
@@ -414,3 +430,36 @@ def get_view_name(self):
...
+
+class RemoteOrganizationViewSet(APIv3Settings, ListModelMixin, GenericViewSet):
+    model = RemoteOrganization
+    serializer_class = RemoteOrganizationSerializer
+    filterset_class = RemoteOrganizationFilter
+    queryset = RemoteOrganization.objects.all()
+    permission_classes = (IsAuthenticated,)
+
+    def get_queryset(self):
+        return super().get_queryset().api(self.request.user)
```
humitos, line 465: I'd create a `RemoteQuerySetMixin` in `api/v3/mixins.py` with this line, and use it in both views. We are doing that for the other views.

`RemoteRepositoryViewSet` may need to overwrite it tho, to add the `.annotate`, but that's fine.

readthedocs/api/v3/views.py
```diff
@@ -414,3 +430,36 @@ def get_view_name(self):
...
+                RemoteRepositoryRelation.objects.filter(
+                    remote_repository=OuterRef('pk'),
+                    user=self.request.user,
+                    admin=True
+                )
+            )
+        )
+        return queryset.select_related('organization').order_by(
+            'organization__name', 'full_name'
+        ).distinct()
```
humitos, line 454: `.select_related` will be only required if `?expand=organization` is passed.

humitos (changes requested):
Looks great! I left some comment with a few changes I'd like to see before merging this PR.

I'm also pinging @agjohnson here who is the one using this endpoint for the search-as-you-type in the new theme.

docs/api/v3.rst
saadmk11: I don't think so, `vcs_provider` should be enough, but I saw it in the API v2, so I added it as I did not know the specific requirements for the API v3. :)

docs/api/v3.rst
saadmk11: We are not using anything like `__contains`, for example:

https://github.com/readthedocs/readthedocs.org/blob/901100594bf6549eda973be84b1ab7ad44379d64/readthedocs/api/v3/filters.py#L20


https://github.com/readthedocs/readthedocs.org/blob/901100594bf6549eda973be84b1ab7ad44379d64/docs/api/v3.rst#L480

humitos: Gotcha!

docs/api/v3.rst
```diff
@@ -1681,7 +1681,7 @@ Remote Organization listing
 
     The ``results`` in response is an array of remote organizations data.
 
-    :query string name: return remote organizations with matching name
+    :query string name__contains: return remote organizations with containing the name
```
humitos, line 1684: Oh, sorry. If we are already using `name=` to filter by `icontains`, we should use the same here. No need to change it to `__contains`. I'm sorry for the confusion here.

docs/api/v3.rst
```diff
@@ -1759,10 +1805,12 @@ Remote Repository listing
 
     The ``results`` in response is an array of remote repositories data.
 
-    :query string name: return remote repositories with matching name
-    :query string vcs: return remote repositories for specific vcs (``git``, ``svn``, ``hg`` or ``bzr``)
+    :query string name__contains: return remote repositories containing the name
     :query string vcs_provider: return remote repositories for specific vcs provider (``github``, ``gitlab`` or ``bitbucket``)
-    :query string organization: return remote repositories for specific remote organization (Using remote organization ``pk``)
+    :query string organization: return remote repositories for specific remote organization (Using remote organization ``slug``)
```
humitos, line 1810:
```suggestion
    :query string organization: return remote repositories for specific remote organization (using remote organization ``slug``)
```

I think that should be lower case :)

readthedocs/api/v3/views.py
```diff
@@ -449,17 +465,26 @@ def get_queryset(self):
                 )
             )
         )
-        return queryset.select_related('organization').order_by(
-            'organization__name', 'full_name'
-        ).distinct()
+
+        if is_expanded(self.request, 'organization'):
+            queryset = queryset.select_related('organization')
```
humitos, line 470: Nice! I wasn't aware of the `is_expanded` function!

humitos (commented):
This is great!

We only need to revert the `name__contains` :angel:

humitos (approved):
:tada:

humitos:
I'm blocking this PR only on @agjohnson 's review.

agjohnson:
I don't have any particular feedback here, the PR looks great. If there is anything that I need changed it can be updated later.

humitos:
Thanks @saadmk11 for all your work here! I'm merging this PR.

PR #7536, opened 2020-10-05
Add Initial Modeling with Through Model and Data Migration for RemoteRepository Model

saadmk11:
This is the initial modeling and data migration for the ``RemoteRepository`` Model.
This represents a simple implementation to start the discussion for the implementation of #7169 

- The first migration ``0011`` decouples the ``RemoteRepository`` model with a through model ``RemoteRelation`` using  ``SeparateDatabaseAndState`` so that Django does not create a new table for this and uses the table that already exists for ``ManyToMany`` relation.

- The second migration ``0012`` adds the required fields for the ``RemoteRelation`` Model. (uses ``JSONField`` instead of  ``models.TextField`` for the ``json`` field)

- The third migration ``0013`` is added to perform data migration between ``RemoteRepository`` and ``RemoteRelation``

- The fourth migration ``0014`` removes the transferred fields from the ``RemoteRepository`` model.

Ref: #2759 and #7169


readthedocs/oauth/migrations/0013_data_migration_for_remote_relation_model.py
```diff
@@ -0,0 +1,49 @@
...
+    while True:
+        batch = list(islice(remote_relations, batch_size))
+
+        if not batch:
+            break
+        RemoteRelation.objects.bulk_update(
+            batch,
+            ['account', 'active', 'admin', 'pub_date', 'json'],
+            batch_size
+        )
```
saadmk11, line 38: Taken from Django's example https://docs.djangoproject.com/en/3.0/ref/models/querysets/#bulk-create

readthedocs/oauth/migrations/0013_data_migration_for_remote_relation_model.py
```diff
@@ -0,0 +1,49 @@
...
+            try:
+                relation.json = json.loads(relation.remoterepository.json)
+            except json.decoder.JSONDecodeError:
+                pass
+
+            yield relation
+
+    relations_queryset = RemoteRelation.objects.all().select_related('remoterepository')
+    remote_relations = remote_relations_generator(relations_queryset)
+    batch_size = 5000
```
saadmk11, line 27: Not sure about what the size can be for each batch. but if we do all of them at once it will create a huge SQL Query

readthedocs/oauth/migrations/0011_add_remote_relation_model.py
```diff
@@ -0,0 +1,48 @@
...
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        migrations.swappable_dependency(settings.AUTH_USER_MODEL),
+        ('oauth', '0010_index_full_name'),
+    ]
+
+    operations = [
+        migrations.SeparateDatabaseAndState(
```
humitos, line 16: I never used this so I will need some time to process how it works. The documentation on 2.2 is not enough, but I found that [3.1 docs](https://docs.djangoproject.com/en/3.1/ref/migration-operations/#separatedatabaseandstate) has an example of what we want to do exactly: https://docs.djangoproject.com/en/3.1/howto/writing-migrations/#changing-a-manytomanyfield-to-use-a-through-model

readthedocs/oauth/migrations/0011_add_remote_relation_model.py
```diff
@@ -0,0 +1,48 @@
...
+            state_operations=[
+                migrations.CreateModel(
+                    name='RemoteRelation',
+                    fields=[
+                        ('id', models.AutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
+                    ],
+                    options={
+                        'db_table': 'oauth_remoterepository_users',
```
humitos, line 24: The example does not use `db_table` here but uses `database_operations` and renames the table. Does this have the same effect?

readthedocs/oauth/migrations/0012_add_fields_to_remote_relation_model.py
```diff
@@ -0,0 +1,49 @@
...
+        migrations.AddField(
+            model_name='remoterelation',
+            name='account',
+            field=models.ForeignKey(blank=True, null=True, on_delete=django.db.models.deletion.SET_NULL, related_name='remote_relations', to='socialaccount.SocialAccount', verbose_name='Connected account'),
+        ),
+        migrations.AddField(
+            model_name='remoterelation',
+            name='active',
+            field=models.BooleanField(default=False, verbose_name='Active'),
+        ),
```
humitos, line 26: I'm not sure what `active` means :smile:. Do you? Why do you think it's a field from the `RemoteRelation` and not for the `RemoteRepository`? I think it's not even used (it's the first time I see it), but I'm not sure.

readthedocs/oauth/migrations/0013_data_migration_for_remote_relation_model.py
```diff
@@ -0,0 +1,49 @@
...
+
+from django.db import migrations
+
+
+def move_data_to_remote_relations(apps, schema_editor):
+    RemoteRelation = apps.get_model('oauth', 'RemoteRelation')
+
+    def remote_relations_generator(relations, batch_size):
+        for relation in relations.iterator(chunk_size=batch_size):
+            relation.account_id = relation.remoterepository.account_id
```
humitos, line 14: Shouldn't be `relation.account` here?

readthedocs/oauth/migrations/0013_data_migration_for_remote_relation_model.py
```diff
@@ -0,0 +1,49 @@
...
+            relation.active = relation.remoterepository.active
+            relation.admin = relation.remoterepository.admin
+            relation.pub_date = relation.remoterepository.pub_date
+            try:
+                relation.json = json.loads(relation.remoterepository.json)
+            except json.decoder.JSONDecodeError:
+                pass
```
humitos, line 21: It would be good to log something here with some data. At least to be warned that we couldn't migrate some data.

readthedocs/oauth/migrations/0013_data_migration_for_remote_relation_model.py
humitos: Yeah, this is one of my concerns of this migrations. I don't have a good answer yet and we may ask somebody else for advice here.

The `RemoteRepostory` table is incredible huge. I already had to trigger the creation of an INDEX in this table manually without blocking things (concurrently) because it took hours.

In this case, we will need to find a way to not block the DB while deploying these changes.

readthedocs/oauth/migrations/0013_data_migration_for_remote_relation_model.py
```diff
@@ -0,0 +1,49 @@
...
+
+def move_data_to_remote_relations(apps, schema_editor):
+    RemoteRelation = apps.get_model('oauth', 'RemoteRelation')
+
+    def remote_relations_generator(relations, batch_size):
+        for relation in relations.iterator(chunk_size=batch_size):
+            relation.account_id = relation.remoterepository.account_id
+            relation.active = relation.remoterepository.active
+            relation.admin = relation.remoterepository.admin
+            relation.pub_date = relation.remoterepository.pub_date
```
humitos, line 17: I think we should update the `modified_date` as well, to keep the original. Don't you think?

readthedocs/oauth/models.py
```diff
@@ -206,3 +195,35 @@ def matches(self, user):
...
+        null=True,
+        blank=True,
+        on_delete=models.SET_NULL,
+    )
+    active = models.BooleanField(_('Active'), default=False)
+    admin = models.BooleanField(_('Has admin privilege'), default=False)
+    json = JSONField(_('Serialized API response'))
+
+    pub_date = models.DateTimeField(_('Publication date'), auto_now_add=True)
+    modified_date = models.DateTimeField(_('Modified date'), auto_now=True)
```
humitos, line 224: For new models, I prefer if we can use the `TimeStampedModel` from django extensions.

readthedocs/oauth/migrations/0013_data_migration_for_remote_relation_model.py
humitos: Would be good to have it as a comment in the code.

humitos: Would be good to have it as a comment in the code.

readthedocs/oauth/models.py
```diff
@@ -90,14 +91,7 @@ class RemoteRepository(models.Model):
         User,
         verbose_name=_('Users'),
         related_name='oauth_repositories',
-    )
-    account = models.ForeignKey(
```
humitos, line 94: We should probably add the `remote_id` field here as well and populate it with the `id` that it's in `remoterepository.json`.

If we are not doing it in this PR, we should create an issue to keep track of. We will need it to remove the usage of `full_name` in many places.

I created a project at https://github.com/orgs/readthedocs/projects/74?add_cards_query=is%3Aopen where we can track all of the related PR and issues. Let me know if you have permissions.

readthedocs/oauth/migrations/0014_remove_field_from_remote_repository_model.py
```diff
@@ -0,0 +1,29 @@
...
+
+    dependencies = [
+        ('oauth', '0013_data_migration_for_remote_relation_model'),
+    ]
+
+    operations = [
+        migrations.RemoveField(
+            model_name='remoterepository',
+            name='account',
+        ),
```
humitos, line 16: Now that we are moving the `SocialAccount` to the `RemoteRelation`, shouldn't we have a `provider` or similar in the `RemoteRepository` to know to which service this repository belongs to?

humitos (commented):
This is going in a good direction IMO. The PR description was super useful to understand what you did :100: 

I think there is no more big modeling changes here probably. The rest of the work is more about "how to deploy it without breaking things".

We will need a plan for:

- do not kill the db while running this migrations
  - how we can improve the migration?
  - is it enough just by lowering the `batch_size`?
  - will it block the db while running the migration?
- keeping compatibility with all our existing features that use this models
  - are we deploying the model changes and data migrations together with the python code changes that use these models? (https://github.com/readthedocs/readthedocs.org/pull/7169/files#diff-12aacc84bda4107192b54a46c0922fb2R63-R79)
  - doing everything at once sounds risky to me (we need to update .org & .com) code all at once
  - is it possible to find a way of doing incremental changes here (migrate feature by feature)?

Unfortunately, I don't have the answer to most of those questions yet, but it's definitely something we need to talk about and have a plan for.

I'm not sure if you already did a test locally and ran this queries, but it would be useful if you can connect your Social accounts (github, gitlab **and** bitbucket) to your local instance and double check that the data migration and structure is all in its place.

Thanks for helping us on this, really appreciated! :heart:

readthedocs/oauth/migrations/0011_add_remote_relation_model.py
saadmk11: It would have been a lot simpler if I had seen the example but What we are doing here is quite similar.

readthedocs/oauth/migrations/0011_add_remote_relation_model.py
saadmk11: Yes, It should have the same effect, but After seeing the example from Django it seems better to rename the model. I will test that out.

readthedocs/oauth/migrations/0012_add_fields_to_remote_relation_model.py
saadmk11: I have not seen it reference anywhere, Maybe we could just remove it. I assumed that it was used to make the repository unimportable by the user or something.

readthedocs/oauth/migrations/0013_data_migration_for_remote_relation_model.py
saadmk11: We only need to make the ``ForeinKey`` relation here, so we can just use the ``_id`` to directly reference the ID and store that in the ``RemoteRelation`` model. It will make it so that no extra query is made to retrieve the ``account`` object and we don't need to add it to ``select related()`` also.

https://docs.djangoproject.com/en/3.1/ref/models/fields/#database-representation

readthedocs/oauth/migrations/0013_data_migration_for_remote_relation_model.py
saadmk11: Won't the ``modified_date`` change after we save the remote relation objects after the data migration?

readthedocs/oauth/models.py
saadmk11: Okay sure, maybe we can do data migration for this and https://github.com/readthedocs/readthedocs.org/pull/7536#discussion_r501637990 first. 

This PR will remove the json field from ``RemoteRepository`` model so I think its better to do this first in a different PR or make it the first migration on this PR.
@humitos What do you say?

readthedocs/oauth/migrations/0014_remove_field_from_remote_repository_model.py
saadmk11: Yes! Great idea :100:  It will be better to add a ``provider`` field so we don't have to do an extra query to know the provider name or if the user is deleted or disconnected we don't lose that data.

Some thought on migration of this data here https://github.com/readthedocs/readthedocs.org/pull/7536#discussion_r501799277

readthedocs/oauth/models.py
saadmk11: Okay! I'll update the PR with ``TimeStampedModel``

readthedocs/oauth/models.py
saadmk11: > I created a project at https://github.com/orgs/readthedocs/projects/74?add_cards_query=is%3Aopen where we can track all of the related PR and issues. Let me know if you have permissions.

@humitos I can not see some of the cards maybe they are from a private repo or something, otherwise it all good :+1:

saadmk11:
> I'm not sure if you already did a test locally and ran this queries, but it would be useful if you can connect your Social accounts (github, gitlab **and** bitbucket) to your local instance and double check that the data migration and structure is all in its place.

I have connected my GitHub account and imported all my repositories and ran the migrations. In addition to that I created ~53k ``RemoteRelation`` objects and ran migrations on them It took around ~50 seconds to run.

saadmk11:
Updated the PR!

Used ``TimeStampedModel`` for the ``RemoteRelationModel`` also uses the table rename method showed in Django docs to create the ``RemoterelationModel``. Added some small data migration optimization and other suggested changes.

saadmk11:
Learned something interesting, It seems ``SQLite`` has restrictions on the number of variables used in a query so it needs multiple queries for ``bulk_update()``, for my case it was doing ``41`` queries per ``5000`` row updates, So, I changed to ``Postgresql`` and it was only doing ``1`` query per ``5000`` row updates (Consumes more memory), it's the same case for ``10000`` rows together.


https://docs.djangoproject.com/en/3.0/ref/models/querysets/#bulk-update

> The batch_size parameter controls how many objects are saved in a single query. The default is to update all objects in one batch, except for SQLite and Oracle which have restrictions on the number of variables used in a query.

humitos:
I got another idea that I wanted to share here after talking about it with @saadmk11 today. Keeping in mind that we can't just run migrations on this table because is so huge that it will take down the DB.

In a few words, I'm thinking on:

- re-sync the data _we are using_ immediately (~700k `RemoteRepository`)
- migrate the data not frequently used by chunks (~6.5M `RemoteRepository`)

To accomplish this, this is the plan we've got:

1. use this PR to modify the DB schema decoupling `RemoteRepository` to create `RemoteRelation`
2. modify all the Python code to use the new modeling in .org and .com (will help us to find out bugs locally in an easier way)
3. QA this locally with test data
4. write a management command to re-sync (from GitHub) `RemoteRepository` for active users (logged in in the last 30 days). Active users will have updated data immediately (currently we have ~6k active users and we have a rate limit of 5k requests/hour to GitHub --we will need to adapt these numbers or run it multiple times on different hours)
5. enable Django signal to re-sync `RemoteRepository` on login async (we already have this in .com). New active users will have updated data immediately

At this point, we could do the first deploy. Active users and new active users will have updated data immediately after they log in. From here, we can continue migrating not frequently used data (or stale) following these steps:

6. write a periodic celery task to migrate ~1k records (from the old table to the new modeling) every ~1h to not kill the db (some of this code is written in a migration on this PR already)
  - needs to consider deleting old duplicated `RemoteRepository` records (new modeling will have only one `RemoteRepository` for all the users)
  - may needs to re-sync the `RemoteRepository` for all the users that have access to this `RemoteRepository`

> Optionally, we can spin up a test DB in production and an new instance with the new code to do as much QA as we want.

This will give us the best of two worlds: active/new users will have updated data & old users will be migrated slowly without killing the db. However, immediately one of those old users log in, their data will be automatically re-synced from GitHub via the Django signal.

7. remove all the extra fields we have in the old table.

I left 7) for the end and alone because I'm not sure if removing this fields will take the DB down as well, considering the amount of data we have in that table.

ericholscher:
@humitos This makes sense to me. I think the main thing we need to worry about is GH rate limits, but there's no reason we couldn't ship the modeling a week in advance, and then migrate the data over a longer period of time, then deploy the code to use it, right?

humitos:
> but there's no reason we couldn't ship the modeling a week in advance, and then migrate the data over a longer period of time, then deploy the code to use it, right?

We can't really ship the new modeling and continue using the old one in the application Python code. The new modeling modifies relationships between models and renames a table, so the queries will need to be changed as well. In my proposal, we are deploying the new modeling _and_ application Python code changes to use it at the same time --together with a script to re-sync active users in the same deploy.

I'm still not 100% convinced that _we have to migrate all the data at the same time_, considering that we can re-fetch it (and up to date!) from GitHub when we need it.

readthedocs/oauth/migrations/0012_add_fields_to_remote_relation_model.py
humitos: I think we can just avoid it from the migration and remove it.

readthedocs/oauth/models.py
```diff
@@ -206,3 +210,30 @@ def matches(self, user):
                 },
             ),
         } for project in projects]
+
+
+class RemoteRelation(TimeStampedModel):
```
humitos, line 215: We may need to rename this model to something like `RemoteRepositoryRelation` because we will also need a relation model for `RemoteOrganization` as well and we will have conflicting names.

Does it make sense to add make `remoterepository` field nullable and add `remoteorganization` field as well to this `RemoteRelation` model and use it for both? It seems it's exactly the same data but `RemoteOrganization` does not has `admin` field.

humitos (commented):
Changes look good to me. I'm surprised that tests keep passing after changing the `RemoteRepository.users` field. I'd expected the code to fail because of this considering that field is not valid anymore and it has to be accessed via the `remoterelation`.

We need to do exactly the same migration for `RemoteOrganization` as well.

PR #7545, opened 2020-10-07
Include month-ago day to traffic data

### LINKED ISSUE #7544: Always zero value for the first day in traffic analytics.
https://github.com/readthedocs/readthedocs.org/blob/78e6c08e8155f98a976ac7eeeac6ba06ef9cbf61/readthedocs/analytics/models.py#L102-L116

I think because `date__gt` filter is used (instead of `date__gte`) the first value (a day exactly month ago) doesn't get to the output dictionary `count_dict`. So `count_dict.get(date) or 0` is evaluates to zero every time.

kuzmoyev:
closes #7544

### DIFF
--- a/readthedocs/analytics/models.py
+++ b/readthedocs/analytics/models.py
@@ -104,7 +104,7 @@ class PageView(models.Model):
 
         queryset = cls.objects.filter(
             project__slug=project_slug,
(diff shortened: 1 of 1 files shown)


stsewd (approved):
Thanks!

PR #7565, opened 2020-10-15
Add our `readthedocs_processor` data to our notifications

humitos:
This fixes an issue where we were sending emails with `Contact us at {{
SUPPORT_EMAIL }}` but the support email variable was not set and the result was
an empty email.

Using our own `readthedocs_processor` here we are passing all the same variables
to the notification that to all the Django templates, avoiding this kind of confusions.

### DIFF
--- a/readthedocs/notifications/notification.py
+++ b/readthedocs/notifications/notification.py
@@ -3,6 +3,7 @@
 """Support for templating of notifications."""
 
 import logging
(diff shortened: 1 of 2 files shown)


ericholscher (approved):

agjohnson:
Tests appear to be breaking here stil. I'm pushing this to the next release.

PR #7579, opened 2020-10-19
Undeprecate APIv2 in docs

agjohnson:
APIv2 is still available, we still use it for application operations,
and APIv3 is not quite ready as a replacement for a lot of applications.
I think the docs should reflect that, even if we want new users using
APIv3.

### DIFF
--- a/docs/api/v2.rst
+++ b/docs/api/v2.rst
@@ -1,5 +1,5 @@
-API v2 (deprecated)
-===================
+API v2
+======
 
 The Read the Docs API uses :abbr:`REST (Representational State Transfer)`.
 JSON is returned by all API responses including errors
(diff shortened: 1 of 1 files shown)


docs/api/v2.rst
```diff
@@ -1,19 +1,21 @@
...
 JSON is returned by all API responses including errors
 and HTTP response status codes are to designate success and failure.
 
 .. note::
 
-    A newer and better API v3 is ready to use.
-    Some improvements coming in v3 are:
+    API v2 will be soon completely deprecated. If you are integrating with Read
+    the Docs, you should try to use :doc:`API v3 </api/v3>` instead. API v2 is
+    still used by some application operations still.
```
humitos, line 12: I think we should strongly not recommend the usage of APIv2: "you should try" doesn't sound too strong to me :smile: 

I'd say something like "we don't recommend the usage of APIv2 because we will remove it soon without notification. You should use APIv3 and if there is something that you can't do with APIv3, please contact us"

I don't want new users using APIv2 to do things that are supported on APIv3 and I want users that are forced to use APIv2 to contact us and let us know what's the use case.

humitos (commented):
I'm not 100% sure that we should remove the deprecation note. I do agree that there are things that are not possible in APIv3 yet, but I think we should improve the message we are communicating here.

docs/api/v2.rst
agjohnson: Yeah, a more aggressive stance would be fine here. I just have an issue calling it deprecated.

We need to go through a formal deprecation process with APIv2 still. Right now, it's in a limbo where it says it's deprecated, but it's still fully functional, and APIv3 is not yet a full replacement for APIv2. When APIv3 is ready to be our primary API, then we can call it deprecated, go through a formal process of alerting users and doing rolling brown outs of APIv2, and finally call it deprecated. Right now, we're sending mixed messages and I was a bit confused by this.

agjohnson: Added firmer language in e776174

humitos (approved):
Message is better and clearer now to me! :100:

agjohnson:
Great, and I definitely want to keep up pressure to move to APIv3, but understand it's been hard to get everything through review. Soon!

PR #7586, opened 2020-10-20
Ads JS hotfix

davidfischer:
- Fix an issue with the publisher not being set correctly
- Don't make the API call at all if ads are disabled


readthedocs/core/static-src/core/js/doc-embed/sponsorship.js
```diff
@@ -177,7 +185,7 @@ function init() {
         error: function () {
             console.error('Error loading Read the Docs user and project information');
 
-            if (!rtd.ad_free && rtd.api_host === 'https://readthedocs.org' && detect_adblock()) {
```
davidfischer, line 180: This isn't needed because it's checked on line 139 above.

ericholscher (approved):

PR #7593, opened 2020-10-22
Use correct Cache-Tag (CDN) and X-RTD-Project header on subprojects

### LINKED ISSUE #7590: .inv files are over-cached?
## Details

* Read the Docs project URL: https://edx.readthedocs.io/projects/edx-guide-for-students/en/nedbat-fix-learner-links/objects.inv
* Build URL (if applicable):
* Read the Docs username (if applicable): edx

## Expected Result

We have a multi-book docset, with references between the books.  We use an intersphinx_mapping to resolve these references:
```
intersphinx_mapping = {
    'opencoursestaff' : ('https://edx.readthedocs.io/projects/open-edx-building-and-running-a-course/en/latest/', None),
    'data' : ('https://edx.readthedocs.io/projects/devdata/en/latest/', None),
    'partnercoursestaff': ('https://edx.readthedocs.io/projects/edx-partner-course-staff/en/latest/', None),
    'insights' : ('https://edx.readthedocs.io/projects/edx-insights/en/latest/', None),
    'xblockapi' : ('https://edx.readthedocs.io/projects/xblock/en/latest/', None),
    'xblocktutorial' : ('https://edx.readthedocs.io/projects/xblock-tutorial/en/latest/', None),
    'installation' : ('https://edx.readthedocs.io/projects/edx-installing-configuring-and-running/en/latest/', None),
    'olx' : ('https://edx.readthedocs.io/projects/edx-open-learning-xml/en/latest/', None),
    'learners' : ('https://edx.readthedocs.io/projects/edx-guide-for-students/en/nedbat-fix-learner-links/', None),
    'openlearners' : ('https://edx.readthedocs.io/projects/open-edx-learner-guide/en/latest/', None),
    'opendevelopers' : ('https://edx.readthedocs.io/projects/edx-developer-guide/en/latest/', None),
    'opendataapi' : ('https://edx.readthedocs.io/projects/edx-data-analytics-api/en/latest/', None),
    'openreleasenotes' : ('https://edx.readthedocs.io/projects/open-edx-release-notes/en/latest/', None),
    'partnerreleasenotes': ('https://edx.readthedocs.io/projects/edx-release-notes/en/latest/', None),
    '2014releasenotes' : ('https://edx.readthedocs.io/projects/edx-2013-2014-release-notes/en/latest/', None),
    'retirement' : ('https://user-retirement-guide.readthedocs.io/en/latest/', None),
}
```
(The "learners" reference is to a particular branch so that my new references will be found as I work on the pull request.)

When I update one book, I find the next local build still can't resolve the new references.  Using wget to pull the .inv file, it seems that the old content is still being served, but if I add a nonsense query parameter, I get the new content:
```
$ wget -q -O - https://edx.readthedocs.io/projects/edx-guide-for-students/en/nedbat-fix-learner-links/objects.inv | wc -c
    2923
$ wget -q -O - https://edx.readthedocs.io/projects/edx-guide-for-students/en/nedbat-fix-learner-links/objects.inv\?bust | wc -c
    3821
$ wget -q -O - https://edx.readthedocs.io/projects/edx-guide-for-students/en/nedbat-fix-learner-links/objects.inv | wc -c
    2923
```

I saw the problem yesterday, and then new references seemed to resolve this morning, so it seems that waiting long enough will bring the new content, as if a cache expires.  I don't know how long that time is.

humitos:
Currently, when accessing a subproject (eg. `edx-guide-for-students` being
subproject of `edx`) we are generating `Cache-Tag: edx,edx-latest`. Then, when
a build for `edx-guide-for-students` on version `latest` succeeds, we are purging
the cache tags `edx-guide-for-students-latest`.

This commit makes the `Cache-Tag` generated for subprojects to use the
subproject's slug instead of the parent's/superproject's slug and matching the
cache tag that we are purging when the build succeeds.

Besides, it fixes the `X-RTD-Project` header as well, since it was using the
same parent's/superproject's slug as value.

Fixes #7590

### DIFF
--- a/readthedocs/proxito/middleware.py
+++ b/readthedocs/proxito/middleware.py
@@ -120,7 +120,7 @@ class ProxitoMiddleware(MiddlewareMixin):
     def add_proxito_headers(self, request, response):
         """Add debugging headers to proxito responses."""
 
-        project_slug = getattr(request, 'host_project_slug', '')
+        project_slug = getattr(request, 'path_project_slug', '')
         version_slug = getattr(request, 'path_version_slug', '')
         path = getattr(response, 'proxito_path', '')
 
--- a/readthedocs/proxito/tests/test_headers.py
+++ b/readthedocs/proxito/tests/test_headers.py
@@ -38,6 +38,22 @@ class ProxitoHeaderTests(BaseDocServing):
         self.assertEqual(r['X-RTD-version-Method'], 'path')
         self.assertEqual(r['X-RTD-Path'], '/proxito/media/html/project/latest/index.html')
 
+    def test_subproject_serve_headers(self):
+        r = self.client.get('/projects/subproject/en/latest/', HTTP_HOST='project.dev.readthedocs.io')
+        self.assertEqual(r.status_code, 200)
+        self.assertEqual(r['Cache-Tag'], 'subproject,subproject-latest')
+        self.assertEqual(r['X-RTD-Domain'], 'project.dev.readthedocs.io')
+        self.assertEqual(r['X-RTD-Project'], 'subproject')
+
+        # I think it's not accurate saying that it's `subdomain` the method
+        # that we use to get the project slug here, since it was in fact the
+        # URL's path but we don't have that feature built
+        self.assertEqual(r['X-RTD-Project-Method'], 'subdomain')
+
+        self.assertEqual(r['X-RTD-Version'], 'latest')
+        self.assertEqual(r['X-RTD-version-Method'], 'path')
+        self.assertEqual(r['X-RTD-Path'], '/proxito/media/html/subproject/latest/index.html')
+
     def test_404_headers(self):
         r = self.client.get('/foo/bar.html', HTTP_HOST='project.dev.readthedocs.io')
         self.assertEqual(r.status_code, 404)
--- a/readthedocs/proxito/views/utils.py
+++ b/readthedocs/proxito/views/utils.py
@@ -92,7 +92,8 @@ def _get_project_data_from_request(
     # * Subproject
     # * Translations
 
-    # Set the version slug on the request so we can log it in middleware
+    # Set the project and version slug on the request so we can log it in middleware
+    request.path_project_slug = final_project.slug
     request.path_version_slug = version_slug
 
     return final_project, lang_slug, version_slug, filename


readthedocs/proxito/tests/test_headers.py
```diff
@@ -38,6 +38,22 @@ def test_serve_headers(self):
...
+        r = self.client.get('/projects/subproject/en/latest/', HTTP_HOST='project.dev.readthedocs.io')
+        self.assertEqual(r.status_code, 200)
+        self.assertEqual(r['Cache-Tag'], 'subproject,subproject-latest')
+        self.assertEqual(r['X-RTD-Domain'], 'project.dev.readthedocs.io')
+        self.assertEqual(r['X-RTD-Project'], 'subproject')
+
+        # I think it's not accurate saying that it's `subdomain` the method
+        # that we use to get the project slug here, since it was in fact the
+        # URL's path but we don't have that feature built
+        self.assertEqual(r['X-RTD-Project-Method'], 'subdomain')
```
humitos, line 51: This is not related to this PR, but I wanted to make a note here, tho. Maybe a TODO where we set this header is better? https://github.com/readthedocs/readthedocs.org/blob/a81d82ed736b611491050f39649bd668ef212c0f/readthedocs/proxito/middleware.py#L141-L146

This header is wrong to all the subprojects when they are accessed via the domain of their parent's, eg. `project.readthedocs.io/projects/subproject/en/latest/`

readthedocs/proxito/middleware.py
```diff
@@ -120,7 +120,7 @@ class ProxitoMiddleware(MiddlewareMixin):
     def add_proxito_headers(self, request, response):
         """Add debugging headers to proxito responses."""
 
-        project_slug = getattr(request, 'host_project_slug', '')
+        project_slug = getattr(request, 'path_project_slug', '')
```
ericholscher, line 123: Do we not need to use `host_project_slug` sometimes as well? It seems like this is breaking existing logic, unless `path_project_slug` is getting set other places?

humitos: > Do we not need to use `host_project_slug` sometimes as well? 

As a fallback, you mean? In case that `path_project_slug` is not available? Why it would not be available?

> It seems like this is breaking existing logic, unless `path_project_slug` is getting set other places?

What existing logic do you think it's breaking? If you have an example on mind I can try to write a test case.

humitos: We are using the same logic for `version_path_project` that follows this line. It's used here and set in `_get_project_data_from_request` (https://github.com/readthedocs/readthedocs.org/blob/a81d82ed736b611491050f39649bd668ef212c0f/readthedocs/proxito/views/utils.py#L96) which is the first function that is called when serving a page at https://github.com/readthedocs/readthedocs.org/blob/a81d82ed736b611491050f39649bd668ef212c0f/readthedocs/proxito/views/serve.py#L72-L79

The `host_project_slug` is returned by `map_host_project_slug` which is accurate, https://github.com/readthedocs/readthedocs.org/blob/a81d82ed736b611491050f39649bd668ef212c0f/readthedocs/proxito/middleware.py#L159. However, the project slug we get from the host _sometimes_ (in subprojects or translations, for example) is not the project slug being served.

ericholscher: Gotcha, are other uses of `host_project_slug` unaffected by this issue? It just seems bad to have 2 different variables for the project_slug, where we use them in different places.

ericholscher (approved):
This makes sense to me, though we should audit to make sure we aren't creating additional bugs by not using the same method of resolving the `project_slug` everywhere.

ericholscher:
Gonna merge this, but @humitos please check the other places this might be needed.

nedbat:
Thanks so much, I love the responsiveness!

readthedocs/proxito/middleware.py
humitos: > It just seems bad to have 2 different variables for the project_slug, where we use them in different places.

Well, if I'm understand correctly this they _are different things_:

* `host_project_slug` is the project slug that we infer from the host/domain's URL
  * not accurate for subdomain/translations
  * it's fine for custom domains and `.readthedocs.io`
  * it's set at the beginning of the request in the middleware
* `path_project_slug` is the project slug that we can infer from the path's URL mainly
  * it's good for subdomain/translations
  * it's the same as what we have been calling `final_project` (maybe it's better to call it `final_project_slug`?)
  * it's set in the `_get_project_data_from_request` which is called from inside the `ServeDocs` view in Proxito as the first step of the request after all the middlewares

It seems that _both_ are required. We use `host_project_slug` as first step trying to get the `Project` object (`map_project_slug` decorator). Then, inside the `_get_project_data_from_request` we use `host_project_slug` to check single project, translation and subproject and set the `path_project_slug` which in those cases could be different than `host_project_slug`.

> Gotcha, are other uses of `host_project_slug` unaffected by this issue? 

Grepping the code I didn't find anything that clearly needs to be adapted. Maybe there is something deeper that I'm not seeing, tho

ericholscher: This makes sense to me. We should probably document all these variables (as you have done above) in the code/docs, so we don't get confused in the future.

PR #7594, opened 2020-10-23
Add APIv3 version edit URL

agjohnson:
For use in dashboard client code, having a link to the dashboard for
some resources would be helpful, and would avoid exposing the url
resolver through Django to the JS side.

Requires https://github.com/readthedocs/readthedocs.org/pull/7595

### DIFF
--- a/docs/api/v3.rst
+++ b/docs/api/v3.rst
@@ -464,6 +464,7 @@ Version detail
             },
             "urls": {
                 "documentation": "https://pip.pypa.io/en/stable/",
+                "edit": "https://readthedocs.org/dashboard/pip/version/stable/edit/",
                 "vcs": "https://github.com/pypa/pip/tree/19.0.2"
             },
             "_links": {
--- a/docs/api/v3.rst
+++ b/docs/api/v3.rst
@@ -463,8 +463,10 @@ Version detail
                 "epub": "https://pip.readthedocs.io/_/downloads/epub/pip/stable/"
             },
             "urls": {
+                "dashboard": {
+                    "edit": "https://readthedocs.org/dashboard/pip/version/stable/edit/"
+                },
                 "documentation": "https://pip.pypa.io/en/stable/",
-                "edit": "https://readthedocs.org/dashboard/pip/version/stable/edit/",
                 "vcs": "https://github.com/pypa/pip/tree/19.0.2"
             },
             "_links": {
--- a/readthedocs/api/v3/serializers.py
+++ b/readthedocs/api/v3/serializers.py
@@ -222,13 +222,8 @@ class VersionLinksSerializer(BaseLinksSerializer):
         return self._absolute_url(path)
 
 
-class VersionURLsSerializer(BaseLinksSerializer, serializers.Serializer):
-    documentation = serializers.SerializerMethodField()
+class VersionDashboardURLsSerializer(BaseLinksSerializer, serializers.Serializer):
     edit = serializers.SerializerMethodField()
-    vcs = serializers.URLField(source='vcs_url')
-
-    def get_documentation(self, obj):
-        return obj.project.get_docs_url(version_slug=obj.slug,)
 
     def get_edit(self, obj):
         path = reverse(
@@ -238,6 +233,15 @@ class VersionURLsSerializer(BaseLinksSerializer, serializers.Serializer):
                 'version_slug': obj.slug,
             })
         return self._absolute_url(path)
+
+
+class VersionURLsSerializer(BaseLinksSerializer, serializers.Serializer):
+    documentation = serializers.SerializerMethodField()
+    vcs = serializers.URLField(source='vcs_url')
+    dashboard = VersionDashboardURLsSerializer(source='*')
+
+    def get_documentation(self, obj):
+        return obj.project.get_docs_url(version_slug=obj.slug,)
 
 
 class VersionSerializer(FlexFieldsModelSerializer):
--- a/readthedocs/api/v3/serializers.py
+++ b/readthedocs/api/v3/serializers.py
@@ -222,12 +222,22 @@ class VersionLinksSerializer(BaseLinksSerializer):
         return self._absolute_url(path)
 
 
-class VersionURLsSerializer(serializers.Serializer):
+class VersionURLsSerializer(BaseLinksSerializer, serializers.Serializer):
     documentation = serializers.SerializerMethodField()
+    edit = serializers.SerializerMethodField()
     vcs = serializers.URLField(source='vcs_url')
 
     def get_documentation(self, obj):
         return obj.project.get_docs_url(version_slug=obj.slug,)
+
+    def get_edit(self, obj):
+        path = reverse(
+            'project_version_detail',
+            kwargs={
+                'project_slug': obj.project.slug,
+                'version_slug': obj.slug,
+            })
+        return self._absolute_url(path)
 
 
 class VersionSerializer(FlexFieldsModelSerializer):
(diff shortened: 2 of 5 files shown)


readthedocs/api/v3/serializers.py
```diff
@@ -222,13 +222,23 @@ def get_project(self, obj):
         return self._absolute_url(path)
 
 
-class VersionURLsSerializer(serializers.Serializer):
+class VersionURLsSerializer(BaseLinksSerializer, serializers.Serializer):
     documentation = serializers.SerializerMethodField()
+    edit = serializers.SerializerMethodField()
```
humitos, line 227: Shouldn't this be called `detail` to keep the same naming than the view?

humitos (approved):
Code changes look good! I suppose we need to add the new field in the version list JSON response as well to make test to pass.

readthedocs/api/v3/serializers.py
agjohnson: Good point. It's not a detail view though, it's a form view. I think the view is probably just poorly named.

agjohnson: Should we use `update` instead perhaps? `edit` is still probably a bit clearer than `update`, but i don't have strong opinions

humitos: I think we have list (eg. `/objects/`, or https://readthedocs.org/projects/docs/versions/) and detail views (eg. `/objects/<pk>/`, https://readthedocs.org/dashboard/docs/version/latest/) --which is how Django REST Framework calls them (the action is communicated by the HTTP verb). I don't think we should have `edit` or `update` views.

For example, https://readthedocs.org/dashboard/docs/edit/, is not correct to me and it should be something like https://readthedocs.org/dashboard/docs/admin/, where all the other views to edit/update something else related to the project are nested under `/admin/`. Then, https://readthedocs.org/dashboard/docs/domains/ should be https://readthedocs.org/dashboard/docs/admin/domains/

So, I wouldn't try to replicate the `/edit` postfix pattern here as well because I don't think it's correct. IMHO, in our case for edit version, it should be called `detail`. So, under the version's detail page, you can see all the details of the versions and perform some actions as well (eg. edit) which will be translated in the proper HTTP verb.

agjohnson: I'd probably disagree on edit vs detail distinction here. We use this as an edit view in both our current theme and it is strictly an edit view in the new theme. That is, there is no view that users go to see the details of a version, admin users go to this view to edit the version. We really only have maybe 2 detail views -- build detail and project detail. Ultimately, it will be the edit view URL that I need, to point users from the version "create" view to the update view of the selected version.

I'm not sure we want to move version edit back under the admin dashboard again. Either way, we use the same pattern under admin dashboard pages -- `/edit/` postfix on URLS -- for edit views on model objects. This was copying that pattern.

So for this use case `detail = serializers.SerializerMethodField()` doesn't seem correct

humitos: I prefer to follow the conventions that are established already in Django and DRF. As I mentioned before, there are `list` and `detail` URLs (which maps to the template suffix file: `_detail.html` and `_list.html`); then the actions (e.g. create, update) are HTTP verbs --which can be done from from those templates. Usually, create/POST is done from the `_list.html` and update/PATCH is done from `_detail.html` one.

Considering this is why prefer to keep calling `list` and `detail` URLs (without adding the `/edit/` suffix to URLs as done in #7595).

Depending on each object in particular, the `detail` URL may allow you to edit the object or not. The same happens with the `list` URL, where you may not be allowed to create new objects from that listing.

agjohnson: Roger, I think the confusion here might be between DRF patterns and standard Django patterns. It's a bit different in django vanilla views and django based views.

A `list` view is used via `GET` and create/POST is not performed from this view. `create` is the form that creates/POST the object (not applicable to this model/view), `detail` is a view that only *displays* information about the model instance, and `edit` is a view that *displays editable* information about the model instance.

The current/future version view is not a detail view -- it's a form view and that is really only useful for admin users. Our other URLs on the project admin dashboard all use `../edit/` in the URL. I think this is correct and this URL change follows that pattern.

agjohnson: Trying to make sense of outstanding PRs that I'm blocked on, I realized that the underlying for `/edit/` is not clear. I added more context here: https://github.com/readthedocs/readthedocs.org/pull/7595#issuecomment-812731544

It does seem the conflict here is DRF naming conventions vs Django URL patterns. I have matched what we use in our dashboard in other views, so feel safe with this naming convention. The underlying change is because a URL without an additional verb or separator conflicts with the URL for ProjectVersionCreate.

agjohnson:
I'd like to push this PR forward if possible. Merging this would help reduce the volume of the rat's nest of code I've created to make templates work.

docs/api/v3.rst
```diff
@@ -464,6 +464,7 @@ Version detail
             },
             "urls": {
                 "documentation": "https://pip.pypa.io/en/stable/",
+                "edit": "https://readthedocs.org/dashboard/pip/version/stable/edit/",
```
humitos, line 467: Maybe worth to have a different field, `dashboard`, for all the URLs under our own platform instead of mixing them with other non-related URLs that go to external sites?

* `urls.dashboard.edit`
* `dashboard.edit`

agjohnson: Are we expecting that many URLs in `urls` that we need the disambiguation or further nesting? It doesn't seem like API returns are going to have more the 4 or 5 urls at a time -- documentation links for some, perhaps CRUD view URLs in addition? I always found it worked well as a flat map. The implementation seems easy to understand and easy to use, and it also matches other providers' APIs.

agjohnson: Coming back to this discussion, do we have any other API returns that we're sending back dashboard links? This could be a pattern we see more as more of our expensive interactions get pushed through JS, but this is hard for me to estimate right now.

If `urls` is a common field in API return, and it is mostly used for external site urls, then perhaps that is a good reason to nest dashboard links. I'm feeling maybe -0 on the change though, but I also don't have a good grasp on how we're using external links through the rest of the API currently. This feels like a pattern we wouldn't be using a lot, but I'm not certain of this.

agjohnson: Also, if we want to really expand usage of dashboard links in our URL, that could maybe push us towards a dedicated structure. For example, this is a single case that we need for the dashboard, but maybe it's helpful to include any dashboard link we can in the API response. I think GitHub or Stripe does something similar, but they perhaps also don't have a tertiary data structure for external site URLS on top of API URLs and dashboard URLs.

humitos: I always prefer nested/dictionary fields in those cases where we don't _exactly_ know how our API is going to be used and when we know that it's not _100% finished_ and we may accept/add more fields. Using dictionaries allow us to add these "future" fields without breaking the structure and keeping the semantic of the field name consistent over time. That's why I was suggesting to split dashboard and external URLs.

This example sounds good to me: https://readthedocs.org/api/v3/projects/docs/versions/stable/

```js
    "urls": {
        "dashboard": {
            "edit": "https://readthedocs.org/dashboard/pip/version/stable/edit/"
        },
        "documentation": "https://docs.readthedocs.io/en/stable/",
        "vcs": "https://github.com/readthedocs/readthedocs.org/tree/5.14.3/"
    },
```

Maybe we can't think about more dashboard URLs for the `Version` object, but we may have more for `Project` (e.g. admin, edit, versions, etc) and I think it makes sense to have the same structure for all the resources.

agjohnson:
Updated with the feedback on API return data structure. I think I did it right :shrug:

humitos (approved):
Like a champ! 🏅

Issue #6238, opened 2019-10-02
Move inactivate version list to standard CRUD UI pattern


agjohnson:
We are addressing a bug with our version display with #6224. This is an intermediate fix for a bug in the display, the final version to move towards would move us to a standardized pattern for our CRUD interfaces and probably isn't something we'd take on until we are using SUI and have a strong UI base to work from.

We have UI standardization notes about moving towards common UI patterns, so I won't reiterate here. The point of this next step for version admin would be to remove the pattern we're using in the UI here in favor of a UI/UX that is predictable and expected from the user's perspective.

In this case, standardization would involve:

* Add a button to the top of the list view
* This button leads to the create view
* The create view, which is the same form and view as the update view, is where a user selects the version from a filter search list
* The user can also select project privacy while editing the version
* The list view drops the second list view for inactive versions

### Why?

* The user is coming to the page with one of maybe two motives: they want to activate a version or they want to update/deactivate a version that is active (or they simply want to see the list)
  * If the user is looking to see or edit versions, they are interested in a list view of active versions, it will be the first thing they see as the list view is prominent
  * If the user is looking to activate a new version, the button, also at the top of the page, will be the first thing they see. They don't need to scroll to the bottom of the page (or worst case don't see the listing of inactive versions at all because they have so many active versions)
* Filter view of the inactive versions is moved to the create view, and the tabular listing that we're using now for inactive versions is removed in favor of a simple search list, because:
  * It greatly reduces the amount of information that the user needs to process to find the version they are looking for -- no scanning redundant buttons, commit id, or version slugs. The user just searches for the branch name they are used to (also similar to the branch/tag search in GitHub/GitLab)
  * The bug of 5xx/timeout on list view display is caused by displaying a tabular view from our templates. Moving to a JS/API driven implementation separates the page load and data load operations. We might load all version from an API endpoint, or might figure out how to do a filtered API query as the user updates the search field, reducing the number of versions the API would have to process.
* Standardization of UI and template code. The UI/UX is standardized for the user. We might end up with a reusable template pattern for development as well. This would mean greatly reducing the amount of redundant template code that we're maintaining.

### Examples

Mockups, using live loaded SUI, so the CSS is all wonky. You get the idea.

![version-list-view](https://user-images.githubusercontent.com/1140183/66065795-70b4b900-e505-11e9-92bf-7fe8969372b8.png)

[...]

PR #7595, opened 2020-10-23
Add separate version create view and create view URL

agjohnson:
This is not currently used outside the new theme, but enables the
solution for #6238 -- latest `master` of the new theme already expects
this code.

For now, the create URL does throw and exception, because we are
rendering template code inside the version form. I'm not going to
address this as the create view is not linked and unused, and we
probably shouldn't be rendering templates inside form instantiation
anyways.

The edit version URL does change with this, to append the `/edit/`
postfix. This was done to match similar URL patterns, and to give a path
for the create view URL that does not match the edit URL pattern.

### DIFF
--- a/readthedocs/api/v3/tests/responses/projects-builds-detail.json
+++ b/readthedocs/api/v3/tests/responses/projects-builds-detail.json
@@ -13,7 +13,7 @@
     "urls": {
         "build": "https://readthedocs.org/projects/project/builds/1/",
         "project": "https://readthedocs.org/projects/project/",
-        "version": "https://readthedocs.org/dashboard/project/version/v1.0/"
+        "version": "https://readthedocs.org/dashboard/project/version/v1.0/edit/"
     },
     "project": "project",
     "state": {
--- a/readthedocs/api/v3/tests/responses/projects-detail.json
+++ b/readthedocs/api/v3/tests/responses/projects-detail.json
@@ -29,7 +29,7 @@
                 "urls": {
                     "build": "https://readthedocs.org/projects/project/builds/1/",
                     "project": "https://readthedocs.org/projects/project/",
-                    "version": "https://readthedocs.org/dashboard/project/version/v1.0/"
+                    "version": "https://readthedocs.org/dashboard/project/version/v1.0/edit/"
                 },
                 "project": "project",
                 "state": {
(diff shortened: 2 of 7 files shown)


ericholscher (approved):
Seems simple enough 👍

agjohnson:
I seem to recall @humitos having feedback on the URL structure, I'll wait for his feedback here.

agjohnson:
Found the conversation, it was here: https://github.com/readthedocs/readthedocs.org/pull/7594#discussion_r512148924

agjohnson:
The additional URL for the create view was causing errors with tests, so i did the more correct thing and made the URL conditional. Test look good now :+1:

humitos:
I don't want to block this work, but I made a comment with maybe a better explanation about why I think the `/edit/` on the URL shouldn't be used at https://github.com/readthedocs/readthedocs.org/pull/7594#discussion_r528623867

It's not wrong using it, but I just find it easier to follow the conventions that Django and DRF are already using for these URLs.

agjohnson:
This is blocking continuing to work on templates so I'd like to get this merged. To resurrect this discussion, some added context: the reason for the `/edit/` postfix is because we need a corresponding URL for the create version view. Following patterns for other create/form views, there is a project version create URL:

`/dashboard/pip/version/create/`

However, if we don't change the edit form URL to `../edit/` postfix, the above URL for ProjectVersionCreate collides with the URL for ProjectVersionDetail where `version=create`. Adding an explicit `../edit/` postfix removes this overlap.

Using the `../edit/` postfix for update views is also an existing pattern we use, so nothing really new here.

If we want this to be even clearer, ProjectVersionDetail should really be ProjectVersionUpdate. This view is not a detail view at all, it is an admin only form view. I won't take this chunk on just yet, but can if it seems important to getting this merged.

ericholscher:
I don't think the URL's matter nearly as much for the dashboard as the API, so I'm fine with whatever is most obvious, and hopefully we can standardize on it.

humitos (approved):
I re-read the whole conversation and I think Anthony is right :smile: 

As an example of this, Django Admin uses:

* `/admin/auth/user/` to list all the users
* `/admin/auth/user/add/` to create a new user
* `/admin/auth/user/<pk>/change/` to edit a user

I understand that we are going to follow the same pattern but using `/create/` and `/edit/` instead of `/add/` and `/change/`, so :+1:

Issue #7583, opened 2020-10-20
Builds fail trying to `create_container` and/or `remove_container`


minrk:
## Details:

* Project URL: https://readthedocs.org/projects/jupyterlab/
* Build URL(if applicable): https://readthedocs.org/projects/jupyterlab/builds/12139704/
* Read the Docs username(if applicable): minrk

## Expected Result

Build succeeds, docs are published

## Actual Result

Build succeeds, docs are not published. Or an error somewhere is not reported. The build status is

> Error
> There was a problem with Read the Docs while building your documentation. Please try again later.

but no errors are reported in the logs.

The build log ends with:

```
The HTML pages are in _build/html.
Updating searchtools for Read the Docs search... /home/docs/checkouts/readthedocs.org/user_builds/jupyterlab/checkouts/latest/docs/source/conf.py:281: RemovedInSphinx40Warning: The app.add_stylesheet() is deprecated. Please use app.add_css_file() instead.
  app.add_stylesheet('custom.css')  # may also be an URL
Command time: 739s Return: 0
```

This started failing after an update to how the project builds docs:

1. typedoc-built HTML files are built and staged to the output directory (adds an expensive npm build that could create memory pressure, but succeeds before sphinx builds continue)
2. conda is used for installation, which included bumping the nodejs runtime to 14.x on $PATH and sphinx from 1.8 to 3.2.1 from the last successful build (12139664)

Daltz333:
Use ``add_css_file`` instead. https://github.com/jupyterlab/jupyterlab/blob/master/docs/source/conf.py

Daltz333:
But that's just a deprecation, not sure why it's failing on it.

humitos:
It seems it was a temporal error. It failed when trying to cleanup the docker container used to build the documentation.

I just trigger a new build or `latest`: https://readthedocs.org/projects/jupyterlab/builds/12146392/ (this may fail because it's using a smaller builder)

Also, it could be that you are hitting the max time allowed to build (your build reports 1311 seconds). So, increasing this time may help here as well. Let's see.

@minrk I think I know what happened. Where you using `pip` to create the virtualenv and now you are using `conda`? I guess it was the case. I'm moving your project a bigger builder so it doesn't time out or fail because of memory usage.

I triggered another build https://readthedocs.org/projects/jupyterlab/builds/12146492/ (this should success because it's running in a bigger server).

Please, let me know if the following builds are building OK.

humitos:
[...]

PR #7618, opened 2020-10-29
Reserve 1Gb for Application Memory

humitos:
We have noticed some strange behavior when trying to `create_container` and
`remove_container` using the Docker API. For some reason we started getting
ReadTimeout and I'm supposing that our Docker daemon has been restarted (because
of OOM) or in a dumb state.

This commit increases from 750 to 1000 Mb of RAM memory reserved for our
application and OS processes.

The user will get 7000Mb instead of 7200Mb in build-large instances and
in build-default instances they will get 2400 intead of 2700Mb.

Reference: https://github.com/readthedocs/readthedocs.org/issues/7583#issuecomment-718702683

### DIFF
--- a/readthedocs/settings/base.py
+++ b/readthedocs/settings/base.py
@@ -495,7 +495,7 @@ class CommunityBaseSettings(Settings):
                 "free -m | awk '/^Mem:/{print $2}'",
                 shell=True,
             ))
-            return total_memory, round(total_memory - 750, -2)
+            return total_memory, round(total_memory - 1000, -2)
         except ValueError:
             # On systems without a `free` command it will return a string to
             # int and raise a ValueError


readthedocs/settings/base.py
```diff
@@ -495,7 +495,7 @@ def _get_docker_memory_limit(self):
                 "free -m | awk '/^Mem:/{print $2}'",
                 shell=True,
             ))
-            return total_memory, round(total_memory - 750, -2)
+            return total_memory, round(total_memory - 1000, -2)
```
ericholscher, line 498: This should probably be a constant, or at least have a comment explaining the logic.

ericholscher (approved):
Seems reasonable 👍

readthedocs/settings/base.py
humitos: Agree. I'm merging it for now so it goes out tomorrow, tho.

PR #7620, opened 2020-10-29
Some thoughts on what versioned images look like

agjohnson:
Opening this up just for discussion, it was easier to edit here.

Based on #7566

### DIFF
--- a/docs/development/design/build-images.rst
+++ b/docs/development/design/build-images.rst
@@ -51,75 +51,92 @@ Goals
 * reduce size on builder VM disks by sharing Docker image layers
 * deprecate ``stable``, ``latest`` and ``testing``
 * allow use custom images for particular users/customers by sharing most layers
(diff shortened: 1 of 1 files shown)


docs/development/design/build-images.rst
```diff
@@ -51,88 +51,101 @@ Goals
...
-Custom images
--------------
+``readthedocs/build:conda12``
+    Installs 1.2 release of Conda
+
+``readthedocs/build:node12``
+    Installs Node 12.x release
+
+``readthedocs/build:python+node``
+    Installs latest Python and latest Node
```
agjohnson, line 105: This is where things get screwy. I'm not a huge fan, but this seems better than arbitrarily selecting a version of node or treating all languages that != python as second class. Python is the big one for us though of course.

humitos:
Closing as this is already implemented. We are versioning the OS and installing tools versions on the fly.

PR #7623, opened 2020-10-30
Position the ad at the top of our docs

davidfischer:
- Always use a text-only ad at the top of our docs
- Allocate space for the ad

## Screenshot

Screen Shot 2020-10-29 at 7 19 41 PM

### DIFF
--- a/docs/_templates/ethicalads.html
+++ b/docs/_templates/ethicalads.html
@@ -1,4 +1,3 @@
 <div class="rtd-sponsorship">
-  <div class="raised" data-ea-publisher="readthedocs" data-ea-type="text"></div>
+  <div class="raised" data-ea-publisher="readthedocs" data-ea-type="text" data-ea-manual="true"></div>
 </div>
-
--- a/docs/_templates/ethicalads.html
+++ b/docs/_templates/ethicalads.html
@@ -1,0 +1,4 @@
+<div class="rtd-sponsorship">
+  <div class="raised" data-ea-publisher="readthedocs" data-ea-type="text"></div>
+</div>
+
--- a/docs/_templates/ethicalads.html
+++ b/docs/_templates/ethicalads.html
@@ -1,3 +1,3 @@
 <div class="rtd-sponsorship">
-  <div class="raised" data-ea-publisher="readthedocs" data-ea-type="text" data-ea-manual="true"></div>
+  <div class="raised" data-ea-publisher="readthedocs" data-ea-type="text"></div>
 </div>
--- a/docs/advertising/ad-customization.rst
+++ b/docs/advertising/ad-customization.rst
@@ -28,7 +28,11 @@ The ad will be inserted into this container wherever this element is in the docu
 
 .. code-block:: html
 
-    <div id="ethical-ad-placement"></div>
+    <div data-ea-publisher="readthedocs"></div>
+
+For additional options, see the `EthicalAds client documentation`_.
+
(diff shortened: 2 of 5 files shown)


davidfischer:
You can also see the ad placement in the [preview](https://docs--7623.org.readthedocs.build/en/7623/)!

docs/advertising/ad-customization.rst
```diff
@@ -28,7 +28,11 @@ The ad will be inserted into this container wherever this element is in the docu
 
 .. code-block:: html
 
-    <div id="ethical-ad-placement"></div>
+    <div data-ea-publisher="readthedocs"></div>
```
ericholscher, line 31: I wonder what happens if users set this to something that isn't `readthedocs` -- it'll just fail and not show an ad right? :/

docs/conf.py
```diff
@@ -139,7 +140,3 @@ def get_version():
     # This page is under login
     r'https://readthedocs\.org/accounts/gold',
 ]
-
-
-def setup(app):
-    app.add_css_file('css/sphinx_prompt_css.css')
```
ericholscher, line 145: Is this automatically included with the extension now?

docs/_templates/ethicalads.html
```diff
@@ -0,0 +1,4 @@
+<div class="rtd-sponsorship">
+  <div class="raised" data-ea-publisher="readthedocs" data-ea-type="text"></div>
```
ericholscher, line 2: I wonder if we should have some content in here for when ads don't load/show (eg. for Gold users?) so they aren't confused by the space.

ericholscher (approved):
Looks great!

docs/_templates/ethicalads.html
davidfischer: This is actually trickier than I thought. The goal of pre-allocating the space is so we don't get the page re-flowing after the fact. However, there isn't currently a way to hook into the response from the sustainability data API.

docs/advertising/ad-customization.rst
davidfischer: Or the ads JS should re-set the values (publisher, keywords, etc.) on this element based on the response from the sustainability data API.

davidfischer:
I think I'm going to fix this up in two stages. The first stage is done and just involved setting the placement to manual. This should make this PR good to ship.

Step 2 will be a bit more involved and involve changes to the RTD embed JS. If the embed JS sees an existing ethical ads placement (`[data-ea-publisher]`), it should set the values on that DOM element (manual, publisher, keywords, etc.) to make sure they are correct and get the up-to-date values.

Step 3, which is optional, would allow hooking into more parts of the workflow. We've wanted this in a few places -- for search, for the flyout menu, and other reasons. This would allow displaying a message to gold users like "Thank you for supporting Read the Docs in the pre-allocated space".

Thoughts?

ericholscher:
Sounds like a good plan. I'd call #3 a stretch goal, but definitely something worthwhile on the RTD side 👍

davidfischer:
Now that Step 2 ([above](https://github.com/readthedocs/readthedocs.org/pull/7623#issuecomment-719730738)) is done and out, I updated this PR to use the ad client standard way of positioning the ad.

I think this PR is ready for another look.

ericholscher:
I think this still needs a bit more tuning -- the page still shifts when an ad has 2 lines, and for gold users without ads, the page has a weird blank space at the top. I'm not sure there is anything we can do about the 2nd one, but we should probably fine a way to at least fix the page moving.

PR #7669, opened 2020-11-16
Set ad configuration values if using explicit placement

davidfischer:
This only applies if the project is setting an explicit placement. This sets the values on the ad element sanely.

Ref: https://github.com/readthedocs/readthedocs.org/pull/7623#issuecomment-719730738


readthedocs/core/static-src/core/js/doc-embed/sponsorship.js
```diff
@@ -32,6 +32,11 @@ function create_ad_placement() {
     var offset;
 
     if ($(EXPLICIT_PLACEMENT_SELECTOR).length > 0) {
+        $(EXPLICIT_PLACEMENT_SELECTOR).attr("data-ea-publisher", "readthedocs");
```
ericholscher, line 35: Do we not need this to respect the value coming back from the sustainability API?

ericholscher (approved):

readthedocs/core/static-src/core/js/doc-embed/sponsorship.js
davidfischer: That will set it later.

ericholscher: Ah, reading through the code, sounds like that's set dynamically after the call returns.

ericholscher: Cool, should get this merged for deploy 👍

PR #7670, opened 2020-11-16
Fix for out of order script loading

davidfischer:
There are two issues here:

- The first issue is that a Sphinx theme sets defer or async when loading jQuery. This can cause jQuery to not be available before readthedocs-doc-embed.js sets its own `$(document).ready` event listener.
- The second issue is caused by the sustainability API loading faster than the ad client. This should be rare since the ad client is loaded via a CDN but it can happen especially on slow networks.

I was able to reproduce these issues by setting throttling in the browser developer tools.


ericholscher (approved):
Looks simple enough 👍

PR #7672, opened 2020-11-17
Remove Moved Fields from RemoteRepository Model After Data Migration

saadmk11:
This should be the last part of the normalization process, as I already have the code, so making a PR for this. This can be reviewed and merged at the last.

***This also removes the ``active`` field from the ``RemoteRepository`` model which is not used.


stsewd (commented):
We should make these fields nullable before deleting them, so we avoid downtime.

saadmk11:
This is not required anymore as we are creating a new table for `RemoteRepository`. See https://github.com/readthedocs/readthedocs.org/pull/7777 :)

PR #7675, opened 2020-11-18
Update OAuth App to Use The RemoteRepositoryRelation Model

saadmk11:
This does not take into account for duplicate RemoteRepository for now.


readthedocs/oauth/utils.py
```diff
@@ -31,9 +31,17 @@ def update_webhook(project, integration, request=None):
...
+        remote_relations = project.remote_repository.remote_relations.filter(
+            account__isnull=False
+        )
+
+        for relation in remote_relations:
+            service = service_cls(request.user, relation.account)
+            updated, __ = service.update_webhook(project, integration)
+
+            if updated:
+                break
```
saadmk11, line 43: I'm not entirely sure if we want to loop through all the remote relations and try to use the accounts for each user to activate the webhook, I think this will make sure webhooks are updated even if the user who imported the repo doesn't have a valid account. would like to know your thoughts?

readthedocs/oauth/services/github.py
```diff
@@ -97,21 +100,19 @@ def create_repository(self, fields, privacy=None, organization=None):
...
-                )
-                repo.users.add(self.user)
+
+            repo, _ = RemoteRepository.objects.get_or_create(
+                full_name=fields['full_name']
+            )
+            remote_relation, _ = RemoteRepositoryRelation.objects.get_or_create(
+                remoterepository=repo,
+                user=self.user,
+                account=self.account
```
saadmk11, line 113: This might fail if the user has multiple connected account and more than one of them has access to the same repo, we current have unique_together(``remoterepository``, ``user``) so this will raise ``IntegrityErrot`` while trying to create a relation with the same user and remore_repo but with different account,

So, either we need to allow only one connected account for a user to have access to the same repo or have different logic around the uniqueness.

Though this is an edge case

readthedocs/oauth/utils.py
humitos: Yeah. I think it's weird that if the user that imports the project doesn't have access to setup the webhook, it's done using a different user. That sounds bad, IMO.

I suppose we need to use the current logged in user for this action, and if it fails, that user should contact an admin of that repository to "re-sync" the webhook. I'd say that we should filter the previous query by using `user=request.user`.

readthedocs/rtd_tests/tests/test_oauth.py
```diff
@@ -147,7 +146,7 @@ def test_multiple_users_same_repo(self):
         )
         self.assertIsInstance(github_project, RemoteRepository)
         self.assertIsInstance(github_project_2, RemoteRepository)
-        self.assertNotEqual(github_project_2, github_project)
+        self.assertEqual(github_project_2, github_project)
```
humitos, line 149: :100:

humitos (commented):
Changes look GOOD! I left some comments to keep talking, but there isn't anything too big to change.

> This does not take into account for duplicate RemoteRepository for now.

Is this comment old now? The code seems to handle duplicated RemoteRepository properly now.

readthedocs/api/v2/views/model_views.py
```diff
@@ -397,10 +397,10 @@ def get_queryset(self):
             )
 
         query = query.filter(
-            account__provider__in=[
+            remote_relations__account__provider__in=[
```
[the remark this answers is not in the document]
saadmk11: Sure! I'll make it ``remote_repository_relations``

readthedocs/oauth/services/base.py
```diff
@@ -206,7 +206,7 @@ def sync(self):
             .exclude(
                 Q(full_name__in=repository_full_names) | Q(project__isnull=False)
             )
-            .filter(account=self.account)
+            .filter(remote_relations__account=self.account)
```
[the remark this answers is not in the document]
saadmk11: Oh, my bad, Thanks for pointing it out :+1:

readthedocs/oauth/services/bitbucket.py
```diff
@@ -155,15 +165,18 @@ def create_repository(self, fields, privacy=None, organization=None):
 
             repo.html_url = fields['links']['html']['href']
             repo.vcs = fields['scm']
-            repo.account = self.account
```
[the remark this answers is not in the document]
saadmk11: Actually, I think we talked about adding the ``remote_id`` fields later, but now that I think of it it would be better to do it now, but I'll add it in another PR after this is merged.

readthedocs/oauth/utils.py
```diff
@@ -31,9 +31,17 @@ def update_webhook(project, integration, request=None):
 
     updated = False
     try:
-        account = project.remote_repository.account
-        service = service_cls(request.user, account)
-        updated, __ = service.update_webhook(project, integration)
+        remote_relations = project.remote_repository.remote_relations.filter(
+            account__isnull=False
```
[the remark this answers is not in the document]
saadmk11: If the user disconnects the Oauth account from their RTD profile, I haven't tested it out yet. I'll test it out and confirm it :)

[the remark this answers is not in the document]
saadmk11: @humitos Confirmed! if the user disconnects a social account then the ``account`` field will be ``NULL``, Let me know if you want to handle this in a different way :)

saadmk11:
> > This does not take into account for duplicate RemoteRepository for now.
> 
> Is this comment old now? The code seems to handle duplicated RemoteRepository properly now.

Unfortunately, it does not handle it. We can only filter ``RemoteRepository`` with the full name so that there is no duplication in the future, but the old/current duplication is still there so this will return multiple objects on ``get_or_create()`` which will result in an error

```python
    full_name=fields['full_name']
)
```

We need to figure out a way to remove the current DB duplicates

readthedocs/oauth/services/github.py
saadmk11: @humitos  Just tested it, Users can add multiple GitHub accounts to their profile and if both of them have access to the same GitHub Repository then it will cause the error.

This is the error user will get:

``There was an error with your request: UNIQUE constraint failed: oauth_remoterepositoryrelation.remoterepository_id, oauth_remoterepositoryrelation.user_id``

Let me know your thought on this :)

readthedocs/oauth/services/bitbucket.py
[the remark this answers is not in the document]
humitos: Yeah. I said that in our conversation. However, I think it's better to have it _before_ we do the initial resync so we already populate this data from the beginning and then we can start making use of it. I'm fine if it's in another PR.

readthedocs/oauth/utils.py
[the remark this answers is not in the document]
humitos: If the user disconnects the social account, we should cascade and delete all the `RemoteRelations` for that social account, right?

humitos:
> We can only filter `RemoteRepository` with the full name so that there is no duplication in the future, but the old/current duplication is still there so this will return multiple objects on `get_or_create()` which will result in an error

I understand. We may not need to care about this anymore, if the what I'm going to describe next makes sense :sweat_smile: 

Talking to Eric last week, we thought that we can deploy all of these changes in a slightly different way than we originally described in https://github.com/readthedocs/readthedocs.org/pull/7536#issuecomment-724102640:

* between point 4 and 5 we could spin up a new instance with the new Python code and modelling (using `RemoteRepository.Meta.db_table = "oauth_remoterepositoy_202011"` so it creates a new table and doesn't touch the current one) and re-sync from VCS providers into that table. This new instance won't attend users requests, it will be used only for fetching data from VCS providers.
* after some days, once the re-sync is finished and we have all the data in the new table, we can dump the `Project <-> RemoteRepository` relations and load into the new table `oauth_remoterepository202011`. * finally, we do a full deploy for all our production instances with the new Python code and modelling

readthedocs/oauth/services/base.py
```diff
@@ -202,11 +202,12 @@ def sync(self):
         all_remote_repositories = remote_repositories + remote_repositories_organizations
         repository_full_names = [r.full_name for r in all_remote_repositories if r is not None]
         (
-            self.user.oauth_repositories
+            self.user.remote_repository_relations
             .exclude(
-                Q(full_name__in=repository_full_names) | Q(project__isnull=False)
+                Q(remoterepository__full_name__in=repository_full_names) |
+                Q(remoterepository__project__isnull=False)
```
humitos, line 208: This restriction is no needed anymore.

We were doing this to avoid disconnecting Projects from RemoteRepository. However, now that we are deleting RemoteRepositoryRelation we can delete the relation without care too much about if the Project is connected to a RemoteRepository or not

readthedocs/oauth/services/github.py
humitos: Hrm... So, I see some options here:

- remove the unique constraint here
- change it to be `unique(remoterepository_id, socialaccount_id)` instead

The second one seems more accurate to me since we don't really need to filter by user since the social account is already connected to a user.

readthedocs/oauth/services/base.py
saadmk11: Okay, Got it! :+1:

readthedocs/oauth/utils.py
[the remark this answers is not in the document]
saadmk11: Good point! :100: I agree with you, as we do not do it currently, I think this new functionality can be added in a different PR

readthedocs/oauth/services/github.py
saadmk11: Yes, I have updated the PR with the second option, now the error I mentioned above has gone away :)

saadmk11:
> I understand. We may not need to care about this anymore, if the what I'm going to describe next makes sense 
> 
> Talking to Eric last week, we thought that we can deploy all of these changes in a slightly different way than we originally described in [#7536 (comment)](https://github.com/readthedocs/readthedocs.org/pull/7536#issuecomment-724102640):
> 
> * between point 4 and 5 we could spin up a new instance with the new Python code and modelling (using `RemoteRepository.Meta.db_table = "oauth_remoterepositoy_202011"` so it creates a new table and doesn't touch the current one) and re-sync from VCS providers into that table. This new instance won't attend users requests, it will be used only for fetching data from VCS providers.
> * after some days, once the re-sync is finished and we have all the data in the new table, we can dump the `Project <-> RemoteRepository` relations and load into the new table `oauth_remoterepository202011`. * finally, we do a full deploy for all our production instances with the new Python code and modelling

Sounds good :100:

humitos (approved):
I re-review this PR and everything looks fine to me. Thanks for your help on this! We can now move forward with the similar changes required for `RemoteOrganization`

PR #7716, opened 2020-12-01
Ensure there is space for sidebar ads

davidfischer:
Already we check whether a sidebar ad would be visible before deciding whether to put an ad in the sidebar or footer. However, in some cases, only a small bit of the ad is visible. This ensures that there is enough space for an ad to be displayed in the sidebar.


ericholscher (approved):
Simple enough 👍

Issue #7667, opened 2020-11-16
Add `remote_id` Field to RemoteRelation Model


saadmk11:
We should probably add the `remote_id` field here as well and populate it with the `id` that it's in `remoterepository.json`.

We will need it to remove the usage of `full_name` in many places.

ref: https://github.com/readthedocs/readthedocs.org/pull/7536#discussion_r501636916

PR #7722, opened 2020-12-02
Add remote_id and vcs_provider field to RemoteRepository model and Remove Data Migration

saadmk11:
closes: https://github.com/readthedocs/readthedocs.org/issues/7667


readthedocs/oauth/constants.py
```diff
@@ -0,0 +1,9 @@
+GITHUB_BRAND = 'GitHub'
+GITLAB_BRAND = 'GitLab'
+BITBUCKET_BRAND = 'Bitbucket'
+
+VCS_PROVIDER_CHOICES = (
+    (GITHUB_BRAND, 'GitHub'),
```
humitos, line 6: Shouldn't this be `(slug, display_name)`, like `('github', 'GitHub')` instead?

readthedocs/oauth/models.py
```diff
@@ -168,8 +169,17 @@ class RemoteRepository(models.Model):
         null=True,
         blank=True,
     )
-
     json = models.TextField(_('Serialized API response'))
+    # TODO: Make remote_id and vcs_provider not nullable and
+    #  unique together after migration
+    remote_id = models.CharField(max_length=128, blank=True, null=True)
```
humitos, line 175: I think we should index this field, since it's the one that we are going to use most. We indexed `full_name` for this reason.

readthedocs/oauth/models.py
```diff
@@ -168,8 +169,17 @@ class RemoteRepository(models.Model):
         null=True,
         blank=True,
     )
-
     json = models.TextField(_('Serialized API response'))
+    # TODO: Make remote_id and vcs_provider not nullable and
+    #  unique together after migration
```
humitos, line 174: Would be good to create an issue or a note card in the project to not forget about this.

humitos (commented):
I think we should index `remote_id`. I'm not sure about `vcs_provider`, but I don't think it's super important.

Also, I'm not sure if you just want to add these fields in this PR, but they are not used nor populated in the code. Are you thinking on doing that in another PR?

readthedocs/oauth/constants.py
saadmk11: Yes, you are right, but I have intensionally added it like this because this requires the constants to be in this camel case (``'GitHub'``)

https://github.com/readthedocs/readthedocs.org/blob/811890894f72c14be05ddd443a20adc017656c5c/readthedocs/projects/tasks.py#L1916

@humitos I would like to know your thoughts on this.

readthedocs/oauth/models.py
saadmk11: Done! :)

readthedocs/oauth/constants.py
humitos: Good catch! I _personally_ think `*_BRAND` should only be used when we need to show them to users and internally, we should use a slug.

saadmk11: Then we need to keep them separated, My original intention to merge the both was to not make it confusing in the future as they are almost the same. But I get your point. :)

readthedocs/oauth/services/bitbucket.py
```diff
@@ -147,6 +148,8 @@ def create_repository(self, fields, privacy=None, organization=None):
 
             repo.organization = organization
             repo.name = fields['name']
+            repo.remote_id = fields['uuid']
```
saadmk11, line 151: Bitbucket returns a weird id (includes curly braces). Don't know if it should be stripped or be this way.

``"uuid": "{39c6eaba-9b71-47f1-a88c-249a656a60f8}"``

humitos: Weird... I wouldn't touch it and rely on their weirdness thing. I hope it's always returned in the same way :)

humitos (approved):
:tada: :100:

PR #7728, opened 2020-12-04
Update parts of code that were using the old RemoteRepository model fields

saadmk11:
This PR will change references to the old modeling

### DIFF
--- a/readthedocs/oauth/services/github.py
+++ b/readthedocs/oauth/services/github.py
@@ -139,9 +139,7 @@ class GitHubService(Service):
             repo.save()
 
             remote_repository_relation.json = fields
-            remote_repository_relation.admin = fields.get(
-                'permissions', {}
-            ).get('admin', False)
+            remote_repository_relation.admin = fields.get('permissions', {}).get('admin', False)
             remote_repository_relation.save()
 
             return repo
--- a/readthedocs/oauth/services/github.py
+++ b/readthedocs/oauth/services/github.py
@@ -108,10 +108,12 @@ class GitHubService(Service):
             repo, _ = RemoteRepository.objects.get_or_create(
                 full_name=fields['full_name']
             )
-            remote_relation, _ = RemoteRepositoryRelation.objects.get_or_create(
-                remoterepository=repo,
-                user=self.user,
-                account=self.account
+            remote_repository_relation, _ = (
+                RemoteRepositoryRelation.objects.get_or_create(
+                    remoterepository=repo,
+                    user=self.user,
+                    account=self.account
+                )
             )
 
             if repo.organization and repo.organization != organization:
@@ -143,9 +145,11 @@ class GitHubService(Service):
 
             repo.save()
 
-            remote_relation.json = fields
-            remote_relation.admin = fields.get('permissions', {}).get('admin', False)
-            remote_relation.save()
+            remote_repository_relation.json = fields
+            remote_repository_relation.admin = fields.get(
+                'permissions', {}
+            ).get('admin', False)
+            remote_repository_relation.save()
 
             return repo
 
--- a/readthedocs/oauth/services/github.py
+++ b/readthedocs/oauth/services/github.py
@@ -107,7 +107,7 @@ class GitHubService(Service):
             repo, _ = RemoteRepository.objects.get_or_create(
                 full_name=fields['full_name']
             )
-            remote_repository_relation = self.get_remote_relation(repo)
+            remote_repository_relation = self.get_remote_repository_relation(repo)
 
             if repo.organization and repo.organization != organization:
                 log.debug(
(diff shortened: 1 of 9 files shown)


saadmk11:
Not sure about what to do here, we can try to use all the connected accounts to ``RemoteRepository`` through ``RemoteRepositoryRelation`` to send the status. but I don't think it's a good way to do it.

@humitos Would like to hear your thoughts on this?

https://github.com/readthedocs/readthedocs.org/blob/0d59ef5700da5c838f30d09368ae9048d552d1e6/readthedocs/projects/tasks.py#L1921-L1925

readthedocs/oauth/services/github.py
```diff
@@ -143,9 +138,11 @@ def create_repository(self, fields, privacy=None, organization=None):
 
             repo.save()
 
-            remote_relation.json = fields
-            remote_relation.admin = fields.get('permissions', {}).get('admin', False)
-            remote_relation.save()
+            remote_repository_relation.json = fields
+            remote_repository_relation.admin = fields.get(
+                'permissions', {}
+            ).get('admin', False)
```
humitos, line 144: nit: I prefer the one line style for this

humitos (commented):
> Not sure about what to do here, we can try to use all the connected accounts to `RemoteRepository` through `RemoteRepositoryRelation` to send the status

We have touched this code already. The current behavior is:

1. try using user who imported the project's account
1. try using any of the users' maintainer accounts

We changed this because sometime the original author (1) removed their account or doesn't have permissions anymore to the original repository. So, it makes sense to try reporting the build status back using any of the social account from its maintainers. In my opinion, we should keep the same logic.

If I understand correctly, the query you proposed,

> ```python
> build.project.remote_repository.remote_repository_relations.filter(
>     account__isnull=False,
>     user__projects=build.project
> ).select_related('account', 'user').only('user', 'account')
> ```

does not restrict these social accounts to _only maintainers of the project_ (on Read the Docs). So, I think we should add that filter here and we should be fine. It seems it's missing something like `user__in=build.project.users.all()`, right?

saadmk11:
> We have touched this code already. The current behavior is:
> 
> 1. try using user who imported the project's account
> 2. try using any of the users' maintainer accounts
> 
> We changed this because sometime the original author (1) removed their account or doesn't have permissions anymore to the original repository. So, it makes sense to try reporting the build status back using any of the social account from its maintainers. In my opinion, we should keep the same logic.

The issue I see here is we do not have this kind of relationship anymore ``build.project.remote_repository.account``, now we have multiple social accounts per remote repo. So, I was thinking we can iterate over all the project maintainers (RTD) ``remote_relation`` object for the ``remote_repo`` and try all of them to send the build status. Does it make sense?  Maybe I'm not getting something here, please feel free to correct me. :)

> If I understand correctly, the query you proposed,
> 
> > ```python
> > build.project.remote_repository.remote_repository_relations.filter(
> >     account__isnull=False,
> >     user__projects=build.project
> > ).select_related('account', 'user').only('user', 'account')
> > ```
> 
> does not restrict these social accounts to _only maintainers of the project_ (on Read the Docs). So, I think we should add that filter here and we should be fine. It seems it's missing something like `user__in=build.project.users.all()`, right?

@humitos  Here I'm using ``user__projects=build.project`` instead of `user__in=build.project.users.all()` which results in one extra query (we can do subquery to work around it) but I prefered this ``user__projects=build.project`` which generates the same results (Filter remote relations with users that are maintainers to ``build.project``) with one less query.

humitos:
> The issue I see here is we do not have this kind of relationship anymore `build.project.remote_repository.account`, now we have multiple social accounts per remote repo. So, I was thinking we can iterate over all the project maintainers (RTD) `remote_relation` object for the `remote_repo` and try all of them to send the build status. Does it make sense?

@saadmk11 gotcha, I understand. You are right, and this makes sense to me. Basically we are not doing (1) but directly (2).

> @humitos Here I'm using `user__projects=build.project` instead of `user__in=build.project.users.all()` which results in one extra query (we can do subquery to work around it) but I prefered this `user__projects=build.project` which generates the same results (Filter remote relations with users that are maintainers to `build.project`) with one less query.

I think this is correct. However, I just realized that this is not going to work in commercial because _user is not related to projects directly_, they relate to them via the `Team` objects instead. In fact, this is currently working in commercial _only if the user who imported the project has permissions_ (using `build.project.remote_repository.users.first()`). So, if we loose that information, we need to find a workaround for commercial. We have some helpers for these cases, like `readthedocs.core.permissions.AdminPermission.members(project)` for example that we may use here.

I don't think we need to fix this issue in this PR, but we should write a `# TODO` commend and open a note/issue in the GH project we have to not forget about this and come back before deploying this.

saadmk11:
> I don't think we need to fix this issue in this PR, but we should write a `# TODO` commend and open a note/issue in the GH project we have to not forget about this and come back before deploying this.

Added ``# TODO`` and a card for this in the project. Also Updated the PR :)

humitos (approved):
:tada:

PR #7734, opened 2020-12-08
Use remote_id and vcs_provider Instead of full_name to Get RemoteRepository

saadmk11:


readthedocs/oauth/services/base.py
```diff
@@ -34,6 +34,7 @@ class Service:
 
     adapter = None
     url_pattern = None
+    PROVIDER_SLUG = None
```
humitos, line 37: Why don't we call it `VCS_PROVIDER` to match the field's name?

humitos (approved):

readthedocs/oauth/services/base.py
saadmk11: @humitos This class is a VCS provider service class so It felt like I was referring to this class if I use ``VCS_PROVIDER``, so I used ``PROVIDER_SLUG``. but we can use ``VCS_PROVIDER_SLUG`` here. Or should I just use ``VCS_PROVIDER``?

humitos: Either one is fine. It just called me the attention that we were using a different name here. Also, we can make it lowercase to match the pattern that we have with the other class variables used here as well.

humitos (approved):

PR #7746, opened 2020-12-10
Use "path_with_namespace" for GitLab RemoteRepository full_name Field

saadmk11:
For GitLab we use
``repo.full_name = fields['name_with_namespace'] # ("Maksudul Haque / test")``
for ``full_name`` and not
``repo.full_name = fields['path_with_namespace'] # ("saadmk11/test")``
Here ``fields['path_with_namespace']`` is more appropriate for the ``full_name`` field as it looks similar to the other VCS providers (GitHub, BitBucket) ``full_name`` field value.


humitos (approved):
Excellent!

PR #7749, opened 2020-12-14
Register ShutdownBuilder task

humitos:
Register the task to shutdown the builders by themselves.

Required by: https://github.com/readthedocs/readthedocs-ext/pull/368


ericholscher (approved):
This is a hacky way to do this, we should find a better pattern :/

PR #7753, opened 2020-12-15
Set The Right Permissions on GitLab OAuth RemoteRepository

saadmk11:
ref: https://github.com/readthedocs/readthedocs.org/issues/7743

### DIFF
--- a/readthedocs/oauth/services/gitlab.py
+++ b/readthedocs/oauth/services/gitlab.py
@@ -122,8 +122,13 @@ class GitLabService(Service):
                 remote_organizations.append(remote_organization)
 
                 for repo in org_repos:
+                    # TODO: Optimize this so that we don't re-fetch project data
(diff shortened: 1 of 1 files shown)


readthedocs/oauth/services/gitlab.py
```diff
@@ -122,11 +122,35 @@ def sync_organizations(self):
                 remote_organizations.append(remote_organization)
 
                 for repo in org_repos:
-                    remote_repository = self.create_repository(
-                        repo,
-                        organization=remote_organization,
-                    )
-                    remote_repositories.append(remote_repository)
+                    try:
+                        # Get response for a single repository with permission data
```
humitos, line 126: It would be good to expand this comment here about _why_ this is required and making a TODO note to optimize this --since it's data that we already fetched from `sync_repositories` method.

humitos (approved):
Looks good to me! I just left a comment to improve the comment.

stsewd (approved):

PR #7754, opened 2020-12-15
Router project using mamba to build:default

humitos:
`mamba` consumes a lot less memory than `conda` and the common case should not
require 8Gb to perform the build.

This will help us to reduce the build-large queue backing up, save some money
and also give users a better UX.

Besides, we can still forcing a specific queue for mamba projects if needed by
setting `project.build_queue` as usual.


ericholscher:
I don't think we need this optimization yet. Lets confirm that mamba works well before we start doing more work around it.

PR #7773, opened 2020-12-21
Use Python3.7 on conda base environment when using mamba

humitos:
Mamba 0.1.0 is the latest version that supports Python2 (which is our default in
the conda environment --as we use miniconda2) while the current stable version
is 0.7.6

Using the older version of mamba is causing different resolution issues and
making the build to fail.

Environment still can be created with python2 or python3 versions.

See https://github.com/mamba-org/mamba/issues/640 for more context.

### DIFF
--- a/readthedocs/doc_builder/python_environments.py
+++ b/readthedocs/doc_builder/python_environments.py
@@ -529,6 +529,7 @@ class Conda(PythonEnvironment):
             '--quiet',
             '--name=base',
             '--channel=conda-forge',
+            'python=3.7',
             'mamba',
             cwd=self.checkout_path,
         )


ericholscher (approved):
New tools always add additional support overhead :( This is a simple change, but I really wish this stuff wasn't so fiddly.

wolfv:
Thanks @humitos!

@ericholscher the underlying problem is that Miniconda 2 ships Python 2.7 and that's deprecated.
Conda & Mamba have some special handling to keep the (minor) Python version so the base-environment Python 2.7 stays untouched. 

I see no reason for readthedocs to switch to Miniconda 3 (that would come with Python 3.7).
There are also no recent conda versions released for Python 2.7 afaik.

PR #7778, opened 2020-12-22
Updated RemoteRepository API and import.js to use new RemoteRepository modeling

saadmk11:


readthedocs/api/v2/serializers.py
```diff
@@ -193,6 +194,19 @@ def get_matches(self, obj):
         if request.user is not None and request.user.is_authenticated:
             return obj.matches(request.user)
 
+    def is_admin(self, obj):
+        request = self.context['request']
+
+        # Use cached value
+        if hasattr(obj, 'admin'):
+            return obj.admin
```
humitos, line 202: Can we use `functools.lru_cache` for this?

readthedocs/api/v2/serializers.py
```diff
@@ -193,6 +194,19 @@ def get_matches(self, obj):
...
+
+        if request.user is not None and request.user.is_authenticated:
```
humitos, line 204:
```suggestion
        if request.user and request.user.is_authenticated:
```

readthedocs/api/v2/views/model_views.py
```diff
@@ -381,7 +377,27 @@ class RemoteRepositoryViewSet(viewsets.ReadOnlyModelViewSet):
     pagination_class = RemoteProjectPagination
 
     def get_queryset(self):
-        query = self.model.objects.api(self.request.user)
+        user = (
+            self.request.user
+            if self.request.user.is_authenticated
+            else None
+        )
```
humitos, line 384: If the user is not authenticated, I think we should shortcircuit this and just return an empty queryset, right?

readthedocs/api/v2/views/model_views.py
```diff
@@ -381,7 +377,27 @@ class RemoteRepositoryViewSet(viewsets.ReadOnlyModelViewSet):
...
+        query = self.model.objects.api(self.request.user).annotate(
+            admin=Case(
+                When(
```
humitos, line 387: This query looks really good! I think it was a good idea to keep the compatibility. However, I'm a little worry about its performance. I never used `Case` and `When` and we will have a lot of relationships here.

We may need to put a TODO here with a message to checkout this after deploying it. 

Also, we can ask someone else what they think; but probably a TODO comment and a note in the GH project is fine for now.

humitos (approved):
Great! Thanks!

readthedocs/api/v2/views/model_views.py
saadmk11: we can also use ``Exists`` with a subquery here. But I think this is quite efficient here. I'll add a ``# TODO`` and a note in the GH project. :)

readthedocs/api/v2/serializers.py
saadmk11: I'm not that much familiar with ``functools.lru_cache`` but I tried it and it doesn't seem to work here. :( And I don't think we will have many subsequent calls to the same objects ``is_admin`` method in the API. But let me know what you think.

humitos: Honestly, I don't really understand how the logic you write here works :sweat_smile:. I didn't see where we are setting `obj.admin` in our code to use it as a cached value. Can you point me to that?

humitos: > I don't think we have any subsequent calls to the same objects `is_admin` method in the same API request

This is basically what I don't understand. If we don't have subsequent calls, why do we need to cache something here?

saadmk11: `obj.admin` attribute is coming from the queryset annotation on the API view

https://github.com/readthedocs/readthedocs.org/blob/0c04f9f2db8763a64ecf9d600f4e3f1663b4d0e5/readthedocs/api/v2/views/model_views.py#L386

I went with this approach for database optimization. We require ``admin`` field in the API response and it is used in `import.js`

We have a few similar approaches in the codebase, for example, 

https://github.com/readthedocs/readthedocs.org/blob/7b53485714ba30ce5cfe63129eb9ab555f46abba/readthedocs/projects/models.py#L1176-L1188

humitos: Thanks for the explanation! Your code is correct :smile:

PR #7783, opened 2020-12-23
Add migration to create a new table for RemoteOrganization and RemoteOrganizationRelation and Update Code

saadmk11:


saadmk11:
After this is merged I can add PR's to update the python code according to the new modeling. :)

readthedocs/oauth/migrations/0012_create_new_table_for_remote_organization_normalization.py
```diff
@@ -0,0 +1,62 @@
...
+        ('oauth', '0011_add_default_branch'),
+    ]
+
+    operations = [
+        migrations.CreateModel(
+            name='RemoteOrganization',
+            fields=[
+                ('id', models.AutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
+                ('pub_date', models.DateTimeField(auto_now_add=True, verbose_name='Publication date')),
+                ('modified_date', models.DateTimeField(auto_now=True, verbose_name='Modified date')),
```
humitos, line 24: Is it possible to change these and use `TimeStampedModel` instead as you did with `RemoteRepositoryRelation`?

Same question for `RemoteRepository` :)

readthedocs/oauth/models.py
```diff
@@ -59,17 +50,54 @@ class RemoteOrganization(models.Model):
...
+        User,
+        related_name='remote_organization_relations',
+        on_delete=models.CASCADE
+    )
+    account = models.ForeignKey(
+        SocialAccount,
+        verbose_name=_('Connected account'),
+        related_name='remote_organization_relations',
+        null=True,
+        blank=True,
```
humitos, line 90: What would be the case where we do have a relation with user, but the account is null? I think this is not possible.

humitos (approved):
Looks good to me!

readthedocs/oauth/models.py
saadmk11: Yes you are right, maybe we did it because of the previous plan

humitos (approved):

humitos:
Some tests are failing. We can merge it after they pass.

saadmk11:
> Some tests are failing. We can merge it after they pass.

@humitos I did not change the related code according to the new modeling of ``RemoteOrganization`` That's why the tests are failing. Should I add the related code changes to this PR as well?

> It seems it's because we removed the `null=True` in `account` and there are old tests code that depends on that. If it's too complicated to make them pass, we can keep it with the `null=True`.

I will look into this. :)

humitos:
> @humitos I did not change the related code according to the new modeling of `RemoteOrganization` That's why the tests are failing. Should I add the related code changes to this PR as well?

I don't have a strong preference here. Whatever works best for you is fine.

humitos (approved):
Looks great to me! I left some small comments, but I'm happy to merge it as it is. Let me know!

Tip: next time avoid doing force-push on PRs that have already an instance of review. This helps reviewers to only review "new changes" instead of everything again. Doing incremental reviews is easier and fast. When the push is forced, GH does not known what was the last thing I reviewed and show me all the changes together instead of only the changes from last time I saw them.

humitos (approved):

PR #7796, opened 2021-01-05
Update Re-connect RemoteRepository Management command with new modeling

saadmk11:


readthedocs/oauth/management/commands/reconnect_remoterepositories.py
```diff
@@ -56,20 +56,19 @@ def _connect_repositories(self, organization, no_dry_run, only_owners):
             Q(clone_url__in=Subquery(organization.projects.values('repo')))
         )
         for remote in RemoteRepository.objects.filter(remote_query).order_by('created'):
-            admin = json.loads(remote.json).get('permissions', {}).get('admin')
+            admin = remote.remote_repository_relations.filter(
+                user__owner_organizations=organization,
+                admin=True
+            ).exists()
 
-            if only_owners and remote.users.first() not in organization.owners.all():
```
saadmk11, line 61: I'm not sure what to do here and if it's required anymore or not, I'm not familiar with Organization code :/

humitos:
@saadmk11 don't worry about this command for now, it won't be needed anymore after the DB normalization. Currently, we use it as a helper to reconnect `Project -> RemoteRepository` in case the user deletes his account. As this problem will be fixed by the normalization we are working on, this command can be deleted completely in the near future.

PR #7797, opened 2021-01-05
Change query on `send_build_status` task for compatibility with .com

humitos:
We can't query over `user__projects=` since in .com User objects are not related
directly to Project but instead they are related via an intermediate model
called Team or via SSO.

Use `user__in=AdminPermission.members(project)` instead, where `AdminPermission`
is overloaded in .com to behaves as we need here.

Closes this comment: https://github.com/readthedocs/readthedocs.org/pull/7728#issuecomment-739977236

### DIFF
--- a/readthedocs/projects/tasks.py
+++ b/readthedocs/projects/tasks.py
@@ -46,6 +46,7 @@ from readthedocs.builds.constants import (
 from readthedocs.builds.models import APIVersion, Build, Version
 from readthedocs.builds.signals import build_complete
 from readthedocs.config import ConfigError
+from readthedocs.core.permissions import AdminPermission
 from readthedocs.core.resolver import resolve_path
 from readthedocs.core.utils import send_email
 from readthedocs.doc_builder.config import load_yaml_config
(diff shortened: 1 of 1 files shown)


readthedocs/projects/tasks.py
```diff
@@ -1899,15 +1900,16 @@ def send_build_status(build_pk, commit, status, link_to_build=False):
     if provider_name in [GITHUB_BRAND, GITLAB_BRAND]:
         # get the service class for the project e.g: GitHubService.
         service_class = build.project.git_service_class()
-        users = build.project.users.all()
```
saadmk11, line 1902: If the `try` block does not return or raise an exception then the `except` block won't be executed and Local variable 'users' will be referenced before assignment here

https://github.com/readthedocs/readthedocs.org/blob/be8282ad75e8f9632a2b055ec6ba00940566e6fb/readthedocs/projects/tasks.py#L1965

humitos: Thanks! I shouldn't have changed that :)

Moving it back where it was.

saadmk11 (approved):
Looks Great :+1:

Issue #7791, opened 2021-01-04
Management command to trigger VCS provider data re-sync


humitos:
Once we have the new `RemoteRepository` and `RemoteOrganization` modeling using `*Relation` models in production, we will need a management command to trigger the re-sync of this data.

Taking into account that we may hit rate limit API (*) we will need to do this in chunks and have some way to trigger re-sync with some filters

- re-sync only specific users: `--users humitos --users readthedocs`
- skip users: `--skip-users humitos`
- use a specific celery queue for the tasks (only the instance with new modeling will be listen to this queue): `--queue resync-oauth`
- limit the amount of tasks triggered: `--max-tasks 100`
- continue from an offset: `--offset 100`
- print total tasks required to complete and current position: `Re-syncing 100-199 / 60000`

(*) I'm not super worried about this since the rate limit is based on user, and it's 5k per user on GH, IIRC.

saadmk11:
closed on https://github.com/readthedocs/readthedocs.org/pull/7803

PR #7803, opened 2021-01-06
Add management command to Sync RemoteRepositories and RemoteOrganizations

saadmk11:
This is the simple implementation of the management command that can be used in production to Sync `RemoteRepositories` and `RemoteOrganizations` for users. This will be useful to populate the new table created for the normalization.

**Options**
- `--queue` name of the celery queue for the tasks to be triggered with. (Default: resync-oauth)
- `--skip-users` Skip re-sync VCS provider data for specific users. (Default: [])
- `--users` Re-sync VCS provider data for specific users only. (Default: [])
- `--max-users` maximum number of users that should be synced. (Default: 100)
- `--force` Force re-sync VCS provider data even if the users are already synced.
 - `--dry-run` Do not trigger tasks for VCS provider re-sync.

**Example:** 
`python manage.py sync_vcs_data --queue resync-oauth --max-users 500`
`python manage.py sync_vcs_data --users saadmk11 test_user --skip-users another_user`

This management command triggers 1 `sync_remote_repositories` task per user.

closes https://github.com/readthedocs/readthedocs.org/issues/7791


readthedocs/oauth/management/commands/sync_vcs_data.py
```diff
@@ -0,0 +1,74 @@
...
+        sync_users = options.get('users')
+        skip_users = options.get('skip_users')
+        max_users = options.get('max_users')
+
+        # Filter users who have social accounts connected
+        # and has no remote repository relations
+        users = User.objects.filter(
+            socialaccount__isnull=False,
+            remote_repository_relations__isnull=True
+        ).distinct()
```
humitos, line 49: I think we shouldn't apply these filters if we pass `sync_users` so we can force a re-sync for a user that was already synced some time ago.

Or... maybe another parameter `--force` that avoid these two filters is more explicit here.

humitos (commented):
This looks great to me!

readthedocs/oauth/management/commands/sync_vcs_data.py
```diff
@@ -0,0 +1,74 @@
...
+        if skip_users:
+            users = users.exclude(username__in=skip_users)
+
+        users_to_sync = users.values_list('id', flat=True)[:max_users]
+
+        self.stdout.write(
+            self.style.SUCCESS(
+                'Found %s user(s) with the given parameters' % users.count()
+            )
+        )
```
humitos, line 63: Can we also show the total number of users that need to be resync before taking the `max_users` slice?

saadmk11: `users.count()` shows the total number of users that needs to be re-synced according to the filters. The next count is the count after the slice. Did you mean something else?

readthedocs/oauth/management/commands/sync_vcs_data.py
saadmk11: @humitos Shouldn't we keep `socialaccount__isnull=False` for all the filters? As there is no point trying to sync and triggering a task for a user who doesn't have any social account connected.

saadmk11:
I think we can add a `--dry-run` option which will not trigger any tasks but show us the number of users found with the given parameters. What are your thoughts on this?

readthedocs/oauth/management/commands/sync_vcs_data.py
humitos: Yeah, I didn't express myself correctly. Only the `remote_repository_relations__isnull=True` is the one that I wanted to remove.

readthedocs/oauth/management/commands/sync_vcs_data.py
```diff
@@ -0,0 +1,84 @@
...
+        # Filter users who have social accounts connected to their RTD account
+        users = User.objects.filter(
+            socialaccount__isnull=False
+        ).distinct()
+
+        if not force_sync:
+            users = users.filter(
+                remote_repository_relations__isnull=True
+            ).distinct()
+
```
humitos, line 60:
```suggestion
        if force_sync:
            users = users.filter(
                remote_repository_relations__isnull=False
            ).distinct()

```

Seems easier to read to me.

readthedocs/oauth/management/commands/sync_vcs_data.py
```diff
@@ -0,0 +1,84 @@
...
+            users = users.filter(username__in=sync_users)
+
+        if skip_users:
+            users = users.exclude(username__in=skip_users)
+
+        users_to_sync = users.values_list('id', flat=True)[:max_users]
+
+        self.stdout.write(
+            self.style.SUCCESS(
+                'Found %s user(s) with the given parameters' % users.count()
```
humitos, line 71: Why not using `.format` of f-string here?

readthedocs/oauth/management/commands/sync_vcs_data.py
humitos: Yeah, I wanted to know the total of users that needs to be synced without applying `sync_users` or `skip_users`. So, maybe those filters need to be applied to `users_to_sync` instead of `users`. Don't you think?

In that case we will have:

- Found TOTAL user(s) to sync in total
- Re-syncing VCS Providers for FILTERED_USERS user(s) with given parameters

This gives us a clear number about "where we are" when running this command. Every time we run it we know exactly how many users still need re-syncing.

humitos:
> I think we can add a `--dry-run` option which will not trigger any tasks but show us the number of users found with the given parameters. What are your thoughts on this?

:100:

readthedocs/oauth/management/commands/sync_vcs_data.py
```diff
@@ -40,13 +40,20 @@ def add_arguments(self, parser):
             default=False,
             help='Force re-sync VCS provider data even if the users are already synced.',
         )
+        parser.add_argument(
+            '--no-dry-run',
```
humitos, line 44: I think it's more common to use the opposite here: `--dry-run`

humitos (approved):
Looks good! I just suggest to use `--dry-run` instead of `--no-dry-run`

readthedocs/oauth/management/commands/sync_vcs_data.py
saadmk11: Okay, I added this so that we don't accidentally trigger tasks

humitos (approved):
Excellent! Thanks for your help on this!

PR #7812, opened 2021-01-08
Use S3 (MinIO emulator) as storage backend

humitos:
- remove Azurite storage backend
- update all settings to use S3
- update docs to mention set "Read Only" on buckets
- disable S3 console logs
- bring `s3_storage.py` from -ext
- delete `azure_storage.py` from .org
- update django-storages to install boto3 dependencies


readthedocs/settings/docker_compose.py
```diff
@@ -18,10 +18,14 @@ class DockerBaseSettings(CommunityDevSettings):
     PRODUCTION_DOMAIN = 'community.dev.readthedocs.io'
     PUBLIC_DOMAIN = 'community.dev.readthedocs.io'
     PUBLIC_API_URL = f'http://{PRODUCTION_DOMAIN}'
+
     SLUMBER_API_HOST = 'http://web:8000'
+    SLUMBER_USERNAME = 'admin'
+    SLUMBER_PASSWORD = 'admin'
```
agjohnson, line 24: These look to be new, are these settings intentional?

requirements/pip.txt
```diff
@@ -101,11 +101,19 @@ django-cors-middleware==1.4.0  # pyup: ignore
...
-# private buckets
-git+https://github.com/jschneier/django-storages@d0f027c98a877f75615cfc42b4d51c038fa41bf6#egg=django-storages[azure]==1.9.1
+# private buckets.
+#
+# Besides, support for the corresponding AWS_BUCKET_ACL and
+# AWS_AUTO_CREATE_BUCKET settings have been removed in 1.10. We depend on this
+# in our Docker setup. We can upgrade it but we need to add a
+# `create_buckets.sh` to be called on `--init` as we used to do for Azurite
+# https://github.com/jschneier/django-storages/pull/636
+git+https://github.com/jschneier/django-storages@d0f027c98a877f75615cfc42b4d51c038fa41bf6#egg=django-storages[boto3]==1.9.1
```
agjohnson, line 115: We can track this with a separate issue perhaps, we probably don't want to get too far behind on django-storages releases

readthedocs/storage/s3_storage.py
```diff
@@ -0,0 +1,90 @@
...
+    bucket_name = getattr(settings, 'S3_BUILD_ENVIRONMENT_STORAGE_BUCKET')
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        if not self.bucket_name:
+            raise ImproperlyConfigured(
+                'AWS S3 not configured correctly. '
+                'Ensure S3_BUILD_ENVIRONMENT_STORAGE_BUCKET is defined.',
+            )
```
agjohnson, line 90: Note: we'll want a PR on commercial/ops/etc in the near future to make sure we're using this class instead of the ext class

Also, I assumed this file is just forked and didn't spend much time at all on review here.

agjohnson (commented):
Seems to look good to me!

requirements/pip.txt
humitos: I'm not sure to follow you here.

We are not changing the version in this PR. I just changed `azure` by `boto3` + add a comment explaining _why_ we can't upgrade.

readthedocs/settings/docker_compose.py
humitos: These are required for builds to hit the API. Otherwise, we get 403 and builds don't work.

This problem is not new, but I suppose that you may have these settings overwritten in a `local_settings.py` file?

They are required in docker environment to be able to build docs.

humitos: We can remove them from this PR and figure it out in another one, tho.

ericholscher: Yea, seems odd this would be related to this PR. It must have been working before somehow?

ericholscher (approved):
This looks good to me. Will be good to get it testing locally 👍

readthedocs/settings/docker_compose.py
humitos: It's completely unrelated to this PR, yes. I will remove it from it and add it to another one so we can discuss it there. We will see if it fails for you as well when building.

PR #7817, opened 2021-01-11
Register MetricsTask to send metrics to AWS CloudWatch

humitos:


readthedocs/builds/apps.py
```diff
@@ -28,3 +28,9 @@ def ready(self):
             app.tasks.register(AutoscaleBuildersTask)
         except (ModuleNotFoundError, ImportError):
             log.info('AutoscaleBuildersTask task could not be imported.')
+
+        try:
+            from readthedocsext.monitoring.metrics.tasks import MetricsTask
```
ericholscher, line 33: This makes me think maybe the monitoring app should be moved into RTD core, to avoid this weird magic. No reason reason not to.

ericholscher (approved):

readthedocs/builds/apps.py
humitos: My only reason to not include it in the .org repository was that I didn't think this could be useful to anyone else -at least at this moment, that it's new and we are still in a testing phase. But yeah, there is nothing private currently in our monitoring code.

agjohnson: I agree with @humitos here and thing the separation between core/ext for me is whether the code is important to running a RTD instance or if it's just for running our prod RTD instance. This is very particular to running on AWS and would extra burden to publicly maintain, for little benefit to us -- we'd never accept supporting other providers here or customizing this much for other platforms, so ext feels like an applicable place for this code.

ericholscher: Makes sense. I just don't like having these weird hooks in the OSS codebase, but it's fine I guess.

agjohnson (approved):

PR #7820, opened 2021-01-12
Log Stripe Resource fallback creation in Stripe

humitos:
I hope this helps to track down an issue we are having with subscriptions.


agjohnson (approved):
Looks good for hotfix :+1:

PR #7839, opened 2021-01-18
Speed up concurrent builds by limited to 5 hours ago

### LINKED ISSUE #7822: Improve performance for concurrency builds DB query
We start getting timeout when calling `/api/v2/builds/concurrent/` from builders to know `(limit_reached, concurrent, max_concurrent)` for a particular project.

We need to improve the queryset that works over the `Build` model because sometimes it takes ~4s or more. It usually takes ~200ms, so I guess this is because the project is/has subprojects/translations since in that case the query gets more complex.

This is the query we need to optimize,

https://github.com/readthedocs/readthedocs.org/blob/2e9cb8dd85c5f8a55ab085abe58082fe3f2c6799/readthedocs/builds/querysets.py#L129-L177

Sentry issues: https://sentry.io/organizations/read-the-docs/issues/1719520575/?project=148442&query=is%3Aunresolved&statsPeriod=14d

humitos:
Instead of querying the whole table, we filter builds up to 5 hours ago when
looking for concurrent builds.

The query goes from ~4s (in the worst case) to ~6ms with this filter.

Closes #7822

### DIFF
--- a/readthedocs/builds/querysets.py
+++ b/readthedocs/builds/querysets.py
@@ -1,8 +1,10 @@
 """Build and Version QuerySet classes."""
+import datetime
 import logging
 
 from django.db import models
 from django.db.models import Q
+from django.utils import timezone
 
 from readthedocs.core.utils.extend import SettingsOverrideObject
 from readthedocs.projects import constants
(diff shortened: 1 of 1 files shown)


humitos:
There is an unrelated problem installing `node` on CircleCI. However, `tox -e lint` passes locally.

agjohnson:
I reran tests and looks like probably just a temporary failure

ericholscher (approved):

PR #7841, opened 2021-01-19
Metric tasks for community

humitos:
Override default base -ext metrics to add `build-large` queries.

This file is missing from PR https://github.com/readthedocs/readthedocs.org/pull/7817

### DIFF
--- a/readthedocs/metrics/tasks.py
+++ b/readthedocs/metrics/tasks.py
@@ -1,0 +1,39 @@
+"""
+Specific metric tasks for community.
+
+Override the base metric tasks to add specific ones only required on community.
+"""
+
+# Disable import error because -ext is not available on pylint
+# pylint: disable=import-error
+from readthedocsext.monitoring.metrics.database import RunningBuildsMetric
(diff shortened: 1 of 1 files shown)


ericholscher (approved):

readthedocs/metrics/tasks.py
```diff
@@ -0,0 +1,39 @@
+"""
+Specific metric tasks for community.
+
+Override the base metric tasks to add specific ones only required on community.
+"""
+
+# Disable import error because -ext is not available on pylint
+# pylint: disable=import-error
+from readthedocsext.monitoring.metrics.database import RunningBuildsMetric
```
ericholscher, line 9: Will this not break when the user doesn't have ext installed?

humitos: I shouldn't break because nobody import this file, but yeah. I don't like it here either. I didn't know where to put them, tho. `metrics/community.py` in -ext sounds better than here. I'll move this code there after the deploy.

PR #7889, opened 2021-02-02
Use Stripe Checkout for Gold Users

humitos:
Initial work to make Stripe Checkout to work with out Gold Users workflow.

1. User goes to Gold User admin tab
2. Select the amount per month to subscribe to and send it (Django Form)
   - Javascript POST request with the PlanID
   - Django creates a Stripe Checkout Session and returns the SessionID
   - Javascript redirects the user to the Stripe Checkout (*it seems this step can't be done on the backend*)
3. User sees all the subscription details and enters CC info directly on Stripe
4. The user is redirected to success or cancel URLs

The next time the user goes to Gold User admin tab, they will see small details
about the subscription and a button "Manage subscription" that will redirect
them directly to Stripe Customer Portal to update/cancel their subscription.

### Gold Membership tab (without a subscription)

Screenshot_2021-03-03_17-38-02

### Gold Membership tab (**with** a subscription)

Screenshot_2021-03-03_17-37-39

#### TODO

- [ ] remove more Gold Users related code
- [ ] remove more `readthedocs/payments` code
- [ ] review/improve Javascript code and handle exceptions
- [x] verify we are handling the Stripe webhook and associating the new customer created to the `GoldUser` object
- [x] show "Manage subscription" button once the user has a subscription and redirect to Stripe Customer Portal when clicked
- [x] create/update success and cancel pages (maybe just show the detail page and add Django message)
- [x] remove "Update quantity" from Stripe Customer Portal
- [ ] add TAX ID now it's supported
- [x] tests?

Based on https://github.com/stripe-samples/checkout-single-subscription/tree/master/server/python


ericholscher:
Would love to get this merged prior to next deploy. 🚀 ❤️

readthedocs/gold/views.py
```diff
@@ -52,6 +60,7 @@ def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
         domains = Domain.objects.filter(project__users=self.request.user)
         context['domains'] = domains
+        context['stripe_publishable'] = settings.STRIPE_PUBLISHABLE
```
ericholscher, line 63: Is there a reason we need this if we're POST'ing to our URL then redirecting? Can we not just have the key in our backend and pass it along?

readthedocs/gold/views.py
```diff
@@ -152,3 +139,149 @@ def post(self, request, *args, **kwargs):
...
+
+
+class GoldSubscriptionPortal(GenericView):
+    http_method_names = ['post']
+
+    def post(self, request, *args, **kwargs):
+        user = request.user
+
+        # TODO: review if User <-> GoldUser being ``ForeignKey`` makes
+        # sense no that we are removing the one time donation
```
ericholscher, line 189:
```suggestion
        # sense now that we are removing the one time donation
```

Do we need to remove one-time donations? I feel like they are useful, but it might also just make more sense to direct those people to GH sponsors or OpenCollective or something? If possible, we should probably keep it tho, because some users want a one-time way of paying to remove ads.

readthedocs/gold/views.py
```diff
@@ -152,3 +139,149 @@ def post(self, request, *args, **kwargs):
...
+                    # TODO: add user notification saying it failed
+                    log.exception(
+                        'Gold User payment failed. username=%s customer=%s',
+                        username,
+                        stripe_customer,
+                    )
+
+                elif event.type == self.EVENT_CUSTOMER_SUBSCRIPTION_UPDATED:
+                    # TODO: check if the subscription was canceled, past due,
+                    # etc and take the according action
```
ericholscher, line 261: Yea, that seems useful. I believe we haven basically just never cancelled a Gold sub, even when payments stopped. Would be useful to do more state tracking if it isn't too difficult.

ericholscher (approved):
This seems like a huge win for me. The only notable thing is it would be useful to keep one-time payments if possible. Can we just direct folks to Stripe in some fashion for one-time payments? Seems like we should be able to?

readthedocs/gold/views.py
humitos: We _don't have to_. I know that we decided to remove one-time payments, but I'm not sure if it was because it was too complex to maintain or we wanted to removed completely for other reasons. If we want to keep it, we may not be able re-use the code from `Subscription` (in the medium term) or we may need to do something extra to mark them as `subscribed=False` (as the current `GoldUser` is doing). However, with the latest changes to our Gold Membership, I'm not sure if doing a one-time donation has extra benefits for the user, so we may not need to track this in the database. IIRC to remove ads you need to be subscribed and pay USD 5 (user) or 50 (company) per project.

readthedocs/gold/views.py
humitos: This was tricky to me as well. I wanted _our backend_ to redirect directly to Stripe Checkout, but for some reason, that has to be done in frontend via Javascript :shrug: 

https://stripe.com/docs/api/checkout/sessions/object#checkout_session_object-id

The official Python example does it in this way also, so I thought it was not possible to do it from the backend.

Docs: https://stripe.com/docs/billing/subscriptions/checkout

However, for the Stripe Customer Portal _it is_ possible to do the redirect from Python.

readthedocs/gold/views.py
humitos: Shouldn't be too difficult. Besides, we already have this code written for .com --I want to bring `subscriptions` app into .org eventually and share all this code if possible.

readthedocs/gold/views.py
ericholscher: Hrm yea -- I think in the past we've done one-time donations them manually marked projects as Ad-free. That should still work fine. 

It would be useful to continue supporting one-time donations. We support it here: https://readthedocs.org/sustainability/ -- maybe that code will still "just work", because it isn't tied to the Gold stuff?

readthedocs/gold/views.py
ericholscher: Yea, that would be ideal, it seems.

readthedocs/gold/views.py
ericholscher: Interesting.

readthedocs/gold/views.py
humitos: > I'd prefer if the form submitted to Stripe and you put your CC info directly there, so we don't ever have a chance to see it, but probably fine to keep the existing workflow if it still works.

The idea is to remove (or replace with Stripe Checkout) this workflow completely. They are not going to be included in the new theme.

ericholscher: Yea, I think we need a way to do one-time payments then. It doesn't need to be hooked up to the DB at all I don't think. I'm 👍  on using another platform for this, so we don't need to maintain code. If we can easily do it with Stripe, that would be great, but I don't care too much.

humitos:
I think we can merge this PR as it is now. It handles Gold User Membership by using Stripe Checkout (create subscription) and Stripe Customer Portal (manage subscription).

@ericholscher Note that we may need to upgrade our Stripe account to be able to use the Stripe Customer Portal.

One-time donations are not touched and keep using the same approach we had. This is written in `readthedocsext.donate` application and I think we can migrate it in another PR (requires a PR in .org to remove old code and -ext for the logic)

ericholscher:
I'm 👍 on shipping this next week. It should be a much better experience for al our gold users. 💯

PR #7891, opened 2021-02-04
Add publicly visible env vars

### LINKED ISSUE #7413: Set environment variables in PR builds of private repos
## Actual Result

In PR builds, environment variables (=secrets) don’t get set to prevent people from sending them somewhere in plain text just by making a PR.

## Expected Result

In a private repo, the above wouldn’t be necessary as all contributors are trusted. There’s still incentive not to add the secrets to the code though, e.g. in case the code becomes public at some point. Therefore it would be great to be able to access the secrets in PR builds of private repos.

flying-sheep:
I have no idea what I’m doing but maybe this eventually fixes #7413

Also IDK how to restructure the API to support this.

Since `Version`s are always associated with a `Project`, maybe the better abstraction would be to add a `environment_variables` property to the `Version` class. That way we wouldn’t have to do `all=version.type != EXTERNAL` in two places but could derive it in the property body.


flying-sheep:
Huh, why is the field not found? I’m adding it in the migration!

readthedocs/projects/models.py
```diff
@@ -1233,8 +1233,7 @@ def show_advertising(self):
 
         return True
 
-    @property
-    def environment_variables(self):
+    def environment_variables(self, *, all=True):
```
humitos, line 1236: I would use `False` as default value, which I think it's safer.

Also, I may use a different name than `all`. Maybe `only_public=True` to be explicit about what's returning

readthedocs/projects/models.py
```diff
@@ -1334,8 +1334,9 @@ def show_advertising(self):
         """Whether this project is ad-free (don't access the database)."""
         return not self.ad_free
 
-    @property
-    def environment_variables(self):
+    def environment_variables(self, *, all=True):
```
humitos, line 1337: I think this should keep being a property. I understand it will be populated with what's returned by the API, so there is no need to re-filter by "only public" environment vars.

readthedocs/projects/models.py
```diff
@@ -1825,6 +1826,11 @@ class EnvironmentVariable(TimeStampedModel, models.Model):
         on_delete=models.CASCADE,
         help_text=_('Project where this variable will be used'),
     )
+    private = models.BooleanField(
```
humitos, line 1829: Hrm... In some places you are calling it `public` and in other `private` :smile: --that's why test are failing.

humitos (commented):
Hi, thanks for the PR!

I left some initial comments about this PR. We may also need to update the APIv3 and its documentation to expose `.public` field

I also think that at some point we will want to decide if the env variable will be available only in PRIVATE or PUBLIC versions.

flying-sheep:
> if the env variable will be available only in PRIVATE or PUBLIC versions

sorry for being nitpicky, just making sure: you mean “only in private or both in private and public versions”. to make them available only in public versions doesn’t make sense

readthedocs/projects/models.py
flying-sheep: hahaha whoops! It was `private` before, then /u/stsewd suggested calling it `public`!

readthedocs/projects/models.py
flying-sheep: But isn’t the idea that this is a drop-in replacement for the real model? where is it used?

readthedocs/projects/models.py
flying-sheep: OK! Since `only_public` means flipping the meaning, flipping the default as well means it stays `True`

flying-sheep: If we use “return only publy” as a default, then the API would change behaviors, as the serializer only supports calling methods with their default parameters.

readthedocs/projects/models.py
flying-sheep: OK, done by leaving the default as “return all”, see above

flying-sheep:
ha! tests work. why doesn’t the migration check though? the output doesn’t speak to me

stsewd (commented):
Thanks for the PR!

So, in the build process we don't have direct access to the db for security reasons. Instead we get the information we need for the build from the API (v2).

As you have already seen, we use a wrapper class for data returned from the API.  
https://github.com/readthedocs/readthedocs.org/blob/f18c584d9e6ba71105a0c42753ed6e9fbe4ebf0a/readthedocs/projects/models.py#L1288-L1288

To be able to pass only a subset of env variables into the build process (public env vars),
we need to be able to exclude them from the API first.

stsewd:
> ha! tests work. why doesn’t the migration check though? the output doesn’t speak to me

That's because you changed the model but didn't update the migration, you can delete the migration and generate it again to have it match the changes.

readthedocs/projects/models.py
```diff
@@ -1825,6 +1828,11 @@ class EnvironmentVariable(TimeStampedModel, models.Model):
         on_delete=models.CASCADE,
         help_text=_('Project where this variable will be used'),
     )
+    public = models.BooleanField(
+        _('Public'),
+        default=False,
+        help_text=_('Expose this environment variable in PR builds?')
+    )
```
stsewd, line 1835: When adding a new field we mark it a `null=True`, this is to avoid downtime when doing the deploy. For example https://github.com/readthedocs/readthedocs.org/blob/f18c584d9e6ba71105a0c42753ed6e9fbe4ebf0a/readthedocs/projects/models.py#L1379-L1384

And when adding it to the form we need to remove the none option

https://github.com/readthedocs/readthedocs.org/blob/f18c584d9e6ba71105a0c42753ed6e9fbe4ebf0a/readthedocs/projects/forms.py#L229-L231

flying-sheep: Ah, so that’s what’s missing from the migration!

flying-sheep:
OK, done!

stsewd:
@flying-sheep as mentioned in my comment (https://github.com/readthedocs/readthedocs.org/pull/7891#pullrequestreview-586156426), you'll need to change the API in order to make this work. The privacy level should be returned in the API, so we can filter them from the builders.

https://github.com/readthedocs/readthedocs.org/blob/7a117b630094d0d795349235ee65dfa48093f89d/readthedocs/api/v2/serializers.py#L49-L53

flying-sheep:
Hmm, maybe this is a good point for someone to take this over, as

1. I don’t know how things interact here.

   E.g. that method isn’t called anywhere, so where and how does it come into play?

2. Decisions need to be made. 

   E.g. if that code does what I think it does, it results in JSON like this being created:

   ```typescript
   { [name: string]: string }
   ```

   For the better abstraction, we’d need to break the API. Something like

   ```typescript
   {
     [name: string]: {
       value: string,
       visibility: 'private' | 'public',
     },
   }
   ```

   Should that be done? Is the only consumer of the API in the same project? Where?

As said, maybe it’s better someone takes this over, but maybe the explanations are quick …

stsewd:
> So the tests pass for ce7dc1f, but if we want db5d9c7, I need help to make the serializer call it with the correct parameter.

Not sure to understand. I think we want to set it to `public_only=True` by default. The API response is filtered in 
https://github.com/readthedocs/readthedocs.org/blob/db5d9c755b441cadad28ab8ec5d964f741c054c4/readthedocs/projects/models.py#L1338-L1343

We need to add some tests here for public env vars. And, since the new field is nullable, we need to change the form like this

https://github.com/readthedocs/readthedocs.org/blob/a3f3705953427a68f96773ac71d38e3f4b2b35dc/readthedocs/projects/forms.py#L229-L231

stsewd:
Let me know if you need more help here, thanks for taking this :)

readthedocs/rtd_tests/tests/test_api.py
```diff
@@ -707,16 +707,21 @@ def test_init_api_project(self):
         self.assertEqual(api_project.features, [])
         self.assertFalse(api_project.ad_free)
         self.assertTrue(api_project.show_advertising)
-        self.assertEqual(api_project.environment_variables, {})
+        self.assertEqual(api_project.environment_variables(public_only=False), {})
```
humitos, line 710: Would be good to add a test case for `public_only=True` as well.

humitos (commented):
@flying-sheep This is looking good to me! We are close to have it merged :smile: 

Let us know if there is something else you need from our side to wrap up this PR.

readthedocs/rtd_tests/tests/test_api.py
flying-sheep: done!

flying-sheep:
All done!

readthedocs/rtd_tests/tests/test_api.py
```diff
@@ -681,6 +681,7 @@ def test_project_environment_variables(self):
             EnvironmentVariable,
             name='TOKEN',
             value='a1b2c3',
+            public=True,
```
flying-sheep, line 684: This could be a problem: You said I should set `public_only` to `True`, which changes behavior of this API. (As the default serializer calls the method without arguments)

If we want to maintain behavior, the `environment_variables` method needs to be made a property of `Version`, where it has the necessary context to infer what variables are visible for that version, or we need to customize the serializer.

flying-sheep:
Ha, got it! The Admin API now still returns all env variables, by calling the method with `public_only=False`

flying-sheep:
And v3 is done! I hope this is ready now!

stsewd:
We are almost there!

We need to show the privacy level in the list of env vars

Screenshot_2021-03-01 Environment Variables Read the Docs

https://github.com/readthedocs/readthedocs.org/blob/master/readthedocs/templates/projects/environmentvariable_list.html
Something like this using the right and quiet classes https://github.com/readthedocs/readthedocs.org/blob/29c23600d7ef2f2b63d90c900f0d13fff44eaec6/readthedocs/templates/projects/project_dashboard_base.html#L70-L84

And also show this in the detail view

Screenshot_2021-03-01 Environment Variables Read the Docs(1)


https://github.com/readthedocs/readthedocs.org/blob/master/readthedocs/templates/projects/environmentvariable_detail.html

Maybe also allow to edit the privacy level, right now users need to re-create the env var, value can't be changed as isn't show (maybe this can be done in another PR?) /cc @humitos

stsewd:
> OK! Trying to recreate the translations made it touch every single line of the file, so IDK how to do that.

Translations are managed via transifex, so you don't need to update that here.

readthedocs/templates/projects/environmentvariable_detail.html
```diff
@@ -18,9 +18,15 @@
 {% block project_edit_content %}
 
   <p>
-    {% blocktrans trimmed %}
-      The value of the environment variable is not shown here for security purposes.
-    {% endblocktrans %}
+    {% if environmentvariable.public %}
+      {% blocktrans trimmed with value=environmentvariable.value %}
+        Value of the environment variable: {{ value }}
```
stsewd, line 23: I meant to show something like "privacy level: public" but maybe showing the value isn't a bad idea p:

flying-sheep: Yup, that’s what I was thinking. And an overview about which ones are public and which ones are private exists.

humitos:
> Maybe also allow to edit the privacy level, right now users need to re-create the env var, value can't be changed as isn't show (maybe this can be done in another PR?) /cc @humitos

I think it's fine as it is: without showing the value. That's on purpose. I'd say that if you want to change the privacy level, you should create a new environment variable for security reasons.

I'm :+1: showing the privacy level on the list and detail page, tho, but not the value.

flying-sheep:
I’m curious: Why not the value? It’s public after all!

Happy to change however you like it (as long as we get this out soon), I’d just like to know.

readthedocs/templates/projects/environmentvariable_detail.html
stsewd: I think we should probably add a note that the variable would be available in builds from PRs or not

stsewd:
I'm fine showing the value for public vars, but we should use the same style we use in similar parts of the site instead of plain html. And also, we should show a message that this var would be available in builds from PRs or not.

flying-sheep:
OK, I followed the style of the SSH key detail view which is a `<dl/>` with a `<pre/>` for the value.

readthedocs/templates/projects/environmentvariable_detail.html
```diff
@@ -16,12 +16,39 @@
...
+        <p>
+          {% blocktrans trimmed %}
+            The environment variable is available in pull request builds and shown here.
+          {% endblocktrans %}
+        </p>
+      {% else %}
+        <p><strong>{% trans "Private" %}</strong></p>
+        <p>
+          {% blocktrans trimmed %}
+            The environment variable is not available in pull request builds and hidden here.
```
stsewd, line 33:
```suggestion
            The environment variable is not available in pull request builds.
```

readthedocs/templates/projects/environmentvariable_detail.html
```diff
@@ -16,12 +16,39 @@
...
-    {% endblocktrans %}
-  </p>
+  <dl>
+    <dt>{% trans "Privacy Level" %}</dt>
+    <dd>
+      {% if environmentvariable.public %}
+        <p><strong>{% trans "Public" %}</strong></p>
+        <p>
+          {% blocktrans trimmed %}
+            The environment variable is available in pull request builds and shown here.
```
stsewd, line 26: This is redundant with the message below

```suggestion
            The environment variable is available in pull request builds.
```

readthedocs/api/v2/serializers.py
```diff
@@ -31,6 +31,12 @@ class Meta:
         )
 
 
+class AllEnvVarsField(serializers.ReadOnlyField):
+    def get_attribute(self, instance):
+        assert self.source_attrs == ['environment_variables']
+        return instance.environment_variables(public_only=False)
+
```
stsewd, line 38: Not sure if this is needed

readthedocs/api/v2/serializers.py
```diff
@@ -46,9 +52,14 @@ class ProjectAdminSerializer(ProjectSerializer):
         slug_field='feature_id',
     )
 
+    environment_variables = AllEnvVarsField()
```
stsewd, line 55: I think this can be `serializers.SerializerMethodField()`

readthedocs/api/v2/serializers.py
flying-sheep: It is, that’s why painstakingly stepped through the code to add it :laughing: 

It’s `public_only=False` as opposed to the default `public_only=True`, which means that the V2 API keeps its behavior.

readthedocs/api/v2/serializers.py
flying-sheep: Ah, so that’s what `get_environment_variables` is for! Since we don’t want to change the V2 API’s behavior (right?) that means reverting the change to that method.

readthedocs/api/v2/serializers.py
flying-sheep: I see now! Done.

readthedocs/api/v2/serializers.py
stsewd: We need to change it, since it's used by the builders to filter no-public vars. It's a breaking change, but this is used internally only, and hasn't been documented yet, so I think it should be fine to change it. /cc @humitos

readthedocs/api/v2/serializers.py
```diff
@@ -52,14 +46,12 @@ class ProjectAdminSerializer(ProjectSerializer):
...
-    environment_variables = AllEnvVarsField()
+    environment_variables = serializers.SerializerMethodField()
 
     def get_environment_variables(self, obj):
+        """Get all environment variables, including public ones."""
         return {
-            variable.name: dict(
-                value=variable.value,
-                public=variable.public,
-            )
```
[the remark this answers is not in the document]
flying-sheep: Ah, because the environment_variable endpoint for the v2 API isn’t documented for external use? OK.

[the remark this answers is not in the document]
stsewd: yeah, I think we are fine breaking compatibility there

flying-sheep:
All done!

stsewd (commented):
Everything looks good to me. I'm not sure how to improve the UI for the detail view, maybe using a read-only form? This is how it looks like right now

Screenshot_2021-03-09 Environment Variables Read the Docs

readthedocs/templates/projects/environmentvariable_detail.html
```diff
@@ -16,12 +16,39 @@
...
+        <pre>{{ environmentvariable.value }}</pre>
+      {% else %}
+        <p>
+          {% blocktrans trimmed %}
+            The value of the environment variable is not shown here for security purposes.
+          {% endblocktrans %}
+        </p>
+      {% endif %}
+    </dd>
+  </dl>
```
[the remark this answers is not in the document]
flying-sheep: Sorry, but I don’t know how those work in order to make them look nice, and make the form have no associated action and make it accessible for screen readers.

If you want the form, could I please ask you to do it yourself? That’s much quicker than me setting up a running environment for everything.

I added some CSS that should do the job I think. Feel free to revert the last commit and do anything you like for the display.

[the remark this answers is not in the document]
stsewd: Sure, we can take it from here. Thanks a lot for what you did so far!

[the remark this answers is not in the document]
flying-sheep: great, thank you!

stsewd:
Ended up doing this with a readonly form. Not sure about how to handle the value from private env vars. Maybe show `****` instead of an empty value? Or maybe remove the field?

Screenshot_2021-03-10 Environment Variables Read the Docs
Screenshot_2021-03-10 Environment Variables Read the Docs(1)

flying-sheep:
Hello? Could anyone please have a look? My company could really use this.

readthedocs/api/v2/serializers.py
humitos: I'm not sure I follow. Are we just adding `public` field in the API response? If that's the case, that's not a breaking change.

humitos:
> Maybe show `****` instead of an empty value? Or maybe remove the field?

@stsewd I think it's better if we show stars so we keep the message "The value isn't show for security reasons".

readthedocs/projects/forms.py
```diff
@@ -759,12 +759,16 @@ class EnvironmentVariableForm(forms.ModelForm):
...
-        fields = ('name', 'value', 'project')
+        fields = ('name', 'value', 'public', 'project')
 
     def __init__(self, *args, **kwargs):
         self.project = kwargs.pop('project', None)
         super().__init__(*args, **kwargs)
 
+        # Remove the nullable option from the form
+        self.fields['public'].widget = forms.CheckboxInput()
+        self.fields['public'].empty_value = False
```
humitos, line 770: Shouldn't we write a data migration that set `public=False` to all the existing ones instead?

readthedocs/projects/forms.py
```diff
@@ -793,3 +797,19 @@ def clean_name(self):
...
+
+class EnvironmentVariableReadOnlyForm(EnvironmentVariableForm):
+
+    """Make all fields readonly and hide the value from private variables."""
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        if not self.instance.public:
+            self.initial['value'] = ''
```
humitos, line 810:
```suggestion
            self.initial['value'] = '**********'
```

readthedocs/projects/models.py
```diff
@@ -1826,6 +1830,12 @@ class EnvironmentVariable(TimeStampedModel, models.Model):
         on_delete=models.CASCADE,
         help_text=_('Project where this variable will be used'),
     )
+    public = models.BooleanField(
+        _('Public'),
+        default=False,
+        null=True,
```
humitos, line 1836: I think it's better if we write a data migration instead (we can remove the form code as well). This way the data will be valid in our DB.

readthedocs/projects/views/private.py
```diff
@@ -940,7 +941,12 @@ class EnvironmentVariableCreate(EnvironmentVariableMixin, CreateView):
 
 class EnvironmentVariableDetail(EnvironmentVariableMixin, DetailView):
 
-    pass
+    form_class = EnvironmentVariableReadOnlyForm
+
+    def get_context_data(self, **kwargs):
+        form = self.get_form(instance=self.object)
+        context = super().get_context_data(form=form, **kwargs)
+        return context
```
humitos, line 949: Isn't this done automatically by Django when defining `form_class`?

humitos (approved):
I think this is fine.

I'm not sold on showing the value of the environment variable in any case. CircleCI does not show it either. It shows 4 stars and the last 5 chars of the value.

Regarding the UI, we could copy the pattern we have in other places and only have a listing page (without detail page per variable) with the "Delete" button at the right of each element.

The logic is fine to me and I think we can merge this PR. The UI is going to change in the new theme anyways and I think that @agjohnson already built these pages; so worth to ping him here as well.

It seems that some tests are failing. We should take a look at them as well.

readthedocs/api/v2/serializers.py
humitos: Oh, I just saw that we are changing the structure from:

```js
{
 name: value
}
```

to 

```js
{
 name: {
  value: str
  public: bool
 } 
}
```

However, as @stsewd said, I don't think it's super important in this case since it's used internally only.

readthedocs/projects/forms.py
stsewd: yeah, but we need to do the migration in the next deploy so we don't have downtime. There are a lot of these missing a migration, may add a PR migrating all of them

stsewd:
> Regarding the UI, we could copy the pattern we have in other places and only have a listing page (without detail page per variable) with the "Delete" button at the right of each element.

Yeah, that option is good as well, but the value won't be shown, which I think is fine.

stsewd:
I went ahead and deleted the detail view, the delete button is on the listing now.

stsewd:
@humitos let me know what you think now, I can add this to https://github.com/orgs/readthedocs/projects/9 to track the template changes

humitos (approved):
@flying-sheep @stsewd looks good to me! :tada: Thanks! :pray: 

@stsewd Please, create a note in the "New theme" project so we can track this.

stsewd:
@flying-sheep thanks for working on this! This should be live by next Tuesday.

flying-sheep:
Thank you for helping me figure it all out! How does the rollout work for rtd.org vs rtd.com? Do they get updated in unison?

stsewd:
> Thank you for helping me figure it all out! How does the rollout work for rtd.org vs rtd.com? Do they get updated in unison?

yes, we deploy them the same day, first rtd.org and if everything is good we move to rtd.com.

humitos:
This is already deployed and it's working.

PR #7895, opened 2021-02-05
Doc fix: two endpoints had 'pip' for the project_slug

nedbat:
The description of the end point should have `(str:project_slug)`, not "pip"


stsewd (approved):
Thanks!

PR #7899, opened 2021-02-09
Allow HTTPS without a valid certificate when hitting APIv2

humitos:
This is required in the transition period while moving from Azure to AWS. It has to be reverted once we are fully deployed on AWS.

### DIFF
--- a/readthedocs/api/v2/client.py
+++ b/readthedocs/api/v2/client.py
@@ -26,6 +26,7 @@ class DrfJsonSerializer(serialize.JsonSerializer):
 
 def setup_api():
     session = requests.Session()
+    session.verify = False
     if settings.SLUMBER_API_HOST.startswith('https'):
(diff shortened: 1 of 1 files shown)


agjohnson (approved):
I think we should be safe just including this in `rel`, and not doing a full release with this commit. I've set to block so we don't merge this PR by accident. This PR can stick around until after migration so that we remember to cut another hotfix release with this PR reverted.

humitos:
This is not required anymore. Closing.

ericholscher:
@humitos did you unhotfix it?

humitos:
@ericholscher I didn't do it, finally. It seems we still need it for now in production. @agjohnson suggested to use a setting like `SLUMBER_INSECURE` for now while we decide how to fix it properly. I suppose I should open a PR with that change.

PR #7905, opened 2021-02-13
Proxito: Keep storage class at the class level

ericholscher:
Doing storage.url on a new instance does a lot of overhead tasks.

This fixed the huge overhead issue we were having on our webs. Still need to fix tests to merge it, but it's hotfixed for now to keep the servers happy.


ericholscher:
Hotfix removed

Issue #7815, opened 2021-01-08
External builds don't use the same builder as normal builds (defaullt vs large)


astrojuanlu:
## Details

* Read the Docs project URL: https://readthedocs.org/projects/poliastro/
* Build URL (if applicable): https://readthedocs.org/projects/poliastro/builds/12722868/
* Read the Docs username (if applicable): Juanlu001

## Expected Result

If my pull request succeeds:

https://github.com/poliastro/poliastro/pull/1069, https://readthedocs.org/projects/poliastro/builds/12722362/

I'd expect the build to succeeds after merging to `master`.

## Actual Result

`master` is consistently failing. Last build: https://readthedocs.org/projects/poliastro/builds/12722868/

---

At the beginning, we suspected of a particular notebook that might take too much time or RAM to render. However, we did not observe long running times or excessive RAM usage locally. Then, the PR builder succeeded to render that notebook just fine. `latest` builds have been failing for a while now, and we don't know what else to do to debug the issue.

humitos:
My theory here is that nbsphinx opens a ipython kernel and execute some code on it. The kernel consumes more memory than the limit (~4Gb) and there is OOM. Then, the OS kernel kills the ipython kernel. Finally, nbsphinx keeps waiting for a response from ipython kernel and it fails due to timeout.

One thing that I noticed is that this PR build, https://readthedocs.org/projects/poliastro/builds/12722362/, was built in ` buildlarge000JOV` (8Gb RAM) but a normal version build, https://readthedocs.org/projects/poliastro/builds/12735901/, is built in ` builddefault000HPO`. This is _another_ issue because the same project should always build in the same queue (`build:default` or `build:large`)

For now, I put this project into the `build:large` queue to always use VMs with 8Gb. I triggered a build for `latest` version and it passed https://readthedocs.org/projects/poliastro/builds/12737987/

I also left a comment in https://github.com/jupyter/jupyter_client/issues/154#issuecomment-758013354 in case they know more about this.

humitos:
I found another issue with this:

- PR version on build large: https://readthedocs.org/projects/python-docs-es/builds/12750712/
- normal version on build default: https://readthedocs.org/projects/python-docs-es/builds/12750710/

Is our router doing something weird here?

humitos:
I think I found the problem:

[...]

PR #7912, opened 2021-02-15
Router PRs builds to last queue where a build was executed

humitos:
Currently, every time a PR is triggered to be built we are building it under our
build:large queue. This is because we have no data about this version and we
build them on build:large.

This commit checks where the last build was built when a PR is triggered to be
built and use that queue to build the PR.

Closes https://github.com/readthedocs/readthedocs.org/issues/7815


ericholscher (approved):
Makes sense 👍

PR #7925, opened 2021-02-17
Use `admin` user for SLUMBER API on local environment

humitos:
By default on local docker environment we were usign `test` user for SLUMBER
API. However, this user is not created at `--init` and it has to be created
manually.

To avoid this extra step, we are defining `admin` user for SLUMBER which is
created automatically on init.

### DIFF
--- a/readthedocs/settings/docker_compose.py
+++ b/readthedocs/settings/docker_compose.py
@@ -20,6 +20,8 @@ class DockerBaseSettings(CommunityDevSettings):
     PUBLIC_API_URL = f'http://{PRODUCTION_DOMAIN}'
 
     SLUMBER_API_HOST = 'http://web:8000'
+    SLUMBER_USERNAME = 'admin'
+    SLUMBER_PASSWORD = 'admin'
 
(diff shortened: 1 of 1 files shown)


ericholscher:
Is there a strong reason to change the actual login? Seems like it will just break things, when the `test` user seems fine for now?

humitos:
Currently, there is no `test` user; that's my point. We need either create another super user in https://github.com/rtfd/readthedocs-common/blob/5ff15ec196a50ac538632e52eef1763ee0caa045/dockerfiles/scripts/createsuperuser.py#L4 _or_ use the `admin` one that we have already created. This PR does the later one.

In case we want to create `test` as well, it needs to be superuser also, so there is no big difference in using the `admin` one.

humitos:
It's basically the same. However, if we want to have `test` user for builders separated from `admin` user for general usage, I opened the required PR in `common/`: https://github.com/readthedocs/common/pull/83 :smile:

ericholscher:
Ah, I see. I do have an `admin` user, so that should be good then :) I didn't realize we were already creating it.

ericholscher (approved):

PR #7929, opened 2021-02-18
Add a support form to the website

ericholscher:
This will allow us to capture user information on form submission,
verifying that users are who they say.
This will save a few round trips in our support process,
and also be more user-friendly.

## Screenshot

Screen Shot 2021-02-17 at 5 26 04 PM


docs/support.rst
```diff
@@ -14,34 +14,24 @@ Good questions for Stack Overflow would be:
...
+These provide step-by-step solutions to common user requirements.
+
 User Support
 ------------
 
 If you need a specific request for your project or account,
 like more resources, change of the project's slug or username.
-Send an email to support@readthedocs.org.
-We will get to you as soon as possible.
+Please fill out the form at https://readthedocs.org/support/,
```
ericholscher, line 25: Should this be a tab display with links for both Community & Business?

readthedocs/templates/support.html
```diff
@@ -1,76 +0,0 @@
...
-
-{% load i18n %}
-{% load core_tags %}
-
-
-{% block title %}Support{% endblock %}
-
-{% block content %}
-<div id="support_page">
-{% filter restructuredtext %}
```
humitos, line 11: I think this was the only place were this filter was used. Maybe it's time to kill it?

docs/support.rst
humitos: Yeah, that's the pattern we have already. It's not the best, but keeping consistency is better.

readthedocs/templates/support/index.html
```diff
@@ -0,0 +1,105 @@
...
+you can file bug reports on our <a href="https://github.com/readthedocs/readthedocs.org/issues">GitHub issue tracker</a>.
+You can also <a href="http://docs.readthedocs.io/page/contribute.html">contribute</a> to Read the Docs,
+as the code is open source.
+</p>
+
+<h2>Commercial Support</h2>
+
+<p>
+We offer commercial support for Read the Docs
+and commercial hosting with <a href="https://readthedocs.com">Read the Docs for Business</a>,
```
humitos, line 36:
```suggestion
and commercial hosting with <a href="https://readthedocs.com">Read the Docs for Business</a>.
```

readthedocs/templates/support/index.html
```diff
@@ -0,0 +1,105 @@
...
+If you need a specific request for your project or account,
+like more resources, change of the project's slug or username.
+Fill out the form below,
+and we will get back to you as soon as possible.
+</p>
+
+<form
+  method="POST"
+  name="fa-form-1"
+  action="https://webhook.frontapp.com/forms/036c4169294f3b04abaa/yuarKDj5Z5LtxfjgyX2G9hA1pZwV8qPBIe1MFXS-nS8TGNSEBTHXHr0D2nsvNb1JltXQKscj55VbOs1Adbli6c-zFS5-rlkrLZZttjBX68XMZX2sACCUmkiBtI0"
```
humitos, line 52: This could be a setting.

readthedocs/urls.py
```diff
@@ -20,7 +20,13 @@
 
 basic_urls = [
     url(r'^$', HomepageView.as_view(), name='homepage'),
-    url(r'^support/', TemplateView.as_view(template_name='support.html'), name='support'),
+    url(r'^support/success/$',
+        TemplateView.as_view(template_name='support/success.html'),
+        name='support_success'),
+    url(r'^support/error/$',
+        TemplateView.as_view(template_name='support/error.html'),
+        name='support_error'),
```
humitos, line 28: How the users are redirected to these pages? Is this configured in front?

docs/support.rst
humitos: Also, we have plenty of places where we say "contact as <email>" --we probably want to replace them all with the link to this new page.

humitos (approved):
This is good. I'd happy if we have the same in commercial as well.

readthedocs/templates/support/index.html
stsewd: definitely this should be a setting, we don't want to receive support requests from custom installations p:

docs/support.rst
ericholscher: @humitos I wonder if it's possible to create a URL that checks for a session on .com & .org, and redirect the user accordingly? If there is no active session, then we just show a small page with a selection between them. This seems useful if it isn't impossible with cookies.

readthedocs/templates/support/index.html
ericholscher: Aye, good point :)

readthedocs/urls.py
ericholscher: Yep. I will note a comment.

docs/support.rst
agjohnson: We've thrown around a few additional options here too. We could also be doing this client side with javascript and a query param or something -- ie `../en/latest/?source=commercial`. JS could read this an do some version of a CSS class change or something, though i'm not familiar with sphinx-tabs and what option we have here.

Long term, I would love to have an example of session inspection on a third party site -- hitting an external site via JS and using api return in string replacements. I've always imagined this would mostly be a sphinx extension though.

readthedocs/templates/support/index.html
```diff
@@ -0,0 +1,142 @@
...
+        {% endif %}
+            <option value="low">Low</option>
+            <option value="medium">Medium</option>
+            <option value="high">High</option>
+        </select>
+        {% if request.user.gold.exists or request.user.goldonce.exists %}
+        <p> Please rate the severity of this event. </p>
+        {% else %}
+        <p>This option is only enabled for <a href="{% url "gold_subscription" %}">Gold users</a>.</p>
+        {% endif %}
```
agjohnson, line 120: This should be enabled for commercial plans too. Is there a good way to do that here and not require an overridden template?

ericholscher: Yea, I want an abstraction that works for both .org & .com around this. I don't think we have one, and didn't know if it was in scope for this PR :)

I haven't ported this to commercial yet, was planning to do that today and figure out a plan.

humitos (approved):

humitos:
Not super relevant, but we could update our Twitter account to point to this form as well instead of mention github: 

Screenshot_2021-02-23_08-47-50

readthedocs/templates/support/index.html
agjohnson: Good point, yeah we might want some more information on this page in commercial.

I suppose we could also unconditionally ask for the severity level and maybe avoid overriding the template. We wouldn't need to really use this information for anything but data collection, except for upper tier plans where response time varies by severity levels.

PR #7933, opened 2021-02-18
Route external versions to the queue were default version was built

humitos:
When a build for an external version is triggered, we find the the last build
for the default version and use the same queue for this version.

This should make the build to always succeed when it's merged into the default
version since it's using the same build queue.

I thought this was fixed originally in
https://github.com/readthedocs/readthedocs.org/pull/7912/ but that PR was wrong.

### DIFF
--- a/readthedocs/builds/tasks.py
+++ b/readthedocs/builds/tasks.py
@@ -82,6 +82,24 @@ class TaskRouter:
             )
             return project.build_queue
 
+        # Use last queue used by the default version for external versions
+        if version.type == EXTERNAL:
+            last_build_for_default_version = (
+                project.builds
+                .filter(version__slug=project.get_default_version())
+                .order_by('-date')
+                .first()
+            )
+            if 'default' in last_build_for_default_version.builder:
+                routing_queue = self.BUILD_DEFAULT_QUEUE
+            else:
+                routing_queue = self.BUILD_LARGE_QUEUE
+            log.info(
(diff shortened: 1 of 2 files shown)


readthedocs/builds/tasks.py
```diff
@@ -82,6 +82,24 @@ def route_for_task(self, task, args, kwargs, **__):
             )
             return project.build_queue
 
+        # Use last queue used by the default version for external versions
+        if version.type == EXTERNAL:
+            last_build_for_default_version = (
+                project.builds
+                .filter(version__slug=project.get_default_version())
```
ericholscher, line 89: I wonder if we should filter `success=True` here? Though I guess that isn't really a good check, and might miss ones without a successful build on the default project. 

I wonder if what we really want is an attribute on the project that keeps track of what queue it should use? I don't know if there's a "correct" way of handling this for every case though.

ericholscher (approved):
Heh, I reviewed the last one and thought it was correct. This makes sense tho.

readthedocs/builds/tasks.py
humitos: > I wonder if what we really want is an attribute on the project that keeps track of what queue it should use?

I thought about something like this as well but I wasn't sure it was a good idea. We could talk a little more about this and share our thoughts. I think this PR is fine for now, tho.

> I don't know if there's a "correct" way of handling this for every case though.

Yeah, :disappointed:

PR #7948, opened 2021-02-24
Add proxito healthcheck

ericholscher:
This is just a simple health check that confirms 
that proxito is still up and responding to requests.

## Example
Screen Shot 2021-02-23 at 6 38 20 PM

### DIFF
--- a/readthedocs/core/views/__init__.py
+++ b/readthedocs/core/views/__init__.py
@@ -10,7 +10,7 @@ import logging
 from django.conf import settings
 from django.http import Http404, JsonResponse
 from django.shortcuts import render, get_object_or_404, redirect
-from django.views.generic import TemplateView
+from django.views.generic import View, TemplateView
 
 from readthedocs.builds.models import Version
 from readthedocs.core.utils.general import wipe_version_via_slugs
@@ -21,6 +21,11 @@ log = logging.getLogger(__name__)
 
 class NoProjectException(Exception):
     pass
+
+
+class HealthCheckView(View):
+    def get(self, request, *args, **kwargs):
+        return JsonResponse({'status': 200}, status=200)
 
 
 class HomepageView(TemplateView):
--- a/readthedocs/proxito/tests/test_full.py
+++ b/readthedocs/proxito/tests/test_full.py
@@ -36,7 +36,7 @@ class TestFullDocServing(BaseDocServing):
     # Test the full range of possible doc URL's
 
     def test_health_check(self):
-        url = '/_/health_check/'
+        url = reverse('health_check')
         host = 'project.dev.readthedocs.io'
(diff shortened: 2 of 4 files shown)


readthedocs/proxito/tests/test_full.py
```diff
@@ -35,6 +35,13 @@
 class TestFullDocServing(BaseDocServing):
     # Test the full range of possible doc URL's
 
+    def test_health_check(self):
+        url = '/_health_check/'
+        host = 'project.dev.readthedocs.io'
```
agjohnson, line 40: The health check from the ALB comes in without a host header. Doesn't seem like it would be too much difference, but in case the host name could determine the view being used, maybe best to test without

readthedocs/core/views/__init__.py
```diff
@@ -23,6 +23,11 @@ class NoProjectException(Exception):
     pass
 
 
+class HealthCheckView(View):
+    def get(self, request, *args, **kwargs):
+        return JsonResponse({'ok'}, status=200)
```
agjohnson: Probably a dict here instead? I'm assuming it's getting encoded to an array in JSON output

```suggestion
        return JsonResponse({'status': 'ok'}, status=200)
```

readthedocs/proxito/tests/test_full.py
```diff
@@ -35,6 +35,13 @@
...
+        resp = self.client.get(url, HTTP_HOST=host)
+        self.assertEqual(resp.status_code, 200)
+        self.assertEqual(resp.json, {'ok'})
```
agjohnson: Will probably need to update this too. I don't think encoding above will work though, given it's a set.

agjohnson (commented):
Instead of a separate path, we could also put it under `/_/healthcheck` to keep custom stuff in one place

ericholscher:
This has been hotfixed on RTD for Business 👍

readthedocs/proxito/urls.py
```diff
@@ -50,6 +51,13 @@
 
 DOC_PATH_PREFIX = getattr(settings, 'DOC_PATH_PREFIX', '')
 
+health_check_urls = [
+    url('^{DOC_PATH_PREFIX}health_check/$'.format(DOC_PATH_PREFIX=DOC_PATH_PREFIX),
```
humitos, line 55: It seems we have this health check for El Proxito process. Does it worth to have the same health check for Web process as well?

humitos (approved):
I left a suggestion to use `reverse` instead of the hardcoded URL.

readthedocs/proxito/urls.py
ericholscher: @humitos The health checks currently are at the server level, so proxito is all we need since it's the default in the nginx config.

Issue #7806, opened 2021-01-07
Management command to export/load `Project -> RemoteRepository` relations


humitos:
#7791 will help us to re-sync all the VCS data coming from GH, GL and BB. However, it won't re-sync the relations between `Project`s and `RemoteRepository`s.

We will need a command to export these relationships from our old table and import them into the new table. I'm imaging something like this:

```python
# Export query for current production database
Project.objects.filter(remote_repository__isnull=False).values('slug','remote_repository__full_name')  # about 38k records
```

```python
# Load code at the moment of doing the swap to `oauth_*_2020` tables
for project_slug, remote_repository__full_name in relations:
    project = Project.objects.get(slug=project_slug)
    remote_repository = RemoteRepository.objects.get(full_name=remote_repository__full_name)

    project.remote_repository = remote_repository
    project.save()
```

It may happen the `full_name` does not match in case it was updated in the VCS provider and not re-synced in our database. We could use `remote_repository__json__id` instead (or similar depending on the VCS provider) to be 100% sure.

saadmk11:
I like the idea of using `remote_repository__json__id` as this is more robust approach and less likely to have issues. but we need to check if all the `RemoteRepository` objects have data in `json` field with `id`.

The looping might be really slow and put stress on the DB as it will do 3 queries per loop. So, for 38k projects, there will be 100k+ queries. I don't know many different ways to make it faster. But something like this may help a little bit to decrease the query count. @humitos Let me know what you think. :)

```python
for project_slug, remote_repository__full_name in relations:
    subquery = Subquery(
        Project.objects.filter(slug=project_slug).values('id')[:1]
    )
    RemoteRepository.objects.filter(
        full_name=remote_repository__full_name
    ).update(project=subquery)
```

humitos:
> I like the idea of using `remote_repository__json__id` as this is more robust approach and less likely to have issues. but we need to check if all the `RemoteRepository` objects have data in `json` field with `id`.

:+1: --I think they all have the `json` field. We can give this a try, but I think it will be more complex.

> The looping might be really slow and put stress on the DB as it will do 3 queries per iteration

Yeah. The looping was the easiest way to describe what we need here.

Your proposal, will do only 2 queries per iteration, right? One for the `SELECT` (together with the subquery) and another one for the `UPDATE`?

Question, if instead of `project_slug` we save the `project_pk` when exporting can we remove the subquery by doing the following?:

```python
[...]

PR #7957, opened 2021-02-25
Add Management Command to Dump Project and RemoteRepository Relationship in JSON format

saadmk11:
This is one of two PR's for dumping/loading Project and RemoteRepository Relationship data.

This PR adds a management command to dump the project -> RemoteRepository relationship data in a json file. So, that It can be loaded by another command to update the relationship for the new tables.

ref: https://github.com/readthedocs/readthedocs.org/issues/7806

### DIFF
--- a/readthedocs/projects/management/commands/dump_project_remote_repo_relation.py
+++ b/readthedocs/projects/management/commands/dump_project_remote_repo_relation.py
@@ -1,0 +1,45 @@
+import json
+
+from django.core.management.base import BaseCommand
+
+from readthedocs.projects.models import Project
+
+
+class Command(BaseCommand):
+    help = "Dump Project and RemoteRepository Relationship in JSON format"
+
+    def handle(self, *args, **options):
+        data = []
+
+        queryset = Project.objects.filter(
+            remote_repository__isnull=False,
+        ).values_list('id', 'remote_repository__json').distinct()
+
+        for project_id, remote_repository__json in queryset:
+            try:
+                json_data = json.loads(remote_repository__json)
+                # GitHub and GitLab uses `id` and Bitbucket uses `uuid`
+                # for the repository id
+                remote_id = json_data.get('id') or json_data.get('uuid')
+
+                if remote_id:
+                    data.append({
+                        'remote_id': remote_id,
+                        'project_id': project_id
+                    })
+                else:
+                    self.stdout.write(
+                        self.style.ERROR(
+                            f'Project {project_id} does not have a remote_repository remote_id'
+                        )
+                    )
+            except json.decoder.JSONDecodeError:
+                self.stdout.write(
+                    self.style.ERROR(
+                        f'Project {project_id} does not have a valid remote_repository__json'
+                    )
+                )
+
+        # Dump the data to a json file
+        with open('project-remote-repo-dump.json', 'w') as f:
+            f.write(json.dumps(data))
--- a/readthedocs/projects/management/commands/dump_project_remote_repo_relation.py
+++ b/readthedocs/projects/management/commands/dump_project_remote_repo_relation.py
@@ -24,12 +24,13 @@ class Command(BaseCommand):
             username=Subquery(users.values("username")[:1])
         ).values_list(
             'id',
+            'slug',
             'remote_repository__json',
             'remote_repository__html_url',
             'username'
         ).distinct()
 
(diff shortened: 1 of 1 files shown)


readthedocs/projects/management/commands/dump_project_remote_repo_relation.py
```diff
@@ -0,0 +1,45 @@
...
+
+class Command(BaseCommand):
+    help = "Dump Project and RemoteRepository Relationship in JSON format"
+
+    def handle(self, *args, **options):
+        data = []
+
+        queryset = Project.objects.filter(
+            remote_repository__isnull=False,
+        ).values_list('id', 'remote_repository__json').distinct()
```
saadmk11, line 16: We are not able to use the json lookup to just get the `remote_id` in the `values_list()` as we are currently using `TextField` to store the json data. Also, by getting the whole field we have more flexibility to check for inconsistent data at the dump stage so that we have clean and consistent data in the load stage.

readthedocs/projects/management/commands/dump_project_remote_repo_relation.py
```diff
@@ -0,0 +1,45 @@
...
+                json_data = json.loads(remote_repository__json)
+                # GitHub and GitLab uses `id` and Bitbucket uses `uuid`
+                # for the repository id
+                remote_id = json_data.get('id') or json_data.get('uuid')
+
+                if remote_id:
+                    data.append({
+                        'remote_id': remote_id,
+                        'project_id': project_id
+                    })
```
humitos, line 29: Can we save the `RemoteRepository.users.first().username` here as well? This is not _required_ but could help us to re-sync users with linked projects quickly.

readthedocs/projects/management/commands/dump_project_remote_repo_relation.py
```diff
@@ -0,0 +1,45 @@
...
+    help = "Dump Project and RemoteRepository Relationship in JSON format"
+
+    def handle(self, *args, **options):
+        data = []
+
+        queryset = Project.objects.filter(
+            remote_repository__isnull=False,
+        ).values_list('id', 'remote_repository__json').distinct()
+
+        for project_id, remote_repository__json in queryset:
```
humitos, line 18:
```suggestion
        for project_id, remote_repository__json in queryset.iterator():
```

I think we need iterator here because we will probably have a lot of projects.

readthedocs/projects/management/commands/dump_project_remote_repo_relation.py
```diff
@@ -0,0 +1,45 @@
...
+                    )
+            except json.decoder.JSONDecodeError:
+                self.stdout.write(
+                    self.style.ERROR(
+                        f'Project {project_id} does not have a valid remote_repository__json'
+                    )
+                )
+
+        # Dump the data to a json file
+        with open('project-remote-repo-dump.json', 'w') as f:
```
humitos, line 44: Could we use a datetime `%y%m%d-%h%m%s` or similar in the filename?

readthedocs/projects/management/commands/dump_project_remote_repo_relation.py
humitos: This query takes about ~10s, just to COUNT:

```
In [2]: queryset.count()
Out[2]: 39468
```

readthedocs/projects/management/commands/dump_project_remote_repo_relation.py
humitos: Also, let's save the `full_name` and `provider` as well. That will help us to debug things in case something breaks.

humitos (changes requested):
Looks good to me!

I'm requesting some small changes.

humitos:
We need to change the base branch to be `remote-repository-normalization`

saadmk11:
Output JSON file example:

`project-remote-repo-dump-2021-02-25-14:45.json`
```json
[
    {
        "project_id": 2,
        "slug": "test",
        "remote_id": 123456789,
        "html_url": "https://github.com/saadmk11/test",
        "username": "saadmk11"
    },
    {
        "project_id": 4,
        "slug": "django-test",
        "remote_id": 123456789,
        "html_url": "https://github.com/saadmk11/django-todo",
        "username": "saadmk11"
    }
]

```

humitos (approved):
Excellent!

PR #7966, opened 2021-02-28
Add management command to Load Project and RemoteRepository Relationship from JSON file

saadmk11:
This is the second PR for dumping/loading Project and RemoteRepository Relationship data.

Data dump PR: https://github.com/readthedocs/readthedocs.org/pull/7957

This PR adds a management command to load the project -> RemoteRepository relationship data from a json file which is created by #7957.

ref: #7806

### DIFF
--- a/readthedocs/oauth/management/commands/load_project_remote_repo_relation.py
+++ b/readthedocs/oauth/management/commands/load_project_remote_repo_relation.py
@@ -36,9 +36,19 @@ class Command(BaseCommand):
 
         for item in data:
             try:
-                RemoteRepository.objects.filter(
+                update_count = RemoteRepository.objects.filter(
                     remote_id=item['remote_id']
                 ).update(project_id=item['project_id'])
+
+                if update_count < 1:
+                    self.stdout.write(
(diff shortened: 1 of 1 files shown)


readthedocs/oauth/management/commands/load_project_remote_repo_relation.py
```diff
@@ -0,0 +1,50 @@
...
+                    f'Exception: {e}.'
+                )
+            )
+            return
+
+        for item in data:
+            try:
+                RemoteRepository.objects.filter(
+                    remote_id=item['remote_id']
+                ).update(project_id=item['project_id'])
```
humitos, line 41: I think it will be good to log if the updated wasn't done for any reason. Here, if `remote_id` is not found, we won't notice it and we will think that everything worked correctly, even if there is 0 updates.

So, we should check if there was any update and if not, log that the item was not updated.

humitos (changes requested):
Cool! I just left a comment to log when there is no updates. That's the only required change here.

humitos (approved):

PR #7967, opened 2021-03-01
Upgrade `postgres-client` to v12 in Docker image

humitos:
This is required to have a fully compatible client with the new features from
v12, which is the one that we are using on our docker setup, but also in
production.

As a example, the following command fails with the old client:

```
psql (10.15 (Ubuntu 10.15-0ubuntu0.18.04.1), server 12.5 (Debian 12.5-1.pgdg100+1))
WARNING: psql major version 10, server major version 12.
         Some psql features might not work.
Type "help" for help.

docs_db=# \d oauth_remoterepository
ERROR:  column c.relhasoids does not exist
LINE 1: ...riggers, c.relrowsecurity, c.relforcerowsecurity, c.relhasoi...
```

(note the differences between client and server versions)

We will need to update this also in our production images (cc @ericholscher we may need this to run DB queries from `dbshell` for the constraints)

Required by https://github.com/readthedocs/readthedocs.org/pull/7949#issuecomment-787988840

### DIFF
--- a/dockerfiles/Dockerfile
+++ b/dockerfiles/Dockerfile
@@ -20,13 +20,22 @@ RUN apt-get -y install \
         build-essential \
         python3-pip \
         python3-dev \
-        postgresql-client \
         libmysqlclient-dev \
         libfreetype6 \
         libjpeg-dev \
         sqlite \
         netcat \
         telnet
+
+# https://www.postgresql.org/download/linux/ubuntu/
+# Use PostgreSQL client 12 from official PostgreSQL Ubuntu's repository to avoid this WARNING
+# WARNING: psql major version 10, server major version 12.
+# Some psql features might not work.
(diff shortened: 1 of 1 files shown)


stsewd (approved):

ericholscher:
Hrm, not sure the best way to update this in prod. Is it not possible with a PPA from ubuntu, or should we use a similar approach here?

humitos:
@ericholscher I didn't find a PPA from Ubuntu and as this PPA is the official from PostgreSQL itself, I thought we can trust on it. I'd say to follow the same steps than here for now until we upgrade to Ubuntu 20.04 LTS which this won't be required anymore.

humitos:
I'm merging this PR and I will see if I can try to open a new PR for servers. We can discuss the approach there.

PR #7970, opened 2021-03-02
refactor footer, add jobs & status page

ericholscher:
We probably need to change these links up a little,
but this should be good for now.

### DIFF
--- a/readthedocs/templates/base.html
+++ b/readthedocs/templates/base.html
@@ -159,14 +159,11 @@
         </div>
 
         <div class="column-about">
-          <h4>{% trans 'Get Involved' %}</h4>
+          <h4>{% trans 'Resources' %}</h4>
 
           <ul>
             <li>
               <a href="https://docs.readthedocs.io/page/intro/getting-started-with-sphinx.html">{% trans 'Getting Started Guide' %}</a>
-            </li>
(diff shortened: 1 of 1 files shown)


readthedocs/templates/base.html
```diff
@@ -192,8 +195,7 @@ <h4>{% trans 'Business Info' %}</h4>
               <a href="https://readthedocs.com">{% trans 'Read the Docs for Business' %}</a>
             </li>
             <li>
-              {% url "donate" as donate_url %}
-              <a href='{{ donate_url }}'>{% trans 'Supporters' %}</a>
+              <a href='http://status.readthedocs.com/'>{% trans 'Site Status' %}</a>
```
humitos, line 198: Hrm, the HTTPS version is missing a certificate. We should get one, probably.

humitos (approved):

readthedocs/templates/base.html
ericholscher: We can't get one unless we proxy it via something we control, unfortunately. I had it on a domain proxied by Cloudflare which worked, but doesn't for .com

davidfischer (approved):
This references https://github.com/readthedocs/blog/pull/96

PR #7993, opened 2021-03-08
Remove json field from RemoteRepositoryRelation and RemoteOrganizationRelation model

saadmk11:
This PR removes `json` field from RemoteRepositoryRelation and RemoteOrganizationRelation model as the `json` field is no longer required or used in the codebase, we are storing the `remote_id` separately in another field.


readthedocs/oauth/services/gitlab.py
```diff
@@ -55,7 +55,7 @@ def _get_repo_id(self, project):
         # The ID or URL-encoded path of the project
         # https://docs.gitlab.com/ce/api/README.html#namespaced-path-encoding
         try:
-            repo_id = json.loads(project.remote_repository.json).get('id')
+            repo_id = project.remote_repository.remote_id
         except Exception:
```
humitos, line 59: We could probably remove the try/except clause here as well.

humitos (approved):
Great! Thanks a lot!

readthedocs/oauth/services/gitlab.py
saadmk11: if the project was imported manually it would throw an exception as stated here.

https://github.com/readthedocs/readthedocs.org/blob/349ba5a3f83efacb3b7a5c1005d81dc3d7332185/readthedocs/oauth/services/gitlab.py#L60

humitos: He, I didn't read the following line.

However, the previous case would have failed because the `json.loads` would have failed. Now, accessing `remote_id` will return just `None` and won't fail. I think we should check `if project.remote_repository.remote_id is None` instead of the try/except, right?

saadmk11: In this case the `project.remote_repository` wont exist and throw `RemoteRepositroy.DoesNotExist` Exception.

humitos: Good point! Can we catch that exception here then instead of the generic `Exception`?

PR #8001, opened 2021-03-10
Add __str__ to RemoteRepositoryRelation and RemoteOrganizationRelation and Use raw_id_fields in Admin

saadmk11:
ref: https://github.com/orgs/readthedocs/projects/74#card-56545525

### DIFF
--- a/readthedocs/oauth/admin.py
+++ b/readthedocs/oauth/admin.py
@@ -19,6 +19,14 @@ class RemoteRepositoryAdmin(admin.ModelAdmin):
     raw_id_fields = ('users',)
 
 
+class RemoteRepositoryRelationAdmin(admin.ModelAdmin):
+
+    """Admin configuration for the RemoteRepositoryRelation model."""
+
+    raw_id_fields = ('account', 'remote_repository', 'user',)
+    list_select_related = ('remote_repository', 'user',)
+
+
 class RemoteOrganizationAdmin(admin.ModelAdmin):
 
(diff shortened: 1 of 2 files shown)


readthedocs/oauth/admin.py
```diff
@@ -19,14 +19,30 @@ class RemoteRepositoryAdmin(admin.ModelAdmin):
     raw_id_fields = ('users',)
```
humitos, line 19: I think we need to update these `raw_id_fields` as well, because is timing out when accessing to a specific `RemoteRepository`. Not sure which ones we need to add here.

humitos (approved):

readthedocs/oauth/admin.py
```diff
@@ -16,17 +16,26 @@ class RemoteRepositoryAdmin(admin.ModelAdmin):
...
+
+    """Admin configuration for the RemoteOrganizationRelation model."""
+
+    raw_id_fields = ('account', 'remote_organization', 'user',)
+    list_select_related = ('remote_organization', 'user',)
 
 
 admin.site.register(RemoteRepository, RemoteRepositoryAdmin)
-admin.site.register(RemoteRepositoryRelation)
-admin.site.register(RemoteOrganization, RemoteOrganizationAdmin)
```
saadmk11, line 31: Removed `RemoteOrganizationAdmin` because the only attribute it modifies (`raw_id_fields = ('users',)`) is no longer required because of the `through` model.

humitos (approved):

PR #8002, opened 2021-03-10
Stripe Checkout: handle duplicated wehbook

humitos:
If we receive a webhook multiple times, we fetch the GoldUser already created and update its fields.

This PR also removes some old and unused payment code.


readthedocs/gold/models.py
```diff
@@ -52,10 +52,6 @@ class GoldUser(models.Model):
     stripe_id = models.CharField(max_length=255)
     subscribed = models.BooleanField(default=False)
 
-    # TODO: these are managed completely by Stripe now; we could remove them
-    last_4_card_digits = models.CharField(max_length=4, null=True, blank=True)
-    business_vat_id = models.CharField(max_length=128, null=True, blank=True)
```
agjohnson, line 57: I assume we probably want to make sure that the users that did specify a VAT actually have this configured in Stripe on their subscription as well.

readthedocs/payments/forms.py
```diff
@@ -1,234 +0,0 @@
...
-        group = {
-            'is_cc_group': True,
-            'fields': [],
-        }
-        for field in self:
-            if field.name in ['cc_number', 'cc_expiry', 'cc_cvv']:
-                group['fields'].append(field)
-            else:
-                yield field
-        yield group
```
agjohnson, line 234: :wastebasket:

agjohnson (commented):
It seemed the duplicate webhook thing was probably not going to be common, but :+1: on being more defensive with the webhook.

readthedocs/gold/models.py
humitos: Good point! I checked this with a small script:

```python
users = []
for g in GoldUser.objects.exclude(business_vat_id__isnull=True).exclude(business_vat_id=''):
    customer = stripe.Customer.retrieve(g.stripe_id)
    stripe_vat = customer.tax_ids.data[0].value
    if stripe_vat != g.business_vat_id:
        subscriptions = stripe.Subscription.list(customer=g.stripe_id)
        users.append((g.user.username, stripe_vat, g.business_vat_id, len(subscriptions.data)))
```

and I can say that all the users we have `business_vat_id` for has the same tax id on Stripe.

ericholscher (approved):

PR #8016, opened 2021-03-15
Stripe checkout: handle events

humitos:
- register payment failed in our handled events
- return unhandled event if it's not known
- remove one level of indentation

Based on https://github.com/readthedocs/readthedocs.org/pull/8002


readthedocs/gold/views.py
```diff
@@ -221,80 +222,82 @@ class StripeEventView(APIView):
...
-                    (
-                        GoldUser.objects
-                        .filter(stripe_id=stripe_customer)
-                        .update(
-                            level=level,
-                            modified_date=timezone.now(),
-                        )
+            stripe_customer = event.data.object.customer
+
+            if event.type == self.EVENT_CHECKOUT_COMPLETED:
```
humitos, line 234: All this code is the same. I only remove one level of indentation.

ericholscher (approved):

PR #8017, opened 2021-03-15
RemoteRepository: Improvements to `sync_vcs_data.py` script

humitos:
* Add `--logged-in-days-ago` argument to resync only users who logged in some days ago
* Add `--skip-revoked-users` argument to avoid resyncing users we already known they revoked our access token

### DIFF
--- a/readthedocs/oauth/management/commands/sync_vcs_data.py
+++ b/readthedocs/oauth/management/commands/sync_vcs_data.py
@@ -78,7 +78,7 @@ class Command(BaseCommand):
             socialaccount__isnull=False
         ).distinct()
 
-        if logged_in_days_ago:
+        if logged_in_days_ago > 0:
             users = users.filter(
                 last_login__gte=timezone.now() - datetime.timedelta(days=logged_in_days_ago),
             )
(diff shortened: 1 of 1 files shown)


readthedocs/oauth/management/commands/sync_vcs_data.py
```diff
@@ -60,6 +78,11 @@ def handle(self, *args, **options):
             socialaccount__isnull=False
         ).distinct()
 
+        if logged_in_days_ago:
```
saadmk11, line 81:
```suggestion
        if logged_in_days_ago > 0:
```

Maybe we should check `if logged_in_days_ago > 0` ? As we have a default for this argument, and if it's `0` or Not provided, it will just filter `last_login__gte=timezone.now()` which I think we don't want. :)

saadmk11 (approved):
Looks Great. I really like the way you are using sentry's data to filter revoked users :100:

readthedocs/oauth/management/commands/sync_vcs_data.py
humitos: Well, if it's 0 (or not provided) it will do `if 0:` and won't enter to the if.

I'm fine adding it explicitly, tho.

saadmk11: Oops! My bad :)

PR #8031, opened 2021-03-18
Make Build models default to `triggered`

ericholscher:
I'm not sure why we were defaulting them to Finished,
but this should help not confuse the badge code,
and just be more correct in general.


humitos (approved):

stsewd (commented):
Looks like tests need to be updated, this also discovered a bug (just a missing none check) :D

btw, we already change the value to triggered https://github.com/readthedocs/readthedocs.org/blob/655a34a4b4189fcbd4180f599190fb450d93fef9/readthedocs/core/utils/__init__.py#L118-L125 but better to have it by default.

ericholscher:
@humitos feel free to finish it up if you have bandwidth. I don't think I'll have time for a little while, unfortunately.

humitos:
I'm adding this PR to the Roadmap Q3 so we can finish it. It would be good to have our data being consistent.

humitos:
@stsewd would you like (if you have some time 😄 ) to take a look at this PR and update it so we can move forward and merge it? 🙏🏼

stsewd (approved):
Did a quick test locally, I didn't see anything weird.

ericholscher (commented):
👍 Can't approve this since I originally opened it, but looks great. Thanks for pushing this to done!

PR #8062, opened 2021-03-31
Sign Up: limit the providers allowed to sign up

humitos:
If `signup_providers` is passed to `account/signup.html` template we only show
the "Sign up with ..." button for those providers.

This is useful to send invitation links for organizations using Single Sign-On
to avoid unauthed access because they sign up with a different provider.


ericholscher (approved):

humitos:
We are still having a conversation in -corporate and we are not yet ready to merge this without the PR in .com

readthedocs/templates/account/signup.html
```diff
@@ -13,6 +13,22 @@ <h1>{% trans "Sign Up" %}</h1>
   <small>{% blocktrans %}Already have an account? Then please <a href="{{ login_url }}">sign in</a>.{% endblocktrans %}</small>
 </p>
 
+{% if signup_providers %}
+<p>
+    You have been invited to join
```
agjohnson, line 18: This should be a localized string.

readthedocs/templates/account/signup.html
```diff
@@ -13,6 +13,22 @@ <h1>{% trans "Sign Up" %}</h1>
...
+    {% if organization %}
+    the organization "{{ organization.name }}"
+    {% else %}
+    an organization
+    {% endif %}
+    at Read the Docs.
+</p>
+<p>
+    Sign up using the following provider.
+    If you don't have an account on this provider,
```
agjohnson, line 28:
```suggestion
    If you don't have an account with this provider,
```

PR #8080, opened 2021-04-08
Add page and page_size to server side api documentation

cocobennett:
The `page` and `page_size` query parameters were missing from the Server Side API documentation, so this PR should fix https://github.com/readthedocs/readthedocs.org/issues/8006


stsewd (approved):
Thanks!

PR #8106, opened 2021-04-14
Add proposal for new Sphinx and RTD tutorials

astrojuanlu:
After reflecting a lot on [Daniele Procida's thoughts on documentation](https://diataxis.fr), gathering information from several existing Sphinx tutorials and drawing inspiration from documentation of other projects, I tried to conceive the tutorial as a syllabus, a learning path that we propose to the reader, that should be ordered and structured in a meaningful way. In that vein, it starts with the creation of a small Python project, setting up a virtual environment, and using `git`. I left this last bit as optional, but I actually think it should be used all throughout the tutorial. Would love to know your thoughts on that.

We intend these new guides to be useful for the broad Python community, without forgetting about scientific users and their needs. In particular, I'd like to write most of the material in Markdown rather than reStructuredText, and pay special attention to the integration of Jupyter notebooks.

I devoted more effort to break out the first sections into smaller steps up front, so the intent and the overall philosophy are more clear to those reviewing, and left the final sections more open ended. Besides, I also proposed several topics we could cover as How-to guides, especially for people that already know the basics of Sphinx but want to address particular problems.

Rendered version: https://docs--8106.org.readthedocs.build/en/8106/development/design/new-sphinx-guides.html


docs/development/design/new-sphinx-guides.rst
```diff
@@ -0,0 +1,186 @@
...
+         do this every time we change something)
+
+   5. Simplify documentation building by using Make
+
+      -  Install Make (and its Windows counterpart)
+      -  Make another trivial change
+      -  Build HTML doing ``cd doc && make html``
+      -  Observe that the HTML docs have changed (yay!)
+
+   6. Converting our documentation to PDF
```
ericholscher, line 90: This seems almost impossible to do well. I don't think we should include this in the tutorial. I almost never created PDFs locally, and we shouldn't expect users to manage this kind of task.

docs/development/design/new-sphinx-guides.rst
```diff
@@ -0,0 +1,186 @@
...
+3. Customizing Sphinx configuration
+
+   1. Changing the HTML theme
+
+      -  Install https://pypi.org/project/furo/
+      -  Change the ``html_theme`` in ``conf.py``
+      -  Rebuild the HTML documentation and observe that the theme has
+         changed (yay!)
+
+   2. Changing the PDF appearance
```
ericholscher, line 108: If we want another format, we should do something with a ton less dependencies, like a man page or epub. I'm not sure we want to dive so deeply into customizing anything other than HTML though, based on what I've seen users care about.

docs/development/design/new-sphinx-guides.rst
```diff
@@ -0,0 +1,186 @@
...
+
+      -  Activate our virtual environment
+      -  Install Sphinx inside the virtual environment
+      -  Check that ``sphinx-build --help`` works (yay!)
+
+   2. Creating the project layout
+
+      -  Run the sphinx-quickstart wizard: separate source and build
+         (``y``), project name, author name, and go
+      -  Check that the correct files are created (yay!)
```
ericholscher, line 64: I feel like we probably need an explainer on how sphinx works and how to think about all the various files. Basically a longer version of this: https://sphinx-tutorial.readthedocs.io/start/#sphinx-philosophy

Similarly, do we want to introduce people to basic markdown? I'd be much more 👍  on explaining markdown to folks (and how Myst markdown is different) -- but we could also point folks to the Myst docs for that (https://myst-parser.readthedocs.io/en/latest/using/intro.html).  Perhaps a hoverxref of their page? :)

Should we start from a cookiecutter template instead of sphinx-quickstart? quickstart is kinda overwhelming.

docs/development/design/new-sphinx-guides.rst
```diff
@@ -0,0 +1,186 @@
...
+
+5. Using Jupyter notebooks inside Sphinx
+6. Describing code in Sphinx
+
+   -  Explain the Python domain as part of narrative documentation to
+      interleave code with text, include doctests, and justify the
+      usefulness of the next section.
+
+7. Autogenerating documentation from code in Sphinx
+
```
ericholscher, line 140: We should include an RTD section here, or at least a pointer to an RTD continuation of the tutorial? It seems like this would be the natural end-point with deploying.

docs/development/design/new-sphinx-guides.rst
```diff
@@ -0,0 +1,186 @@
...
+.. note 
+
+   - Looks like MathJax is enabled by default now? Can't see a reference in the docstrings
+   - The fact that the ``index.rst`` created by ``sphinx-quickstart`` is written in
+     reStructuredText gets in the way of teaching Markdown (MyST).
+     On the other hand, users need to know reST anyway to write Python docstrings,
+     so it is necessary to teach both.
+
+How-to Guides
```
ericholscher, line 149: I don't quite understand what this section is meant to include? Is this just additional ideas for our existing guide section? 

https://docs.readthedocs.io/en/latest/guides/

docs/development/design/new-sphinx-guides.rst
```diff
@@ -0,0 +1,186 @@
+Proposed contents for new Sphinx guides
+=======================================
+
+This classification follows the `Diátaxis Framework <https://diataxis.fr/>`__.
```
ericholscher, line 4: I feel like we basically just want a tutorial here. I don't think we need to think about this project as having 4 sections, but more about building a tutorial for Sphinx & RTD. This would be fulfilling the "Tutorial" value for these projects. 

Are we thinking about expanding this work into a refactor of the entire RTD or Sphinx docs? That's where I could see the framework coming into play more.

docs/development/design/new-sphinx-guides.rst
```diff
@@ -0,0 +1,186 @@
...
+   **Objective**: Write a tutorial, some examples, and reference
+   documentation for our Python code.
+
+Every step should lead to a “yay!” moment, which marks the end of the
+section. This keeps the learner motivated.
+
+Appendixes are optional, i.e. not required to follow the tutorial, but
+highly recommended.
+
+1. Getting started
```
ericholscher, line 18: This seems useful, but I don't know if we want to have so much introductory content in our tutorial. This seems like something we can link out to existing docs/tutorials for. I feel like we want to maintain content focused on Sphinx, and external sources will be better canonical references for virtualenv, adding dependencies, etc. 

We could consider including some of this, but I'd want to do this last if we have user requests, instead of first.

docs/development/design/new-sphinx-guides.rst
```diff
@@ -0,0 +1,186 @@
...
+   6. Converting our documentation to PDF
+
+      -  Install TeX Live to create PDF output (should work on every
+         operative system)
+      -  Build LaTeX using ``make latexpdf``
+      -  See that the PDF appeared (yay!)
+
+   7. Appendix: PDF without LaTeX using rinoh (beta)
+
+3. Customizing Sphinx configuration
```
ericholscher, line 99: This section is great 👍

docs/development/design/new-sphinx-guides.rst
```diff
@@ -0,0 +1,186 @@
...
+
+      -  Add a string to the ``extensions`` list in ``conf.py`` for
+         ``sphinx.ext.duration``
+      -  Rebuild the HTML docs ``make html`` and notice that now the
+         times are printed (yay!)
+
+4. Writing narrative documentation with Sphinx
+
+   -  First focus on ``index.rst``, gently introducing the learner to
+      reST and mentioning Semantic Line Breaks.
```
ericholscher, line 124: I think we probably just want to introduce markdown here, if that's the plan, and never do much of an introduction to rst.

docs/development/design/new-sphinx-guides.rst
```diff
@@ -0,0 +1,186 @@
...
+   -  Then add another ``.rst`` file to teach how ``toctree`` works.
+   -  Then enable Markdown adding ``myst_parser`` to the extensions and
+      authoring a Markdown file, to show that both can exist at the same
+      time.
+   -  Then continue introducing elements of the syntax to add pictures,
+      cross-references, and the like.
```
ericholscher, line 130: This is a huge thing, and could possibly become a larger sphinx/md reference. Myst has a large reference here, but it's not very beginning friendly: https://myst-parser.readthedocs.io/en/latest/using/syntax.html#block-tokens-summary

Need to figure out how to combine these various approaches though -- perhaps we contribute more examples to their repos and point users there?

docs/development/design/new-sphinx-guides.rst
```diff
@@ -0,0 +1,186 @@
...
+
+5. Using Jupyter notebooks inside Sphinx
```
ericholscher, line 132: This seems important 👍

ericholscher (commented):
This is a great start, but I think we definitely need more thought around what we want to focus on and audience of this content. I think we definitely want to think through where external resources can be used here, and what we want to "own" in terms of our content. We will be responsible for keeping this information up to date, so we definitely want to keep the scope smaller to start, and link out to externally maintained tutorials where possible.

I think we also need to think about where this content will live. Are we doing to add it to the RTD docs? Sphinx docs? Both? Neither? That will help scope what we want to handle.

I think the biggest goals for me are:

* Building a tutorial for Sphinx, which will likely live in their docs
  * Introducing people to all the various Sphinx features in order of value/complexity, in a narrative format
* Building a tutorial for RTD, which will live in our docs
  * This is similar to our guides, but in an overarching narrative, perhaps continued from the Sphinx tutorial? 

But we need to discuss a bit more in depth the overall goals of this work I think. I don't have a full understanding of how it should all fit together, but it's definitely something to think through.

docs/development/design/new-sphinx-guides.rst
astrojuanlu: You think? I still did not explore how complicated it is to install TeX Live on Windows and macOS, but on Linux it was fairly straightforward. It required a number of dependencies, that's true.

docs/development/design/new-sphinx-guides.rst
astrojuanlu: Yes, the intention was to introduce a second format, so it's clear that it can be done and also where does the configuration go.

In line with my previous comment, I still think PDF is valuable because it's a format everyone is familiar with, and it can be opened either with a web browser or with any PDF viewer. Also, I've seen comments of people turning down other documentations systems because they lacked PDF export, so this is perhaps one of Sphinx strengths, even if HTML is used more. According to our Google Analytics, there is a small but nonzero amount of people downloading PDFs from Read the Docs at least.

As a second option, https://github.com/brechtm/rinohtype supports PDF generation without *TeX - I tested it and works nicely. However, the README says that it's still beta.

docs/development/design/new-sphinx-guides.rst
astrojuanlu: > I feel like we probably need an explainer on how sphinx works and how to think about all the various files. Basically a longer version of this: https://sphinx-tutorial.readthedocs.io/start/#sphinx-philosophy

Yes, in line with the Diátaxis Framework I considered this part of the "Explanations" (see below in the document).

> Similarly, do we want to introduce people to basic markdown? I'd be much more +1 on explaining markdown to folks (and how Myst markdown is different) -- but we could also point folks to the Myst docs for that (https://myst-parser.readthedocs.io/en/latest/using/intro.html). Perhaps a hoverxref of their page? :)

I think (and this applies to other comments as well), we should strike a balance between reinventing the wheel and adding too much content about things that are already explained elsewhere (because, well, they are already explained elsewhere) having too much indirection and pointing the learner to external resources too often, because this would reduce consistency and harm the learning experience, I believe. See for example what Astropy does with virtual environments:

https://docs.astropy.org/en/stable/development/workflow/virtual_pythons.html#virtual-envs

This is perhaps a bit too much for my taste, but I am not very happy with the other extreme, i.e. telling the user "before you continue, read this <here>" and not having any control over that material.

> Should we start from a cookiecutter template instead of sphinx-quickstart? quickstart is kinda overwhelming.

I also considered this, and I was inclined to use `sphinx-quickstart` instead. Years ago it asked you about every other plugin and it was quite intimidating as you say, but these days it only asks about five questions.

docs/development/design/new-sphinx-guides.rst
astrojuanlu: Exactly, they are additional ideas to what we have in https://docs.readthedocs.io/en/latest/guides/tools.html. Our guides there are "intermediate level" already I think, the ones I presented are probably more basic.

docs/development/design/new-sphinx-guides.rst
astrojuanlu: It helped me conceive this as a separate site with its own structure, but after reading

> Are we thinking about expanding this work into a refactor of the entire RTD or Sphinx docs? That's where I could see the framework coming into play more.

I realized that it might be too ambitious. The simplest path would be then:

- Upstream the tutorial to Sphinx
- Write a tutorial on Read the Docs (which I hadn't start working on yet)
- Expand our guides (see my comment above), with the option of upstreaming them.

And talking about upstreaming, I did not reach out yet to Sphinx developers about this. Will do when we are more settled.

docs/development/design/new-sphinx-guides.rst
astrojuanlu: Absolutely!

docs/development/design/new-sphinx-guides.rst
astrojuanlu: I was also hesitant to introduce it, but [as I said above](https://github.com/readthedocs/readthedocs.org/pull/8106#discussion_r614136579), having a short preparation like this would make the tutorial more consistent, I think. Otherwise, it's might end up being a series of random sentences, "lorem ipsum"-like text, or stock photos, which is what many other Sphinx tutorials do.

For example, Brandon tutorial https://brandons-sphinx-tutorial.readthedocs.io/en/latest/ creates a fictitious library called `trianglelib`. It omits lots of details because it's meant for a 3-hour PyCon tutorial, but we can afford to do much more.

docs/development/design/new-sphinx-guides.rst
astrojuanlu: Yeah, I am a bit conflicted. From the notes in the document:

> The fact that the ``index.rst`` created by ``sphinx-quickstart`` is written in reStructuredText gets in the way of teaching Markdown (MyST). On the other hand, users need to know reST anyway to write Python docstrings, so it is necessary to teach both.

In other words: I feel bad ignoring the `index.rst` that is created by the quickstart, and on the other hand reST is needed for docstrings anyway. Sure, we don't need to do a full reST introduction (see comments above), but even though I'd love to focus only on Markdown/MyST, perhaps it's not realistic for Python projects.

...Unless we don't have a Python library in mind, and this is geared towards narrative documentation only. But I think that would be too limiting.

docs/development/design/new-sphinx-guides.rst
astrojuanlu: These sections are not very polished, I don't have a strong opinion yet.

I wonder which of these strategies could work better:

- We work somewhat independently, and upstream small bits as we go, with the option of moving the docs altogether if the upstream projects like them.
- We work from the beginning with the upstream projects.

I see pros and cons in both, and also I think the strategy in part constraints what we can or can't do. But I lean towards the former.

docs/development/design/new-sphinx-guides.rst
astrojuanlu: Fair enough - I will think of a way of making this optional, but very visible at the same time. Scientific users will appreciate this a lot.

docs/development/design/new-sphinx-guides.rst
astrojuanlu: Just in case, I bit the bullet and tried to build an EPUB locally. The multi-file output was confusing at first, but `pdflatex` has the same issue. I viewed the `.epub` result with https://calibre-ebook.com/ and it worked.

man pages are Linux/UNIX-specific I guess, so I would like to use something more generic. I would reluctantly use EPUB instead of PDF as a secondary format, as long as we keep it at least as an optional section.

astrojuanlu:
About

> I think we definitely want to think through where external resources can be used here, and what we want to "own" in terms of our content. We will be responsible for keeping this information up to date, so we definitely want to keep the scope smaller to start, and link out to externally maintained tutorials where possible.

This applies mainly to virtual environments and version control. We agreed to:

- Leave version control out (too confusing, even though it's a recommended good practice - and git in particular is known to be tough for beginners)
- Keep virtual environments as part of the workflow, but don't delve into details and send users to external resources instead

astrojuanlu:
About the mismatch of the markup we want to use (Markdown/MyST) and `sphinx-quickstart` producing a `index.rst` file by default:

Proposing a flag to `sphinx-quickstart` to produce a `index.myst` file instead might be contentious, in that it would probably require Sphinx to depend on MyST-Parser. Instead of going that way, we agreed to use a cookiecutter. More details to follow in the document.

docs/development/design/new-sphinx-guides.rst
astrojuanlu: - Added a section at the beginning give more background on Sphinx, Markdown, and MyST
- Changed proposal to use a cookiecutter (we would need to create it)
- Removed mentions to reST in favor of Markdown

docs/development/design/new-sphinx-guides.rst
astrojuanlu: Added section on deployment, deferring to the actual RTD tutorial which I still have to propose a TOC for.

docs/development/design/new-sphinx-guides.rst
astrojuanlu: Rephrased this as ideas for new how-to guides covering use cases detected during the interviews.

docs/development/design/new-sphinx-guides.rst
astrojuanlu: Greatly simplified the introduction.

docs/development/design/new-sphinx-guides.rst
astrojuanlu: Switched to Markdown (by avoiding the `index.rst` created by `sphinx-quickstart`)

docs/development/design/new-sphinx-guides.rst
astrojuanlu: Might be too early to decide course of action here, we should in any case have a pedagogical section with the important parts that complements the list of tokens in the MyST docs.

docs/development/design/new-sphinx-guides.rst
astrojuanlu: Moved this to the last appendix.

astrojuanlu:
- Addressed most of the comments (replied inline to those)
- Left unresolved the conversation about PDFs (I still think they are important and useful)
- Missing: proposal for Read the Docs tutorial

docs/development/design/new-sphinx-guides.rst
ericholscher: It adds multiple GB of downloads, at least to do it at the RTD level. It also has a lot of issues with non-english languages, terrible error output, and lots of other drawbacks. I'm definitely not convinced it's a value-add to any tutorial.

What is the goal of showing additional output formats? In my mind, it shows the value of writing in RST and then outputting in another format. If that's the goal, lets just do text output or something.

docs/development/design/new-sphinx-guides.rst
```diff
@@ -0,0 +1,175 @@
...
+      -  Rebuild the HTML docs ``make html`` and notice that now the
+         times are printed (yay!)
+
+5. Writing narrative documentation with Sphinx
+
+   -  First focus on ``index.md``, diving more into Markdown
+      and mentioning Semantic Line Breaks.
+   -  Then add another ``.md`` file to teach how ``toctree`` works.
+   -  Then continue introducing elements of the syntax to add pictures,
+      cross-references, and the like.
```
ericholscher, line 120: 💯

docs/development/design/new-sphinx-guides.rst
```diff
@@ -0,0 +1,175 @@
...
+   -  Explain the Python domain as part of narrative documentation to
+      interleave code with text, include doctests, and justify the
+      usefulness of the next section.
+
+7. Autogenerating documentation from code in Sphinx
+8. Deploying a Sphinx project online
+
+   - A bit of background on the options: GitHub/GitLab Pages,
+     custom server, Netlify, Read the Docs
+   - Make reference to Read the Docs tutorial
```
ericholscher, line 133: 👍

ericholscher (commented):
This is feeling a lot closer. Excited to bring it up with other folks in the community, so we should discuss that plan for that tomorrow. 👍

docs/development/design/new-sphinx-guides.rst
```diff
@@ -1,102 +1,95 @@
...
       -  Observe that the HTML docs have changed (yay!)
 
-   6. Converting our documentation to PDF
+   3. Converting our documentation to PDF
 
-      -  Install TeX Live to create PDF output (should work on every
-         operative system)
+      -  Install TeX (MiKTeX looks very straightforward on Windows, TeX Live
+         is an alternative)
       -  Build LaTeX using ``make latexpdf``
```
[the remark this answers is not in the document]
astrojuanlu: 😮 TIL about these... Good enough to showcase the different formats 👍🏽

astrojuanlu:
The Sphinx part is essentially there. I uploaded the RTD tutorial. In the same spirit, I added more detail to the initial chapters and left the final ones more open (also, because I still have to figure out many things, like translations!)

humitos:
> because I still have to figure out many things, like translations

Let me know when you are ready to jump into this, and I can share what I know from `python-docs-es`.

docs/development/design/new-sphinx-guides.rst
```diff
@@ -17,19 +17,19 @@ Sphinx tutorial
 Appendixes are optional, i.e. not required to follow the tutorial, but
 highly recommended.
 
-0. The Sphinx way
+1. The Sphinx way
```
ericholscher, line 20:
```suggestion
#. The Sphinx way
```

Believe you can auto-number them ;)

docs/development/design/new-sphinx-guides.rst
```diff
@@ -73,23 +73,24 @@ highly recommended.
       -  Open http://localhost:8000 in a web browser, and see the HTML
          documentation (yay!)
 
-   2. Simplify documentation building by using Make [3]_
+   2. Converting our documentation to other formats
+
+      -  Install TeX (MiKTeX looks very straightforward on Windows, TeX Live
```
ericholscher, line 78:
```suggestion
```

docs/development/design/new-sphinx-guides.rst
```diff
@@ -73,23 +73,24 @@ highly recommended.
...
+         is an alternative)
```
ericholscher, line 79:
```suggestion
```

docs/development/design/new-sphinx-guides.rst
```diff
@@ -73,23 +73,24 @@ highly recommended.
...
+      -  Build PseudoXML using ``make pseudoxml``
+      - Build Text using ``make text``
+      -  See that the PDF appeared (yay!)
```
ericholscher, line 82:
```suggestion
      -  See how the various formats change the output (yay!)
```

docs/development/design/new-sphinx-guides.rst
```diff
@@ -73,23 +73,24 @@ highly recommended.
       -  Open http://localhost:8000 in a web browser, and see the HTML
          documentation (yay!)
 
-   2. Simplify documentation building by using Make [3]_
+   2. Converting our documentation to other formats
+
+      -  Install TeX (MiKTeX looks very straightforward on Windows, TeX Live
+         is an alternative)
+      -  Build PseudoXML using ``make pseudoxml``
```
ericholscher, line 80: This is a particularly neat one, because it shows the docutils AST. Could be worth diving into in some more depth in an appendix.

docs/development/design/new-sphinx-guides.rst
```diff
@@ -111,33 +112,110 @@ highly recommended.
...
+     the development documentation.
 
 .. note 
 
    - Looks like MathJax is enabled by default now? Can't see a reference in the docstrings
 
+Read the Docs tutorial
+----------------------
+
+1. What is Read the Docs
```
ericholscher, line 151: Not "The Read the Docs Way"? :D

docs/development/design/new-sphinx-guides.rst
```diff
@@ -111,33 +112,110 @@ highly recommended.
...
+2. Getting started
+
+   1. Preparing our project on GitHub
+
+      - Fork a starter GitHub repository (notice that
+        even an empty project is ready for build the first time without changes,
+        resulting in a nice "welcome to read the docs" dummy page,
+        but using something like `our demo template
```
ericholscher, line 159: I think we plan to get rid of the dummy page, so probably should not include this :)

docs/development/design/new-sphinx-guides.rst
```diff
@@ -111,33 +112,110 @@ highly recommended.
...
+        <https://github.com/readthedocs/template>`_,
+        as a starting point
+        helps mimicking the `sphinx-quickstart` or `cookiecutter` step
+        without having to checkout the code locally)
+
+   2. Importing our project to Read the Docs
+
+      - Sign up on RTD
+      - Connect your GitHub account (is this step necessary
+        if the user signed up with GitHub?)
```
ericholscher, line 169: Nope -- we should definitely do "sign up with GH" as this step.

docs/development/design/new-sphinx-guides.rst
```diff
@@ -111,33 +112,110 @@ highly recommended.
...
+      - Connect your GitHub account (to avoid manual import)
+      - Import the project (don't "Edit advanced project options", we
+        will do this later)
+      - The project is created on RTD
+      - Browse "builds", wait a couple of minutes, open the docs (yay!)
```
ericholscher, line 174: You can watch the build output :)

docs/development/design/new-sphinx-guides.rst
```diff
@@ -111,33 +112,110 @@ highly recommended.
...
+
+   3. Basic configuration changes
+
+      - Add a description, homepage, and tags
```
ericholscher, line 178: Not sure if this is even useful, other than tags.

docs/development/design/new-sphinx-guides.rst
```diff
@@ -111,33 +112,110 @@ highly recommended.
...
+      - Browse other project with similar tags as yours
+        (this way we encourage tagging projects, as well as discovery)
```
ericholscher, line 180: This is not a great experience, unfortunately. Not sure if we want to promote it, unless we curate the tags they use -- but that will just show a bunch of test projects haha

docs/development/design/new-sphinx-guides.rst
```diff
@@ -111,33 +112,110 @@ highly recommended.
...
+
+   1. Rendering our documentation from pull requests
+
+      - Enable "build pull requests for this project" in the advanced settings
+      - Edit a file from the GitHub UI as part of a new branch, and open a pull request
+      - See the RTD check on the GitHub PR UI, wait a few minutes, open result (yay!)
+
+   2. Using custom domains
+
+      - Essentially the steps already described in our Custom Domains reference
```
ericholscher, line 194: This is probably a bit advanced for a tutorial (DNS editing!). Maybe an appendix, or just a mention pointing at the guide?

docs/development/design/new-sphinx-guides.rst
```diff
@@ -111,33 +112,110 @@ highly recommended.
...
+
+4. Versioning documentation
+
+   - Explain how to manage versions on RTD: create release branches,
+     activate the corresponding version, browse them in the version selector,
+     selectively build versions
+   - Intermediate topics: hide versions, create Automation Rules
```
ericholscher, line 201: 👍  -- Automation rules is probably an advanded/appendix topic.

docs/development/design/new-sphinx-guides.rst
```diff
@@ -111,33 +112,110 @@ highly recommended.
...
+
+5. Customizing the build process
+
+   - Use `.readthedocs.yaml` (rather than the web UI) to customize build formats,
+     change build requirements and Python version, enable fail-on-warnings
```
ericholscher, line 206: I'd probably pull this earlier? We generally want to promote config via YAML over the web UI. 

Need to solve https://github.com/readthedocs/readthedocs.org/issues/4221

docs/development/design/new-sphinx-guides.rst
```diff
@@ -111,33 +112,110 @@ highly recommended.
...
+
+6. Getting insights from your projects
+
+   - Move around the project, explore results in Traffic Analytics
+   - Play around with server-side search, explore results in Search Analytics
+
+7. Managing translations
```
ericholscher, line 213: I probably wouldn't even include this in the tutorial, tbh -- it's definitely super advanced and not relevant for most users (unfortunately)

ericholscher (commented):
Looks great. I think the RTD tutorial is missing a couple things:

* Show simple edits building docs via webhook (early in the process()
* Alternative outputs (Look how simple PDF is on RTD!)

docs/development/design/new-sphinx-guides.rst
astrojuanlu: 😃

docs/development/design/new-sphinx-guides.rst
astrojuanlu: Fair enough!

docs/development/design/new-sphinx-guides.rst
astrojuanlu: Yeah, my idea was to use that as an excuse to add the tags. I recall discussing with someone from the team that these could be useful for ads - but now that we have the keyword detection, I'm not sure if that's still necessary.

I also admit that this was the first time I realized I could browse other projects with the same tags. But if we don't want to promote those, should we get rid of them? (That's a separate discussion of course)

In any case, removing this from here

docs/development/design/new-sphinx-guides.rst
astrojuanlu: Cool! Added an appendix about it 👍🏽

docs/development/design/new-sphinx-guides.rst
astrojuanlu: I do have a simple use case in mind - automating something I've been doing manually in poliastro for years 😄 but I'd also want to think through @humitos ideas here https://github.com/readthedocs/blog/pull/74

ericholscher:
* https://numpy.org/numpy-tutorials/
* https://gitpod.io

Binder for tutorial environments, gitpod for a "devy" type environment.

astrojuanlu:
Comments from https://github.com/sphinx-doc/sphinx/issues/9165 (will keep updating):

- More detailed treatment of `toctree`?
- Mention of intersphinx
- Contentious choice of Markdown/MyST (to be discussed separately?)
- ...

ericholscher:
We should probably just merge this, I think?

ericholscher (approved):
This is in progress.

astrojuanlu:
@ericholscher What about adding a note at the top saying "this is in progress, see https://github.com/sphinx-doc/sphinx/issues/9165 and linked pull requests" and merging it then?

ericholscher:
@astrojuanlu Works for me 👍

astrojuanlu:
Added, and merging 👍🏽

PR #8113, opened 2021-04-19
Improve contributing docs, take 2

astrojuanlu:
- Force Python 3.6 on tox, for those using it (to avoid installation issues of dependencies and also to match the version we are currently using)
- Add more detail to MinIO instructions (do not assume that the user knows how MinIO works)
- Add post-install, verification steps

~This should be ready to go, but at the moment I am not able to complete the steps because of a local error:~ _(Edit: Figured it out, fix to follow)_

### DIFF
--- a/docs/development/install.rst
+++ b/docs/development/install.rst
@@ -70,6 +70,20 @@ Set up your environment
 #. go to http://community.dev.readthedocs.io to access your local instance of Read the Docs.
 
 
+Check that everything works
+---------------------------
+
+#. go to http://community.dev.readthedocs.io and check that the appearance and style looks correct
+   (otherwise the MinIO buckets might be misconfigured, see above)
+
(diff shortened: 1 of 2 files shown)


docs/development/install.rst
```diff
@@ -62,15 +62,26 @@ Set up your environment
 
    * go to http://localhost:9000/ (MinIO S3 storage backend)
    * login as ``admin`` / ``password``
-   * click "..." next to the bucket name and then "Edit Policy"
-   * give "Read Only" access on all the buckets (``static`` and ``media``)
+   * click "..." next to the ``static`` bucket name and then "Edit Policy"
```
cocobennett, line 65: These new instructions are very helpful! Thank you for doing this!

astrojuanlu:
Thanks @humitos and @cocobennett for having a look! 🙌🏽

PR #8125, opened 2021-04-21
Logging: use %s to format the variable

humitos:
Using %d was causing an issue when trying to format `None`:

```
 celery-web celery.redirected:227[19247]: WARNING Message: '(Sync Versions) Deleted Versions: project=%s versions_count=[%d]'
 celery-web Arguments: ('***************', None)
```

This commit uses %s and defaults to 0 when no versions were deleted.

### DIFF
--- a/readthedocs/api/v2/utils.py
+++ b/readthedocs/api/v2/utils.py
@@ -212,9 +212,9 @@ def delete_versions_from_db(project, tags_data, branches_data):
         .exclude(active=True)
     )
     _, deleted = to_delete_qs.delete()
-    versions_count = deleted.get('builds.Version')
+    versions_count = deleted.get('builds.Version', 0)
     log.info(
-        '(Sync Versions) Deleted Versions: project=%s versions_count=[%d]',
+        '(Sync Versions) Deleted Versions: project=%s versions_count=%s',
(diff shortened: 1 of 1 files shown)


readthedocs/api/v2/utils.py
```diff
@@ -212,9 +212,9 @@ def delete_versions_from_db(project, tags_data, branches_data):
         .exclude(active=True)
     )
     _, deleted = to_delete_qs.delete()
-    versions_count = deleted.get('builds.Version')
+    versions_count = deleted.get('builds.Version', 0)
```
astrojuanlu, line 215: If the default is now `0`, doesn't that mean that `%d` can be kept? Or am I missing something?

stsewd (approved):

readthedocs/api/v2/utils.py
humitos: I added the 0 as default to avoid None and render it nicely (eg. `versions_count=0` instead of `versions_count=None`). I'm using `%s` to avoid the same issue we hit here in the feature because the usage of `%d`.

Technically speaking, `%d` is correct _and_ better (?). Practically speaking, `%s` doesn't break because something dumb as this 😄

PR #8127, opened 2021-04-21
Include aria-label in flyout search box

jonels-msft:
Screen readers use aria-label.

Fixes part of https://github.com/readthedocs/sphinx_rtd_theme/issues/970


stsewd (approved):
Thanks!

PR #8131, opened 2021-04-22
web-celery: add logging for OOM debug on suspicious tasks

humitos:
Add some `log.info` with the local context for this task so we can debug OOM a
little more in production code.


agjohnson (approved):
`locals()` might be noisy, we can see what the output looks like though.

PR #8142, opened 2021-04-28
Add project/build filters

agjohnson:
Most of these filters are currently only used in views when the new
templates are enabled. The project dashboard view context data was
replaced with a filter, which does enable sorting on the dashboard view
using request arguments (like `?sort=built_last`), but this isn't
exposed in the UI.

### DIFF
--- a/readthedocs/builds/filters.py
+++ b/readthedocs/builds/filters.py
@@ -1,0 +1,43 @@
+import logging
+
+from django.forms.widgets import HiddenInput
+from django.utils.translation import ugettext_lazy as _
+from django_filters import CharFilter, ChoiceFilter, FilterSet
+
+from readthedocs.builds.constants import BUILD_STATE_FINISHED
+
+log = logging.getLogger(__name__)
+
+
+class BuildListFilter(FilterSet):
+
+    STATE_ACTIVE = 'active'
+    STATE_SUCCESS = 'succeeded'
+    STATE_FAILED = 'failed'
+
+    STATE_CHOICES = (
+        (STATE_ACTIVE, _('Active')),
+        (STATE_SUCCESS, _('Build finished')),
+        (STATE_FAILED, _('Build failed')),
+    )
+
+    # Attribute filter fields
+    version = CharFilter(field_name='version__slug', widget=HiddenInput)
+    state = ChoiceFilter(
+        label=_('State'),
+        choices=STATE_CHOICES,
+        empty_label=_('Any'),
+        method='get_state',
+    )
+
+    def get_state(self, queryset, name, value):
+        if value == self.STATE_ACTIVE:
+            queryset = queryset.exclude(state=BUILD_STATE_FINISHED)
+        elif value == self.STATE_SUCCESS:
+            queryset = queryset.filter(state=BUILD_STATE_FINISHED, success=True)
+        elif value == self.STATE_FAILED:
+            queryset = queryset.filter(
+                state=BUILD_STATE_FINISHED,
+                success=False,
+            )
+        return queryset
--- a/readthedocs/builds/views.py
+++ b/readthedocs/builds/views.py
@@ -4,24 +4,22 @@ import logging
 import textwrap
 from urllib.parse import urlparse
 
+from django.conf import settings
 from django.contrib import messages
 from django.contrib.auth.decorators import login_required
-from django.http import (
-    HttpResponseForbidden,
-    HttpResponseRedirect,
-)
+from django.http import HttpResponseForbidden, HttpResponseRedirect
 from django.shortcuts import get_object_or_404
 from django.urls import reverse
 from django.utils.decorators import method_decorator
 from django.views.generic import DetailView, ListView
 from requests.utils import quote
 
+from readthedocs.builds.filters import BuildListFilter
 from readthedocs.builds.models import Build, Version
 from readthedocs.core.permissions import AdminPermission
 from readthedocs.core.utils import trigger_build
 from readthedocs.doc_builder.exceptions import BuildEnvironmentError
 from readthedocs.projects.models import Project
-
 
 log = logging.getLogger(__name__)
 
@@ -97,7 +95,13 @@ class BuildList(BuildBase, BuildTriggerMixin, ListView):
         context['project'] = self.project
         context['active_builds'] = active_builds
         context['versions'] = self._get_versions(self.project)
-        context['build_qs'] = self.get_queryset()
+
+        builds = self.get_queryset()
+        if settings.RTD_EXT_THEME_ENABLED:
+            filter = BuildListFilter(self.request.GET, queryset=builds)
+            context['filter'] = filter
+            builds = filter.qs
+        context['build_qs'] = builds
 
         return context
 
--- a/readthedocs/projects/filters.py
+++ b/readthedocs/projects/filters.py
@@ -36,8 +36,8 @@ class VersionSortOrderingFilter(OrderingFilter):
     def filter(self, qs, value):
         # This is where we use the None value for this custom filter. This
         # doesn't work with a standard model filter. Note: ``value`` is always
-        # an iterable.
-        if value is None:
+        # an iterable, but can be empty.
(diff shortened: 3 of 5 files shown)


agjohnson:
I found a missing commit that was supposed to be part of this branch. Going back to WIP.

agjohnson:
Ready for review!

readthedocs/projects/filters.py
```diff
@@ -0,0 +1,158 @@
...
+
+from django.db.models import Count, F, Max
+from django.forms.widgets import HiddenInput
+from django.utils.translation import ugettext_lazy as _
+from django_filters import CharFilter, ChoiceFilter, FilterSet, OrderingFilter
+
+log = logging.getLogger(__name__)
+
+
+class SortOrderingFilter(OrderingFilter):
```
humitos, line 11: I didn't find _where_ you use this custom filter.

readthedocs/projects/filters.py
```diff
@@ -0,0 +1,158 @@
...
+
+    """
+    Version list sort ordering django_filters filter.
+
+    Django-filter is highly opionated, and the default model filters do not work
+    well with empty/null values in the filter choices. In our case, empty/null
+    values are used for a default query. So, to make this work, we will use a
+    custom filter, instead of an automated model filter.
```
humitos, line 19: By reading this comment, I'm not sure to understand what's the problem that this filter solves 😅 --I only get that it does "something" with null values, but not sure what fixes from the default filter.

readthedocs/projects/views/private.py
```diff
@@ -100,6 +101,13 @@ def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
         # Set the default search to search files instead of projects
         context['type'] = 'file'
+
+        filter = ProjectListFilterSet(self.request.GET, queryset=self.get_queryset())
+        context['filter'] = filter
+        context['project_list'] = filter.qs
+        # Alternatively, dynamically override super()-derived `project_list` context_data
+        # context[self.get_context_object_name(filter.qs)] = filter.qs
```
humitos, line 109: Shouldn't this be behind `if settings.RTD_EXT_THEME_ENABLED`?

humitos (approved):
The code looks good to me. However, I haven't QAed this yet because it seems it's not ready from the UI. This is not used yet, so I think it's fine to merge and leave it ready for the new templates to make usage of it.

readthedocs/projects/filters.py
agjohnson: Hrm, then I might have busted something in all of the merges. Let me check

readthedocs/projects/filters.py
agjohnson: :+1: I can expand on this

readthedocs/projects/views/private.py
agjohnson: I think you're right. I assumed that I wrote this to replace existing filtering functionality on `project_list`, but digging through code I don't see any similar filter mechanism.

agjohnson:
I fixed some goofs, but that probably means I should check the application again with the new templates because I'm sure I've changed something incorrectly. I will test this today.

readthedocs/builds/filters.py
```diff
@@ -0,0 +1,43 @@
...
+
+    STATE_ACTIVE = 'active'
+    STATE_SUCCESS = 'succeeded'
+    STATE_FAILED = 'failed'
+
+    STATE_CHOICES = (
+        (STATE_ACTIVE, _('Active')),
+        (STATE_SUCCESS, _('Build finished')),
+        (STATE_FAILED, _('Build failed')),
+    )
```
ericholscher, line 22: Should these live in a constants file somewhere? Seems useful outside of just the filter.

readthedocs/builds/views.py
```diff
@@ -97,7 +95,13 @@ def get_context_data(self, **kwargs):  # pylint: disable=arguments-differ
         context['project'] = self.project
         context['active_builds'] = active_builds
         context['versions'] = self._get_versions(self.project)
-        context['build_qs'] = self.get_queryset()
+
+        builds = self.get_queryset()
+        if settings.RTD_EXT_THEME_ENABLED:
```
ericholscher, line 100: I think we can just always run it through the filter, no need for the setting. It just won't get filtered, if the user doesn't have the GET args.

readthedocs/projects/filters.py
```diff
@@ -0,0 +1,163 @@
...
+            ('-recent', _('Most recently built')),
+            ('recent', _('Least recently built')),
+        ]
+
+    def filter(self, qs, value):
+        # This is where we use the None value for this custom filter. This
+        # doesn't work with a standard model filter.
+        if value is None:
+            value = ['relevance']
+        return qs.annotate(
```
ericholscher, line 41: Don't we only want to annotate the value that we're actually filtering? Seems like this will add query overhead.

readthedocs/projects/filters.py
```diff
@@ -0,0 +1,163 @@
...
+
+class ProjectSortOrderingFilter(OrderingFilter):
+
+    """
+    Project list sort ordering django_filters filter.
+
+    Django-filter is highly opionated, and the default model filters do not work
+    well with empty/null values in the filter choices. In our case, empty/null
+    values are used for a default query. So, to make this work, we will use a
+    custom filter, instead of an automated model filter.
```
ericholscher, line 61: Seems like that's what the null_choice is? https://django-filter.readthedocs.io/en/stable/guide/tips.html#solution-2-using-choicefilter-s-null-choice

ericholscher (commented):
This looks like a good start. I think we should just always use the filter, and that will let us figure out if there's performance issues with it. 

Seems like this also needs some tests. I'm a little worried about performance here, so it might be good to make sure the queries we're generating aren't going to kill the DB, especially around the Build & Version tables.

readthedocs/builds/filters.py
agjohnson: These are very particular to the filter UI, and I wouldn't expect to reuse them.

readthedocs/builds/views.py
agjohnson: That was what I originally had here, but updated on feedback already. I think the conditional is fine, we'll test this in the beta on production data.

readthedocs/projects/filters.py
agjohnson: I wasn't able to accomplish what I was looking for with null_choice

readthedocs/projects/filters.py
agjohnson: Yeah, that's an easy optimization to make. I'll update this

agjohnson:
> Seems like this also needs some tests.

Perhaps. I'm probably not the best person to write these tests however. It would probably be better if someone else wants to dive into test and/or we revisit this later. My major concern is continuing to block this work, it's holding up my progress on templates.

ericholscher (approved):
I'm fine merging and adding a TODO somewhere to test this code, as long as we actually get back to it :)

I'd love to have you working w/ @humitos as a small team, with him doing backend & API changes needed. I'd generally like to see us doing this more, but we all have close to the same skillsets which makes it hard to find space for it.

agjohnson:
Yup, that's the workflow I was describing.

I've already pushed off API changes, which are easier to collaborate on, though differences in priority slow progress a good deal.

Changes like this PR are hard, and require experimentation on multiple repositories to nail down the UI. This process would have to be more of a pairing process, as there was experimentation on multiple repositories to get this UI correct. It would have been an awful amount of back and forth to collaborate on this work. This is where I'm looking for process improvements. We will want to figure out how to split up backend and frontend development into digestible work.

agjohnson:
Moved test coverage to #8238

agjohnson:
I have tested these changes and they look correct. Build/version/project filters are all working on new templates. I squeezed a small fix for unpacking the dictionary, I didn't get a chance to test the previous commit because my dev environment was broken again. Looks good now however.

PR #8145, opened 2021-04-29
Disable FLOC by introducing permissions policy header

davidfischer:
This PR introduces the Permissions-Policy header ([MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Feature_Policy)) which controls the features that a webpage is allowed to use. For example, Permissions-Policy could be used to disable access to the JS geolocation APIs or the accelerometer.

This PR, however, only uses Permissions-Policy to disable [FLOC](https://github.com/WICG/floc) a new JS API in Chrome that can be used to categorize users together into cohorts with machine learning so they can be targeted with ads. FLOC will be disabled on the RTD dashboard as well as on documentation sites. There are a few reasons for this but the biggest is probably that FLOC has some privacy issues and can aid in fingerprinting users (see the [EFF's coverage](https://www.eff.org/deeplinks/2021/03/googles-floc-terrible-idea)). While Read the Docs does display advertising on our documentation sites, we are quite explicit in the [methods we use to target advertising](https://docs.readthedocs.io/en/stable/advertising/advertising-details.html#our-targeting-details) -- basically just contextual targeting -- and FLOC definitely crosses into behavioral targeting and targeting ads based on past actions rather than content/context.

### DIFF
--- a/dockerfiles/nginx/proxito.conf
+++ b/dockerfiles/nginx/proxito.conf
@@ -66,6 +66,8 @@ server {
         add_header Cache-Tag $cache_tag always;
         set $referrer_policy $upstream_http_referrer_policy;
         add_header Referrer-Policy $referrer_policy always;
+        set $permissions_policy $upstream_http_permissions_policy;
+        add_header Permissions-Policy $permissions_policy always;
     }
 
     # Serve 404 pages here
--- a/readthedocs/settings/base.py
+++ b/readthedocs/settings/base.py
@@ -99,6 +99,12 @@ class CommunityBaseSettings(Settings):
         "/admin/",
     )
 
+    # Permissions Policy
+    # https://github.com/adamchainz/django-permissions-policy
+    PERMISSIONS_POLICY = {
+        "interest-cohort": [],
+    }
+
     # Read the Docs
     READ_THE_DOCS_EXTENSIONS = ext
     RTD_LATEST = 'latest'
@@ -228,6 +234,7 @@ class CommunityBaseSettings(Settings):
         'corsheaders.middleware.CorsMiddleware',
         'csp.middleware.CSPMiddleware',
         'readthedocs.core.middleware.ReferrerPolicyMiddleware',
+        'django_permissions_policy.PermissionsPolicyMiddleware',
     )
 
     AUTHENTICATION_BACKENDS = (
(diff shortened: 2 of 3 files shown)


davidfischer:
Note, if you are a documentation author and for some reason you would like to have FLOC enabled on your documentation, please get in touch. I would like to hear from you and we would consider making this a configurable option on projects.

readthedocs/settings/base.py
```diff
@@ -99,6 +99,12 @@ class CommunityBaseSettings(Settings):
         "/admin/",
     )
 
+    # Permissions Policy
+    # https://github.com/adamchainz/django-permissions-policy
+    PERMISSIONS_POLICY = {
```
ericholscher, line 104: Is this something we can easily test? Would be good to have at least an integration test to make sure the setting, etc. is correct.

ericholscher (approved):
Looks good to me 👍

readthedocs/settings/base.py
davidfischer: Are you thinking a validatehttp check? Or a pagerduty/CF check? I think those would be the best places for it.

dockerfiles/nginx/proxito.conf
```diff
@@ -66,6 +66,8 @@ server {
         add_header Cache-Tag $cache_tag always;
         set $referrer_policy $upstream_http_referrer_policy;
         add_header Referrer-Policy $referrer_policy always;
+        set $permissions_policy $upstream_http_permissions_policy;
+        add_header Permissions-Policy $permissions_policy always;
```
humitos, line 70: We should probably add the same in `web.conf` to make it work int the same way on the dashboard. It's not _required_ sine this is locally, but I think it makes sense to keep consistency with prod.

readthedocs/settings/base.py
humitos: I think a regular Django test + validatehttp are 💯

humitos (approved):
Looks good to me!

The description of the PR is 💯

dockerfiles/nginx/proxito.conf
davidfischer: I don't think I understand this proposal. `web.conf` ([here](https://github.com/readthedocs/readthedocs.org/blob/master/dockerfiles/nginx/web.conf)) doesn't have any nginx configuration to include headers -- either here or in the private operations repo.

humitos: Oh, yeah. You are right! I got confused about what was the case were this was needed. Thanks!

PR #8178, opened 2021-05-13
APIv2: disable listing endpoints

humitos:
We have experienced DOS in some opportunities when hitting these endpoints
without any kind of filter. They may require a big DB query causing the site
going down.

This commit disable all listing endpoint without removing the functionability we
are currently using:

- filter versions by project slug
- getting a specific build by commit hash

On those cases where it's disabled, we return a nice message to the user
communicating that it's disabled and APIv3 should be used instead.

### DIFF
--- a/readthedocs/api/v2/views/model_views.py
+++ b/readthedocs/api/v2/views/model_views.py
@@ -66,6 +66,49 @@ class PlainTextBuildRenderer(BaseRenderer):
         return data.encode(self.charset)
 
 
+class DisableListEndpoint:
+
+    """
+    Helper to disable APIv2 listing endpoint.
+
+    We are disablng the listing endpoint because it could cause DOS without
+    using any type of filtering.
+
+    This class disables these endpoints except:
+
+     - version resource when passing ``?project__slug=``
+     - build resource when using ``?commit=``
+
+    All the other type of listings are disabled and return 409 CONFLICT with an
+    error message pointing the user to APIv3.
+    """
+
+    def list(self, *args, **kwargs):
+        disabled = True
+
+        # NOTE: keep list endpoint that specifies a resource
+        if any([
+                self.basename == 'version' and 'project__slug' in self.request.GET,
+                self.basename == 'build' and 'commit' in self.request.GET,
+        ]):
+            disabled = False
+
+        if not disabled:
+            return super().list(*args, **kwargs)
+
+        return Response(
+            {
+                'error': 'disabled',
+                'msg': (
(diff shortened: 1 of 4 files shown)


readthedocs/api/v2/views/model_views.py
```diff
@@ -66,6 +66,49 @@ def render(self, data, accepted_media_type=None, renderer_context=None):
...
+            disabled = False
+
+        if not disabled:
+            return super().list(*args, **kwargs)
+
+        return Response(
+            {
+                'error': 'disabled',
+                'msg': (
+                    'List endpoint have been disabled: please provide a resource identifier. '
```
ericholscher, line 103:
```suggestion
                    'List endpoint have been disabled due to heavy resource usage: please provide a resource identifier. '
```

I have no idea what a resource identifier is. We should provide examples or something.

readthedocs/api/v2/views/model_views.py
```diff
@@ -66,6 +66,49 @@ def render(self, data, accepted_media_type=None, renderer_context=None):
...
+                    'Take into account than APIv2 is planned to be deprecated soon. '
+                    'Please use APIv3: https://docs.readthedocs.io/page/api/v3.html'
+                )
+            },
+            status=status.HTTP_409_CONFLICT,
```
ericholscher, line 108: I wish there was a better status code. I googled for this -- but we should probably comment here that there isn't a proper one to use.

ericholscher (approved):
Thanks for getting on this 👍

readthedocs/api/v2/views/model_views.py
humitos: A resource identifier is the object pk, or commit, or the unique id we use for that particular object: `/api/v2/projects/` -> `/api/v2/projects/1/`

We may want to remove that part of the message tho and just point to APIv2.

readthedocs/api/v2/views/model_views.py
humitos: I will `410 GONE`, instead. Which seems to be exactly what we need.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/410

ericholscher: Cool, we should try and get this released this week.

readthedocs/api/v2/views/model_views.py
humitos: I changed the message and remove the part of the resource id.

PR #8179, opened 2021-05-13
Build: use `is_active` method to know if the build should be skipped

humitos:
Currently we are using only `project.skip` to know if the build for this
particular project has to be skipped after the task was triggered. However, if
the if the project belongs to an organization and the organization is disabled
those builds already queued will be built.

This commit uses `is_active` for `project.skip` field in the API response, that
takes care of this case.

This helps us if we found that an organization _already_ triggered many builds
and we need to cancel all of them. With this change, just by doing
`organization.disable=True` will make kill them all at built time instead of
processing them.

### DIFF
--- a/readthedocs/api/v2/serializers.py
+++ b/readthedocs/api/v2/serializers.py
@@ -47,6 +47,7 @@ class ProjectAdminSerializer(ProjectSerializer):
     )
 
     environment_variables = serializers.SerializerMethodField()
+    skip = serializers.SerializerMethodField()
 
     def get_environment_variables(self, obj):
         """Get all environment variables, including public ones."""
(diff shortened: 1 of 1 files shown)


stsewd (approved):

readthedocs/api/v2/serializers.py
```diff
@@ -58,6 +59,9 @@ def get_environment_variables(self, obj):
             for variable in obj.environmentvariable_set.all()
         }
 
+    def get_skip(self, obj):
```
ericholscher, line 62: Could use a docstring explaining why we're overriding it.

ericholscher (approved):

PR #8215, opened 2021-05-26
Design doc: magical builder maintainability and builder expanding API

humitos:
Design document to discuss how to make the "magical" builder maintainable over
time and also how to extend the builder process to support other doctools eventually.

> Note that this document is not finished yet, but it's a good starting point to start receiving feedback and continue the discussion.

Rendered version: https://docs--8215.org.readthedocs.build/en/8215/development/design/builder-maintainability.html

### DIFF
--- a/docs/development/design/builder-maintainability.rst
+++ b/docs/development/design/builder-maintainability.rst
@@ -1,0 +1,308 @@
+Builder Maintainability
+=======================
+
+.. contents::
+   :local:
+   :depth: 2
+
+This document extends the proposal done by Manuel at `Future Builder`_ to replace the last 3 goals:
+
+* Allow us to add a feature with a defined contract without worry about breaking old builds
+* Introduce ``build.builder: 2`` config (does not install pre-defined packages) for these new features
+* Motivate users to migrate to ``v2`` to finally deprecate this magic by educating users
+
+
+It proposes a way to version the current "magical" builder so we can freeze its current status and
+extend it if needed by doing a new release of it.
+Besides, it removes the needing of using ``build.builder: 2`` since the *new builder* will extend the "magical" builder behavior
+and just override what's required.
+
+.. _Future Builder: https://github.com/readthedocs/readthedocs.org/pull/8190
+
+
+Goals
+-----
+
+* Keep the current "magical" builder working as-is (freeze this version)
+* Make "magical" builder maintainable over time
+* Allow users to specify a version of the "magical" builder
+* Define a code API to split in a modularized way the builder's code
+* Use the API to extend builder's code to support the "new builder"
+* Split the builders' code in: ``BuilderBase``, ``SphinxMagicalBuilder`` and ``SphinxDummyBuilder``
+* Fail the build when mixing config for commands (e.g. ``sphinx.fail_on_warning``, ``submodules``, etc)
+  with the new builder (e.g. ``build.jobs`` and ``build.commands``)
+* Fail the build of new projects if users doesn't pin the version of the builder
+* Fail the build if new builder is used without pinning the builder version
+* Set current "magical" builder version as default for current projects
+* Move the build process out from Read the Docs core on their own packages
+
+
+"Magic" currently executed by our builder
+-----------------------------------------
+
+This section describes all the "magic" that Read the Docs does on behalf of the user in the build process.
+The magic is split in two different scenarios: environment and tool.
+
+.. note::
+
+   This section is strongly related with a similar research done by Santos at `Explicit Builders`_
+
+   .. _Explicit Builders: https://github.com/readthedocs/readthedocs.org/pull/8103/
+
+
+Environment: virtualenv
+~~~~~~~~~~~~~~~~~~~~~~~
+
+Magic happening at virtualenv setup process:
+
+- Creates virtualenv
+- Tries possible pip's requirement files if not defined
+- Install package: repository itself
+- Updates pip and installs setuptools
+  (uses feature flag: ``DONT_INSTALL_LATEST_PIP``)
+- Installs *core* dependencies
+  (uses feature flag: ``DEFAULT_TO_MKDOCS_0_17_3``, ``USE_MKDOCS_LATEST``, ``USE_SPHINX_LATEST``, ``USE_SPHINX_RTD_EXT_LATEST``)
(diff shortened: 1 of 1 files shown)


docs/development/design/builder-maintainability.rst
```diff
@@ -0,0 +1,308 @@
...
+
+* Allow us to add a feature with a defined contract without worry about breaking old builds
+* Introduce ``build.builder: 2`` config (does not install pre-defined packages) for these new features
+* Motivate users to migrate to ``v2`` to finally deprecate this magic by educating users
+
+
+It proposes a way to version the current "magical" builder so we can freeze its current status and
+extend it if needed by doing a new release of it.
+Besides, it removes the needing of using ``build.builder: 2`` since the *new builder* will extend the "magical" builder behavior
+and just override what's required.
```
humitos, line 18: I thought I had this idea clear in my mind when I started writing this, but then I realized that I didn't have a good and clear way as a user to disable _all the magic_. This may be a good topic to cover in our meeting.

docs/development/design/builder-maintainability.rst
```diff
@@ -0,0 +1,308 @@
...
+* How do we allow people to *remove all our magic* without ``build.version: 2``?
+* Should ``build.commands`` do not execute *any* ``readthedocs_`` methods?
+* If ``build.jobs.install`` is overwritten with ``conda env create -f environment.yaml``,
+  should we execute our ``readthedocs_pre_install`` that appends our *core* requirements?
+* Overwritting the job itself should remove automatic pre/post hooks?
+* If we decide to use ``build.version: 2`` to remove all the magic,
+  how we communicate to users what's the magic removed?
+* How users will decide what builder to use?
+  Should this be a config like ``build.builder: readthedocs-sphinx-magical-builder==1.0.0``?
+  Maybe a Python class path ``readthedocs.builder.sphinx.SphinxMagicalBuilder`` that we can import?
```
humitos, line 308:
```suggestion
* How users will decide what builder to use?
  Should this be a config like ``build.builder: readthedocs-sphinx-magical-builder==1.0.0``?
  Maybe a Python class path ``readthedocs.builder.sphinx.SphinxMagicalBuilder`` that we can import?
* How do we audit code for new builders (e.g. `PelicanBuilder` done by the community)? The builder is currently executed _outside_ the Docker container.
* Does it make sense to *isolate* the build process from our Python code and ran it *inside* the container instead? (e.g. ``$ readthedocs build --builder sphinx --source docs/``) This code won't have access to the db nor our app code.
```

agjohnson:
We talked through this design document and some of these ideas today on a call. The main takeaway we had is that this isn't as straight forward as we were considering, but versioning our builder would be an incredibly useful first step to changing builder logic.

### Goals

To distill down some higher level points:

* Our goals with changing builder process are, in order of descending importance:
  1. Make builders more maintainable by core team
  2. Increase configuration of build process
  3. Allow for a build process without magic
  4. Enable arbitrary build tools with good UX

### Versioned builders

We are discussing versioning builders as a first step to this work. This would allow us to make incremental, backwards incompatible changes to builder logic without negatively affecting users.

We poked holes in our plan for versioned builders and came to the following conclusions:

* A versioned builder class can't be used in the application
  * The application code only runs one installed version
    * Users can't select the builder package version then
  * The application code shouldn't run untrusted code
    * We can't extend a pattern like this to fit arbitrary builder classes later (goal 4)
  * It doesn't really help to version the application builder that we have now

* Versioned builder classes could be executed in the docker container environment
  * Workflow would be:
    * We get builder package version from config
    * Versioned builder is installed in environment first
    * Application logic calls through docker to execute builder class
      * Builder class executes command directly?
  * This pattern gets us really close to providing a wrapper cli script to execute all of the same build commands locally for users
  * Very neat idea, but this defeats the purpose of versioning our builders to assist in making large changes to our builders
    * We have to drastically alter the build process to acheive this.

* A third option is to version our implementation outside packaging
  * Use a `build.builder: 2.1` to denote builder in configuration
  * Our application logic is where we keep idea of versioned build classes, not packaging
    * Similar to Stripe API and client interaction
      * User specifies the version of the API to use
    * Hopefully not accomplished with lots of conditional logic
      * Probably accomplished with a lot of class inheritance instead
  * This does not yet address arbitrary builder classes (goal 4), but we view this as a later priority anyways

The plan in this design doc to create a magic Sphinx builder and a separate non-magic Sphinx builder still stands. Without versioning figured out, we don't know how this looks technically.

### Changes to other steps

Using the Sphinx builder as an example, we thought through how we might alter the build steps. I believe we came to the following conclusions:

* We likely want to keep a number of steps the same for the current Sphinx builder
  * Search indexing is going to be better using Sphinx data than it would HTML
  * Overriding things like `html_context` does provide a lot of benefit
    * Users/themes are using `html_context`, so this would be a hostile change
    * If we move to another data structure, we're probably still translating that to `html_context` anyways
    * `html_context` is not the issue, the extraneous logic in `conf.py.tmpl` is the larger issue

We did not list "remove magic" as a goal of this work, because a good deal of the magic that helps the build process also provides really good UX to a lot of users. "Remove magic" is also maybe misleading, because we can't make the RTD build process match the local build process without dropping features like our flyout, search, etc.

So, in this scenario, we are mostly keeping the current Sphinx builder code, instead of our plan to throw away all of this functionality in search of a build process that we don't control.

### Priorities and potential next steps

The priority order for work likely looks something like this:

* Figure out if versioned builders is possible
* Freeze current builder in a versioned builder. Version Sphinx builder as we change/remove features, options, flags, etc.
* We add all options for `build.jobs`, these override default Sphinx builder commands
* We add a separate builder class to remove all magic, this is independently selectable via config
* We figure out a pattern to allow for community contributions and good UX around arbitrary builders
  * The build pack concept for code executed inside the docker container

humitos:
We are in a lot better situation now. We've removed a lot of the magic and we are migrating all our projects to addons (see https://github.com/readthedocs/website/pull/308)

The only relevant part from this document is the ability from the community to "package a builder" --but we've discussed we don't want to implement something custom and, if we are going to in that direction, using buildpacks or support running GitHub actions in our builders would be the path we would follow somehow.

It's also worth noting the "magic" part of our build process, was migrated to a Sphinx extension in case people want to keep the old/magical behavior for some reason: https://github.com/readthedocs/sphinx-build-compatibility/

PR #8216, opened 2021-05-26
Remove IRC from our docs

ericholscher:
We haven't used this actively in years,
but freenode is now lost :(


agjohnson (approved):
Looks good. I have reserved Libera.chat `#readthedocs` as well, but I would probably vote to emphasize gitter rooms instead anyways.

Issue #7117, opened 2020-05-22
Add more context data to embed API


agjohnson:
While implementing embed docs topics via the API for our project admin pages, I got to play with the API and found some pieces that I wanted.

These might make good additions as API return or potentially we just do this as a javascript client side (which is ultimately how I'm manipulating the content now).

- More metadata around headers, such as heading level. I'd like to display the topics in a nested menu, as they don't make as much sense in sequential order
- Embed API should accept a language argument. I wasn't able to get this working and had to fall back to using a full URL
- Embed API could use accept language header of the client to determine the translation to give, this would be a great middle ground. That way topics embedded are also translated. For instance, we request embed from `https://docs.readthedocs.io/en/stable` always, but the embed API would use accept language header to lookup a translation, if it exists, and return the embed data from that translation.
- Same, but include an explicit `translation` argument or something for the accepted language
- Probably more of a client side feature, it would be great to break up the content block a lot more. Parsing out content from the sphinx output was too tedious, but it is possible. Data could include a list of sections with section header, section header nesting level, perhaps a truncated version of the content?

We should decide if any of these are worth pursuing and decide where they might land -- either as API features or on the client side javascript library TBD.

humitos:
> Embed API should accept a language argument. I wasn't able to get this working and had to fall back to using a full URL

In the v3 of this API we are only allowing `?url=` argument and removing all the other ones to avoid Sphinx-specific concept here. So, I think this point is outdated now.

> Embed API could use accept language header of the client to determine the translation to give, this would be a great middle ground. That way topics embedded are also translated. For instance, we request embed from `https://docs.readthedocs.io/en/stable` always, but the embed API would use accept language header to lookup a translation, if it exists, and return the embed data from that translation.
> 
> Same, but include an explicit `translation` argument or something for the accepted language

What would be the use case for this? I imagine that a set of English documentation should use Embed API with English URLs, Spanish docs should use Spanish URLs, and so on. What's a useful case to do `http://docs.readthedocs.io/en/development/install.html?translation=ja` from a project's English documentation. In this case, the tooltip will be in Japanese but the whole documentation in English, which is not good for me.

In case the front-end shows different languages on the same URL, the client could tweak the URL to change `/en/` to `/ja/` instead of adding `?translation=ja` to an URL for an English document.

Besides, this is a Read the Docs-specific feature that won't have any effect on external sites.

[...]

PR #8222, opened 2021-05-31
Design doc: Embed APIv3

### LINKED ISSUE #7963: Improvements for the embed app

humitos:
I made too many changes to the original document that I'm creating a new PR based on Santos' PR: #8039. It basically extends and replaces it.

📝 read the rendered version at: https://docs--8222.org.readthedocs.build/en/8222/development/design/embed-api.html

Closes #7963 
Closes #8039


docs/development/design/embed-api.rst
```diff
@@ -0,0 +1,211 @@
...
+* Support external documents hosted outside Read the Docs
+* Do not depend on Sphinx ``.fjson`` files
+* Query and parse the ``.html`` file directly (from our storage or from an external request)
+* Rewrite all links returned in the content to make them absolute
+* Always return valid HTML structure
+* Delete HTML tags from the original document if needed
+* Support ``?nwords=`` and ``?nparagraphs=`` to return chunked content
+* Require a valid HTML ``id`` selector
+* Handle special cases for particular doctools (e.g. Sphinx requires to return the ``.parent()`` element for ``dl``)
+* Make explicit the client is asking to handle the special cases (e.g. send ``?doctool=sphinx&version=4.0.1``)
```
stsewd, line 54: I think we voted against handling special cases, and as mentioned in my other comment this isn't a special case for sphinx, but following the semantic representation of html tags.

docs/development/design/embed-api.rst
```diff
@@ -0,0 +1,211 @@
...
+the inclusion of new features and the definition of a contract that works the same for all,
+this document set the following goals for the new version of this endpoint:
+
+* Support external documents hosted outside Read the Docs
+* Do not depend on Sphinx ``.fjson`` files
+* Query and parse the ``.html`` file directly (from our storage or from an external request)
+* Rewrite all links returned in the content to make them absolute
+* Always return valid HTML structure
+* Delete HTML tags from the original document if needed
+* Support ``?nwords=`` and ``?nparagraphs=`` to return chunked content
```
stsewd, line 51: I think we still were discussing this. I don't think we can support it without having to handle specific cases, not all returned html will have `p` tags or the word will end in a valid tag, we also need to skip tags from the lenght


```
<p>Cut <strong>here?</strong></p>
```

Having the client using css to "hide" long texts is more easy to handle.

docs/development/design/embed-api.rst
```diff
@@ -0,0 +1,211 @@
...
+   .. sourcecode:: json
+
+      {
+         "project": "docs",
+         "version": "latest",
+         "language": "en",
+         "path": "development/install.html",
+         "title": "Development Installation",
+         "url": "https://docs.readthedocs.io/en/latest/install.html#set-up-your-environment",
+         "id": "#set-up-your-environment",
```
stsewd, line 81: We shouldn't include the `#` in the response, that's part of the anchor spec, not the id itself.

docs/development/design/embed-api.rst
```diff
@@ -0,0 +1,211 @@
...
+         "content": "<div class=\"section\" id=\"development-installation\">\n<h1>Development Installation<a class=\"headerlink\" href=\"https://docs.readthedocs.io/en/stable/development/install.html#development-installation#development-installation\" title=\"Permalink to this headline\">¶</a></h1>\n ..."
+      }
+
+
+   When used together with ``?expand=identifiers`` the follwing field is also returned:
```
stsewd, line 86: I'm not sure about returning this information on the same endpoint, is confusing returning the content of a section plus the list of sections, we should have a different endpoint to return data about the page itself.

docs/development/design/embed-api.rst
humitos: Good point!

docs/development/design/embed-api.rst
humitos: This is definitely a special case considering how our contract was defined: "return the HTML tag with the `id=` requested". However, this case is "return the **parent** of HTML tag with the `id=` requested". It may not be a Sphinx-specific special case, but we need a way to communicate the backend we want to benefit from that extra manipulation: "return the parent _and_ remove the siblings"

I see this as some "magic" the backend is doing for us and, as a client, we will need a way to opt-in/out to this "magic". Adding a config to opt-in/out will make this endpoint maintainable overtime when we discover more similar cases.

docs/development/design/embed-api.rst
humitos: The example you share would return:

1. with `?nwords=1`: `<p>Cut</p>`
2. with `?nwords=2`: `<p>Cut <strong>here?</strong></p>`
3. with `?nparagraphs=1`: `<p>Cut <strong>here?</strong></p>`

docs/development/design/embed-api.rst
humitos: Can you expand here why it may cause confusion and how having a different endpoint to only show the available `id=` for a page is better? I think I'm not against a different endpoint, but I want to understand your position better.

docs/development/design/embed-api.rst
stsewd: We are committed to return valid html, so I don't see that as magic, is just following a well-defined case for a html tag.

docs/development/design/embed-api.rst
stsewd: yeah, but the problem is calculating the words and taking the tags into account, you'll be removing tags from the original html, possible breaking some styles like with tables/definition lists. And also how would you know if there is a paragraph when you have content surrounded in other tags like lists or divs.

docs/development/design/embed-api.rst
stsewd: Because you are requesting information for a section, but getting information about the page too, you don't want to have to query a random section just to get information about the page. This also allow us to return more information about the page itself like the title.

docs/development/design/embed-api.rst
humitos: We have been playing with this already in https://github.com/readthedocs/readthedocs-ext/pull/304/. _How to do it_ is an implementation detail and I'm sure there is going to be some problems we will need to solve. However, I don't think it's impossible.

I've been playing a little with BeautifulSoup already and this seems to work close enough:

```python
# nparagraphs.py
from bs4 import BeautifulSoup

soup = BeautifulSoup(open('install.html'))
nparagraphs = 3
for element in soup.find('div', attrs={'id': 'development-installation'}).findAll():
    if nparagraphs == 0:
        element.replaceWith('')

    if element.name == 'p' and nparagraphs > 0:
        nparagraphs -= 1
```

```bash
$ wget https://docs.readthedocs.io/en/stable/development/install.html
$ python nparagraph.py
```

<details>
  <summary>Click to see the output</summary>

```html
<div class="section" id="development-installation">
 <h1>
  Development Installation
  <a class="headerlink" href="#development-installation" title="Permalink to this headline">
   ¶
  </a>
 </h1>
 <p>
  These are development setup and
  <a class="hoverxref tooltip reference internal" data-doc="development/install" data-docpath="/development/install.html" data-project="docs" data-section="core-team-standards" data-version="stable" href="#core-team-standards">
   <span class="std std-ref">
    standards
   </span>
  </a>
  that are adhered to by the core development team while
developing Read the Docs and related services. If you are a contributor to Read the Docs,
it might a be a good idea to follow these guidelines as well.
 </p>
 <div class="admonition note">
  <p class="admonition-title">
   Note
  </p>
  <p>
   We do not recommend to follow this guide to deploy an instance of Read the Docs for production usage.
Take into account that this setup is only useful for developing purposes.
  </p>
 </div>
</div>
```
</details>

docs/development/design/embed-api.rst
humitos: Pinging @agjohnson here that he wrote:

> More metadata around headers, such as heading level. I'd like to display the topics in a nested menu, as they don't make as much sense in sequential order

in https://github.com/readthedocs/readthedocs.org/issues/7117

@agjohnson can you take a look at this proposal and suggest that would be the ideal response to you and how you would like to use it?

humitos:
Notes to myself: it seems I missed some points from https://github.com/readthedocs/readthedocs.org/issues/7117 that @agjohnson wants, like language as argument and/or HTTP header. So, I'll come back to this document and update it with those points.

docs/development/design/embed-api.rst
humitos: @agjohnson returning headers may not be useful if those headers don't have an `id=` that we can find. Also, I realized that it's not trivial to "find the title for a particular `id=`" ([we have this hack currently implemented that only works over `a` tags](https://github.com/readthedocs/readthedocs.org/blob/0340604163a15fccc8b786ce0a5e589bf78f6f24/readthedocs/embed/views.py#L283-L286). The `id=` could be in the `div` surrounding the `h1` tag or the `dl` tag or any other.

```html
<div id="configuration">
  <h1>Configuration<a class="headerlink" href="#configuration" title="Permalink to this headline">¶</a></h1>
  <p>...</p>
</div>
```

For this case, we could get the `.next()` from the `id=configuration`. However, we can't guarantee that this will always work. Even if we get it, we will want to remove the trailing `¶` from its name as well in this case (but it could be a different char)


```html
<dt id="confval-hoverxref_modal_class">
  <code class="sig-name descname">hoverxref_modal_class</code>
  <a class="headerlink" href="#confval-hoverxref_modal_class" title="Permalink to this definition">¶</a>
</dt>
```

This one is similar, but the `.next()` element is exactly the title we are looking for. It does not contain the `¶` char.

However, if the title we want is exactly on the `h1` or `dl` (instead of their child), we will fail to detect it and we would return something invalid.

----

So, I think it's still useful to return all the available `id=`s from the page so developers can explore them, but it's not easy (maybe impossible) to know the exact title for that particular `id=` and the exact hierarchy as well.

docs/development/design/embed-api.rst
```diff
@@ -0,0 +1,211 @@
...
+
+   This leaves the door open to be able to support more special cases (e.g. for other doctools) without breaking the actual behavior.
+
+
+Support for external documents
+------------------------------
+
+When the ``?url=`` argument passed belongs to a documentation page not hosted on Read the Docs,
+the endpoint will do an external request to download the HTML file,
+parse it and return the content for the identifier requested.
```
agjohnson, line 175: I'm a bit hesitant on this concept. Requesting remote content does open up some security concerns. It seems like this is primarily just to get around CORS headers? Otherwise, this seems like it would be better implemented as a client side function, not an RTD hosted service.

To add to the security concerns you outlined, anything purposely or accidentally tying up our web processes, like latent servers and large file requests, and proxying abusive traffic through our server requests to mask origins.

docs/development/design/embed-api.rst
```diff
@@ -0,0 +1,211 @@
...
+* Do not depend on Sphinx ``.fjson`` files
+* Query and parse the ``.html`` file directly (from our storage or from an external request)
+* Rewrite all links returned in the content to make them absolute
+* Always return valid HTML structure
+* Delete HTML tags from the original document if needed
+* Support ``?nwords=`` and ``?nparagraphs=`` to return chunked content
+* Require a valid HTML ``id`` selector
+* Handle special cases for particular doctools (e.g. Sphinx requires to return the ``.parent()`` element for ``dl``)
+* Make explicit the client is asking to handle the special cases (e.g. send ``?doctool=sphinx&version=4.0.1``)
+* Accept only ``?url=`` request GET argument to query the endpoint
```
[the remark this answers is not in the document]
stsewd: I think we should support both forms, only url and `project` + `version` + `section` as query params. For the translation I'm not sure, maybe that's a client implementation (you know what translations your project has)?

docs/development/design/embed-api.rst
stsewd: >  It seems like this is primarily just to get around CORS headers?

Nope, is required for the grant to make the api work with sites hosted outside RTD. I'm also worried about supporting external sites, don't know if we already committed to have that feature or not /cc @ericholscher

docs/development/design/embed-api.rst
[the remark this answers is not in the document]
humitos: > I think we should support both forms, only url and `project` + `version` + `section` as query params

@stsewd Why do you think we should support both? What are the arguments in favor of this? What are the benefits over the `?url=` argument?

docs/development/design/embed-api.rst
humitos: @agjohnson I was asking about how is the ideal output you would like to get from the API. Besides, I raised some issues that will make it hard to do it _right_ with the output that I imagined you need. Can you take a look at my previous comments and tell me your thoughts on this?

docs/development/design/embed-api.rst
```diff
@@ -0,0 +1,211 @@
+Embed APIv3
+===========
+
+The Embed API allows to embed content from documentation pages in other sites.
+It has been treated as an *experimental* feature without public documentation or real applications,
+but recently it started to be used widely (mainly because we created a Sphinx extension).
+
+The main goal of this document is to design a new version of the Embed API to be more user friendly,
+make it more stable over time, support documentation pages not hosted at Read the Docs,
```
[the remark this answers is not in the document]
humitos: Hey @agjohnson! 

> this language still sounds like "support for Sphinx"

I don't understand why it sounds like referring to Sphinx to you. Can you clarify this?

Also, I had some trouble understanding your suggestion "support embedding _outside_ documentation _hosted at_ Read the Docs". Mainly because it says "hosted at Read the Docs" _but_ with "outside" used as an adjective for "documentation". In any case, I found confusing how the sentence is written. I like the suggestion about using "embedding" but I'd like to keep it explicitly that it's "not hosted at Read the Docs" if that's possible.

I asked @astrojuanlu about this and he was also confused about how the sentence was written.

Is it possible to use something like the following? 

- "make it more stable over time, support embedding outside documentation not hosted at Read the Docs," (note that I re-added the _not_) or
- "make it more stable over time, support embedding documentation not hosted at Read the Docs," (note that I removed the _outside_ and added the _not_)

docs/development/design/embed-api.rst
```diff
@@ -0,0 +1,211 @@
...
+
+
+Goals
+-----
+
+Considering the problems mentioned in the previous section,
+the inclusion of new features and the definition of a contract that works the same for all,
+this document set the following goals for the new version of this endpoint:
+
+* Support external documents hosted outside Read the Docs
```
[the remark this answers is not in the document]
humitos: I have the same problem that I mentioned in https://github.com/readthedocs/readthedocs.org/pull/8222#discussion_r647310946 understanding this sentence 😅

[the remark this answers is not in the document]
humitos: What about re-adding the _not_ as in "Support outside documentation not hosted on Read the Docs"?

docs/development/design/embed-api.rst
humitos: There are some security concerns we should manage very carefully, for sure.

The main goal of this feature is to be able to "embed the whole Internet" 😄, not to get around CORS. As an example, currently, if you are using `sphinx-hoverxref` together with intersphinx for external projects and the official Python documentation is one of those projects, you have to disable intersphinx tooltips _only_ for the Python docs because they are not hosted on RTD and tooltips doesn't work. By supporting embedding content from external sites, we could support this case.

docs/development/design/embed-api.rst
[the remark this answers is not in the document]
stsewd: @humitos the original document has that I think, this is so users don't depend on the docs' domain, but on the slug that is more explicit and that it won't change.

docs/development/design/embed-api.rst
ericholscher: We haven't committed to this for the grant. It would likely start out as an allow list, not allowing any random site on the internet. The main goal here is CDN caching to improve the UX, and making it work across the docs ecosystem, not just things hosted on RTD.

docs/development/design/embed-api.rst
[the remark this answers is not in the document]
ericholscher: I think Anthony is talking about using a client that isn't hoverxref to embed docs content in applications. I think the original copy was referring to having the Embed API embed content that we don't host.

docs/development/design/embed-api.rst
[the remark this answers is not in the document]
ericholscher: Yea, same here. I think Anthony is changing the meaning of this goal.

docs/development/design/embed-api.rst
humitos: No, the contract is not about valid/invalid HTML. We are committing ourselves to return the HTML tag for the `id` requested. Any other response that is outside this contract should be managed as a special case.

Also, we can't commit ourselves to "always return valid HTML" because we don't know if the HTML we are parsing is valid or not. In fact, I just removed that from the "Goals" section.

I updated the design doc with a new section, "The API contract", that exposes our commitment.

docs/development/design/embed-api.rst
humitos: I moved this to a particular endpoint `/api/v3/embed/identifiers/` and define the initial response: "return all the available identifiers for a specific page".

We can expand it later if we need something else and there is a good way to do it.

docs/development/design/embed-api.rst
```diff
@@ -0,0 +1,211 @@
...
+* Rewrite all links returned in the content to make them absolute
+* Always return valid HTML structure
+* Delete HTML tags from the original document if needed
+* Support ``?nwords=`` and ``?nparagraphs=`` to return chunked content
+* Require a valid HTML ``id`` selector
+* Handle special cases for particular doctools (e.g. Sphinx requires to return the ``.parent()`` element for ``dl``)
+* Make explicit the client is asking to handle the special cases (e.g. send ``?doctool=sphinx&version=4.0.1``)
+* Accept only ``?url=`` request GET argument to query the endpoint
+* Add HTTP cache headers to cache responses
+* Allow :abbr:`CORS` from everywhere
```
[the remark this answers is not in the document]
humitos: I updated this by adding "for public projects only".

docs/development/design/embed-api.rst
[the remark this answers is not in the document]
humitos: I didn't find an argument about this in the original document. Please, point it at me if you have one in mind.

Regarding the domain's change, if they change the docs' domain, they will break many other things as well. If they set up a redirect together with the domain change, we could get the new domain's name by making the request and reading `Location:` header as a workaround. That will let us know what's the new domain and we should be able to check our DB again to map the project's slug.

I updated the document to mention this as well.

humitos:
@ericholscher @stsewd @agjohnson I updated the document with the latest conversation we had. I think I covered all the concerns risen here. Let's see if you agree with what I wrote 😄

docs/development/design/embed-api.rst
stsewd: What I mean is that we are committed to return valid html for the well-defined cases, we don't try to fix invalid html, we just don't convert valid html into invalid html.

docs/development/design/embed-api.rst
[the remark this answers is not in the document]
stsewd: https://github.com/readthedocs/readthedocs.org/pull/8039#discussion_r605013162

Checking the location feels overcomplicated, and it also opens our code to SCRF. I think having both ways of querying solves all the issues, people that don't expect the domain to change can use the url, people that want to be more explicit can use the other method.

docs/development/design/embed-api.rst
```diff
@@ -0,0 +1,257 @@
...
+Return the HTML tag (and its children) with the ``id`` selector requested
+and replace all the relative links from its content making them absolute.
+
+.. note::
+
+   Any other case outside this contract will be considered *special* and will be implemented
+   only under ``?doctool=`` and ``?version=`` arguments.
+
+If no ``id`` selector is sent to the request, the content of the first meaningfull HTML tag
+(``<main>``, ``<div role="main">``, etc) identifier found is returned.
```
stsewd, line 71: We should also make it explicit that any requested anchor should be inside the main content, so we don't need to parse the whole html, just things inside the main content.

docs/development/design/embed-api.rst
```diff
@@ -0,0 +1,257 @@
...
+         "language": "en",
+         "path": "development/install.html",
+         "title": "Development Installation",
+         "url": "https://docs.readthedocs.io/en/latest/install.html#set-up-your-environment",
+         "id": "set-up-your-environment",
+         "content": "<div class=\"section\" id=\"development-installation\">\n<h1>Development Installation<a class=\"headerlink\" href=\"https://docs.readthedocs.io/en/stable/development/install.html#development-installation\" title=\"Permalink to this headline\">¶</a></h1>\n ..."
+      }
+
+
+.. http:get:: /api/v3/embed/identifiers/?url=https://docs.readthedocs.io/en/latest/development/install.html
```
stsewd, line 100: Not sure about naming this identifier, I think this is useful to return any meta data about the page itself like the title. We could also have an endpoint to list all pages.

humitos: `/metadata` sounds good to me. @stsewd do you already have in mind some extra data that would be useful to return here?

docs/development/design/embed-api.rst
humitos: I'm not sure to follow what you mean here. If they send an anchor, that maps directly to an `id=` for a particular HTML tag. What would change if it's inside/outside from the "main" content?

docs/development/design/embed-api.rst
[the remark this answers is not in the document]
humitos: > I think having both ways of querying solves all the issues

I'm -1 on having two different ways to use the API.

> it also opens our code to SCRF

I don't think this is a real problem. We can make only one request and cache the `Location:` response for minutes/hours/days if we want.

> people that don't expect the domain to change can use the url, people that want to be more explicit can use the other method.

Nobody expects the domain to change 😄 

In any case, if the problem is how to get the project slug, we could accept a `?project=` argument together with the `?url=`.

docs/development/design/embed-api.rst
stsewd: That we don't need to parse things like the content from the footer, navigation, navbar, etc, content that isn't part of the main content.

docs/development/design/embed-api.rst
stsewd: I have only thought about the title and sections for now. Having a way to list all pages we can build a better UI like the one we have in .org (is hidden and relies on the pages being passed from the template instead of the api)

humitos: 👍🏼 , I changed its name to be `metadata` and returns a dictionary with only one field for now, `identifiers`. This leaves the endpoint opened to return more metadata in the future when we decide what we need exactly.

docs/development/design/embed-api.rst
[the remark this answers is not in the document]
ericholscher: I think `url` is the most user friendly for sure -- whenever I want to use the Embed API, I just grab the URL and use that. The only place I could see the URL being problematic is for some client that knows the RTD data but not the URL, but I don't really think that's a real issue.

docs/development/design/embed-api.rst
ericholscher: I think you are both saying the same thing. We are returning the HTML as it exists on the page. It isn't valid or invalid, it's modified or unmodified, and we plan to serve it unmodified.

docs/development/design/embed-api.rst
ericholscher: You're just suggesting this as a performance optimization right?

stsewd: More or less, so we don't return or try to parse any junk content. But not really strong opinion.

nguyenminhhoang888 (commented):
Good

docs/development/design/embed-api.rst
```diff
@@ -0,0 +1,258 @@
...
+
+
+However, this assumptions may not apply to documentation pages built with a different doctool than Sphinx.
+For this reason, we need to communicate to the API that we want to handle this special cases in the backend.
+This will be done by appending a request GET argument to the Embed API endpoint: ``?doctool=sphinx&version=4.0.1``.
+In this case, the backend will known that has to deal with these special cases.
+
+.. note::
+
+   This leaves the door open to be able to support more special cases (e.g. for other doctools) without breaking the actual behavior.
```
ericholscher, line 188: This seems like a better solution than doing it for all requests, and gives us a way to deprecate things 👍

docs/development/design/embed-api.rst
```diff
@@ -0,0 +1,258 @@
...
+
+Support for external documents
+------------------------------
+
+When the ``?url=`` argument passed belongs to a documentation page not hosted on Read the Docs,
+the endpoint will do an external request to download the HTML file,
+parse it and return the content for the identifier requested.
+
+The whole logic should be the same, the only difference would be where the source HTML comes from.
+
```
ericholscher, line 199:
```suggestion
The whole logic should be the same, the only difference would be where the source HTML comes from.

To start this would be an allowed list of domains for common Sphinx docs projects.
Things like Django & Python, where hoverxref users might commonly want to embed from.
We aren't planning to allow arbitrary HTML from any website.
```

docs/development/design/embed-api.rst
```diff
@@ -0,0 +1,258 @@
...
+At this point there are different possible scenarios:
+
+* The user creates a new ``Domain`` object with ``docs.newdomain.com`` as domain's name.
+  In this case, old links will keep working because we still have the old ``Domain`` object in our database
+  and we can use it to get the project's slug.
+* The user *deletes* the old ``Domain`` besides creating the new one.
+  In this scenario, our query for a ``Domain`` with name ``docs.example.com`` to our database will fail.
+  We will need to do a request to ``docs.example.com`` and check for a 3xx response status code and in that case,
+  we can read the ``Location:`` HTTP header to find the new domain's name for the documentation.
+  Once we have the new domain from the redirect response, we can query our database again to find out the project's slug.
```
ericholscher, line 236: I think mainly you're just saying we should following redirects on the `url` parameter. I think this makes sense (up to a limit of ~5 or so)

docs/development/design/embed-api.rst
```diff
@@ -0,0 +1,258 @@
...
+* Migrate the ``sphinx-hoverxref`` extension to use the new endpoint
+
+Once we have done them, we could check our NGINX logs to find out if there are people still using APIv2,
+contact them and let them know that they have some months to migrate since the endpoint is deprecated and will be removed.
+
+
+Unanswered questions
+--------------------
+
+* How do we distinguish between our APIv3 for resources (models in the database) from these "feature API endpoints"?
```
ericholscher, line 258: Definitely a good question.

ericholscher (commented):
Lots of small language nitpicks. I think in general this is heading in a good direction. I like having a way to know when to apply sphinx magic as needed.

docs/development/design/embed-api.rst
humitos: > I think you are both saying the same thing. We are returning the HTML as it exists on the page.

Yes... and no 😄  We are saying the same about valid/invalid HTML responses.

Apart from that, I'm also saying that I want to have a way to specify "special cases" that don't follow _exctly_ the contract we defined.

docs/development/design/embed-api.rst
```diff
@@ -0,0 +1,258 @@
...
+* Query and parse the ``.html`` file directly (from our storage or from an external request)
+* Rewrite all links returned in the content to make them absolute
+* Require a valid HTML ``id`` selector
+* Accept only ``?url=`` request GET argument to query the endpoint
+* Support ``?nwords=`` and ``?nparagraphs=`` to return chunked content
+* Handle special cases for particular doctools (e.g. Sphinx requires to return the ``.parent()`` element for ``dl``)
+* Make explicit the client is asking to handle the special cases (e.g. send ``?doctool=sphinx&version=4.0.1``)
+* Delete HTML tags from the original document if needed
+* Add HTTP cache headers to cache responses
+* Allow :abbr:`CORS` from everywhere *only* for public projects
```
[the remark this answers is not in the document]
humitos: Hrm, yeah. Good point.

> Delete HTML tags from the original document if needed

This bullet, in particular, refers to this case explained in the document https://github.com/readthedocs/readthedocs.org/pull/8222/files/5063bbc6e900a5c3cbe123e81969f1f2a8e78c80#diff-580a6568f12b1bb9ebf18d8acf51b0e572f1beecdbc94946c81aec8db2ee7944R153

How can I make it more explicit? Maybe adding `(for well-defined special cases)`?

docs/development/design/embed-api.rst
```diff
@@ -0,0 +1,258 @@
...
+Return the HTML tag (and its children) with the ``id`` selector requested
+and replace all the relative links from its content making them absolute.
+
+.. note::
+
+   Any other case outside this contract will be considered *special* and will be implemented
+   only under ``?doctool=`` and ``?version=`` arguments.
+
+If no ``id`` selector is sent to the request, the content of the first meaningfull HTML tag
+(``<main>``, ``<div role="main">``, etc) identifier found is returned.
```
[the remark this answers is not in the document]
humitos: I kept it because it's how it works currently. I don't have a strong opinion and I'm happy to remove it if there is no well-defined use case yet. In fact, I didn't add this originally but then I came back and added this paragraph because I realized that I was removing a feature that we currently have.

docs/development/design/embed-api.rst
humitos: It seems that this changed a little what we talked about. I'm fine starting with an allowed list of domains and grow from there if we find a way to make it secure in the future.

docs/development/design/embed-api.rst
humitos: Yes.

docs/development/design/embed-api.rst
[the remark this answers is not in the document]
ericholscher: Or reference the part of the document where it's explained. It seemed some goals were explained explicitly while others weren't. I think some way of connecting goals to implementation would be good.

docs/development/design/embed-api.rst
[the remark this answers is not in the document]
ericholscher: Yea, I'm not sure. I use this functionality a lot when tested (eg. I just throw a URL at the API and it returns the whole page content) -- but I'm not sure if this is generally useful. It definitely makes the API easier to browser and understand.

I think this functionality probably makes sense, but we should be very clear about what it does (no `etc` :D)

docs/development/design/embed-api.rst
agjohnson: We talked more here, and the use case or feature that I'm describing is currently in a strange place.

More explicitly, I require a list of headings -- basically the toctree on a document -- with heading text, URL to link to the heading, and the heading nesting level or heading nesting as a data structure. This gives me what I need to basically expose the toctree in our application. This is the feature that I had wanted to expose for commercial use -- embedding documentation metadata/headings/etc in customer applications. A hoverxref type extension might be useful in addition here, but is separate. Customers would still need to get metadata out of a particular document in order to inject a toctree into an application view.

So, where I'm at is that I feel like this feature is more of a relic of where the embed API started and it is dragging the direction of the embed API down -- however this might just be my interpretation. The embed API is more focused on having generic support, and so therefore parsing HTML, and what I want is basically exposing the contents of `objects.inv`. I could be talking about a separate feature, or we could still be talking about keeping this as a separate API endpoint.

I am going to research this a bit more, I may be talking about a separate feature entirely at this point.

agjohnson (commented):
Noting a few questions left over after talking through this with @humitos

docs/development/design/embed-api.rst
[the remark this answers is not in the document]
humitos: > but I'm not sure if this is generally useful

I'm not sure we are already using the feature as is right now. However, I opened a PR in sphinx-hoverxref to support `:doc:` that would require this: https://github.com/readthedocs/sphinx-hoverxref/pull/68

> but we should be very clear about what it does (no `etc` :D)

He he, I left the "etc" here because I didn't do the research required to grab the most used ones. The two that I mentioned here are HTML5 and Sphinx default's one, but it may be others. This will be _clear_ in the documentation of the endpoint itself without magic/guessing involved, tho.

docs/development/design/embed-api.rst
```diff
@@ -0,0 +1,260 @@
...
+         "language": "en",
+         "path": "development/install.html",
+         "title": "Development Installation",
+         "url": "https://docs.readthedocs.io/en/latest/install.html#set-up-your-environment",
+         "id": "set-up-your-environment",
+         "content": "<div class=\"section\" id=\"development-installation\">\n<h1>Development Installation<a class=\"headerlink\" href=\"https://docs.readthedocs.io/en/stable/development/install.html#development-installation\" title=\"Permalink to this headline\">¶</a></h1>\n ..."
+      }
+
+
+.. http:get:: /api/v3/embed/metadata/?url=https://docs.readthedocs.io/en/latest/development/install.html
```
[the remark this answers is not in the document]
humitos: > is there a strong need to develop an endpoint like this outside the use case I've described?

Not really. I'm not aware of anyone using it currently.

I originally liked the idea of being able to "see what's available to fetch on a particular page" (list of identifiers), but I'm not sure if that's super useful outside of development workflow.

I may remove it from the document for now. We can come back to this endpoint and re-think how its implementation would be once we have more use cases.

docs/development/design/embed-api.rst
```diff
@@ -0,0 +1,260 @@
...
+
+   <dl class="glossary docutils">
+     <dt id="term-name">Term Name</dt>
+     <dd>Term definition for Term Name</dd>
+   </dl>
+
+
+However, this assumptions may not apply to documentation pages built with a different doctool than Sphinx.
+For this reason, we need to communicate to the API that we want to handle this special cases in the backend.
+This will be done by appending a request GET argument to the Embed API endpoint: ``?doctool=sphinx&version=4.0.1``.
```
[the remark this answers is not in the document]
humitos: Probably, yes.

I didn't dig too much into this, but I wanted to define a way to specify these special cases that require knowing a little more about the HTML structure.

In fact, it may not require the Sphinx's version, but the `docutils` version instead.

> `configuration=html4writer`

Yeah, we will probably need something like this. I'd vote for `writer=html4` and `writer=html5`. I'll update this section with this argument.

docs/development/design/embed-api.rst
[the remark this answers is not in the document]
humitos: Also, this feature combined with `?nparagraphs=` makes more sense.

humitos:
I think we are ready for a final review (cc @stsewd, @ericholscher, @agjohnson) and to start with its implementation. We had discussed a lot of things and it seems that we may add extra features in the future when we have more clear use-cases, but I think that's fine. I'm happy that we are covering the general case in a simple way and the special cases with an explicit argument.

ericholscher:
I think this looks good to get started on. I still have a couple implementation questions, and think we should focus on the core implementation first and then worry about Sphinx edge cases later.

docs/development/design/embed-api.rst
agjohnson: Yup, so we talked about exposing a separate API for that instead.

We need to decide if this feature is going to be a separate product and how users will interact with it. I think we'd be maintaining 1 client JS library that does both embed of hover cards and injection of document header/refs from Sphinx, as the use case is similar.

@humitos Did we talk about what mechanism we'd use if this function was a subpath on the embed API? Are we using Sphinx refs via the SphinxDomain modeling?

docs/development/design/embed-api.rst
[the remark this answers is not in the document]
agjohnson: > Not really. I'm not aware of anyone using it currently.

We haven't promoted it as a feature yet :slightly_smiling_face:

> I'm not sure if that's super useful outside of development workflow

This is a feature for commercial customers and embedding external documentation project output into other sources, like applications.

ericholscher (approved):

docs/development/design/embed-api.rst
```diff
@@ -0,0 +1,290 @@
...
+-----------------------
+
+The v2 is currently widely used by projects using the ``sphinx-hoverxref`` extension.
+Because of that, we need to keep supporting it as-is for a long time.
+
+Next steps on this direction should be:
+
+* Add a note in the documentation mentioning this endpoint is deprecated
+* Promote the usage of the new Embed APIv3
+* Migrate the ``sphinx-hoverxref`` extension to use the new endpoint
```
agjohnson, line 281: If we are specifying deprecation in this document, then the use case for embedding Sphinx refs should also be mentioned here. Deprecation should depend on surfacing an API to expose documentation refs. We need to have a discussion about this end point and how it is implemented/etc next, but that can be separate.

agjohnson (approved):
If we are to separate out the conversation around embedding Sphinx refs/headings, then I think this particular document is very close and we can continue. We need to come back to a separate conversation on a replacement API next.

docs/development/design/embed-api.rst
humitos: > @humitos Did we talk about what mechanism we'd use if this function was a subpath on the embed API? Are we using Sphinx refs via the SphinxDomain modeling?

Nope, we didn't talk about this and I haven't thought too much either. I usually forget that we have `SphinxDomain` model --and I haven't used it yet.

Something like `/api/v3/embed/headers/` may work for this use case. However, we have to keep in mind that:

- it will only work with Sphinx if we use the `SphinxDomain` model
- making it generic by parsing the HTML is probably impossible
- we can use the HTTP request arguments for other doctools, as `?doctool=mkdocs&version=1.0.1` and parse a known HTML structure

docs/development/design/embed-api.rst
humitos: does it make sense to switch the version and API name so we have the version after its name? That way we can release a new version of a particular API without touching the others.

- `/api/search/v3/`
- `/api/footer/v3/`
- `/api/resources/v2/` and `/api/resources/v3/`
- `/api/embed/v3/`
- etc

cc @stsewd @ericholscher

stsewd: That naming feels weird to me, I think we just need to document which endpoints are part of what.

humitos: @stsewd can you expand how that would be and give some examples?

stsewd: this is already done in part for search for example https://docs.readthedocs.io/en/stable/server-side-search.html#api we just need to do the same with the embed api and footer

humitos: @stsewd I don't follow you. That search endpoint is under `/api/v2/` --so we are not differentiating the "Feature APIs" from the "Resources API" there.

```python
print(stsewd.dump("brain").verbose())
```

PR #8224, opened 2021-06-01
Optimize Index time database query

saadmk11:
This change decreases the number of query made to the database for indexing document to elasticsearch by around 50%

### DIFF
--- a/readthedocs/search/documents.py
+++ b/readthedocs/search/documents.py
@@ -168,5 +168,6 @@ class PageDocument(RTDDocTypeMixin, Document):
             queryset
             .internal()
             .exclude(ignore=True)
+            .select_related('version', 'project')
         )
         return queryset


readthedocs/search/documents.py
```diff
@@ -168,5 +168,6 @@ def get_queryset(self):
             queryset
             .internal()
             .exclude(ignore=True)
+            .select_related('version', 'project')
```
saadmk11, line 171:
```suggestion
            .select_related('version', 'project')
            .prefetch_related(
                Prefetch(
                    'sphinx_domains',
                    queryset=SphinxDomain.objects.exclude(
                        domain='std',
                        type__in=['doc', 'label']
                    )
                )
            )
```

We can also add prefetch_related here as well but as we are passing `queryset.iterator()` to `.update()` method of `PageDocument` `.prefetch_related()` will be ignored. this is the same for `ProjectDocument.users` as well

stsewd (approved):
Thanks! Great finding!

PR #8235, opened 2021-06-03
Backport manual indexes

ericholscher:
These indexes were on the .org, but not the .com.
We hit some performance issues around this,
so I checked to see which indexes we had and backported them into Django to standardize them.


stsewd (approved):

readthedocs/builds/models.py
```diff
@@ -693,6 +694,9 @@ class Meta:
             ['version', 'state', 'type'],
             ['date', 'id'],
         ]
+        indexes = [
+            models.Index(fields=['project', 'date']),
+        ]
```
humitos, line 699: I think this is not required if you are already using `db_index=True` in the fields. https://docs.djangoproject.com/en/3.2/ref/models/options/#indexes

humitos (approved):
Should we create these indexes manually to do it `CONCURRENTLY` and avoid killing the db?

readthedocs/builds/models.py
ericholscher: This is indexing them together.

ericholscher:
> Should we create these indexes manually to do it CONCURRENTLY and avoid killing the db?

I don't think these will cause much load -- they are already on .org, and only need to be added to .com.

Issue #8229, opened 2021-06-02
Re-connect a `Project` to a `RemoteRepository`


humitos:
With the work done in Read the Docs for Business about VCS Auth (see https://docs.readthedocs.io/en/stable/commercial/single-sign-on.html#sso-with-vcs-provider-github-bitbucket-or-gitlab) where all the permissions are based on `Project`s being linked/related to `RemoteRepository` and with the work that we are doing to migrate this feature + organization to Read the Docs Community, it will be almost mandatory to support a way for users to "(Re) Connect a Project to a RemoteRepository" to be able to use the VCS Auth feature.

There are some things to consider here:

- we only have `Project.repo` (e.g. `https://github.com/rtfd/sphinx-notfound-page.git`) as a source
- `RemoteRepository` has fields: `clone_url`, `ssh_url` and `html_url` we can compare against
- these URLs won't exactly match in some different scenarios:
  - the GH project is imported living at `rtfd/sphinx-notfound-page` and then moved to `readthedocs/sphinx-notfound-page`
  - the URL used to import the repository has trailing `/`
  - the URL used to import it is SSH and ends without `.git`
  - https:// vs. http://
  - maybe other cases

We will need to manipulate a little the URL to handle these cases:

1. find out the "final URL" for the repository
   - if it's an HTTP URL we can make a request and follow redirects until we get the final URL
2. add/remove the trailing `/`
3. add/remove the final `.git`
4. change http:// to https://
5. others?

Then we can query our database to find out the correct `RemoteRepository`:

```python
# all the generated URLs after manipulating `project.repo`
urls = [
  # https protocol
  'https://github.com/rtfd/sphinx-notfound-page.git',
  'https://github.com/rtfd/sphinx-notfound-page',
  'https://github.com/rtfd/sphinx-notfound-page/',

  # Final URL with https protocol
  'https://github.com/readthedocs/sphinx-notfound-page.git',
  'https://github.com/readthedocs/sphinx-notfound-page',
  'https://github.com/readthedocs/sphinx-notfound-page/',
]

# this query is not super fast
remote_repository = RemoteRepository.objects.filter(
  Q(html_url__in=urls),
  Q(clone_url__in=urls),
  Q(ssh_url__in=urls),
).first()

if remote_repository:
  project.remote_repository = remote_repository
  project.save()
else:
  # show a message to the user that we weren't able to find out the 
  # remote repository and they may need to re-sync their repositories
```

[...]

PR #8259, opened 2021-06-14
Make `Project -> ForeignKey -> RemoteRepository`

### LINKED ISSUE #8232: Change RemoteRepository relationship to Project to a many-to-many relationsihp
In order to allow a repository to be imported twice by the same connected VCS provider account, this relationship should be a many-to-many relationship instead of 1-to-1. We have a few issues blocked on this change, and the migration of our organization structure to community code base is dependent on this as well.

humitos:
Migrate the relation between `Project` and `RemoteRepository` to be a `ForeignKey` from `Project` model, so we can attach multiple projects to the same `RemoteRepository`. This would allow us to import the same repository multiple times and use VCS permissions to manage it.

### 🧑‍🏭 Work done

- add `Project.remote_repository`
- migrate data from `RemoteRepository.project` to `Project.remote_repository`
- drop `RemoteRepository.project`
- change `?expand=project,organization` by `?expand=projects,remote_organization` on `/api/v3/remote/repositories/` endpoint

### 📝 Notes

- `/api/v3/remote/repositories/?expand=project,organization`
  - doesn't make sense to call it `project` in singular now
  - expanding `project` as a RTD object and `organization` as a VCS object looks weird
  - we could use `?expand=projects,organization,remoteorganization` instead
  - we can't `.select_related` anymore
  - we need to update the documentation: https://github.com/readthedocs/readthedocs.org/pull/7510
- all test creating `RemoteRepository` with a `project` attached have to be updated
- `project.remote_repository` does not raise `RemoteRepository.DoesNotExist` anymore; now it just return `None` (e.g. `readthedocs.builds.tasks.send_build_status`)
- require to change all places where `remote_repository.project = project` to `remote_repository.projects.add(project)`

> I just starting collecting some notes here, but I'm sure there will be more things to consider and take care of

Closes #8232 
Related #8229


readthedocs/oauth/migrations/0014_remove_remoterepository_project.py
```diff
@@ -0,0 +1,27 @@
...
+
+from django.db import migrations
+
+
+def migrate_data(apps, schema_editor):
+    RemoteRepository = apps.get_model('oauth', 'RemoteRepository')
+    queryset = RemoteRepository.objects.filter(project__isnull=False).select_related('project')
+    for rr in queryset:
+        rr.project.remote_repository = rr
+        rr.save()
```
humitos, line 11: @agjohnson this data migration from the old field to the new field is an important change to take a look at carefully.

docs/api/v3.rst
```diff
@@ -1814,7 +1814,7 @@ Remote Repository listing
     :query string vcs_provider: return remote repositories for specific vcs provider (``github``, ``gitlab`` or ``bitbucket``)
     :query string organization: return remote repositories for specific remote organization (using remote organization ``slug``)
     :query string expand: allows to add/expand some extra fields in the response.
-                          Allowed values are ``project`` and ``organization``.
+                          Allowed values are ``projects`` and ``remote_organization``.
                           Multiple fields can be passed separated by commas.
```
humitos, line 1818: @agjohnson this a change in the API request/response required to make sense with the new relation. It also reserves `?expand=organization` to be used for `Organization` objects.

readthedocs/oauth/migrations/0014_remove_remoterepository_project.py
humitos: also, this may be slow in production if we have many `RemoteRepositoy` linked to a `Project`.

humitos (commented):
@agjohnson please take a close look at this PR and let me know if it does make sense to you.

readthedocs/oauth/migrations/0014_remove_remoterepository_project.py
```diff
@@ -0,0 +1,27 @@
+# Generated by Django 2.2.22 on 2021-06-14 15:42
+
+from django.db import migrations
+
+
+def migrate_data(apps, schema_editor):
+    RemoteRepository = apps.get_model('oauth', 'RemoteRepository')
+    queryset = RemoteRepository.objects.filter(project__isnull=False).select_related('project')
+    for rr in queryset:
```
saadmk11, line 9:
```suggestion
    for rr in queryset.iterator():
```

saadmk11 (commented):
Changes look great! :100: 

I think we may want to add `remote_repository` to the Project ModelAdmin `raw_id_fields`

`raw_id_fields = ('users', 'main_language_project', 'remote_repository',)`

docs/api/v3.rst
agjohnson: I was going to say we should keep `project` around for backwards compatibility, but this API should really only be internal use anyways. So yeah, agreed this seems fine.

readthedocs/oauth/migrations/0014_remove_remoterepository_project.py
```diff
@@ -0,0 +1,27 @@
...
+
+from django.db import migrations
+
+
+def migrate_data(apps, schema_editor):
+    RemoteRepository = apps.get_model('oauth', 'RemoteRepository')
+    queryset = RemoteRepository.objects.filter(project__isnull=False).select_related('project')
+    for rr in queryset.iterator():
+        rr.project.remote_repository = rr
+        rr.save()
```
agjohnson, line 11: Took me a second to follow this logic, but I *think* this looks good. I suppose I would have concerns about performance here, we will have a few thousand remote repositories to step through individually.

agjohnson (approved):
Looks great! If anyone shares my concern about performance here, we might want to find a way to make the migration faster or operate in chunks. It might be possible with a bulk query :shrug:

readthedocs/oauth/migrations/0014_remove_remoterepository_project.py
humitos: I just checked this: 22k. We should be fine.

humitos: I double-checked this logic and it didn't work.

```
rr.project.remote_repository = rr
```

That doesn't seem to keep the relation.

I also tried, 

```
rr.project.remote_repository_id = rr.pk
```

I don't understand why.

humitos: Meh, I found that I was just saving the wrong object.

We need either:

```
rr.project.save()`
```

or

```
rr.projects.add(rr.project)
```

However, both are producing an error that I'm not sure to understand 😞 

```
database_1  | 2021-06-17 12:05:45.877 UTC [193] ERROR:  cannot ALTER TABLE "projects_project" because it has pending trigger events
database_1  | 2021-06-17 12:05:45.877 UTC [193] STATEMENT:  SET CONSTRAINTS "oauth_remotereposito_project_id_246c363d_fk_projects_" IMMEDIATE; ALTER TABLE "oauth_remoterepository_2020" DROP CONSTRAINT "oauth_remotereposito_project_id_246c363d_fk_projects_"
```

PR #8265, opened 2021-06-15
Sync RemoteRepository for external collaborators

humitos:
When a GitHub user does not belong to a GH organization but does have access to a particular repository inside the GH organization, the `RemoteRepositoryRelation` was not being created. This commit checks for this case and allows the creation of the `RemoteRepositoryRelation`.

Also, if the GH repository was part of a GH organization and then it's moved to a GH user account, we remove the link between ``RemoteRepository`` and ``RemoteOrganization``.

### DIFF
--- a/readthedocs/oauth/services/github.py
+++ b/readthedocs/oauth/services/github.py
@@ -113,14 +113,30 @@ class GitHubService(Service):
                 self.user, self.account
             )
 
-            if repo.organization and repo.organization != organization:
+            # It's possible that a user has access to a repository from within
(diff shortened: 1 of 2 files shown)


humitos:
@saadmk11 hi! if you have some free time, I'd appreciate your review on this PR since it's a kind of edge case that was tricky to find and figure out. Thanks!

saadmk11 (approved):
Quite a tricky edge case, Great Find :100: , The Changes look good to me! :+1:

stsewd (approved):

PR #8282, opened 2021-06-23
oauth webhook: check the `Project` has a `RemoteRepository`

humitos:
Before we were catching the exception because the relationship was a OneToOne.
Now that it's a ForeignKey, the field could be `None` instead.

Fixes this issue: https://sentry.io/organizations/read-the-docs/issues/2470813831/

### DIFF
--- a/readthedocs/oauth/utils.py
+++ b/readthedocs/oauth/utils.py
@@ -30,7 +30,7 @@ def update_webhook(project, integration, request=None):
         return None
 
     updated = False
-    try:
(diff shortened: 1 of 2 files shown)


stsewd (approved):

ericholscher:
Seems this probably needs a test, since it didn't break previously.

humitos:
@ericholscher yes, I'll try to add a test for this. I just did this quickly in case you wanted to hotfix it last Wednesday.

PR #8283, opened 2021-06-23
Add new guide about Jupyter in Sphinx

astrojuanlu:
Still work in progress, but I already decided what sections to cover and created a local project to attach screenshots. I am not done with the tricky bits, but at least I now understood (and hopefully conveyed) the relationship between nbsphinx and MyST-NB for processing `.ipynb` notebooks. Will add a few extra things for completeness.


astrojuanlu:
Rendered version: https://docs--8283.org.readthedocs.build/en/8283/guides/jupyter.html

astrojuanlu:
This is now ready for review. I decided to remove the "Live code execution on HTML documentation" because it's not strictly related to Jupyter notebooks, and the current content is dense enough already.

astrojuanlu:
And by the way, this is the project I used to test everything and generate the screenshots: https://github.com/astrojuanlu/jupyter-on-sphinx

ericholscher:
@astrojuanlu probably better to use https://github.com/readthedocs/test-builds, which is what that repo is for.

docs/guides/jupyter.rst
```diff
@@ -0,0 +1,293 @@
+How to use Jupyter notebooks in Sphinx
+======================================
+
+By default, `Sphinx only supports reStructuredText source files for
+documentation <https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-source_suffix>`_.
+Luckily, there are a few extensions to also allow Jupyter notebooks as source files.
+
```
ericholscher, line 7: This feels like a pretty abrupt introduction. We should cover a bit more about what this tutorial is about and why we're writing it. 

```suggestion
Jupyter notebooks are a common format to represent calculations that have been done on a set of data.
Embedding them in your Sphinx documentation allows an integration between data calculations and documentation,
which can provide a great experience for examples and other pages in your documentation. 
There are a few extensions that allow integrating these two tools,
and this document will explain how to achieve some of the most commonly requested features. 

```

docs/guides/jupyter.rst
```diff
@@ -0,0 +1,293 @@
...
+.. _ipynb-notebooks-sphinx:
+
+Including ``.ipynb`` notebooks in Sphinx documentation
+------------------------------------------------------
+
+There are two main extensions that add support for ``.ipynb`` as source files in Sphinx:
+nbsphinx_ and MyST-NB_. Even though they have some differences,
+their basic mode of operation is very similar. However, they are mutually exclusive,
```
ericholscher, line 15:
```suggestion
their basic mode of operation is very similar. They can't both be used,
```

docs/guides/jupyter.rst
```diff
@@ -0,0 +1,293 @@
...
+so you have to choose one of them.
```
ericholscher, line 16: This feels scary. Are we going to tell users which one to use?! How do they work? 

We just go right into step-by-step content after this introduction, which feels weird.

docs/guides/jupyter.rst
```diff
@@ -0,0 +1,293 @@
...
+   the former renders the cell numbers by default,
+   whereas the latter doesn't.
+
+Rendering interactive widgets
+-----------------------------
+
+You can also embed interactive widgets from Jupyter notebooks created using ipywidgets_
+on HTML Sphinx documentation. This includes basic widgets from ipywidgets_ and also
+more complex ones,
+like `ipyleaflet`_ visualizations.
```
ericholscher, line 88: Why would I want to do this? Without having a lot of context about Jupyter, I can't really parse the goal of interactive widgets -- we should explain them.

docs/guides/jupyter.rst
```diff
@@ -0,0 +1,293 @@
...
+Creating galleries of examples using notebooks
+----------------------------------------------
+
+If you want to create an HTML gallery of examples for your project,
+`Sphinx-Gallery <https://sphinx-gallery.github.io/>`_
+offers a very convenient way of doing so from Sphinx,
+including thumbnail generation.
+However, `it works with Python scripts rather than
+notebooks <https://github.com/sphinx-gallery/sphinx-gallery/issues/245>`_,
+which might not be what you want.
```
ericholscher, line 155: This seems like an odd guide to place this content. Should we have another guide on examples?

ericholscher, line 155: I got confused: "why is this guide telling me about things that don't work in notebooks?" It feels weird to include above `nbsphinx`, since it's talking about something that doesn't work. We should re-order this, and frame it as a way of integrating sphinx & sphinx-gallery instead of saying it doesn't work with notebooks.

docs/guides/jupyter.rst
```diff
@@ -0,0 +1,293 @@
...
+the slider will render correctly in Sphinx.
+
+.. figure:: /_static/images/guides/widget-html.gif
+   :width: 80%
+   :align: center
+   :alt: Interactive widget rendered in HTML by Sphinx
+
+   Interactive widget rendered in HTML by Sphinx
+
+To see more elaborate examples:
```
ericholscher, line 129: This is really good 👍

docs/guides/jupyter.rst
```diff
@@ -0,0 +1,293 @@
...
+  with several interactive backends and `examples in their
+  documentation <https://docs.pyvista.org/index.html#maps-and-geoscience>`_ as well.
+
+.. warning::
+
+   Although widgets themselves can be embedded in HTML as shown above,
+   `events <https://ipywidgets.readthedocs.io/en/latest/examples/Widget%20Events.html>`_
+   require a backend (kernel) to execute.
+   Therefore, ``@interact``, ``.observe``, and related functionalities relying on them
+   will not work as expected.
```
ericholscher, line 144: Great warning.

docs/guides/jupyter.rst
```diff
@@ -0,0 +1,293 @@
...
+   extensions = [
+      'nbsphinx',
+      'sphinx_gallery.load_style',
+   ]
+
+After doing that, there are two ways to create the gallery:
+
+- From a reStructuredText source file, using the ``.. nbgallery::`` directive,
+  `as showcased in the
+  documentation <https://nbsphinx.readthedocs.io/en/latest/a-normal-rst-file.html#thumbnail-galleries>`_.
```
ericholscher, line 175: Should we use intersphinx here?

docs/guides/jupyter.rst
```diff
@@ -0,0 +1,293 @@
...
+- From a Jupyter notebook, adding a ``"nbsphinx-gallery"`` tag to the metadata of a cell.
+  Each editor has a different way of modifying the cell metadata (see figure below).
```
ericholscher, line 177: Do we want to recommend this approach? Seems like the RST way of doing it is way more obvious and nicer -- and doesn't require us talking about editors.

docs/guides/jupyter.rst
```diff
@@ -0,0 +1,293 @@
...
+Using notebooks in other formats
+--------------------------------
+
+Jupyter notebooks in ``.ipynb`` format
+(as described in the `nbformat
+documentation <https://nbformat.readthedocs.io/en/latest/>`_)
+are by far the most widely used for historical reasons.
+
+However, to compensate some of the disadvantages of the ``.ipynb`` format
+(like cumbersome integration with version control systems),
```
ericholscher, line 215: This is a good `why` section. It explains to me why I would want my notebooks in other formats. Is this something that we recommend? Should we?

docs/guides/jupyter.rst
```diff
@@ -0,0 +1,293 @@
...
+
+One of such formats is the `MyST Markdown
+format <https://jupytext.readthedocs.io/en/latest/formats.html#myst-markdown>`_,
+which is based on `MyST`_, an extensible flavor of Markdown
+that includes some features from reStructuredText.
+
+If you are interested in using alternative formats for Jupyter notebooks,
+nowadays there are two main ways to include them
+in your Sphinx documentation:
+
```
ericholscher, line 227:
```suggestion
If you are using alternative formats for Jupyter notebooks,
there are two main ways to include them
in your Sphinx documentation:
```

ericholscher (commented):
This is a good start, but it feels a bit disjoint. Without having a lot of knowledge of Jupyter, I didn't really get much understand of the value of these tasks or what my goals with them would be. I feel like we need a lot more context and explanation of why this is useful.

I'd also really like us to have an opinion more. It seems like Myst NB is simpler and probably the better approach? Why does Myst-NB exist is nbsphinx does the exact same thing? Should I use Jupyterlab or something else for editing? I feel like we could be providing a lot more guidance to people here, or at least making some assumptions to clarify the copy and workflow.

docs/guides/jupyter.rst
astrojuanlu: Yes! In fact, it's done in the next paragraph. Perhaps I can rephrase this to clarify that?

astrojuanlu:
Thanks @ericholscher for the review! Point well taken, this is mostly useful for people that already know Jupyter and they know what they want. It also contains some examples, even though I was trying to avoid sounding too opinionated. Will rework it a bit so it contains more general context and guidance, and more explicit recommendations.

docs/guides/jupyter.rst
astrojuanlu: Good point. I took parts of this and rewrote them slightly.

docs/guides/jupyter.rst
astrojuanlu: It _is_ scary 😄 now seriously, I was trying to avoid getting into too much technical detail, but you're right we should distill that information in a less intimidating way. Rewriting.

docs/guides/jupyter.rst
astrojuanlu: Reworked this section.

docs/guides/jupyter.rst
astrojuanlu: I will add some more technical background on the tradeoffs.

docs/guides/jupyter.rst
astrojuanlu: Committed locally verbatim 👍🏽

docs/guides/jupyter.rst
astrojuanlu: Good point, I rewrote this paragraph and now I think it's more clear.

astrojuanlu: Not necessarily, Sphinx-Gallery workflow can be weird for people used to working with notebooks. I rewrote this paragraph.

docs/guides/jupyter.rst
astrojuanlu: If only I had something like https://webknjaz.github.io/intersphinx-untangled/ to quickly inspect the inventory of a project, to see if this would make sense or not...!

I don't see any labels defined in https://github.com/spatialaudio/nbsphinx/blame/master/doc/a-normal-rst-file.rst, so I suppose we can't? But I might be missing some information here. cc @humitos

docs/guides/jupyter.rst
astrojuanlu: Good point, will add a bit more background at the beginning.

astrojuanlu:
Did a handful of changes, the guide is now longer but hopefully much better.

Pinging also @choldgraf, @chrisjsewell, and @mgeier since I used your input for some parts of it 🚀

docs/guides/jupyter.rst
```diff
@@ -0,0 +1,377 @@
...
+nbsphinx_ and MyST-NB_. Although they have similar intent and basic functionality,
+there are some differences:
+
+- nsphinx uses `pandoc <https://pandoc.org/>`_ to convert the Markdown from Jupyter notebooks
+  to reStructuredText and then to docutils AST,
+  whereas MyST-NB uses `MyST-Parser`_ to directly convert the Markdown text to docutils AST.
+  Both Markdown flavors are mostly equal, but they have some differences.
+- Both nbsphinx and MyST-NB support canonical ``.ipynb`` notebooks.
+  nbsphinx can process any format supported by `jupytext`_,
+  whereas MyST-NB focuses on MyST Markdown Notebooks
```
chrisjsewell, line 32: myst-nb can also support any format: https://myst-nb.readthedocs.io/en/latest/examples/custom-formats.html

docs/guides/jupyter.rst
```diff
@@ -0,0 +1,377 @@
...
+  whereas MyST-NB doesn't.
+
+.. _nbsphinx: https://nbsphinx.readthedocs.io/
+.. _MyST-NB: https://myst-nb.readthedocs.io/
+.. _MyST-Parser: https://myst-parser.readthedocs.io/
+.. _jupytext: https://jupytext.readthedocs.io/
+
+Using which one to use depends on your use case. As general recommendations:
+
+- If you want to use :ref:`other notebook formats <other-formats>`
```
chrisjsewell, line 53: And yeh here again that’s not a
differentiator

docs/guides/jupyter.rst
```diff
@@ -0,0 +1,377 @@
...
+  if you don't need any of nbsphinx or jupytext functionalities.
+
+.. _MyST: https://myst-parser.readthedocs.io/
+
+.. note::
+
+   In summary: both nbsphinx and MyST-NB
+   can parse ``.ipynb`` notebooks and include them in Sphinx documentation.
+   In addition, MyST-NB can read MyST Markdown notebooks,
+   and nbsphinx can read any alternative formats understood by jupytext.
```
chrisjsewell, line 322: And here, a lot of fixation on the alternative formats lol

docs/guides/jupyter.rst
astrojuanlu: Thanks for the clarification! That was a misunderstanding on my side, will fix this (applies to your other comments below)

docs/guides/jupyter.rst
```diff
@@ -0,0 +1,377 @@
...
+- If you want to use :ref:`other notebook formats <other-formats>`
+  or :ref:`generate a thumbnail gallery from your notebooks <notebook-galleries>`,
+  nbsphinx is the right choice.
+- If you will be focusing on MyST Markdown notebooks
+  or want to leverage a more optimized execution workflow
+  and a more streamlined parsing mechanism,
+  you should use MyST-NB.
+
+For the rest of this document we will focus on `nbsphinx`_
+because it is more flexible and offers some extra functionality,
```
chrisjsewell, line 62: I would politely disagree 😜 again myst-nb supports all formats, and there is also things like glue: https://myst-nb.readthedocs.io/en/latest/use/glue.html.  Obviously I’m biased though lol

docs/guides/jupyter.rst
```diff
@@ -0,0 +1,377 @@
...
+
+  - Clear outputs before commit.
+    Minimizes conflicts, but might defeat the purpose of notebooks themselves,
+    since the computation results are not stored.
+  - Use tools like `nbdime <https://nbdime.readthedocs.io/>`_ (open source)
+    or `ReviewNB <https://www.reviewnb.com/>`_ (proprietary)
+    to improve the review process.
+  - Use a different collaboration workflow that doesn't involve notebooks.
+
+- Replace ``.ipynb`` notebooks with `a text-based
```
chrisjsewell, line 285: One think that maybe be worth noting here, is that if you use markdown files in myst-nb, error/warning messages will point to the correct line in the source file: https://myst-nb.readthedocs.io/en/latest/use/start.html#sphinx-error-reporting

docs/guides/jupyter.rst
astrojuanlu: I am obviously biased too, since I've been using nbsphinx for 4 years and I know it much better than MyST-NB :) I will keep the examples with nbsphinx but try to make it look more neutral.

Plus: I didn't know about glue!

docs/guides/jupyter.rst
astrojuanlu: Thanks!

astrojuanlu:
I amended the guide where it said that MyST-NB could not read alternative formats. IMHO the project needs a different name :)

Because of familiarity, and because the examples we showed would be almost identical in both systems, I kept nbsphinx, but toned down our recommendation. I think the main differentiator of nbsphinx is the gallery feature (which is super powerful in my opinion).

docs/guides/jupyter.rst
ericholscher: @astrojuanlu isn't that just a web-based version of the tool built into intersphinx? https://www.sphinx-doc.org/en/master/usage/extensions/intersphinx.html#showing-all-links-of-an-intersphinx-mapping-file

chrisjsewell:
thanks @astrojuanlu 👍 

> IMHO the project needs a different name

Happy to hear recommendations 😄 

> I think the main differentiator of nbsphinx is the gallery feature (which is super powerful in my opinion).

Yeh interesting feedback thanks.
TBH its not something I have really looked at much; but I think it would be pretty easy to add something for myst-nb for this. I guess to tie-in with sphinx-gallery, for people already using that, but also I could envisage having a more "generic" solution with https://sphinx-panels.readthedocs.io

docs/guides/jupyter.rst
ericholscher:
```
	/a-normal-rst-file.rst#thumbnail-galleries Thumbnail Galleries                     : a-normal-rst-file.html#thumbnail-galleries
```

astrojuanlu: TIL about `python -msphinx.ext.intersphinx` 👀 looking into this

astrojuanlu: Right, I had a misconception about how `:ref:` worked that had nothing to do with the inventory. Basically I didn't know I could do this:

```
:ref:`as showcased in the documentation <nbsphinx:/a-normal-rst-file.rst#thumbnail-galleries>`
```

Reviewing the whole guide now. Thanks @ericholscher !

astrojuanlu: Done. I converted some of the references to intersphinx, with these criteria:

- All the "core" references, except for a couple of Sphinx references I didn't know how to link.
- Examples for "social proof" (poliastro, ipyleaflet) _not_ converted, to avoid overloading the intersphinx mapping (not sure if I should worry about this?)

astrojuanlu:
Also, I took the liberty of adding one extra commit to address #8291, let me know if it's too early for it.

docs/guides/jupyter.rst
```diff
@@ -0,0 +1,363 @@
...
+However, there are some differences between them:
+
+- nsphinx uses `pandoc <https://pandoc.org/>`_ to convert the Markdown from Jupyter notebooks
+  to reStructuredText and then to docutils AST,
+  whereas MyST-NB uses `MyST-Parser`_ to directly convert the Markdown text to docutils AST.
+  Both Markdown flavors are mostly equal, but they have some differences.
+- nbsphinx executes each notebook during the parsing phase,
+  whereas MyST-NB can execute all notebooks up front
+  and cache them with `jupyter-cache <https://jupyter-cache.readthedocs.io/>`_.
+  This can result in shorter build times when when notebooks are slightly modified
```
choldgraf, line 35:
```suggestion
  This can result in shorter build times when when notebooks are modified
```

docs/guides/jupyter.rst
```diff
@@ -0,0 +1,363 @@
+How to use Jupyter notebooks in Sphinx
+======================================
+
+`Jupyter`_ notebooks are a popular tool to describe computational narratives
```
choldgraf, line 4: meta point (feel free to disregard) - I am a big fan of [sembr](https://sembr.org/) for deciding when to do line breaks, rather than basing it on character length. If it's too much hassle to think about the "right" place to break lines, I've found that using "one sentence per line" is a nice rubric.

docs/guides/jupyter.rst
```diff
@@ -0,0 +1,363 @@
...
+  whereas MyST-NB doesn't.
+
+.. _nbsphinx: https://nbsphinx.readthedocs.io/
+.. _MyST-NB: https://myst-nb.readthedocs.io/
+.. _MyST-Parser: https://myst-parser.readthedocs.io/
+.. _jupytext: https://jupytext.readthedocs.io/
+
+Using which one to use depends on your use case. As general recommendations:
+
+- If you want to use :ref:`other notebook formats <other-formats>`
```
choldgraf, line 56: I'm leaning towards just not telling people what they "should" do - it feels like that might become semi-arbitrary and opens up the tutorial to debate between the two projects. It seems cleaner to just focus on the "feature matrix" and UI/UX of each tool, rather than to give opinionated recommendations. Does that make sense?

docs/guides/jupyter.rst
```diff
@@ -0,0 +1,363 @@
...
+- The classical Jupyter Notebook interface
+  provides a "Save Notebook Widget State" action in the "Widgets" menu,
+  :ref:`as explained in the ipywidgets
+  documentation <ipywidgets:/embedding.md#embedding-widgets-in-html-web-pages>`.
+  You need to click it before exporting your notebook to HTML.
+- JupyterLab provides a "Save Widget State Automatically" option in the "Settings" menu.
+  You need to leave it checked so that widget state is automatically saved.
+- In Visual Studio Code `it's not possible to save the widget
+  state <https://github.com/microsoft/vscode-jupyter/issues/4404>`_
+  at the time of writing.
```
choldgraf, line 154:
```suggestion
  at the time of writing (June 2021).
```

docs/guides/jupyter.rst
```diff
@@ -0,0 +1,363 @@
...
+
+.. warning::
+
+   Although widgets themselves can be embedded in HTML as shown above,
+   :doc:`events <ipywidgets:examples/Widget Events>`
+   require a backend (kernel) to execute.
+   Therefore, ``@interact``, ``.observe``, and related functionalities relying on them
+   will not work as expected.
+
+.. _notebook-galleries:
```
choldgraf, line 192:
```suggestion
.. warning::

   Some interactive cell outputs expect certain **javascript dependencies** to be loaded in order to correctly display the output (e.g., [`plotly.js`](https://github.com/plotly/plotly.js/) for Plotly outputs).
   In Jupyter interfaces, these libraries are usually loaded via extensions, but building your documentation with notebooks requires you to lead the libraries yourself.

.. _notebook-galleries:
```

choldgraf: I'd also cross-link to some documentation about how people can load their own JS libraries.

docs/guides/jupyter.rst
```diff
@@ -0,0 +1,363 @@
...
+There are two main extensions that add support Jupyter notebooks as source files in Sphinx:
+nbsphinx_ and MyST-NB_. They have similar intent and basic functionality:
+both can read notebooks in ``.ipynb`` and additional formats supported by `jupytext`_,
+and are configured in a similar way.
+However, there are some differences between them:
+
+- nsphinx uses `pandoc <https://pandoc.org/>`_ to convert the Markdown from Jupyter notebooks
+  to reStructuredText and then to docutils AST,
+  whereas MyST-NB uses `MyST-Parser`_ to directly convert the Markdown text to docutils AST.
+  Both Markdown flavors are mostly equal, but they have some differences.
```
choldgraf, line 31: Rather than focusing on the "parser", I'd probably focus on the "flavor" of markdown that each supports. Most users don't really know what a "parser" is anyway, and rather they care more about "what syntax can I write?". I'd re-focus around "CommonMark/Jupyter markdown" with nbsphinx and MyST Markdown with MyST-NB.

docs/guides/jupyter.rst
```diff
@@ -0,0 +1,363 @@
...
+
+   Although widgets themselves can be embedded in HTML as shown above,
+   :doc:`events <ipywidgets:examples/Widget Events>`
+   require a backend (kernel) to execute.
+   Therefore, ``@interact``, ``.observe``, and related functionalities relying on them
+   will not work as expected.
+
+.. _notebook-galleries:
+
+Creating galleries of examples using notebooks
```
choldgraf, line 194: to be honest this one feels a little bit out-of-place given the tutorial's topic. I feel like "generate a gallery from multiple pages" is a generic use-case that is not notebook-specific (though it is a cool feature of nbsphinx). I almost feel like it deserves its own section (that covers `sphinx-gallery` primarily, and links to `nbsphinx` to talk about notebook-specific functionality) rather than to talk about galleries in the context of notebooks.

docs/guides/jupyter.rst
```diff
@@ -0,0 +1,363 @@
...
+
+If you are using alternative formats for Jupyter notebooks,
+you can include them in your Sphinx documentation
+using either `nbsphinx`_ or `MyST-NB`_
+(:ref:`see above <nb-background>`
+for more information about the differences between them).
+
+.. _other-formats:
+
+Using notebooks in other formats
```
choldgraf, line 313: I'd probably move this higher up because it's the information most users will be interested in, and move the "background explanation" stuff about notebook formats beneath it with a cross-link.

docs/guides/jupyter.rst
```diff
@@ -0,0 +1,363 @@
...
+- If you want to use :ref:`other notebook formats <other-formats>`
+  or :ref:`generate a thumbnail gallery from your notebooks <notebook-galleries>`,
+  nbsphinx is the right choice.
+- If you want to leverage a more optimized execution workflow
+  and a more streamlined parsing mechanism,
+  as well as some of the unique MyST-NB functionalities
+  (``glue``, better error reporting)
+  you should use MyST-NB.
+
+For the rest of this document we will focus on `nbsphinx`_,
```
choldgraf, line 65: Just an observation - I think that there are only 2-3 places below where there is tool-specific syntax. And the difference between the two is minimal. What if you included some tabs that let people switch between `nbsphinx` and `myst-nb` configuration?

choldgraf (commented):
A few comments and suggested edits from me! In general this is a nice start that does a good job of covering the basics and several options! I like how you are following up sections with links to more in-depth documentation 👍

docs/guides/jupyter.rst
```diff
@@ -0,0 +1,363 @@
...
+----------------------------------------------
+
+There are two main extensions that add support Jupyter notebooks as source files in Sphinx:
+nbsphinx_ and MyST-NB_. They have similar intent and basic functionality:
+both can read notebooks in ``.ipynb`` and additional formats supported by `jupytext`_,
+and are configured in a similar way.
+However, there are some differences between them:
+
+- nsphinx uses `pandoc <https://pandoc.org/>`_ to convert the Markdown from Jupyter notebooks
+  to reStructuredText and then to docutils AST,
```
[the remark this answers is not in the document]
astrojuanlu: Good catch, thanks! Will add a link to https://docutils.sourceforge.io/docs/ref/doctree.html

docs/guides/jupyter.rst
astrojuanlu: Yeah, even if it doesn't look like it, I'm a fan too! However I still haven't mastered the technique, or often put two sentences in one line. Thanks for the reminder though :)

docs/guides/jupyter.rst
astrojuanlu: I agree that doing specific recommendations might be contentious. However, the two projects are very similar, and I (as a user) was very confused about the differences between them, so if we don't offer at least a small amount of guidance, I think we might leave our users confused or looking for answers. I was very careful here to weigh the pros and cons, and @chrisjsewell corrected me in several places where I was wrong about MyST-NB. What do you think @ericholscher ?

docs/guides/jupyter.rst
astrojuanlu: Good idea! I see we are already using `sphinx_tabs` 👍🏽

docs/guides/jupyter.rst
astrojuanlu: I'm not sure if it's worth adding this warning. I tried creating a project from scratch with `ipywidgets` and it worked without adding anything, and on the other hand Plotly has some rendering modes that either bundle or load using `require.js` the necessary code https://plotly.com/python/renderers/. But of course there might be other cases where some action is actually needed?

astrojuanlu: Maybe

```
.. note::

   If your widgets need some additional JavaScript libraries,
   you can add them using ``Sphinx.add_js_file``
```

?

docs/guides/jupyter.rst
astrojuanlu: I'm not sure. While "generating galleries from notebooks" is something that several projects are doing or seeking to do, "generating galleries from multiple pages" sounds like too abstract of a use case to have a guide dedicated to it.

For me as a user, creating galleries is 40 % of the reason why notebooks are cool, since thumbnails can be automatically generated and notebooks are an excellent choice for creating how-to guides. About sphinx-gallery, I tried using it in a personal project that was using notebooks for examples already https://github.com/poliastro/poliastro/pull/828 but the workflow ended up being more complex, so we abandoned this approach.

docs/guides/jupyter.rst
astrojuanlu: Good point, I will move the "background" to the end of the document (and change "above" by "below" etc)

docs/guides/jupyter.rst
astrojuanlu: I think talking about "which parser" might be interesting technical information difficult to find elsewhere, but you're right that I should more explicitly differentiate what this actually means in practical terms. I will rework this part.

docs/guides/jupyter.rst
choldgraf: I'll leave this one to you - I think it comes down to whether "creating a gallery of notebooks" is a "core part of the notebook experience". In my opinion, it isn't, but I don't feel strongly about it, so if others believe strongly that it _is_, then I'll defer

docs/guides/jupyter.rst
choldgraf: hmm - that sounds like a reasonable first step, though you might get users who are confused because their JS works in JupyterLab but not in the Sphinx HTML (Jupyter Book has a ton of issues opened around this)

astrojuanlu: Done!

astrojuanlu:
Thanks @choldgraf for your comments 🙏🏽 I did another round of changes. Apart from some outstanding questions about scope or intent, I think this is it. cc @readthedocs/core

chrisjsewell:
Thanks @astrojuanlu one question though; where are these guides in the docs lol?!

I tried https://docs--8283.org.readthedocs.build/en/8283/ and could not see them, and even searched for Jupyter and the result was empty

docs/guides/jupyter.rst
ericholscher: I guess I still don't understand the differences, or why they both exist. If there is a short explanation of when one would be better, we should include that. If there are basically just 2 libraries that do the same thing, we should mention that, and clarify when it's better to use one. I don't know if we can do much more than that.

docs/guides/jupyter.rst
ericholscher: I had the same feedback -- I think we probably want a different guide on galleries, and to link from that from this page. https://github.com/readthedocs/readthedocs.org/pull/8283#discussion_r658350299

ericholscher:
@chrisjsewell unfortunately you need to go to https://docs.readthedocs.io/en/latest/guides/tools.html. We should fix this tho, and probably just have a high-level guides section that includes them all on the front page.

docs/guides/jupyter.rst
```diff
@@ -0,0 +1,398 @@
...
+   :alt: Interactive widget rendered in HTML by Sphinx
+
+   Interactive widget rendered in HTML by Sphinx
+
+To see more elaborate examples:
+
+- `ipyleaflet`_ provides several widgets for interactive maps,
+  and renders live versions of them `in their
+  documentation <https://ipyleaflet.readthedocs.io/en/latest/api_reference/velocity.html>`_.
+- `PyVista <https://docs.pyvista.org/>`_ is used for scientific 3D visualization
```
humitos, line 137: pyvista seems to be a Sphinx project as well. We could use intersphinx on it too.

docs/guides/jupyter.rst
```diff
@@ -0,0 +1,398 @@
...
+   :align: center
+   :alt: Interactive widget rendered in HTML by Sphinx
+
+   Interactive widget rendered in HTML by Sphinx
+
+To see more elaborate examples:
+
+- `ipyleaflet`_ provides several widgets for interactive maps,
+  and renders live versions of them `in their
+  documentation <https://ipyleaflet.readthedocs.io/en/latest/api_reference/velocity.html>`_.
```
humitos, line 136: Is there any reason to not use intersphinx on this link?

docs/guides/jupyter.rst
```diff
@@ -0,0 +1,398 @@
...
+   :align: center
+   :alt: Simple thumbnail gallery created using nbsphinx
+
+   Simple thumbnail gallery created using nbsphinx
+
+To see some examples of notebook galleries in the wild:
+
+- `poliastro <https://docs.poliastro.space/>`_ offers tools for interactive Astrodynamics in Python,
+  and features `several examples and how-to guides using
+  notebooks <https://docs.poliastro.space/en/stable/gallery.html>`_
```
humitos, line 281: Why not using intersphinx here as well?

docs/guides/jupyter.rst
```diff
@@ -0,0 +1,398 @@
...
+.. figure:: /_static/images/guides/thumbnail-gallery.png
+   :width: 80%
+   :align: center
+   :alt: Simple thumbnail gallery created using nbsphinx
+
+   Simple thumbnail gallery created using nbsphinx
+
+To see some examples of notebook galleries in the wild:
+
+- `poliastro <https://docs.poliastro.space/>`_ offers tools for interactive Astrodynamics in Python,
```
humitos, line 279: It seems that there some bullets missing from this list. I was expecting to see more examples but there is just one project listed here. I think we should add more bullets or remove the listing and left only one example.

docs/guides/jupyter.rst
```diff
@@ -0,0 +1,398 @@
...
+
+.. _nb-background:
+
+Existing relevant extensions
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+There are two main extensions that add support Jupyter notebooks as source files in Sphinx:
+nbsphinx_ and MyST-NB_. They have similar intent and basic functionality:
+both can read notebooks in ``.ipynb`` and additional formats supported by `jupytext`_,
+and are configured in a similar way.
```
humitos, line 298: I think we already talked about them at the beginning of the document, does it make sense to rephrase this saying something like: "as we have seen, there two main extensions... let's jump into their differences" or similar?

I'm not sure if we presented the extensions previously but only mentioned that there are two ways of doing the guide. It may make sense to make this presentation earlier in the doc and here just refer to that?

docs/guides/jupyter.rst
```diff
@@ -0,0 +1,398 @@
...
+  whereas MyST-NB uses `MyST-Parser`_
+  to directly convert the Markdown text to docutils AST.
+  Therefore, nbsphinx assumes `pandoc flavored Markdown <https://pandoc.org/MANUAL.html#pandocs-markdown>`_,
+  whereas MyST-NB uses `MyST flavored Markdown <https://myst-parser.readthedocs.io/en/latest/syntax/syntax.html>`_.
+  Both Markdown flavors are mostly equal,
+  but they have some differences.
+- nbsphinx executes each notebook during the parsing phase,
+  whereas MyST-NB can execute all notebooks up front
+  and cache them with `jupyter-cache <https://jupyter-cache.readthedocs.io/>`_.
+  This can result in shorter build times when when notebooks are modified
```
humitos, line 313:
```suggestion
  This can result in shorter build times when notebooks are modified
```

docs/guides/jupyter.rst
```diff
@@ -0,0 +1,398 @@
...
+- The visual appearance of code cells and their outputs is slightly different:
+  nbsphinx renders the cell numbers by default,
+  whereas MyST-NB doesn't.
+
+.. _nbsphinx: https://nbsphinx.readthedocs.io/
+.. _MyST-NB: https://myst-nb.readthedocs.io/
+.. _MyST-Parser: https://myst-parser.readthedocs.io/
+.. _jupytext: https://jupytext.readthedocs.io/
+
+Using which one to use depends on your use case. As general recommendations:
```
humitos, line 332:
```suggestion
Deciding which one to use depends on your use case. As general recommendations:
```

(I'm not sure what I'm doing 🐶 , but sounded weird "Using which on to use" to me)

docs/guides/jupyter.rst
```diff
@@ -0,0 +1,398 @@
...
+  nbsphinx is the right choice.
+- If you want to leverage a more optimized execution workflow
+  and a more streamlined parsing mechanism,
+  as well as some of the unique MyST-NB functionalities
+  (``glue``, better error reporting)
+  you should use MyST-NB.
+
+For the rest of this document we will focus on `nbsphinx`_,
+and we will point out the differences with `MyST-NB`_ where appropriate.
+Notice that they can't both be used at the same time.
```
humitos, line 345: It seems we can delete this paragraph, the guide is about to finish and we already presented both options.

docs/guides/jupyter.rst
```diff
@@ -0,0 +1,398 @@
...
+- Pairing ``.ipynb`` notebooks with a text-based format,
+  and putting the text-based file in version control,
+  as suggested in the :doc:`jupytext documentation <jupytext:paired-notebooks>`.
+  This solution has the best of both worlds.
+  In some rare cases you might experience synchronization issues between both files.
+
+These approaches are not mutually exclusive,
+nor you have to use a single format for all your notebooks.
+For the examples in this document, we will use the :ref:`MyST Markdown
+format <jupytext:formats.md#myst-markdown>`.
```
humitos, line 392:
```suggestion
For the examples in this document, we have used the :ref:`MyST Markdown
format <jupytext:formats.md#myst-markdown>`.
```

humitos (commented):
Looks good to me! I left several comments and suggestions that may be addressed before merging.

docs/guides/jupyter.rst
astrojuanlu: That's already done in the document (there's a background section explaining the technical differences between the two). About the "why", I don't think this is the place to clarify that -- the interested readers could head over https://github.com/spatialaudio/nbsphinx/issues/420 for some historical discussion :) But the two projects have evolved since then.

docs/guides/jupyter.rst
astrojuanlu: Fair enough, perhaps I was being too stubborn - I agree to make another guide about example galleries. However, given that I would need to learn sphinx-gallery and investigate if there are other alternatives, do you folks think we can do it in another pull request? That way we can get this closer to being published.

docs/guides/jupyter.rst
```diff
@@ -0,0 +1,398 @@
...
+         :caption: conf.py
+
+         extensions = [
+             "myst_nb",
+         ]
+
+Both extensions will register themselves as processors for ``.ipynb`` notebooks,
+so you don't need to change the
+`source_suffix <https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-source_suffix>`_
+configuration yourself.
```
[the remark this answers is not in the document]
astrojuanlu: You're right, it's too technical. Will remove it.

docs/guides/jupyter.rst
```diff
@@ -0,0 +1,398 @@
...
+      .. code-block:: python
+         :caption: conf.py
+
+         extensions = [
+             "myst_nb",
+         ]
+
+Both extensions will register themselves as processors for ``.ipynb`` notebooks,
+so you don't need to change the
+`source_suffix <https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-source_suffix>`_
```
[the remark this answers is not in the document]
astrojuanlu: Removed the paragraph, see below.

docs/guides/jupyter.rst
astrojuanlu: Yes, my policy was to avoid adding intersphinx links to the examples, and keep them for the "core" explanations of the guide. That way we avoid overloading our intersphinx mapping with too many sites. But perhaps it's a silly premature optimization. What do you think?

docs/guides/jupyter.rst
astrojuanlu: https://github.com/readthedocs/readthedocs.org/pull/8283#discussion_r662533886

docs/guides/jupyter.rst
astrojuanlu: https://github.com/readthedocs/readthedocs.org/pull/8283#discussion_r662533886

docs/guides/jupyter.rst
```diff
@@ -0,0 +1,398 @@
...
+
+The notebook will render as any other HTML page in your documentation
+after doing ``make html``.
+
+.. figure:: /_static/images/guides/example-notebook-rendered.png
+   :width: 80%
+   :align: center
+   :alt: Example Jupyter notebook rendered on HTML by nbsphinx
+
+   Example Jupyter notebook rendered on HTML by nbsphinx
```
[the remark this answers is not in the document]
astrojuanlu: `numfig = True` was introduced in our configuration here https://github.com/readthedocs/readthedocs.org/commit/caf990bc31affb8179fecc0edb8aa2adf75a9abc I don't think we can disable it selectively.

docs/guides/jupyter.rst
astrojuanlu: Given that we are doing that, resolving this comment

docs/guides/jupyter.rst
astrojuanlu: Yes, I will rephrase the beginning and this part so it makes more sense. Otherwise, this introduction comes "too late"

docs/guides/jupyter.rst
astrojuanlu: Good catch!

docs/guides/jupyter.rst
astrojuanlu: Done!

astrojuanlu:
Done with the last round of changes 💪🏽

docs/guides/jupyter.rst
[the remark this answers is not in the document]
humitos: I see. We can either:

- disable it and re-work the section of `numref` from that guide to not use it live
- modify https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-numfig_format to empty string `''` and hope to not show anything

docs/guides/jupyter.rst
humitos: Oh, I see.

I was thinking about using intersphinx whenever is possible. In particular because of these two reasons:

- it promotes the usage of intersphinx + hoverxref
- it's useful for the reader

humitos (approved):
LGTM!

docs/guides/jupyter.rst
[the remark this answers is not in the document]
astrojuanlu: I'll do the latter for now, let's revisit this later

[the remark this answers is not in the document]
astrojuanlu: Nop, it doesn't work:

```
Exception occurred:
  File "/home/juanlu/Projects/RTD/readthedocs.org/.venv/lib/python3.6/site-packages/sphinx/writers/html5.py", line 292, in append_fignumber
    self.body.append(prefix % '.'.join(map(str, numbers)) + ' ')
TypeError: not all arguments converted during string formatting
```

I'll have to do the former.

docs/guides/jupyter.rst
astrojuanlu: Done!

docs/guides/jupyter.rst
astrojuanlu: Done!

docs/guides/jupyter.rst
astrojuanlu: Done!

docs/guides/jupyter.rst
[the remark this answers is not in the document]
astrojuanlu: Done!

astrojuanlu:
Thanks everyone for the comments, I now enabled intersphinx for all projects and @humitos stamped his seal of approval, so I will merge. We can improve it, but I think it is already a valuable resource I wish I had 5 years ago. 🚀

choldgraf:
🚀

humitos:
> @humitos stamped his seal of approval

The power of the 🟢 , 💪🏼

PR #8294, opened 2021-06-28
feat: add json schema

SethFalco:
* Closes https://github.com/readthedocs/readthedocs.org/issues/6363
  * As soon as this is merged, I'll make the PR and get it added to the JSON Schema Store catalog.
* Closes https://github.com/readthedocs/readthedocs.org/pull/6501

Assuming I've understood the YAML Schema correctly (I did read the Yamale README), I believe they should be equivalent (or at least almost equivalent.)

I made a new PR instead of reviewing an older one, as I felt there was a lot to change.

I've tested this against the following:
* https://github.com/pytube/pytube/blob/master/.readthedocs.yml
* https://github.com/Rapptz/discord.py/blob/master/.readthedocs.yml
* https://github.com/deepmind/open_spiel/blob/master/readthedocs.yml
* https://github.com/tensorlayer/hyperpose/blob/master/.readthedocs.yml
* https://github.com/deepchem/deepdock/blob/master/.readthedocs.yml

Note: Before testing I manually changed all `python.version` values to `3.6` as original values (`3.7`/`3.8`) aren't valid against the schema. (Maybe consider using `string` rather than `enum` here?)

Let me know if anything looks off to you!

#### Subtle Differences
* For paths, it just accepts any string. I could validate it against a regex? I thought maybe just leaving it as any string would be fine.

### DIFF
--- a/readthedocs/rtd_tests/fixtures/spec/v2/schema.json
+++ b/readthedocs/rtd_tests/fixtures/spec/v2/schema.json
@@ -92,7 +92,10 @@
             "3.3",
             "3.4",
             "3.5",
-            "3.6"
+            "pypy3.5",
+            "3.6",
+            "3.7",
+            "3.8"
           ],
           "default": "3"
         },
@@ -165,7 +168,6 @@
           "description": "The builder type for the sphinx documentation.",
           "enum": [
             "html",
-            "htmldir",
             "dirhtml",
             "singlehtml"
           ],
--- a/readthedocs/rtd_tests/fixtures/spec/v2/schema.json
+++ b/readthedocs/rtd_tests/fixtures/spec/v2/schema.json
@@ -1,0 +1,288 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "$comment": "https://docs.readthedocs.io/en/stable/config-file/index.html",
+  "title": "Read the Docs",
+  "description": "Read the Docs configuration file.",
+  "type": "object",
+  "properties": {
+    "version": {
+      "title": "Version",
+      "description": "The version of the spec to use.",
+      "type": "string",
+      "enum": [
+        "2"
+      ]
+    },
+    "formats": {
+      "title": "Formats",
+      "description": "Formats of the documentation to be built.",
+      "anyOf": [
+        {
+          "type": "array",
+          "items": {
+            "enum": [
+              "htmlzip",
+              "pdf",
+              "epub"
+            ]
+          }
+        },
+        {
+          "enum": [
+            "all"
+          ]
+        }
+      ],
+      "default": []
+    },
+    "conda": {
+      "title": "Conda",
+      "description": "Configuration for Conda support.",
+      "type": "object",
+      "properties": {
+        "environment": {
+          "title": "Environment",
+          "description": "The path to the Conda environment file from the root of the project.",
+          "type": "string"
+        }
+      },
+      "required": [
+        "environment"
+      ]
+    },
+    "build": {
+      "title": "Build",
+      "description": "Configuration for the documentation build process.",
+      "type": "object",
+      "properties": {
+        "image": {
+          "title": "Image",
+          "description": "The build docker image to be used.",
+          "enum": [
+            "stable",
+            "latest"
+          ],
+          "default": "latest"
+        },
+        "apt_packages": {
+          "title": "APT Packages",
+          "description": "List of packages to be installed with apt-get.",
+          "type": "array",
+          "items": {
+            "title": "APT Package",
+            "type": "string"
+          },
+          "default": []
+        }
+      }
+    },
+    "python": {
+      "title": "Python",
+      "description": "Configuration of the Python environment to be used.",
+      "type": "object",
+      "properties": {
+        "version": {
+          "title": "Version",
+          "description": "The Python version (this depends on the build image).",
+          "type": "string",
+          "enum": [
+            "2",
+            "2.7",
+            "3",
+            "3.3",
+            "3.4",
+            "3.5",
+            "3.6"
+          ],
+          "default": "3"
+        },
+        "system_packages": {
+          "title": "System Packages",
+          "description": "Give the virtual environment access to the global site-packages directory.",
+          "type": "boolean",
+          "default": false
+        },
+        "install": {
+          "title": "Install",
+          "description": "Installation of packages and requiremens.",
+          "type": "array",
+          "items": {
+            "anyOf": [
+              {
+                "properties": {
+                  "requirements": {
+                    "title": "Requirements",
+                    "description": "The path to the requirements file from the root of the project.",
+                    "type": "string"
+                  }
+                },
+                "required": [
+                  "requirements"
+                ]
+              },
+              {
+                "properties": {
+                  "path": {
+                    "title": "Path",
+                    "description": "The path to the project to be installed",
+                    "type": "string"
+                  },
+                  "method": {
+                    "title": "Method",
+                    "description": "Install using python setup.py install or pip.",
+                    "enum": [
+                      "pip",
+                      "setuptools"
+                    ],
+                    "default": "pip"
+                  },
+                  "extra_requirements": {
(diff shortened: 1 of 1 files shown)


readthedocs/rtd_tests/fixtures/spec/v2/schema.json
```diff
@@ -0,0 +1,330 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "$comment": "https://docs.readthedocs.io/en/stable/development/design/yaml-file.html",
```
stsewd, line 3: We should link to https://docs.readthedocs.io/en/stable/config-file/index.html instead, the design doc is outdated now.

readthedocs/rtd_tests/fixtures/spec/v2/schema.json
```diff
@@ -0,0 +1,330 @@
...
+                    },
+                    "default": []
+                  }
+                },
+                "required": [
+                  "path"
+                ]
+              },
+              {
+                "properties": {
```
stsewd, line 154: pip file support isn't going to be implemented, we can just drop this section

readthedocs/rtd_tests/fixtures/spec/v2/schema.json
```diff
@@ -0,0 +1,330 @@
...
+          "default": [
+            "search.html",
+            "search/index.html",
+            "404.html",
+            "404/index.html"
+          ]
+        }
+      }
+    },
+    "redirects": {
```
stsewd, line 314: Redirects aren't implemented yet, either.

stsewd (commented):
Thanks for taking over this!

SethFalco:
> Redirects aren't implemented yet, either.

I interpreted that as "take that same action as with pip file support".
Let me know if that was wrong of me!

<details>
  <summary>Just noting the dropped sections here for future reference.</summary>

### Pipfile (`properties.python.install.items.anyOf[2]`)
```
{
  "properties": {
    "pipfile": {
      "title": "Pipfile",
      "description": "The path to the directory that contains the Pipfile.",
      "type": "string"
    },
    "dev": {
      "title": "Dev",
      "description": "Add the --dev option to pipenv install.",
      "type": "boolean",
      "default": false
    },
    "ignore_pipfile": {
      "title": "Ignore Pipfile",
      "description": "Add the --ignore-pipfile option to pipenv install.",
      "type": "boolean",
      "default": true
    },
    "skip_lock": {
      "title": "Skip Lock",
      "description": "Add the --skip-lock option to pipenv install.",
      "type": "boolean",
      "default": true
    }
  },
  "required": [
    "pipfile"
  ]
}
```

### Redirects (`properties.redirects`)
```
"redirects": {
  "title": "Redirects",
  "description": "Redirects for the current version to be built",
  "properties": {
    "page": {
      "additionalProperties": {
        "type": "string"
      }
    }
  },
  "default": null
}
```

readthedocs/rtd_tests/fixtures/spec/v2/schema.json
```diff
@@ -0,0 +1,288 @@
...
+          "description": "The Python version (this depends on the build image).",
+          "type": "string",
+          "enum": [
+            "2",
+            "2.7",
+            "3",
+            "3.3",
+            "3.4",
+            "3.5",
+            "3.6"
```
stsewd, line 95: We are missing `3.7`, `3.8` and `pypy3.5`

readthedocs/rtd_tests/fixtures/spec/v2/schema.json
```diff
@@ -0,0 +1,288 @@
...
+      "title": "Sphix",
+      "description": "Configuration for sphinx documentation.",
+      "type": "object",
+      "properties": {
+        "builder": {
+          "title": "Builder",
+          "description": "The builder type for the sphinx documentation.",
+          "enum": [
+            "html",
+            "htmldir",
```
stsewd, line 168: This option is deprecated in favor of `dirhtml`, maybe we could omit it from here?

readthedocs/rtd_tests/fixtures/spec/v2/schema.json
```diff
@@ -0,0 +1,288 @@
...
+      "type": "object",
+      "properties": {
+        "configuration": {
+          "title": "Configuration",
+          "description": "The path to the mkdocs.yml file.",
+          "type": "string"
+        },
+        "fail_on_warning": {
+          "title": "Fail on Warning",
+          "description": "Add the --strict optio to mkdocs build.",
```
stsewd, line 199:
```suggestion
          "description": "Add the --strict option to mkdocs build.",
```

stsewd:
> I interpreted that as "take that same action as with pip file support".

yes, that was it

SethFalco:
I did the things you asked for, plus one additional change:

Since `version` only has one possible value, and it's a required field anyway, I figured it might as well also have a default configured to `"2"` as well.

Would you care to give an opinion on that?

Can I also just confirm, did I reference the correct YAML Schema when I created the JSON Schema? 🤔 ([I used this one.](https://github.com/readthedocs/readthedocs.org/blob/master/readthedocs/rtd_tests/fixtures/spec/v2/schema.yml))
As in, are you suggestions fixing mistakes of mine, or are they fresh/new changes to keep things up to date?

Does the YAML Schema need to be updated separately as well?

stsewd:
> Since version only has one possible value, and it's a required field anyway, I figured it might as well also have a default configured to "2" as well.

We do allow `1`, which is the default to use if you don't set a version, so it should be required.

> Can I also just confirm, did I reference the correct YAML Schema when I created the JSON Schema? thinking (I used this one.)
As in, are your suggestions fixing mistakes of mine, or are they fresh/new changes to keep things up to date?

Yeah, that scheme is the one, but it has some ideas that were never implemented, we may remove it and use this json scheme instead of yamale, so no need to update it here.

readthedocs/rtd_tests/fixtures/spec/v2/schema.json
```diff
@@ -0,0 +1,291 @@
...
+  "type": "object",
+  "properties": {
+    "version": {
+      "title": "Version",
+      "description": "The version of the spec to use.",
+      "type": "string",
+      "enum": [
+        "2"
+      ],
+      "default": "2"
```
stsewd, line 15: This should be required, default is 1, which has a different scheme.

SethFalco: It is marked as required in line 289! In a JSON Schema, we mark a property as required in the outer scope. (Unless I have it wrong.)

SethFalco: Just to confirm, I tried with an arbitrary browser based JSON Schema validator.

image

stsewd: oh, I see. We should remove the default value here, then.

stsewd: as the default is 1, and now it's just actually required for this scheme.

SethFalco: Sure, sorry about that! Removed the the default value now.

stsewd (approved):
Thanks!

stsewd:
@SethFalco let us know if you need anything from us to submit this to https://github.com/SchemaStore/schemastore/

SethFalco:
Thanks for that! I think it's simple, so no need to worry.
Just have to add it a reference to the file merged to the`catalog.json` file in the repo.  

* https://github.com/SchemaStore/schemastore/pull/1691

```json
{
  "name": "Read the Docs",
  "description": "Read the Docs configuration file",
  "fileMatch": [
    "readthedocs.yml",
    "readthedocs.yaml",
    ".readthedocs.yml",
    ".readthedocs.yaml"
  ],
  "url": "https://raw.githubusercontent.com/readthedocs/readthedocs.org/master/readthedocs/rtd_tests/fixtures/spec/v2/schema.json"
}
```

PR #8308, opened 2021-07-01
Add link to redirect docs

The-Compiler:
Inspired by what's done on the `projects/domain_list.html` template:

https://github.com/readthedocs/readthedocs.org/blob/e0e6602b23345b65ff67478a8f2d2d4c90ea8b6b/readthedocs/templates/projects/domain_list.html#L16-L18

See https://github.com/readthedocs/readthedocs.org/issues/8264#issuecomment-871115620 for some context.

### DIFF
--- a/readthedocs/templates/projects/project_redirects.html
+++ b/readthedocs/templates/projects/project_redirects.html
@@ -71,7 +71,9 @@
 
 {% block project_edit_content %}
   <p class="help_text">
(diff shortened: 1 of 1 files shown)


astrojuanlu (commented):
Thanks a lot @The-Compiler ! I'm not familiar with how translations work but this looks good to me.

stsewd (approved):
thanks!

PR #8360, opened 2021-07-21
Proxito: return user-defined HTTP headers on custom domains

humitos:
We have had multiple requests from users wanting to add extra HTTP headers to
the responses on their documentation.

This commit allows this by adding a new model `projects.models.HTTPDomain` in
which we (as site administrator) can define addittional HTTP headers to a
particular domain (this objects are linked to `projects.models.Domain`).

When a request is processed, in the middleware executed by El Proxito we check
if there are extra headers for this domain and add them to the response.

Note that we currently support a pre-defined list of extra headers due a
limitation in NGINX (or in our knowledge of NGINX) that we can't include all the
upstream HTTP headers in the response and we have to set them one by one.


humitos:
I've try playing with NGINX and Lua to include all the required headers. I wrote this:

```
        access_by_lua_block {
            ngx.header["X-My-New-Header"] = ngx.var.upstream_http_x_rtd_project;
            ngx.header["X-My-New-Header-Manual"] = "Manual Value";
        }
```

which is close, but we need a way to iterate over _all the `ngx.var.upstream_http_*` variables to avoid pre-defined headers.

> Note that to be able to test this code, the Docker image `firesh/nginx-lua` is required since the official one does not include Lua.

readthedocs/projects/models.py
```diff
@@ -1570,6 +1570,30 @@ def save(self, *args, **kwargs):  # pylint: disable=arguments-differ
...
+    domain = models.ForeignKey(
+        Domain,
+        related_name='http_headers',
+        on_delete=models.CASCADE,
+    )
+    name = models.CharField(max_length=128)
+    value = models.CharField(max_length=256)
+
+    def __str__(self):
+        return self.name
```
ericholscher, line 1594:
```suggestion
        return f"HttpHeader: {self.name} on {self.domain.domain}"
```

Suggestion (nonblocking): I think we probably want a bit more detail here, so we know what the object is when printed.

readthedocs/proxito/middleware.py
```diff
@@ -176,6 +176,51 @@ def add_proxito_headers(self, request, response):
...
 
+    def add_user_headers(self, request, response):
+        """
+        Set specific HTTP headers requested by the user.
+
+        The headers added come from ``projects.models.HTTPHeader`` associated
+        with the ``Domain`` object.
+        """
+        if hasattr(request, 'domain'):
+            domain = request.domain
```
ericholscher, line 187: nitpick: Only need this variable once, probably better to just use it directly.

readthedocs/proxito/middleware.py
```diff
@@ -176,6 +176,51 @@ def add_proxito_headers(self, request, response):
...
+            for http_header in domain.http_headers.all():
+                response[http_header.name] = http_header.value
```
ericholscher, line 189: Idea: Should we warn if the header is already set? Not sure how often this would happen, but based on David's feedback about not double-setting them. I think this will override an existing value, but we probably want to check to see if we're clobbering anything.

ericholscher (approved):
Looks like a simple enough change. I think the `choices` is the main change I'd like to see, just to reduce mistakes.

readthedocs/proxito/middleware.py
humitos: Cool! I'm `log.exception` here so we are aware of this.

readthedocs/projects/models.py
```diff
@@ -1582,16 +1583,28 @@ class HTTPHeader(TimeStampedModel, models.Model):
     expose them all without hardcoding them.
     """
 
+    HEADERS_CHOICES = (
+        ('referrer_policy', 'Referrer-Policy'),
+        ('permissions_policy', 'Permissions-Policy'),
+        ('feature_policy', 'Feature-Policy'),
+        ('access_control_allow_origin', 'Access-Control-Allow-Origin'),
+        ('access_control_allow_headers', 'Access-Control-Allow-Headers'),
```
ericholscher, line 1591: Do we want users to be able to control CORS headers? These are currently controlled in our middleware, and I don't believe these will be exposed on the API without additional changes, so I'm not sure if this will actually work yet (eg. an `OPTIONS` request to `/_/api/v2/`)

readthedocs/proxito/middleware.py
```diff
@@ -184,8 +184,19 @@ def add_user_headers(self, request, response):
         with the ``Domain`` object.
         """
         if hasattr(request, 'domain'):
-            domain = request.domain
-            for http_header in domain.http_headers.all():
+            response_headers = [header.lower() for header in response]
+            for http_header in request.domain.http_headers.all():
+                if http_header.name.lower() in response_headers:
+                    log.exception(
```
ericholscher, line 190: I think we want `error` here, since there isn't an actual exception: https://docs.python.org/3/library/logging.html#logging.Logger.exception
```suggestion
                    log.error(
```

readthedocs/proxito/middleware.py
```diff
@@ -196,6 +207,10 @@ def add_hsts_headers(self, request, response):
         * For custom domains, check the HSTS values on the Domain object.
           The domain object should be saved already in request.domain.
         """
+        if not request.is_secure():
+            # Only set the HSTS header if the request is over HTTPS
+            return response
```
ericholscher, line 212: Good catch. Probably want a test for this 👍

ericholscher (approved):
This looks good. Assuming this is staff-only currently, I'm not too worried about the CORS question, just curious mostly.

readthedocs/projects/models.py
humitos: I'm not thinking of exposing these to users. I added here all the headers that I found that we are using in -ops and -corporate-ops so we can delete the custom proxito nginx files.

readthedocs/proxito/middleware.py
humitos: You are right!

readthedocs/proxito/middleware.py
humitos: I found it because I broke a test that someone's else wrote 😉

readthedocs/projects/models.py
humitos: Also, these CORS headers will only work on documentation pages since the middleware is executed on El Proxito only. I don't think this will affect API calls.

humitos: Right? 😃

ericholscher: I don't believe these CORS headers will be properly sent for the API though, will they? CORS doesn't matter on the documentation pages, only on reply to `OPTIONS` requests when using the API's.

readthedocs/proxito/middleware.py
```diff
@@ -176,6 +176,66 @@ def add_proxito_headers(self, request, response):
...
+        if settings.PUBLIC_DOMAIN_USES_HTTPS and public_domain in host:
+            hsts_header_values = [
+                'max-age=31536000',
+                'includeSubDomains',
+                'preload',
+            ]
+        elif hasattr(request, 'domain'):
+            domain = request.domain
+            # TODO: migrate Domains with HSTS set using these fields to
+            # ``HTTPHeader`` and remove this chunk of code from here.
```
davidfischer, line 226: There's only a few of these. A custom Django data migration may be overkill although it would ensure we get it right =).

readthedocs/proxito/middleware.py
```diff
@@ -176,6 +176,66 @@ def add_proxito_headers(self, request, response):
...
+                        'Overriding an existing response HTTP header. header=%s domain=%s',
+                        http_header.name,
+                        request.domain.domain,
+                    )
+                log.info(
+                    'Adding custtom response HTTP header. header=%s domain=%s',
+                    http_header.name,
+                    request.domain.domain,
+                )
+                response[http_header.name] = http_header.value
```
davidfischer, line 200: Let's just put a comment here that the headers are limited by the validator/choices on the `HTTPHeader` model. Setting arbitrary HTTP headers is potentially dangerous (I realize this is staff-only but things change over time).

readthedocs/projects/models.py
```diff
@@ -1570,6 +1570,43 @@ def save(self, *args, **kwargs):  # pylint: disable=arguments-differ
...
+
+    """
+    Define a HTTP header for a user Domain.
+
+    All the HTTPHeader(s) associated with the domain are added in the response
+    from El Proxito.
+
+    NOTE: the available headers are hardcoded in the NGINX configuration for
+    now (see ``dockerfile/nginx/proxito.conf``) until we figure it out a way to
+    expose them all without hardcoding them.
```
davidfischer, line 1583: I realize there's a desire to be generic but setting arbitrary headers is potentially dangerous. I prefer this set list of headers although we also could probably set any header that starts `X-` as well.

readthedocs/projects/models.py
```diff
@@ -1570,6 +1570,43 @@ def save(self, *args, **kwargs):  # pylint: disable=arguments-differ
...
+        Domain,
+        related_name='http_headers',
+        on_delete=models.CASCADE,
+    )
+    name = models.CharField(
+        max_length=128,
+        choices=HEADERS_CHOICES,
+    )
+    value = models.CharField(max_length=256)
+
```
davidfischer, line 1605: You could generically solve the `Strict-Transport-Security` only when secure by having a "only-set-if-secure" flag here if you want.

davidfischer (approved):

readthedocs/projects/models.py
humitos: I updated the PR adding a new field to handle this case.

PR #8367, opened 2021-07-26
Lock old threads

astrojuanlu:
This enables a bot that locks issues and pull requests that were **closed** more than 30 days ago. It leaves a comment when doing so, which will trigger a notification - we can disable that, at the expense of making it less clear for outsiders why the issue was closed. It also includes a suggestion to have more output in the action logs: https://github.com/dessant/lock-threads/issues/26#issuecomment-877810915

### DIFF
--- a/.github/workflows/lock.yml
+++ b/.github/workflows/lock.yml
@@ -1,0 +1,34 @@
+name: 'Lock old threads'
+
+on:
+  schedule:
+    - cron: '0 0 * * *'
+
+permissions:
+  issues: write
+  pull-requests: write
+
+jobs:
+  action:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: dessant/lock-threads@v2
+        with:
+          issue-lock-inactive-days: '30'
+          pr-lock-inactive-days: '30'
+          issue-lock-comment: >
+            This issue has been automatically locked since there
+            has not been any recent activity after it was closed.
+            Please open a new issue for related bugs.
+          pr-lock-comment: >
+            This pull request has been automatically locked since there
+            has not been any recent activity after it was closed.
+            Please open a new issue for related bugs.
+      - name: Log processed threads
+        run: |
+          if [ '${{ steps.lock.outputs.issues }}' ]; then
+            echo "Issues:" && echo '${{ steps.lock.outputs.issues }}' | jq .
+          fi
+          if [ '${{ steps.lock.outputs.prs }}' ]; then
+            echo "Pull requests:" && echo '${{ steps.lock.outputs.prs }}' | jq .
+          fi


.github/workflows/lock.yml
```diff
@@ -0,0 +1,34 @@
...
+          issue-lock-inactive-days: '30'
+          pr-lock-inactive-days: '30'
+          issue-lock-comment: >
+            This issue has been automatically locked since there
+            has not been any recent activity after it was closed.
+            Please open a new issue for related bugs.
+          pr-lock-comment: >
+            This pull request has been automatically locked since there
+            has not been any recent activity after it was closed.
+            Please open a new issue for related bugs.
```
humitos, line 26: I don't think there is a need to generate a notification to all the people subscribed to the issue/pr.

.github/workflows/lock.yml
```diff
@@ -0,0 +1,34 @@
...
+  pull-requests: write
+
+jobs:
+  action:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: dessant/lock-threads@v2
+        with:
+          issue-lock-inactive-days: '30'
+          pr-lock-inactive-days: '30'
```
humitos, line 18: Is it possible to only lock threads that contain the `Support` label?

humitos (commented):
I don't have a strong opinion on this. I'd probably be -0 on this bot.

I personally like to have all the history of a bug/feature in the same place, even if it's a regression, and be able to re-open the issue if needed.

Splitting "the same" issue into different issues because it's locked, does not help us to have the full context of it easily accessible. Figuring out the whole context by clicking on different issues for the same problem is a worse outcome, IMO.

However, I do think we should keep communicating to our users that "Support" is managed over email and we should try to keep only Bugs/Features/Tasks/Improvements related to the application in our GH repository. Is it possible to enable this bot only for issues with `Support` label?

Finally, being able to comment on an old issue allows all the people involved in that issue to received a notification and continue the discussion in the same place. If users create a new one just linking the old one, that people are not aware that they may be facing the same problem again (regression) or there was any other update.

.github/workflows/lock.yml
astrojuanlu: Nop, we can only exclude certain labels using `issue-exclude-labels` https://github.com/dessant/lock-threads/#inputs

.github/workflows/lock.yml
astrojuanlu: Fair enough, I found the other behavior a bit too "silent", but I'm concerned about this as well.

ericholscher:
I'm +0 on this. We can always unlock the thread as a maintainer, but having users commenting on old threads is often counter-productive, and almost always is the wrong thing to do. I find going to a thread and having a 30+ comment thread is often full of distractions and comments not fully related to the issue, so I think starting a new thread and referencing the old one is usually better and provides a clearer thread on the new issue. 

I don't care enough to fight over it, but I do think it solves a problem, and the imagined issues with it are lesser than the value of it.

agjohnson:
Heh, well I'm about +0 as well. I use octobox to keep up on github notifies, and immediately discard any notifications on merged and closed issues/prs. So this matches my personal policy on issues, but also this doesn't really affect me either way ultimately. Also also, I don't know how much of an issue this is because all of these notification go straight into the trash for me :upside_down_face: 

Though I do also agree with @humitos on keeping issues open for regression. This maybe more often turns into users commenting on issues that stale or off our roadmap though.

If we don't all agree or feel strongly here, we could explore a softer option, perhaps with a comment bot that reminds users that we don't follow closed or merged issues, and you should ping core team only if this is a regression.

ericholscher:
If we don't follow them, we should lock them. Otherwise users are commenting into a void... 

I'm -1 on a bot that adds comments though, because that will notify folks on the issue. I think locking hopefully is a "meta" action that doesn't notify, which is why I wanted this approach.

astrojuanlu:
> I think locking hopefully is a "meta" action that doesn't notify, which is why I wanted this approach.

Correct, see https://github.com/readthedocs/readthedocs.org/pull/8367#discussion_r676639434

humitos:
> Correct, see [#8367 (comment)](https://github.com/readthedocs/readthedocs.org/pull/8367#discussion_r676639434)

I think you wanted to link this comment https://github.com/readthedocs/readthedocs.org/pull/8367#discussion_r676638489 😄

astrojuanlu:
This was more controversial than expected, but at the same time it's trivial to implement and we can resume the discussion at any point. For now, I'm closing.

PR #8378, opened 2021-07-28
Remove clickjacking middleware for proxito

ericholscher:
This was getting set and not passed through nginx,
but with our last deploy it is now getting set.
We can optionally turn this on for docs pages,
but we don't want it on everywhere by default.


davidfischer (approved):
This is not needed while proxying to docs pages. It is necessary on the dashboard (which is set in the base settings).

humitos:
Cool! Remember to add an HTTP header for the habana.ai domain that asked for this originally

ericholscher:
@humitos done.

PR #8437, opened 2021-08-23
Add MyST (Markdown) examples to "cross referencing with Sphinx" guide

astrojuanlu:
This adds MyST (Markdown) examples to "cross-referencing with Sphinx" guide by using https://pypi.org/project/sphinx-tabs/, similarly to how we did in #8283. Basically all our reST samples are in our Sphinx guides - I started with one of them in this pull request to make reviews easier, but I can add a couple more guides. 

While I was at it, I also reworded some parts to use second person and to clarify the role of reStructuredText vs Sphinx and put it on the same ground as MyST.

Got into a couple of rabbit holes while doing it, for example https://github.com/executablebooks/mdit-py-plugins/issues/28. Also, "in-line" references are not supported at the moment, we didn't open an issue for it.

### DIFF
--- a/docs/guides/cross-referencing-with-sphinx.rst
+++ b/docs/guides/cross-referencing-with-sphinx.rst
@@ -66,7 +66,7 @@ that will be rendered as `My target`_.
 
 You can also add explicit targets before paragraphs (or any other part of a page).
 
-Another example, here we add a target to a paragraph:
+Another example, add a target to a paragraph:
 
 .. code-block:: rst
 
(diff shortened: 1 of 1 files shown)


docs/guides/cross-referencing-with-sphinx.rst
```diff
@@ -263,7 +360,7 @@ That will be rendered as:
 Invalid targets
 ---------------
 
-If we reference an invalid or undefined target Sphinx will warn us.
+If you reference an invalid or undefined target Sphinx will warn us.
```
ericholscher, line 363: Guessing we'll do a simple style guide here soon :)

ericholscher (approved):
Simple enough. Definitely a good example of the struggle of the markdown support, since rST stuff is also in the copy in various places.

docs/guides/cross-referencing-with-sphinx.rst
astrojuanlu: Getting closer :)

PR #8445, opened 2021-08-25
Design doc: use GitHub Apps for more granular permissions

humitos:
Design document to communicate how to support GitHub Application for new users
while keeping OAuth Application for old users.

The main benefit is that GitHub Application allows us more granular permissions
than OAuth Application. With GitHub Application most of the permission we
request are Read-only and that's a win for everybody!

Rendered version: https://docs--8445.org.readthedocs.build/en/8445/development/design/github-application.html

### DIFF
--- a/docs/development/design/github-application.rst
+++ b/docs/development/design/github-application.rst
@@ -10,7 +10,7 @@ Read the Docs currently uses the OAuth Application to support "Login with
 GitHub" and "Import Project". They require different level of permissions, but
 OAuth Application does not really allow us to manage them in a granular way.
 
-Based in our needings and the GitHub recommendations in
+Based on our needs and the GitHub recommendations in
 https://docs.github.com/en/developers/apps/getting-started-with-apps/about-apps#determining-which-integration-to-build
 we should use a GitHub Application for Read the Docs::
 
@@ -34,7 +34,7 @@ implementation.
 Problems with current implementation
 ------------------------------------
 
-* We request too permisives scope at Sign Up
+* We request permissions that are too permissive at Sign Up
 
   Even if our application will never require some of these scopes (because it's
   a read user --and will never import a project), we ask for too many permissions.
@@ -65,9 +65,9 @@ Goals
 * Support GitHub Application *and* OAuth Application at the same time during the
   migration time
 * Support all the GitHub features current using OAuth Application
-  * GitHub SSO
-  * Pull Request builder
-  * Trigger a build on push
+   * GitHub SSO
+   * Pull Request builder
+   * Trigger a build on push
 * Don't require SSH key to clone private repositories
 
 
@@ -254,17 +254,12 @@ By doing this, we will keep our ``Remote*`` table very small because we will
 only track repositories that users gave us permissions. Then, only these
 repositories will be shown in the "Import Project" page.
 
-Once they created a new repository under their organization, they will need to
-go to the GitHub Application installation configuration and grant access to the
+Once they created a new repository under their organization, if they haven't given us
+access to "All repositories",
+they will need to go to the GitHub Application installation configuration and grant access to the
 new repository. This will send us a webhook (``installation_repositories``,
 https://docs.github.com/en/developers/webhooks-and-events/webhooks/webhook-events-and-payloads#installation_repositories)
 and we can automatically do a re-sync of ``Remote*`` models.
-
-.. note::
-
-   Users installing our GitHub Application and selecting "All repositories" will
-   always see all the repositories on the "Import Project" page.
-
 
 Need more research
 ------------------
--- a/docs/development/design/github-application.rst
+++ b/docs/development/design/github-application.rst
@@ -1,0 +1,306 @@
+==================
+GitHub Application
+==================
+
+GitHub offers two different ways to access users' data from its own platform:
+OAuth Application and GitHub Application. Depending on the applications
+needings, GitHub recommends using one or the other.
+
+Read the Docs currently uses the OAuth Application to support "Login with
+GitHub" and "Import Project". They require different level of permissions, but
+OAuth Application does not really allow us to manage them in a granular way.
+
+Based in our needings and the GitHub recommendations in
+https://docs.github.com/en/developers/apps/getting-started-with-apps/about-apps#determining-which-integration-to-build
+we should use a GitHub Application for Read the Docs::
+
+  Only as me? -> NO
+  Act as an App -> NO
+  Access Everything? -> NO
+  -> GitHub Application
+
+Or in case we need to act as an App, we end in the same route::
+
+  Only as me? -> NO
+  Act as an App -> YES
+  -> GitHub Application
+
+This document explains some problems we have found in our current implementation
+using OAuth Application and explore the recommended way using GitHub
+Application. It goes through the work required and touches a little about its
+implementation.
+
+
+Problems with current implementation
+------------------------------------
+
+* We request too permisives scope at Sign Up
+
+  Even if our application will never require some of these scopes (because it's
+  a read user --and will never import a project), we ask for too many permissions.
+
+* OAuth Application does not provide granular permissions
+
+  We have to ask for ``read`` scope. It gives us read & write permissions to
+  *all the repositories of this user*. This makes some users/organizations to
+  don't want to user our platform.
+
+  Besides, ``admin:repo_hook`` scope is also required to be able to create a
+  webhook for pushes and trigger a new build.
+
+* Repositories from an GitHub Organization does not appear in the "Import
+  Project" list
+
+  Organizations with OAuth App access restriction enabled, require users using
+  our OAuth Application to request permissions to the GitHub Organization's
+  owner to accept our OAuth App.
+
+
+Goals
+-----
+
+* Reduce as much as possible granted permissions for our GitHub Application
+* Do not require access to *all the users'/organizations' repositories*
+* Allow Read the Docs GitHub Application only to specific repositories
+* Support GitHub Application *and* OAuth Application at the same time during the
+  migration time
+* Support all the GitHub features current using OAuth Application
+  * GitHub SSO
+  * Pull Request builder
+  * Trigger a build on push
+* Don't require SSH key to clone private repositories
+
+
+Non Goals
+---------
+
+* Escalate permissions as they are needed
+* Migrate all current OAuth Application users to GitHub Application users at
+  once
+* Implement something similar to GitHub Application for other VCS providers
+  (GitLab and Bitbucket)
+
+
+Context related to GitHub Application
+-------------------------------------
+
+Considering there are too much confusion around GitHub Application and OAuth
+Application, and copy&paste some context from its official documentation trying
+to mitigate it and have a better understanding of this design doc.
+
+  GitHub Apps are first-class actors within GitHub. A GitHub App acts on its own
+  behalf, taking actions via the API directly using its own identity, which
+  means you don't need to maintain a bot or service account as a separate user.
+
+  GitHub Apps can be installed directly on organizations and user accounts and
+  granted access to specific repositories. They come with built-in webhooks and
+  narrow, specific permissions. When you set up your GitHub App, you can select
+  the repositories you want it to access
+
+  To install a GitHub App, you must be an organization owner or have admin
+  permissions in a repository.
+
+  Don't build an OAuth App if you want your application to act on a single
+  repository. With the repo OAuth scope, OAuth Apps can act on all of the
+  authenticated user's repositories.
+
+(from  https://docs.github.com/en/developers/apps/getting-started-with-apps/about-apps)
+
+  You can install GitHub Apps in your personal account or organizations you own.
+  If you have admin permissions in a repository, you can install GitHub Apps on
+  organization accounts. If a GitHub App is installed in a repository and requires
+  organization permissions, the organization owner must approve the application.
+
+  Account owners can use a GitHub App in one account without granting access to
+  another ... A GitHub App remains installed if the person who set it up leaves
+  the organization.
+
+  The installation token from a GitHub App loses access to resources if an admin
+  removes repositories from the installation.
+
+  Installation access tokens are limited to specified repositories with the
+  permissions chosen by the creator of the app.
+
+  GitHub Apps can request separate access to issues and pull requests without
+  accessing the actual contents of the repository.
+
+  A GitHub App receives a webhook event when an installation is changed or
+  removed. This tells the app creator when they've received more or less access to
+  an organization's resources.
+
+  A GitHub App can request an installation access token by using a private key
+  with a JSON web token format out-of-band.
+
+  An installation token identifies the app as the GitHub Apps bot, such as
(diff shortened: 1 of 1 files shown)


docs/development/design/github-application.rst
```diff
@@ -0,0 +1,306 @@
...
+
+GitHub offers two different ways to access users' data from its own platform:
+OAuth Application and GitHub Application. Depending on the applications
+needings, GitHub recommends using one or the other.
+
+Read the Docs currently uses the OAuth Application to support "Login with
+GitHub" and "Import Project". They require different level of permissions, but
+OAuth Application does not really allow us to manage them in a granular way.
+
+Based in our needings and the GitHub recommendations in
```
ericholscher, line 13:
```suggestion
Based on our needs and the GitHub recommendations in
```

docs/development/design/github-application.rst
```diff
@@ -0,0 +1,306 @@
...
+This document explains some problems we have found in our current implementation
+using OAuth Application and explore the recommended way using GitHub
+Application. It goes through the work required and touches a little about its
+implementation.
+
+
+Problems with current implementation
+------------------------------------
+
+* We request too permisives scope at Sign Up
```
ericholscher, line 37:
```suggestion
* We request permissions that are too permissive at Sign Up
```

docs/development/design/github-application.rst
```diff
@@ -0,0 +1,306 @@
...
+
+* Reduce as much as possible granted permissions for our GitHub Application
+* Do not require access to *all the users'/organizations' repositories*
+* Allow Read the Docs GitHub Application only to specific repositories
+* Support GitHub Application *and* OAuth Application at the same time during the
+  migration time
+* Support all the GitHub features current using OAuth Application
+  * GitHub SSO
+  * Pull Request builder
+  * Trigger a build on push
```
ericholscher, line 70:
```suggestion
   * GitHub SSO
   * Pull Request builder
   * Trigger a build on push
```

docs/development/design/github-application.rst
```diff
@@ -0,0 +1,306 @@
...
+* Don't require SSH key to clone private repositories
+
+
+Non Goals
+---------
+
+* Escalate permissions as they are needed
```
ericholscher, line 77: Because we're asking for a smaller set? Could use a bit more explanation here.

docs/development/design/github-application.rst
```diff
@@ -0,0 +1,306 @@
...
+-------------------------------
+
+:Repository permissions:
+   - Metadata: Read-only (mandatory)
+   - Contents: Read-only
+   - Commit statuses: Read & Write
+   - Pull requests: Read-only
+
+:Organization permissions:
+   - Members: Read-only
```
ericholscher, line 155: Is this for SSO, or something else? Might be useful to escalate in this case, if it lets us not request Org permissions up front.

docs/development/design/github-application.rst
```diff
@@ -0,0 +1,306 @@
...
+for all the installations. The body contains all the information about the
+particular installation that triggered the event. With this data, we will create
+an access token and perform the query/actions we need.
+
+There are some events that need to map to particular ``Project`` in our
+database. For example, "a push to ``main`` branch in repository
+``readthedocs/blog``" should "trigger a build for ``latest`` version on
+``read-the-docs-blog`` project". For these cases we can use use
+``repository.id`` field from the body to get the ``RemoteRepository.remote_id``
+and from there get the ``Project``.
```
ericholscher, line 229: This is just a Project mapping though, and means that if `readthedocs/template` gets a commit, then *every* project on RTD with that `repo_url` will be built? That takes us back to our original webhooks implementation, which is not great :/

docs/development/design/github-application.rst
```diff
@@ -0,0 +1,306 @@
...
+* iterate over each repository checking for user's permission
+
+  * ``https://api.github.com/repos/{owner}/{repo}/collaborators/{username}/permission`` (https://docs.github.com/en/rest/reference/repos#get-repository-permissions-for-a-user)
+
+By doing this, we will keep our ``Remote*`` table very small because we will
+only track repositories that users gave us permissions. Then, only these
+repositories will be shown in the "Import Project" page.
+
+Once they created a new repository under their organization, they will need to
+go to the GitHub Application installation configuration and grant access to the
```
ericholscher, line 258:
```suggestion
Once they created a new repository under their organization, if they haven't given us
access to "All repositories",
they will need to go to the GitHub Application installation configuration and grant access to the
```

docs/development/design/github-application.rst
```diff
@@ -0,0 +1,306 @@
...
+new repository. This will send us a webhook (``installation_repositories``,
+https://docs.github.com/en/developers/webhooks-and-events/webhooks/webhook-events-and-payloads#installation_repositories)
+and we can automatically do a re-sync of ``Remote*`` models.
+
+.. note::
+
+   Users installing our GitHub Application and selecting "All repositories" will
+   always see all the repositories on the "Import Project" page.
+
+
```
ericholscher, line 268:
```suggestion
```

docs/development/design/github-application.rst
```diff
@@ -0,0 +1,306 @@
...
+* How do we use Client ID for GitHub Application for new users and Client ID for
+  OAuth application for current/existing users?
+
+* Should we keep using our OAuth Application for login and GitHub Application
+  for the rest? Is it possible?
+
+  Related: https://docs.github.com/en/developers/apps/getting-started-with-apps/migrating-oauth-apps-to-github-apps
+  Related: https://github.com/readthedocs/readthedocs-ops/issues/532
+
```
ericholscher, line 286: These seem like really important questions. The interaction with allauth in particular worries me. I really don't want to have to hack a lot of stuff, so keeping auth & repo permissions separated might be the best solution. Could even make the GH App approach optional for more security-minded people?

docs/development/design/github-application.rst
```diff
@@ -0,0 +1,306 @@
...
+* How are we going to handle other VCS providers (GitLab and Bitbucket)?
+
+  GitLab and Bitbucket does not offer another option than OAuth Application. We
+  need to maintain the implementation that we currently have for them.
+
+* Does it worth the effort integrating GitHub Application without being able to
+  use the same for other services?
+
+  GitHub is the most VCS provider used in our platform. Because of this, if the
+  we can provide a better UX to most of our users I'd call it a win.
```
ericholscher, line 300: This sounds like something I would say :D

docs/development/design/github-application.rst
```diff
@@ -0,0 +1,306 @@
...
+
+  An installation token identifies the app as the GitHub Apps bot, such as
+  @jenkins-bot.
+
+  GitHub Apps can authenticate on behalf of the user, which is called
+  user-to-server requests. The flow to authorize is the same as the OAuth App
+  authorization flow. User-to-server tokens can expire and be renewed with a
+  refresh token
+
+(from https://docs.github.com/en/developers/apps/getting-started-with-apps/differences-between-github-apps-and-oauth-apps)
```
ericholscher, line 142: This is great to include, good idea 👍

docs/development/design/github-application.rst
```diff
@@ -0,0 +1,306 @@
...
+Use the token generated in the previous step as authorization in the cURL command:
+
+.. code:: bash
+
+   $ curl -H "Authorization: Bearer <JWT>" -H "Accept: application/vnd.github.v3+json" https://api.github.com/app
+
+
+.. note::
+
+   For some reason this is not working and I'm getting:
```
ericholscher, line 195: It says the app needs to be installed on at least 1 organization. That could be it.

ericholscher (commented):
This is a great overview. It definitely feels like we have some more thinking to do, but this is a great first step.

docs/development/design/github-application.rst
humitos: Yes. Once we implement GH Application we won't require to escalate permissions. We won't require `write` permissions or extra `read` permissions at any point in the future.

docs/development/design/github-application.rst
humitos: > This is just a Project mapping though, and means that if `readthedocs/template` gets a commit, then _every_ project on RTD with that `repo_url` will be built?

No.

This will trigger a build _only_ on the projects that were imported by users with permissions on `readthedocs/template` repository. That is, it appeared on the "Import Project" page and they clicked on "+" to import it. This is where the relation between `RemoteRepository -> Project` is created.

All the `readthedocs/template` forks and, all official `readthedocs/template` that were imported manually _won't be associated to any `RemoteRepository`_, so it won't trigger a build for those on commits.

docs/development/design/github-application.rst
humitos: This may be outdated and it's not required 😅  I already solved this problem in "Remote* models re-sync" without requiring this permission.

However, I had an idea in mind with this permission, but I didn't document it here. If we request `Organization - Members: Read-only` we can subscribe to related events (`Member`, `Membership`, `Team` and `Team add`) to immediately sync permissions when they are changed in GitHub.

Currently, we are re-syncing permissions at signup and on each login. So, this permission is not strictly required, but it could be better UX, tho.

We tried to implement the re-sync based on a webhook already with our current code at https://github.com/readthedocs/readthedocs.org/pull/7336

docs/development/design/github-application.rst
humitos: > so keeping auth & repo permissions separated might be the best solution

I'm afraid that this is not possible. At the "Login" page, when the user clicks on "Login with GitHub" we need to redirect them to the URL to authorize the OAuth Application or the GitHub Application --so we have to make that decision before showing the button. I don't really know how to solve this problem yet. This will require a little more thinking and research.

> Could even make the GH App approach optional for more security-minded people?

This question makes me think that we may need to show 2 different buttons: OAuth App and GH App. I know that would be terrible UX and will confuse a lot of users, tho.

I'd say that we should keep OAuth working as long as we have old users using this integration. However, *new users should not have the option to use OAuth App* and they must use GH App for all the good reasons we are migrating to it.

At some point, when no users using OAuth left, we should be able to delete it completely.

The most important thing to solve here is "How do we know that the person on the Login page is old or new to show them the appropriate button?"

agjohnson:
I don't have any strong input here, this all matches our conversations on GHA, as I understand. What are the next steps? Do you need to do more research on the remaining questions in this document?

humitos:
> What are the next steps? Do you need to do more research on the remaining questions in this document?

I put those questions mainly to not forget them and check if we can answer them _before_ the implementation phase, but I don't think we will answer all of them. I think some of them will be tested/answered at that phase (in particular, those related to Django AllAuth).

On the other hand, this question (https://github.com/readthedocs/readthedocs.org/pull/8445/files#r700103315) is important because it's not technical, but requires more UX thinking and I'm not sure how to solve that problem. So, input there would be a good place to collaborate.

docs/development/design/github-application.rst
agjohnson: Agreed having two buttons would be bad UX, we'd confuse users a lot here.

Is there any way we can try both, starting with one and if that attempt fails, try the other automatically? Or do either allow us to query to see if it is installed for the user?

To start, we could attempt authentication with the oauth app, and back down to the GH app -- swapping once we feel more users are using the GH app.

This could leave a single button in tact but for some users there would be an ugly wait to log in as we try to auth twice

> new users should not have the option to use OAuth App and they must use GH App

:+1:

agjohnson:
What is the first thing that we will work on to either start the implementation or to make design decisions around the implementation? I'm having trouble visualizing how this is going to be split up into digestible chunks of work, but the unanswered questions seem like the first place to start experimenting.

docs/development/design/github-application.rst
humitos: Honestly, I'm not finding a good way to solve this problem. I re-read the migration guide that GitHub provides and it's not super useful in this case.

I was thinking of a different approach:

1. implement all the GH application code
2. do not expose GH application to users yet, but make all the URLs to work
3. QA internally and make sure that everything works as expected
4. send emails + add a banner (with "call to action") telling organization owners to install our GH application (https://docs.github.com/en/developers/apps/getting-started-with-apps/migrating-oauth-apps-to-github-apps#direct-users-to-install-your-github-app-on-repositories)
5. send another email + add a banner communicating the EOL of the OAuth and telling them that starting on X date they must have GH App installed to use Read the Docs

As always, dropping support for something at a specific date is not good; but I'm not clearly seeing what else we can do besides forcing users to migrate. This may be doable in commercial, but probably won't in community.

Anyways, I'm open to getting some feedback and ideas on this because I ran out of ideas 😞

agjohnson: You might have the most knowledge and best opinions here. I'm not super familiar with the implementation at a technical level so can only take some wild shots in the dark.

The thing that I find awkward about that approach is that the user will get this notification to use the GHA method, allow the application on their user/organization, but then they need both the GHA and the oauth app installed until we do the cut over to the GHA.

It feels like we want users to have a different login entirely if they want to use the GHA implementation. Does it make sense to discuss using the beta instance for GHA in addition to the new templates?

benjaoming:
@humitos What a fantastic proposal! I love it - been thinking about GitHub Application support, too. Do you know if there is a migration path from the old OAuth App to GitHub App? Or are we indeed forced to either keep the old OAuth (like you suggested) or force all users to re-authorize?

humitos:
@benjaoming 

> Do you know if there is a migration path from the old OAuth App to GitHub App? Or are we indeed forced to either keep the old OAuth (like you suggested) or force all users to re-authorize?

Yeah, this whole thing was a pretty complex topic. We talked about this in many opportunities and different places.

_(there should be issues linked from/to this PR if you want to research a little more and read the whole history)_

I wrote this 1.5 years ago, but I remember that GitHub didn't provide a clear migration path and basically was: "Change this code, and re-authorize all your users" 🤷🏼 which was a no-go for us.

Then we decided to keep both applications (OAuth and GHApp) at the same time, while we perform the migration; but that implies "knowing who the user is _before_ log in" since the backend needs to use one app or the other. Finally, we thought that it could be a good idea to have a two-step login to know this upfront...


That's a small summary; but it's not nearly close to all we discussed 😄

benjaoming:
I wanted to _quickly_ add a user benefit here. But I see that the design doc has a structure where it's a bit hard for me to quickly add this exact part. So I'll just leave it here for later:

With the current OAuth integration, it's difficult for projects with multiple maintainers to know who's access is used for GitHub status callbacks. In a recent support case, it meant that access to setup a webhook was denied for a user that expected it to work since an owner of the GitHub repository was also an owner (maintainer) of the Read the Docs project. Clearly, this situation is fixed with GitHub Apps.

Secondary note: I try to use "GitHub Apps" everywhere since that's the most accurate name. Searching out existing information was a bit tricky, so I renamed a couple of items.

ericholscher:
This is pretty out of date, so we should update our thinking here and decide on a path to move forward with some kind of GH App.

agjohnson:
Work is continuing on #11780, I think we can close this one.

PR #8466, opened 2021-09-06
Docs: Add links to documentation on creating custom classes

nienn:
WIP

I know there might be more than one way to add a custom class, depending on several factors, but it would be useful to add here some reference to where such information could be found.


docs/guides/adding-custom-css.rst
```diff
@@ -77,3 +77,8 @@ you can include the theme's normal CSS using the CSS
     body {
         /* ... */
     }
+
+
+.. seealso::
+
+   You can also add custom classes to your html elements. See `Docutils Class <https://docutils.sourceforge.io/docs/ref/rst/directives.html#class>`_ and this related `Sphinx footnote... <https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#id2>`_ for more information.
```
astrojuanlu, line 84: Good catch! This was painfully difficult to find on the Sphinx docs, I will send a pull request there right now.

I think this information is too important to be in a `seealso` admonition! What about expanding it into its own subsection and adding a small example?

astrojuanlu (changes requested):

docs/guides/adding-custom-css.rst
nienn: Adding its own subsection sounds great. This is information that most people trying to add a custom CSS file will be looking for, so it makes sense to me to expand on it. Also, the information available today on the combined related docs is scarce and insufficient.

This `seealso` could become just a link to that new subsection.

astrojuanlu:
By the way, I opened https://github.com/sphinx-doc/sphinx/pull/9615 in Sphinx.

humitos:
It seems this change looks good. However, it was blocked on writing a new section for it. @nienn do you think you will have the time to write that section or should we merge what we currently have as it is? In that case, can you solve the conflicts?

nienn:
@humitos I think it's safe to merge this for now. The information can be useful independently of having it's own dedicated section.
We can come back here later and add the section, as discussed in the current PR in https://github.com/readthedocs/readthedocs.org/pull/8466#discussion_r703004877, on a separate PR.

👍  Solved the conflicts.

PR #8470, opened 2021-09-07
Docs: update the links to the dependency management content of setuptools docs

pzhlkj6612:
Hi.

Please note that I didn't modify the names of these hyperlinks because I have no idea what their new names should be. For example, the section name of https://setuptools.readthedocs.io/en/latest/userguide/dependency_management.html#optional-dependencies is "Optional dependencies" and the name of the link should match it, but this phrase is in the section about "extra_requirements", so I think it might confuse users.

What's your opinion?


stsewd (approved):
I think the current text is fine, thanks!

PR #8488, opened 2021-09-14
Add virtual env instructions to local installation

astrojuanlu:
Prompted by a conversation with @humitos

### DIFF
--- a/docs/development/docs.rst
+++ b/docs/development/docs.rst
@@ -27,13 +27,13 @@ you may want to verify those changes locally before pushing upstream.
 
 #. clone the `readthedocs.org` repository:
 
-   .. code-block:: bash
+   .. code-block:: console
 
       $ git clone --recurse-submodules https://github.com/readthedocs/readthedocs.org/
 
 #. create a virtual environment with Python 3.6 and activate it:
 
-   .. code-block:: bash
+   .. code-block:: console
 
       $ cd readthedocs.org
       $ python3.6 -m venv .venv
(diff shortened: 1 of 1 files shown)


astrojuanlu:
Rendered version: https://docs--8488.org.readthedocs.build/en/8488/development/docs.html#contributing-from-your-local-machine

docs/development/docs.rst
```diff
@@ -22,33 +22,37 @@ you can verify those changes through the documentation generated when you open a
...
+
+#. create a virtual environment with Python 3.6 and activate it:
+
+   .. code-block:: bash
 
-      git clone --recurse-submodules https://github.com/readthedocs/readthedocs.org/
+      $ cd readthedocs.org
+      $ python3.6 -m venv .venv
+      $ source .venv/bin/activate
+      (.venv) $
```
nienn: I would probably prefer a note here, or even a short explanation, stating that after the `$ source .venv/bin/activate` line the bash would get prefixed with `(.venv)` .
Changing this and the next `.. prompt::` blocks to `.. code-block::` has the disadvantage of the `$` sign now being selectable. Also, it is not immediately clear what is meant by the `(.venv) $` prefix.

nienn: Aha! Ok, this got solved with recent changes!

astrojuanlu: Ueah, right!

nienn:
This is much better!
Tested and it seems to work for me.

humitos (approved):

humitos:
Does make sense to add the virtualenv creation to the "Install common requirements" as well in the installation guide at https://docs.readthedocs.io/en/latest/development/install.html#set-up-your-environment

astrojuanlu:
Thanks @humitos ! Will do right now

astrojuanlu:
Well, on second thought, the virtual environment for https://docs.readthedocs.io/en/latest/development/install.html#set-up-your-environment is not that important, since it's only installing two applications used to orchestrate the rest:

https://github.com/readthedocs/common/blob/5eec3f61b7b0bb1ee39564bdf5d5a6239e8df964/dockerfiles/requirements.txt#L2-L3

Even the Python version is kind of irrelevant.

On the other hand, @nienn opened #8489, so we can continue the discussion about that part of the documentation there.

PR #8489, opened 2021-09-14
WIP add more explicit instructions in install.html

nienn:
**Changes:**
* Added some small steps.
* Moved other steps that felt a bit confusing.
* Added a pre-section on **Working with Docker Compose** to list the common steps to be repeated each time.
* Added a link back to the Front End Development page — there's a link there instructing to follow the docs here, but I don't think Front End Development needs to follow these instructions to the end, so it would be nice to send people back there when the relevant directions are over.

All of this is new to me — I'm also trying to follow the instructions in these docs — so all input is welcome!

_Built docs: https://docs--8489.org.readthedocs.build/en/8489/development/install.html_

### DIFF
--- a/docs/development/install.rst
+++ b/docs/development/install.rst
@@ -29,17 +29,18 @@ Set up your environment
 
    .. prompt:: bash
 
+      cd readthedocs.org
       pip install -r common/dockerfiles/requirements.txt
 
 #. build the Docker image for the servers:
 
-   .. warning::
-
-      This command could take a while to finish since it will download several Docker images.
-
    .. prompt:: bash
 
       inv docker.build
+
+   .. warning::
+
+      This command could take a while to finish since it will download several Docker images.
 
    .. tip::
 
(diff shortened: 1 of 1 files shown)


docs/development/install.rst
```diff
@@ -84,11 +85,34 @@ Check that everything works
...
-save some work while typing docker compose commands. This section explains these ``invoke`` commands:
+After the initial setup you'll need to run a few commands each time.
+
+#. get the containers running:
+
+  .. prompt:: bash
+
+    inv docker.up  # Options: --no-search --no-reload
+
+#. do stuff
```
humitos, line 106: I may call this section "Normal development workflow" or similar. Something that communicates that this is for day-to-day work.

Also, it can be expanded a little more what to do after "do stuff". I don't remember if the fontend guide explains how to check the JS/CSS changes; in that case we could like there.

On the other hand, if the change is for the backend, we could mention that the services are auto-reloaded unless `--no-reload` was passed, and in that case the services should be restarted manually (could link to `inv docker.restart <service>` or `inv docker.down` and `inv docker.up` again).

docs/development/install.rst
```diff
@@ -84,11 +85,34 @@ Check that everything works
...
+
+
+List of available commands
+""""""""""""""""""""""""""
+
+.. admonition:: Note
+
+  We wrote a wrapper with ``invoke`` around ``docker-compose`` to have some shortcuts and
+  save some work while typing docker compose commands.
```
humitos, line 115: I would keep the same title we have here and add an extra paragraph mentioning that this is the extensive list of all available commands (like the manpage for our tool) which is not required to know it completely and there may be commands that the developer will never use. Communicating that this has to be seen as a reference more than something that the developer must know.

humitos (commented):
Looks like a good start to me! Thanks for taking the time to update this guide! 💪🏼 

I left a proposal for the "Working with Docker Compose" section to be considered.

docs/development/install.rst
nienn: I made some changes here. Kept the title and added the paragraph:

> Bellow is the list of all available commands.

Also tried to work in a more extensive paragraph with your suggestions but I think that's mostly implicit by having the previous section and that extending on it here would add unnecessary density.

Let me know what you think of the changes.

docs/development/install.rst
nienn: Made some changes:

* Added the title _"Common development workflow"_ and made this a separate section from the next _"Working with Docker Compose"_.
* Added more info on the `--no-reload` and restarting process, first a short note inside the Warning on the _"Common development workflow"_ section and a bit more expanded bellow on the command list, under the `--no-reload` flag description.


I would also love to expand on the _"do stuff"_. Any suggestions? 👼

humitos:
I'll close this PR because is unfinished, unfortunately 😞

PR #8499, opened 2021-09-15
Set up bump2version

agjohnson:
I mentioned this a while ago in `common`, but bump2version (not bumpversion!)
makes managing version numbers around a lot of files pretty nice. This is copied
from sphinx_rtd_theme.

For example, the version now is 5.24.1. The workflow for versioning would be:

If we only merge bug fixes in for a release, we would run `bump2version patch`.
This would give `5.24.2alpha1`.

If we merged a feature into a release, we would run `bump2version minor`. This
would give `5.25.0alpha1`.

If we merged a backwards incompatible change in, we would run `bump2version major`.
This would give `6.0.0alpha1`.

When we go to release, we would run `bump2version release`, which drops the `alpha`.
We could be bumping the version number up as we merge PRs, or this could be on
our release procedures still too.

Alternatively, we don't even need the alpha designation and can configure this
all as release time procedures instead.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -1,6 +1,5 @@
 import os
 import sys
-from configparser import RawConfigParser
 
 import sphinx_rtd_theme
 
@@ -12,13 +11,6 @@ from django.utils import timezone
 
 import django
 django.setup()
-
-
-def get_version():
-    """Return package version from setup.cfg."""
-    config = RawConfigParser()
-    config.read(os.path.join('..', 'setup.cfg'))
-    return config.get('metadata', 'version')
 
 
 sys.path.append(os.path.abspath('_ext'))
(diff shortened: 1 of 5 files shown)


agjohnson:
I wasn't sure where to put the `bump2version` dependency though :shrug:

agjohnson:
Ping @readthedocs/backend

humitos (commented):
This looks good to me!

I don't have experience using `bump2version` but from the PR's description looks simple enough. I'd suggest to _not call bump2version on each PR_ because that will generate conflicts between PRs and I think it will be tedious to maintain.

I'd would start by calling it just once on releases for now to keep the same pattern that we already have and grow from there if necessary.

Also, this PR should have another one updating the `deploy.sh` script in -ops to use this new method.

Besides, there are some code in -corporate to get the version for Sentry that should be replaced with the new approach.

agjohnson:
Yeah we can scale back and drop the alpha designation from the pattern. We are going to experiment with the workflow above on sphinx_rtd_theme, but the release cadence is much slower there.

I'll try changing the deploy scripting next.

agjohnson:
I've found bumpver is easier to use. Going to close this and make a new one.

PR #8509, opened 2021-09-20
Change newsletter form

astrojuanlu:
Companion to https://github.com/readthedocs/blog/pull/128.


ericholscher (approved):

ericholscher:
I didn't test this, but assume it works 👍 . This is _definitely_ one where we would see value in adding a captcha. We should have an intermediate landing page for that, I think, if we can't do it on mailerlite.

PR #8513, opened 2021-09-21
POC guides landing redesign with Sphinx-design extension

nienn:
This is just a POC.

* Added the **sphinx-design extension** 
* re-made the guides landing page TOC manually using sphinx-design **cards**

* Tried to add a custom css but for some reason it's not compiling...
___
_(I'm still having problems following [this guide](https://docs.readthedocs.io/en/latest/development/docs.html#contributing-from-your-local-machine) as it does not always work. Sometimes it does works.
Haven't been able to check why this happens)_
___

**After** _(with the **custom.css updated file**)_:

Screenshot 2021-09-21 114841

### DIFF
--- a/docs/_static/css/custom.css
+++ b/docs/_static/css/custom.css
@@ -3,3 +3,58 @@
   box-shadow: 0 0 0 7px #fff, 0 0 0 8px #cecece;
   margin: 9px 17px 9px 3px;
 }
+
+:root {
+  --theme-gray: #404040;
+  --theme-light-gray: #777;
+  --theme-blue: #55a5d9;
+  --sd-color-card-border-hover: var(--theme-blue); 
+  --sd-color-shadow: rgba(0,0,0,.12);
+}
+
+#guides h2 {
+  font-size: 19px;
+}
+
+#guides .sd-card {
+  text-align: center;
+  border-top: 6px solid var(--theme-blue);
+}
+#guides .sd-card-large {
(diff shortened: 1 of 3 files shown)


agjohnson:
Using cards with descriptions here could be a good way to maybe visually promote some of our more popular guides. Using cards on all of the guides does seem hard to scan though.

I could see some mix of cards + definition lists used in the same way (title + description) being a good middle ground.

This might also depend a little on how we want the flow through the process to be reflected in that document. That is, if we keep this split up into sections for different parts of the authoring/building process, would we have some promoted guides in each section?

For example:

```markdown
Guides
======

Authoring and Sphinx
--------------------

[ Popular guide card ]  [ Popular guide card ]

Less popular guide
    Description for less popular guide

Less popular guide
    Description for less popular guide

Read the Docs
-------------

[ Popular guide card ]  [ Popular guide card ]

Less popular guide
    Description for less popular guide

Less popular guide
    Description for less popular guide
```

:point_up: This structure might hide popular guides too far below the fold?

humitos:
@nienn @agjohnson what's the status of this PR? Should we merge it or should we close it? It seems like a demo of how we could re-design the home page of our docs, but it doesn't seem like a completed work yet.

This depends on priorities, but I'd say that we should probably close it and come back when we have the time to make it move forward. What do you think?

nienn:
I think this might be closed. It was just a POC. We can come back when it becomes a priority again.

PR #8521, opened 2021-09-22
EmbedAPIv3: updates after QA with sphinx-hoverxref

humitos:
Updates required by https://github.com/readthedocs/sphinx-hoverxref/pull/146 found while doing QA.

### DIFF
--- a/readthedocs/embed/v3/views.py
+++ b/readthedocs/embed/v3/views.py
@@ -6,7 +6,6 @@ from urllib.parse import urlparse
 import requests
 
 from selectolax.parser import HTMLParser
-from pyquery import PyQuery as PQ  # noqa
 
 from django.conf import settings
 from django.core.cache import cache
@@ -71,12 +70,15 @@ class EmbedAPIBase(CachedResponseMixin, APIView):
 
         response = requests.get(url, timeout=settings.RTD_EMBED_API_DEFAULT_REQUEST_TIMEOUT)
         if response.ok:
+            # NOTE: we use ``response.content`` to get its binary
+            # representation. Then ``selectolax`` is in charge to auto-detect
+            # it. We trust more in selectolax for this than in requests.
             cache.set(
                 cache_key,
-                response.text,
+                response.content,
                 timeout=settings.RTD_EMBED_API_PAGE_CACHE_TIMEOUT,
             )
-            return response.text
+            return response.content
 
     def _get_page_content_from_storage(self, project, version_slug, filename):
         version = get_object_or_404(
@@ -205,6 +207,7 @@ class EmbedAPIBase(CachedResponseMixin, APIView):
                     # </dl>
                     node = node.parent
 
+        # FIXME: this is returning '\uf0c1' when it should be '' in the HTML itself
         return node.html
 
     def get(self, request):  # noqa
--- a/readthedocs/embed/v3/views.py
+++ b/readthedocs/embed/v3/views.py
@@ -140,7 +140,11 @@ class EmbedAPIBase(CachedResponseMixin, APIView):
 
         node = None
         if fragment:
-            selector = f'#{fragment}'
+            # NOTE: we use the `[id=]` selector because using `#{id}` requires
+            # scaping the selector since CSS does not support the same
+            # characters than the `id=` HTML attribute
+            # https://www.w3.org/TR/CSS21/syndata.html#value-def-identifier
+            selector = f'[id="{fragment}"]'
             node = HTMLParser(page_content).css_first(selector)
         else:
             html = HTMLParser(page_content)
--- a/readthedocs/embed/v3/views.py
+++ b/readthedocs/embed/v3/views.py
@@ -156,7 +156,10 @@ class EmbedAPIBase(CachedResponseMixin, APIView):
         if doctool == 'sphinx':
             # Handle ``dt`` special cases
             if node.tag == 'dt':
-                if 'glossary' in node.parent.attributes.get('class'):
+                if any([
+                        'glossary' in node.parent.attributes.get('class'),
+                        'citation' in node.parent.attributes.get('class'),
+                ]):
                     # Sphinx HTML structure for term glossary puts the ``id`` in the
                     # ``dt`` element with the title of the term. In this case, we
                     # return the parent node which contains the definition list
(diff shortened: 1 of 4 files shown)


readthedocs/embed/v3/views.py
```diff
@@ -71,12 +70,15 @@ def _download_page_content(self, url):
 
         response = requests.get(url, timeout=settings.RTD_EMBED_API_DEFAULT_REQUEST_TIMEOUT)
         if response.ok:
+            # NOTE: we use ``response.content`` to get its binary
+            # representation. Then ``selectolax`` is in charge to auto-detect
+            # its encoding. We trust more in selectolax for this than in requests.
```
ericholscher, line 75: Why do we trust it more? Is there a specific issue we've seen? I'm guessing that requests has a lot more eyes on it, but perhaps we're doing something different?

readthedocs/embed/v3/views.py
```diff
@@ -138,7 +140,11 @@ def _parse_based_on_doctool(self, page_content, fragment, doctool, doctoolversio
 
         node = None
         if fragment:
-            selector = f'#{fragment}'
+            # NOTE: we use the `[id=]` selector because using `#{id}` requires
+            # scaping the selector since CSS does not support the same
+            # characters than the `id=` HTML attribute
+            # https://www.w3.org/TR/CSS21/syndata.html#value-def-identifier
```
ericholscher, line 146: 💯  comment.

readthedocs/embed/v3/views.py
```diff
@@ -138,7 +140,11 @@ def _parse_based_on_doctool(self, page_content, fragment, doctool, doctoolversio
 
         node = None
         if fragment:
-            selector = f'#{fragment}'
+            # NOTE: we use the `[id=]` selector because using `#{id}` requires
+            # scaping the selector since CSS does not support the same
+            # characters than the `id=` HTML attribute
```
ericholscher, line 145:
```suggestion
            # NOTE: we use the `[id=]` selector because using `#{id}` requires
            # escaping the selector since CSS does not support the same
            # characters as the `id=` HTML attribute
```

readthedocs/embed/v3/views.py
```diff
@@ -163,32 +172,26 @@ def _parse_based_on_doctool(self, page_content, fragment, doctool, doctoolversio
...
+
+                    # Iterate over all the siblings (``.iter()``) of the parent
+                    # node and remove ``dt`` and ``dd`` that are not the ones
+                    # we are looking for. Then return the parent node as
+                    # result.
+                    #
+                    # Note that ``.iter()`` returns a generator and we modify
+                    # the HTML in-place, so we have to convert it to a list
+                    # before removing elements. Otherwise we break the
+                    # iteration before compliting it
```
ericholscher, line 190:
```suggestion
                    # Note that ``.iter()`` returns a generator and we modify
                    # the HTML in-place, so we have to convert it to a list
                    # before removing elements. Otherwise we break the
                    # iteration before completing it
```

readthedocs/settings/base.py
```diff
@@ -818,6 +818,8 @@ def DOCKER_LIMITS(self):
         r'docs\.python\.org',
         r'docs\.scipy\.org',
         r'docs\.sympy\.org',
+        r'www.sphinx-doc.org',
+        r'numpy\.org',
```
ericholscher, line 822: This is likely something that should be defined in the DB, or somewhere else that doesn't require a deploy to modify. It's definitely something we're going to need to iterate on pretty frequently I'm guessing, and perhaps even support customization at a project or organization level.

readthedocs/settings/docker_compose.py
```diff
@@ -66,7 +66,11 @@ def RTD_EXT_THEME_DEV_SERVER(self):
     @property
     def RTD_EMBED_API_EXTERNAL_DOMAINS(self):
         domains = super().RTD_EMBED_API_EXTERNAL_DOMAINS
-        domains.append(r'.*\.readthedocs\.io')
+        domains.extend([
+            r'.*\.readthedocs\.io',
+            r'.*\.org\.readthedocs\.build',
+            r'.*\.com\.readthedocs\.build',
+        ])
```
ericholscher, line 73: Should this include rtd-hosted.com as well?

tox.embedapi.ini
```diff
@@ -1,5 +1,5 @@
 [tox]
-envlist = sphinx-{18,20,21,22,23,24,30,31,32,33,34,35,40,41,latest}
+envlist = sphinx-{18,20,21,22,23,24,30,31,32,33,34,35,40,41,42,latest}
```
ericholscher, line 2: Hopefully someday we can delete some of these :D

ericholscher (approved):
Looks good, just a few small suggestions.

readthedocs/embed/v3/views.py
humitos: @nienn helped me here to understand why this happened 💪🏼

readthedocs/settings/base.py
humitos: I agree. We do have an issue opened for Django settings in the DB at https://github.com/readthedocs/readthedocs-ops/issues/1004 but I don't think that will progress too much. If that's the case, we can create a simple `EmbedDomain` model that we can relate to a Project and/or Organization as well.

humitos: Opened https://github.com/readthedocs/readthedocs.org/issues/8530 to continue the conversation there.

readthedocs/settings/docker_compose.py
humitos: Yeap!

tox.embedapi.ini
humitos: Yeah, I'd like to remove at least 1.x and 2.x; but there are a lot of people that still depend on them --so, I prefer to keep running tests on these versions at least until we start finding issues and making them harder to support.

readthedocs/embed/v3/views.py
humitos: I think we don't have a good amount of data tested yet. However, I've already find myself doing the solution from this SO question: https://stackoverflow.com/a/52615216

```
r = requests.get("https://martin.slouf.name/")
# override encoding by real educated guess as provided by chardet
r.encoding = r.apparent_encoding
# access the data
r.text
```

Then, I read that selectolax also uses the meta-tags from the HTML (https://selectolax.readthedocs.io/en/latest/parser.html#selectolax.parser.HTMLParser) to guess the encoding and it always worked without manual interaction as with `requests`.

Anyways, I don't have strong opinions here. I suppose whatever we choose here, it will fail in some case that the other library wouldn't 🙃

PR #8536, opened 2021-09-28
Requirement: ping django-redis-cache to git tag

humitos:
For some reason django-redis-cache==2.1.3 breaks with:

```
@get_client(write=True)
def get_or_set(self, client, key, func, timeout=DEFAULT_TIMEOUT):
    if not callable(func):
        raise Exception("Must pass in a callable")
```

Looks like an older version gets installed somehow that does not contains the
code we need. So, we force it to be pinned to a git tag.


ericholscher (approved):
Silly, but it works 🤷

PR #8540, opened 2021-09-28
EmbedAPIv3: proxy URLs to be available under `/_/`

humitos:
Follow the same pattern than with APIv2 to access EmbedAPI under `<docs.domain.org>/_/api/v3/embed/`

### DIFF
--- a/readthedocs/api/v3/proxied_views.py
+++ b/readthedocs/api/v3/proxied_views.py
@@ -1,0 +1,14 @@
+from readthedocs.core.utils.extend import SettingsOverrideObject
+from readthedocs.embed.v3.views import EmbedAPIBase
+
+
+class ProxiedEmbedAPIBase(EmbedAPIBase):
+
+    # DRF has BasicAuthentication and SessionAuthentication as default classes.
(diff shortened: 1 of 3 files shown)


readthedocs/api/v3/proxied_views.py
```diff
@@ -0,0 +1,14 @@
+from readthedocs.core.utils.extend import SettingsOverrideObject
+from readthedocs.embed.v3.views import EmbedAPIBase
+
+
+class ProxiedEmbedAPIBase(EmbedAPIBase):
+
+    # DRF has BasicAuthentication and SessionAuthentication as default classes.
+    # We don't support neither in the community site.
+    authentication_classes = []
```
ericholscher, line 9: Probably makes sense to make this a Mixin if we're using it on all proxied API's.

ericholscher (approved):

readthedocs/api/v3/proxied_views.py
humitos: Good point! I opened an issue to track this https://github.com/readthedocs/readthedocs.org/issues/8542

PR #8541, opened 2021-09-28
Script tools cache: fix environment variables

humitos:
`AWS_REGION` is required in production and `--endpoint-url` is not.
Besides, `AWS_BUILD_TOOLS_BUCKET_NAME` is required as well in production.


ericholscher (approved):
Simple enough.

PR #8547, opened 2021-09-30
Document new `build` specification

astrojuanlu:
I decided to add a new section after the current keys to simplify how we present this information, but there are other alternatives.

On the other hand, I have highlighted that "this is beta", since that is my current understanding of the functionality. If this is incorrect, I'll be happy to remove the admonition.

I chose the word "new" because I assume that, at some point in the future, we will reorder this documentation and declare the current one "legacy". However, [it might be a bad idea](https://ddbeck.com/new-is-a-mark/). If we expect this to last long (and it is often the case that we don't expect things to last long, but they do), we should choose a better name.

### DIFF
--- a/docs/config-file/v2.rst
+++ b/docs/config-file/v2.rst
@@ -320,17 +320,14 @@ build.tools
 Version specifiers for each tool. It must contain at least one tool.
 
 :Type: ``dict``
-:Keys: ``python``, ``nodejs``, ``rust``, ``golang``
+:Options: ``python``, ``nodejs``, ``rust``, ``golang``
+:Required: ``true``
 
 build.tools.python
 ``````````````````
 
-Python version to use. Supported values are:
-
-* CPython 2.7 and 3.6+
-* PyPy3.7
-* Miniconda3 4.7
-* Mambaforge 4.10
+Python version to use.
+You can use several interpretes and versions, from CPython, PyPy, Miniconda, and Mamba.
 
 .. note::
 
@@ -339,7 +336,17 @@ Python version to use. Supported values are:
    for more information.
 
 :Type: ``string``
-:Options: ``2.7``, ``3``, ``3.6``, ``3.7``, ``3.8``, ``3.9``, ``3.10``, ``pypy3.7``, ``miniconda3-4.7``, ``mambaforge-4.10``
+:Options:
+  - ``2.7``
+  - ``3``
+  - ``3.6``
+  - ``3.7``
+  - ``3.8``
+  - ``3.9``
+  - ``3.10``
+  - ``pypy3.7``
+  - ``miniconda3-4.7``
+  - ``mambaforge-4.10``
 
 build.tools.nodejs
 ``````````````````
--- a/docs/config-file/v2.rst
+++ b/docs/config-file/v2.rst
@@ -21,6 +21,13 @@ Below is an example YAML file which shows the most common configuration options:
          # Required
          version: 2
 
+         # Set the version of Python and other tools you might need
+         build:
+           os: ubuntu-20.04
+           tools:
+             python: "3.9"
+             nodejs: "16"
+
          # Build documentation in the docs/ directory with Sphinx
          sphinx:
             configuration: docs/conf.py
@@ -29,9 +36,8 @@ Below is an example YAML file which shows the most common configuration options:
          formats:
             - pdf
 
-         # Optionally set the version of Python and requirements required to build your docs
+         # Optionally declare the Python requirements required to build your docs
          python:
-            version: "3.7"
             install:
             - requirements: docs/requirements.txt
 
@@ -46,12 +52,17 @@ Below is an example YAML file which shows the most common configuration options:
          # Required
          version: 2
 
+         # Set the version of Python and other tools you might need
+         build:
+           os: ubuntu-20.04
+           tools:
+             python: "3.9"
+
          mkdocs:
            configuration: mkdocs.yml
 
-         # Optionally set the version of Python and requirements required to build your docs
+         # Optionally declare the Python requirements required to build your docs
          python:
-            version: "3.7"
             install:
             - requirements: docs/requirements.txt
 
--- a/docs/config-file/v2.rst
+++ b/docs/config-file/v2.rst
@@ -159,6 +159,16 @@ Configuration of the Python environment to be used.
          path: another/package
      system_packages: true
 
+python.version
+``````````````
+
+.. warning::
+
+   This option is now deprecated
+   and replaced by :ref:`config-file/v2:build.tools.python`.
+   See :ref:`config-file/v2:python.version (legacy)`
+   for the description of this option.
+
 python.install
 ``````````````
 
--- a/docs/config-file/v2.rst
+++ b/docs/config-file/v2.rst
@@ -359,7 +359,7 @@ You can use several interpretes and versions, from CPython, PyPy, Miniconda, and
 :Type: ``string``
 :Options:
   - ``2.7``
-  - ``3``
+  - ``3`` (last stable CPython version)
   - ``3.6``
   - ``3.7``
   - ``3.8``
--- a/docs/config-file/v2.rst
+++ b/docs/config-file/v2.rst
@@ -688,7 +688,6 @@ Schema
 You can see the complete schema
 `here <https://github.com/readthedocs/readthedocs.org/blob/master/readthedocs/rtd_tests/fixtures/spec/v2/schema.json>`_.
 
-.. _new-build:
 
 Legacy ``build`` specification
 ------------------------------
@@ -775,7 +774,7 @@ Changes
 - The ``build.image`` setting has been replaced by ``build.os``.
   See :ref:`config-file/v2:build.os`.
   Alternatively, you can use the legacy ``build.image``
-  that now only has two options: ``latest`` (default) and ``stable``.
+  that now has only two options: ``latest`` (default) and ``stable``.
 - The settings ``python.setup_py_install`` and ``python.pip_install`` were replaced by ``python.install``.
   And now it accepts a path to the package.
   See :ref:`config-file/v2:Packages`.
--- a/docs/config-file/v2.rst
+++ b/docs/config-file/v2.rst
@@ -686,7 +686,7 @@ Schema
 ------
 
 You can see the complete schema
-`here <https://github.com/readthedocs/readthedocs.org/blob/master/readthedocs/rtd_tests/fixtures/spec/v2/schema.yml>`_.
+`here <https://github.com/readthedocs/readthedocs.org/blob/master/readthedocs/rtd_tests/fixtures/spec/v2/schema.json>`_.
 
 .. _new-build:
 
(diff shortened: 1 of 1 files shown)


astrojuanlu:
🤖 Rendered version: https://docs--8547.org.readthedocs.build/en/8547/config-file/v2.html#new-build-specification

docs/config-file/v2.rst
```diff
@@ -636,6 +640,112 @@ Schema
...
+
+   version: 2
+
+   build:
+     os: ubuntu-20.04
+     tools:
+       python: 3.9
+       nodejs: 16
+       rust: 1.55
+       golang: 1.17
```
humitos, line 673: These should be all strings, with surrounded by `""`

docs/config-file/v2.rst
```diff
@@ -636,6 +640,112 @@ Schema
...
+
+build (new)
+~~~~~~~~~~~
+
+Configuration for the documentation build process.
+
+build.os
+````````
+
+The Docker image used for building the docs.
```
humitos, line 691: It may be worth mentioning Read the Docs specifically in this phrase, like "The operating system the Read the Docs' based Docker image is based on". Otherwise, I think it can be interpreted as the regular `ubuntu:20.04` or any other image could be used.

docs/config-file/v2.rst
```diff
@@ -636,6 +640,112 @@ Schema
...
+
+:Type: ``string``
+:Options: ``ubuntu-20.04``
+:Default: ``ubuntu-20.04``
+
+build.tools
+```````````
+
+Version specifiers for each tool. It must contain at least one tool.
+All tools are managed with `asdf <https://asdf-vm.com/>`_.
```
humitos, line 701: I'm not sure if it's important to users to know how it's implemented. We can probably remove this sentence.

docs/config-file/v2.rst
```diff
@@ -636,6 +640,112 @@ Schema
...
+
+:Type: ``dict``
+:Keys: ``python``, ``nodejs``, ``rust``, ``golang``
+
+build.tools.python
+``````````````````
+
+Python version to use. Supported values are:
+
+* CPython 2.7 and 3.6+
```
humitos, line 711: I think that I missed this in the original review, but `3.6` was not supposed to be supported 🙃 . We talked about the latest 3 releases and the current `-dev`; but it seems that we support the latest 4 releases, now 😄

docs/config-file/v2.rst
```diff
@@ -636,6 +640,112 @@ Schema
...
+* PyPy3.7
+* Miniconda3 4.7
+* Mambaforge 4.10
+
+.. note::
+
+   If you use Miniconda3 or Mambaforge, you can select the Python version
+   using the ``environment.yml`` file. See our :doc:`/guides/conda` guide
+   for more information.
```
humitos, line 720: 💯

humitos (commented):
This is great! I'm happy that we have it documented now! 🥳 

I have opinions on this, tho 😁 . In particular, what's the config we want our users to use by default? I think that our documentation does excellent work as reference documentation, but I think it's very easy to get lost with the different and very detailed ways we have documented.

I'd suggest promoting this `build.tools` and `build.os` more and try to hide (maybe even delete) the _old version_ once it's tested enough.

Also, I'd update the quick n' working example from the top to use the new version, so most users start using the new version.

In general, my question is: what are the benefits of having documented both ways of doing the same at the same level of prominence?

Note that I'm 👍🏼 on merging this, but it seems that I woke up "like that" today 😃

docs/config-file/v2.rst
astrojuanlu: Woops, thanks

astrojuanlu:
Thanks for the comments @humitos ! About promoting this more or not, I think it boils down to how confident we are on it. I saw that you mentioned to some people that "it was in beta", that's why I was a bit more cautious. But I'm +1 on simplifying our docs, declare this the new way, and add a small paragraph about the "legacy" way, which can go away after some time.

I'll wait for others to chime in.

docs/config-file/v2.rst
```diff
@@ -636,6 +640,112 @@ Schema
...
+
+build.tools.python
+``````````````````
+
+Python version to use. Supported values are:
+
+* CPython 2.7 and 3.6+
+* PyPy3.7
+* Miniconda3 4.7
+* Mambaforge 4.10
```
stsewd, line 714: We are listing this values as options, but the "real" options are below, I think we could put this in a paragraph instead.

`You can use several interpretes and versions, from CPython, PyPy, Miniconda, and Mamba`.

Or just keep the list below.

docs/config-file/v2.rst
```diff
@@ -636,6 +640,112 @@ Schema
...
+
+.. note::
+
+   If you use Miniconda3 or Mambaforge, you can select the Python version
+   using the ``environment.yml`` file. See our :doc:`/guides/conda` guide
+   for more information.
+
+:Type: ``string``
+:Options: ``2.7``, ``3``, ``3.6``, ``3.7``, ``3.8``, ``3.9``, ``3.10``, ``pypy3.7``, ``miniconda3-4.7``, ``mambaforge-4.10``
```
stsewd, line 723: You could list them as a list to make it easier to read.

```suggestion
:Options:
   - ``2.7``
   - ``3``
```

And we should explain that using `3` will use the latest stable version from the list (3.9 in this case)

humitos: I think we should consider using a different label before documenting it. I was suggesting using `latest` for this use case instead, which is more explicit

stsewd: I'm not really sure about using latest, if python 4.0 is released you don't want to update to it yet. 3 is more explicit for me.

docs/config-file/v2.rst
```diff
@@ -636,6 +640,112 @@ Schema
 You can see the complete schema
 `here <https://github.com/readthedocs/readthedocs.org/blob/master/readthedocs/rtd_tests/fixtures/spec/v2/schema.yml>`_.
 
+.. _new-build:
+
+New ``build`` specification
+---------------------------
```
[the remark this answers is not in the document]
astrojuanlu: For the reasons stated above, I'll refrain from calling this "new" and use the word "beta" instead. That way, when this stops being beta, I will just remove the word.

docs/config-file/v2.rst
```diff
@@ -636,6 +640,112 @@ Schema
...
+
+.. warning::
+
+   This functionality is in beta. If you find any inconsistencies
+   or have feedback, please `open an
+   issue <https://github.com/readthedocs/readthedocs.org/issues/>`_.
+
+Read the Docs has introduced a more flexible ``build`` specification
```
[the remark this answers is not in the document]
astrojuanlu: Done!

docs/config-file/v2.rst
astrojuanlu: Changed the display of the options, and left `"3"` as an option.

astrojuanlu:
Addressed comments, ready for second round of review.

docs/config-file/v2.rst
```diff
@@ -634,7 +686,77 @@ Schema
 ------
 
 You can see the complete schema
-`here <https://github.com/readthedocs/readthedocs.org/blob/master/readthedocs/rtd_tests/fixtures/spec/v2/schema.yml>`_.
+`here <https://github.com/readthedocs/readthedocs.org/blob/master/readthedocs/rtd_tests/fixtures/spec/v2/schema.json>`_.
+
+.. _new-build:
```
stsewd, line 691:
```suggestion
```

docs/config-file/v2.rst
```diff
@@ -149,24 +159,6 @@ Configuration of the Python environment to be used.
          path: another/package
      system_packages: true
 
-python.version
```
stsewd, line 152: Not sure about removing mention of python.version here, maybe add a note that this is a deprecated option now? (it could be a note in the section above)

docs/config-file/v2.rst
```diff
@@ -650,8 +772,10 @@ Changes
   See :ref:`config-file/v2:Requirements file`.
 - The setting ``conda.file`` was renamed to ``conda.environment``.
   See :ref:`config-file/v2:conda.environment`.
-- The ``build.image`` setting now only has two options: ``latest`` (default) and ``stable``.
-  See :ref:`config-file/v2:build.image`.
+- The ``build.image`` setting has been replaced by ``build.os``.
+  See :ref:`config-file/v2:build.os`.
+  Alternatively, you can use the legacy ``build.image``
+  that now only has two options: ``latest`` (default) and ``stable``.
```
stsewd, line 778:
```suggestion
  that now has only two options: ``latest`` (default) and ``stable``.
```

docs/config-file/v2.rst
astrojuanlu: Damn, I thought I had removed it haha

docs/config-file/v2.rst
astrojuanlu: As it's incompatible with `build.os`, which we promoted to the top, I think it makes sense to move it to the "legacy" section below

stsewd: yeah, but I think we should put a note like

```
warning: the python.version option is now deprecated, see ...
```

otherwise people would be looking for the python.version option under the python section and be confused that it disappeared.

astrojuanlu: I am not sure. What users would look for that option?

- The ones that already have a config working, know trivially how to change it (tweak the number in `.readthedocs.yaml`)
- The ones that are looking for a way to configure it, we want them to use the new method, rather than the old one

am I right?

stsewd: There is the option for users looking to add something new under the `python` option (like a requirement, or just re-read what each option does) and realize that `python.version` isn't longer listed there. They will find it eventually, but I've been there were I remember something used to be there and then one day it isn't and had to re-read the whole thing to know what happened :x

astrojuanlu: Added a section with a deprecation warning and a link to the actual docs:

Screenshot 2021-10-04 at 22-00-48 Configuration File V2 — Read the Docs 5 24 0 documentation

astrojuanlu:
Done 💪🏽

docs/config-file/v2.rst
```diff
@@ -297,43 +294,104 @@ The path to the Conda environment file, relative to the root of the project.
...
+   If you use Miniconda3 or Mambaforge, you can select the Python version
+   using the ``environment.yml`` file. See our :doc:`/guides/conda` guide
+   for more information.
 
 :Type: ``string``
-:Options: ``stable``, ``latest``
-:Default: ``latest``
+:Options:
+  - ``2.7``
+  - ``3``
```
humitos, line 362: Should we mention what `3` means here?

humitos (approved):

docs/config-file/v2.rst
astrojuanlu: Good idea. "Last stable version of CPython"?

astrojuanlu: Done ✔️

astrojuanlu:
Tiny improvement of the example at https://github.com/readthedocs/readthedocs.org/pull/8564

PR #8558, opened 2021-10-05
Build: support Python 3.10.0 stable release

humitos:
Python 3.10.0 was released yesterday. This PR updates the version mapping to
point to that release and makes `python: 3` to point to 3.10 as well.

### DIFF
--- a/readthedocs/settings/base.py
+++ b/readthedocs/settings/base.py
@@ -547,7 +547,7 @@ class CommunityBaseSettings(Settings):
                 '3.7': '3.7.12',
                 '3.8': '3.8.12',
                 '3.9': '3.9.7',
-                '3.10': '3.10.0rc2',
+                '3.10': '3.10.0',
                 'pypy3.7': 'pypy3.7-7.3.5',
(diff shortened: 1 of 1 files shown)


stsewd (approved):

readthedocs/settings/base.py
```diff
@@ -565,7 +565,7 @@ def TEMPLATES(self):
         },
     }
     # Always point to the latest stable release.
-    RTD_DOCKER_BUILD_SETTINGS['tools']['python']['3'] = RTD_DOCKER_BUILD_SETTINGS['tools']['python']['3.9']
+    RTD_DOCKER_BUILD_SETTINGS['tools']['python']['3'] = RTD_DOCKER_BUILD_SETTINGS['tools']['python']['3.10']
```
ericholscher, line 568: Should we wait a little bit to see if any bugs fall out before making this default? That will upgrade a decent number of people I'm guessing, and I don't know if the ecosystem is ready for it.

humitos: This will only affect only people using `build.tools.python: 3` --which at this moment, there isn't anyone probably.

PR #8559, opened 2021-10-05
Build: update ca-certificates before cloning

humitos:
This is a temporal solution while we decide how to fix the real problem. For
now, we are installing a newer version of `ca-certificates` before starting to
clone the repository.

Reference #8555


readthedocs/projects/tasks.py
```diff
@@ -901,6 +901,23 @@ def setup_vcs(self, environment):
...
+            'apt-get', 'update', '--assume-yes', '--quiet',
+            user=settings.RTD_DOCKER_SUPER_USER,
+            record=False,
+        )
+        self.setup_env.run(
+            'apt-get', 'install', '--assume-yes', '--quiet', 'ca-certificates',
+            user=settings.RTD_DOCKER_SUPER_USER,
+            record=False,
+        )
+
```
ericholscher, line 920: This seems a little scary to introduce to every build. Can we at least put it behind a feature flag that we default True or something, so we can easily turn it on/off in prod if we hit issues?

humitos: Done!

PR #8572, opened 2021-10-07
Return a 401 in proxito when a project is banned

ericholscher:
This will allow us to instantly cut off doc serving for banned projects.
It hopefully won't increase the query load in proxito too much,
but seems worth it in this case at least until we figure out deletion.


humitos:
This is already implemented in #8632

ericholscher:
@humitos Thanks.

PR #8585, opened 2021-10-14
Write the modified conda env to a temporary file

Cadair:
This is a stab in the dark at fixing #8201 

It attempts to save the conda env file to a location outside of the checkout so that it doesn't interfere with the checkout / version number.

### DIFF
--- a/readthedocs/doc_builder/python_environments.py
+++ b/readthedocs/doc_builder/python_environments.py
@@ -771,7 +771,8 @@ class Conda(PythonEnvironment):
             dependencies.append(pip_dependencies)
             dependencies.extend(conda_requirements)
             environment.update({'dependencies': dependencies})
-            _, output_filename = tempfile.mkstemp(dir=Path(self.checkout_path) / ".." / "..", suffix=".yml")
+            _, output_filename = tempfile.mkstemp(dir=Path(self.checkout_path) / ".." / "..",
+                                                  suffix=".yml")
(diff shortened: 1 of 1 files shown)


humitos:
Hi all! I think the idea makes sense: use a temporal file _outside_ the git checkout to avoid having a dirty state. However, we do modify other files as well (like `conf.py`) and we should take a look at them also.

Regarding the implementation it self, we should double check where else `self.config.conda.environment` is used and change those for the new location. I may make sense to make `self.config.conda.environment` to return the temporal file instead, tho.

Finally, what happened with the suggestion from this comment https://github.com/readthedocs/readthedocs.org/issues/8201#issuecomment-845277688. Have you tried it?

adrien-berchet:
Hello there,
I proposed something simpler to @Cadair (with a PR on this PR) but I am not sure he saw the notification: https://github.com/Cadair/readthedocs.org/pull/1
I can open a new PR here if you prefer.
Anyway, feel free to comment it, it would be nice to solve this issue :)

humitos:
Thanks for opening this PR. We are implementing pre/post commands in our builders. Once that is done, this issue can be solved with the git command described at https://github.com/readthedocs/readthedocs.org/issues/8201#issuecomment-1018511443

I'm closing this PR because I don't think we will include a solution like this in the core application. We can come back to this and re-open if we feel it. Thanks!

Issue #8529, opened 2021-09-27
mkdocs: Values with special `!` syntax are set to null on mkdocs.yml


j1elo:
## Details

* Read the Docs project URL: https://readthedocs.org/projects/openvidu/
* Build URL (if applicable): https://readthedocs.org/projects/openvidu/builds/14834239/
* Read the Docs username (if applicable):

Our project is using:

* MkDocs 1.2.2
* [Extra Context](https://www.mkdocs.org/dev-guide/themes/#extra-context) option in `mkdocs.yml`, with the [ENV syntax](https://www.mkdocs.org/user-guide/configuration/#environment-variables) to read environment variables, making the `READTHEDOCS` vars available to the template context:

    ```
    extra:
      READTHEDOCS: !ENV [READTHEDOCS, "(no var)"]
      READTHEDOCS_VERSION: !ENV [READTHEDOCS_VERSION, "(no var)"]
      READTHEDOCS_PROJECT: !ENV [READTHEDOCS_PROJECT, "(no var)"]
    ```

* Attempting to read the RtD environment variables as explained in [Build environment](https://docs.readthedocs.io/en/stable/builds.html#build-environment) and [How do I change behavior when building with Read the Docs?](https://docs.readthedocs.io/en/stable/faq.html#how-do-i-change-behavior-when-building-with-read-the-docs).

## Expected Result

1. The `cat mkdocs.yml` build step should show proper values for env vars `READTHEDOCS`, `READTHEDOCS_VERSION`, and `READTHEDOCS_PROJECT`.

2. A test html file with these contents:

    ```
    console.log("config.extra: {{ config.extra }}");
    console.log("config.extra.READTHEDOCS: {{ config.extra.READTHEDOCS }}");
    console.log("config.extra.READTHEDOCS_VERSION: {{ config.extra.READTHEDOCS_VERSION }}");
    ```

   should print something similar to this in the browser console:

    ```
    config.extra: {'READTHEDOCS': True, 'READTHEDOCS_VERSION': 1.2.3, 'READTHEDOCS_PROJECT': openvidu}
    config.extra.READTHEDOCS: True
    config.extra.READTHEDOCS_VERSION: 1.2.3
    ```

## Actual Result

1. The `cat mkdocs.yml` build step shows that `READTHEDOCS`, `READTHEDOCS_VERSION`, and `READTHEDOCS_PROJECT` are all `null`:

    ```
    extra:
      READTHEDOCS: null
[...]

PR #8586, opened 2021-10-14
Documentation names cleanup

### LINKED ISSUE #8573: Clarify webhook and integration documentation
Coming from #8522:

> So, in our docs we have a `webhooks` page...

Seeking some feedback on the terminology here. We could:

- Rename `/webhooks.html` to `/integrations.html`
- Replace `Incoming Webhooks and Automation` by `Integrations and Automation`,
- And use the term "integration" rather than "webhook" throughout the page, unless it refers to a UI option.

This would be more consistent with the UI. However, each of these individual integrations is still called "incoming webhook":

image

Therefore, for full consistency, probably we should rename "incoming webhook" to just "integration". I just checked that this is what GitHub and MailerLite do.

What do folks think?

And then, after renaming `/webhooks.html`, would it be wise to repurpose the URL for outgoing webhooks? Otherwise, we would keep them at https://docs.readthedocs.io/en/stable/build-notifications.html#using-webhook, including the upcoming new features introduced at https://github.com/readthedocs/readthedocs.org/pull/8522.

cc @readthedocs/advocacy @readthedocs/backend

astrojuanlu:
This fixes #8573, fixes a stray broken link, and addresses some confusion that arose at #8529. In summary:

- Renames `/webhooks` to `/integrations`
- Refers to "Integrations" rather than "Webhooks" when a generic term is appropriate or when we refer to our feature, leaves "webhooks" when speaking about a specific incoming webhook
- Removes a confusing admonition on top of the "Edit Source integration" guide, that pointed the users towards a design document that was never implemented
- Renames that "Edit Source" guide from `/vcs` to `/edit-source`
- Improves the wording of that guide, since it was implying that the user had to add some variables to `html_context`. Read the Docs injects them anyway (just checked on [a project using furo](https://readthedocs.org/projects/django-cookiecutter/builds/14990532/)), so developers can focus on using them in their themes.
  - Like other things, it would be cool to have this as an extension instead.

We will need to create appropriate redirections for `/webhooks` and `/vcs`.

I wanted to do this before documenting #8522, since we want to have more in-depth documentation about (outgoing) webhooks.

### DIFF
--- a/docs/integrations.rst
+++ b/docs/integrations.rst
@@ -1,5 +1,5 @@
-Integrations
-============
+VCS Integrations
+================
 
 Read the Docs provides integrations with several VCS providers to detect changes to your
 documentation and versions, mainly using *webhooks*.


astrojuanlu:
Thanks @stsewd for the feedback, addressed 👍🏽

PR #8594, opened 2021-10-18
Design doc: Landing Page User Stories

nienn:
Design document for the landing page user stories.
The purpose is to gather ideas and open the debate about what needs to be built into the community site home page from the perspective of the end user.

This will help us define our standard users and map specific user needs, which will in turn allow us to better define specific features and goals.

----
Edit based on @ericholscher first comment bellow.
>  Am I trying to see if the user stories cover all the various things that users might think

I'm sure I didn't yet cover all the user needs here.
Instead I'd love to see this Design doc being used as a platform for brainstorming on those needs — _in the format of user stories_ — for the landing page. It should be a place where anyone can contribute with an informal story, written from the perspective of a user, a developer or other stakeholder, that needs to accomplish some task or goal on the landing page.

User stories do not translate directly into final features. They have the feeling of ideas in a brainstorm and can help us define future features and goals, but are not yet in any formal format.

### DIFF
--- a/docs/development/design/ext-theme/landing-page/user-stories.rst
+++ b/docs/development/design/ext-theme/landing-page/user-stories.rst
@@ -8,7 +8,7 @@ User sectors
 ***************
 
 +-------------------+--------------+--------------+--------------+--------------+
-| User Levels →     | Potential    | Initiate     | Intermediate | Advanced     |
+| User Levels →     | Potential    | Beginner     | Intermediate | Advanced     |
 +-------------------+              |              |              |              |
 | User Groups       |              |              |              |              |
 +===================+==============+==============+==============+==============+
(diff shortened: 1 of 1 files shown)


nienn:
@ericholscher I edited the initial description so, hopefully, the document objectives are more clear.

humitos:
@nienn if I understand correctly, most of this work is already done in the https://github.com/readthedocs/site-community/ repository and we should probably close this PR. Is that correct? cc @agjohnson

nienn:
@humitos 
This was part of the initial UX research for the landing page, and the work there is mostly done. It's still open because I'd hoped it could work as a place to submit new user needs when they arose, but we've come to the conclusion that it's not the best platform and/or model for that discussion so I'm +1 on closing.

PR #8637, opened 2021-10-30
Drop beta from title of build config option

agjohnson:
The title will eventually change to not be beta, and any existing links to this
anchor will not scroll to this section, which is always confusing. Let's drop
this early, there's already an admonition below the title.


astrojuanlu (approved):
Good point 👍🏽 Thanks @agjohnson !

Issue #8616, opened 2021-10-26
Build Failed. TypeError: 'generator' object is not subscriptable


GilimkhanovDenis:
_**Update from @astrojuanlu**: see [this comment](https://github.com/readthedocs/readthedocs.org/issues/8616#issuecomment-952034858) for a workaround for this issue, [our blog](https://blog.readthedocs.com/build-errors-docutils-0-18/) for a more detailed explanation, and [this comment](https://github.com/readthedocs/readthedocs.org/issues/8616#issuecomment-973994826) for current status of this issue_

---

Success Build: https://readthedocs.org/projects/diadocsdk-1c/builds/15092197/
Failed Build: https://readthedocs.org/projects/diadocsdk-1c/builds/15095749/

It's the same commit

What is changed during 8 hours ?
What can I do for successful build?

```
building [html]: targets for 402 source files that are out of date
updating environment: 402 added, 0 changed, 0 removed
reading sources... [  0%] ComObjects/AcceptanceCertificateBuyerContent
reading sources... [  0%] ComObjects/AcceptanceCertificateContent
reading sources... [  0%] ComObjects/AcceptanceCertificateItem
reading sources... [  0%] ComObjects/AcceptanceCertificateSellerContent

Traceback (most recent call last):
  File "/home/docs/checkouts/readthedocs.org/user_builds/diadocsdk-1c/envs/dev/lib/python3.7/site-packages/sphinx/cmd/build.py", line 304, in build_main
    app.build(args.force_all, filenames)
  File "/home/docs/checkouts/readthedocs.org/user_builds/diadocsdk-1c/envs/dev/lib/python3.7/site-packages/sphinx/application.py", line 341, in build
    self.builder.build_update()
  File "/home/docs/checkouts/readthedocs.org/user_builds/diadocsdk-1c/envs/dev/lib/python3.7/site-packages/sphinx/builders/__init__.py", line 347, in build_update
    len(to_build))
  File "/home/docs/checkouts/readthedocs.org/user_builds/diadocsdk-1c/envs/dev/lib/python3.7/site-packages/sphinx/builders/__init__.py", line 360, in build
    updated_docnames = set(self.read())
  File "/home/docs/checkouts/readthedocs.org/user_builds/diadocsdk-1c/envs/dev/lib/python3.7/site-packages/sphinx/builders/__init__.py", line 468, in read
    self._read_serial(docnames)
[...]

PR #8655, opened 2021-11-08
Update requirements pinning

astrojuanlu:
See https://github.com/readthedocs/readthedocs.org/issues/8616#issuecomment-961883441.

### DIFF
--- a/docs/guides/reproducible-builds.rst
+++ b/docs/guides/reproducible-builds.rst
@@ -47,13 +47,17 @@ A configuration file with explicit dependencies looks like this:
 
    version: 2
 
+   build:
+     os: "ubuntu-20.04"
+     tools:
+       python: "3.9"
+
    # Build from the docs/ directory with Sphinx
    sphinx:
      configuration: docs/conf.py
 
    # Explicitly set the version of Python and its requirements
    python:
-     version: 3.7
      install:
        - requirements: docs/requirements.txt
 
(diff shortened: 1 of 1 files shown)


stsewd (approved):
These are examples, but looks like people are copying/pasting those exact versions

docs/guides/reproducible-builds.rst
```diff
@@ -121,9 +129,9 @@ Some examples:
 
       # File: docs/requirements.txt
 
-      sphinx==3.4.3
-      sphinx_rtd_theme==0.5.1
-      readthedocs-sphinx-search==0.1.0rc3
+      sphinx==4.2.0
+      sphinx_rtd_theme==1.0.1
```
astrojuanlu, line 133:
```suggestion
      sphinx_rtd_theme==1.0.0
```

astrojuanlu:
🤖 Rendered version: https://docs--8655.org.readthedocs.build/en/8655/guides/reproducible-builds.html

humitos:
> These are examples, but looks like people are copying/pasting those exact versions

Yeah 😞  --we may need to add a generic note or something. We can't keep these dependencies updated every time a new version is released.

humitos:
@astrojuanlu also, I understand that the author of the original comment wanted to add the pin for `docutils` in particular, not just updating the version's numbers. This PR does not address that.

astrojuanlu:
@humitos There were two options: pinning docutils, or upgrading Sphinx (that in recent times started pinning docutils). I opted for the latter.

PR #8676, opened 2021-11-11
docs: update faq to mention apt for dependencies

rokroskar:
This PR adds information about using `apt` to install dependencies to the FAQ.

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -37,6 +37,11 @@ which details your dependencies. See our guide on :ref:`guides/reproducible-buil
 You can also set your project documentation to install your project itself
 as a dependency.
 
+Additionally, your build may depend on extensions that require external,
+non-Python libraries to be installed. If you are using a :ref:`configuration
+file <config-file>` you can add libraries to the Ubuntu-based builder with
+``apt``.
+
 If your project or its dependencies rely on C libraries,
 see :ref:`faq:I get import errors on libraries that depend on C modules`.
 
@@ -279,7 +284,7 @@ in the ``.readthedocs.yaml`` config file if it contains::
       - python=3.9
       - scipy
       ...
-      
+
 .. _Anaconda Project: https://anaconda-project.readthedocs.io/en/latest/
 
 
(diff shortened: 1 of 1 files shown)


docs/faq.rst
```diff
@@ -37,6 +37,11 @@ which details your dependencies. See our guide on :ref:`guides/reproducible-buil
 You can also set your project documentation to install your project itself
 as a dependency.
 
+Additionally, your build may depend on extensions that require external,
+non-Python libraries to be installed. If you are using a :ref:`configuration
+file <config-file>` you can add libraries to the Ubuntu-based builder with
+``apt``.
```
agjohnson, line 43: This would be great as a link to the section in config-file for apt install configuration option. It should be addressable using the auto section reference syntax

docs/faq.rst
```diff
@@ -37,6 +37,11 @@ which details your dependencies. See our guide on :ref:`guides/reproducible-buil
 You can also set your project documentation to install your project itself
 as a dependency.
 
+Additionally, your build may depend on extensions that require external,
+non-Python libraries to be installed. If you are using a :ref:`configuration
```
agjohnson, line 41: Dependencies can be executables and various packages as well. Perhaps this would be the most clear describing libraries as system packages or something similar instead?

```suggestion
Additionally, your build may depend on extensions that require additional
system packages to be installed. If you are using a :ref:`configuration
```

Does that sound better?

agjohnson (commented):
Good call, thanks for the addition! This is a great spot to call out to apt installation as another option for missing dependencies.

astrojuanlu (changes requested):
Thanks for this pull request! Can we use this opportunity to reword the sentence that confused you?

> We can't support installing random C binaries on our system, so there is another way to fix these imports.

rokroskar:
> Can we use this opportunity to reword the sentence that confused you?

Ha, indeed I actually started with that sentence. But then I realized that what brought me there was the link from the section above. The part with the controversial sentence is specifically talking about mocking when install of libraries is not possible - and the only way I saw to sneak in a mention of apt was to link back to the section above... which then results in a circular dependency 🤷 But I will try to adapt the wording there to alleviate the confusion a bit

rokroskar:
@astrojuanlu and @agjohnson thanks for your comments - I've tried to address them in 8895b91

astrojuanlu (approved):
Thanks a lot!

PR #8688, opened 2021-11-17
Spam: make admin filters easier to understand

humitos:
Use a `gte` and `lt` applied to the spam filters so we show "_only_ projects
with score between N-M instead of >N"

### DIFF
--- a/readthedocs/projects/admin.py
+++ b/readthedocs/projects/admin.py
@@ -138,7 +138,7 @@ class ProjectSpamThreshold(admin.SimpleListFilter):
         return (
             (
                 self.NOT_ENOUGH_SCORE,
-                _("Not enough score (1-{})").format(
+                _("Not spam (1-{})").format(
                     settings.RTD_SPAM_THRESHOLD_DONT_SHOW_ADS,
                 ),
             ),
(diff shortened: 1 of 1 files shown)


readthedocs/projects/admin.py
```diff
@@ -135,50 +136,79 @@ class ProjectSpamThreshold(admin.SimpleListFilter):
 
     def lookups(self, request, model_admin):
         return (
+            (
+                self.NOT_ENOUGH_SCORE,
+                _("Not enough score (1-{})").format(
```
agjohnson, line 141: "Not enough score" sounds like something other than "not spam". Any reason we don't just want to call these projects "not spam"?

```suggestion
                _("Not spam (1-{})").format(
```

agjohnson (commented):
Makes sense, though it wasn't clear if the first level, "not enough score", was meant to describe legitimate projects or not.

readthedocs/projects/admin.py
ericholscher: Agreed on this change 👍

ericholscher (approved):
This would be useful, I wanted this earlier today 👍

PR #8694, opened 2021-11-17
Make the analytics cookie a session cookie

davidfischer:
This changes the GA analytics cookie to be a session cookie rather than a persistent 30 day cookie. GA still sets two other short term cookies (10 minutes, 1 day) which we can't currently configure as far as I'm aware. Session cookies are deleted when the browser is closed.

I don't think this is *the* long term solution here. The longer term plan is to get rid of GA entirely on RTD (except for projects that set their own GA configuration -- this is user-generated content after all).


astrojuanlu (approved):
🚀🚀🚀

humitos (approved):

ericholscher:
Not sure how this one got missed on merge, but good job @humitos finding it and getting it into the release.

PR #8689, opened 2021-11-17
EmbedAPI: log success requests

humitos:
Log success requests using `django-structlog` and send them New Relic to know
what projects are using EmbedAPI, if they are using sphinx-hoverxref and what
version of it.

This is the minimum configuration required to start using structlog + New Relic.
If we are happy with the results we can migrate the other logs to structlog,
which should be simple, but a little tedious to do:

1. replace `logging` by `structlog`
```python
# from
import logging
log = logging.getLogger(__name__)

# to
import structlog
log = structlog.get_logger(__name__)
```
2. move `key=value` from log messages to keyword attributes

```python
# from
log.info('Domain not allowed. domain=%s url=%s', domain, url)

# to
log.info('Domain not allowed.',  domain=domain, url=url)
```
3. make positional arguments to be keyword attributes (similar to the previous point, but the `%s` is mixed with the message itself)
```python
# from
log.info('Sending team invite for %s to %s', invite.team, invite.email)

# to
log.info('Sending team invite to user.', team=invite.team, email=invite.email)
```

4. migrate to Canonical Log Lines (https://www.structlog.org/en/21.3.0/logging-best-practices.html) by using `log.bind` and context.

In the future, we can also set up django-structlog to log Celery tasks as well (https://django-structlog.readthedocs.io/en/latest/celery.html) which will give us traceability since it logs the `task_id` automatically and other extra features.

### References

* New Relic Log's documentation: https://newrelic.com/blog/how-to-relic/python-structured-logging
* django-structlog: https://django-structlog.readthedocs.io/en/latest/
* structlog: https://www.structlog.org/en/stable/index.html

PR #8705, opened 2021-11-23
structlog: migrate application code to better logging

humitos:
Initial work to migrate our application to structlog for better logging, making
them compatible with New Relic's integration.

There is more work that can be done here to make it better. For example, more
usage of `log.bind` to avoid repetition. Besides, rewriting some log messages and
reducing them to have more Canonical Log
Lines (https://www.structlog.org/en/stable/logging-best-practices.html#canonical-log-lines)
instead.

> **NOTE**: this PR does not really require an extensive review. It's mostly just a search&replace. As far as the tests are passing, it should be fine to merge.

The migration steps followed are described in #8689. Also note that this PR is based on that one.

Screenshot_2021-11-23_19-01-00

### DIFF
--- a/readthedocs/vcs_support/utils.py
+++ b/readthedocs/vcs_support/utils.py
@@ -2,13 +2,13 @@
 
 """Locking utilities."""
 import errno
-import logging
+import structlog
 import os
 import stat
 import time
 
 
-log = logging.getLogger(__name__)
+log = structlog.get_logger(__name__)
 
 
 class LockTimeout(Exception):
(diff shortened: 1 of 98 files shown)


readthedocs/vcs_support/utils.py
```diff
@@ -37,47 +37,41 @@ def __init__(self, project, version, timeout=5, polling_interval=0.1):
...
+                'Lock still locked after some time. Retrying in some seconds',
+                locked_time=timesince,
+                seconds=self.timeout,
             )
         open(self.fpath, 'w').close()
-        log.debug('Lock (%s): Lock acquired', self.name)
+        log.debug('Lock acquired.')
 
     def __exit__(self, exc, value, tb):
+        log.bind(project_slug=self.name)
```
ericholscher, line 65: Does this unbind it, since we're also setting this on `__enter__`? Could definitely use a comment.

ericholscher (approved):
This is a huge change that involved core functionality and massive logging changes, which I was hoping to avoid. I am going to approve and merge it because I assume it's well-tested, but this PR is basically impossible to review as it currently is without a _huge_ amount of time unfortunately. 

I would have loved to at least see the code changes and the mass migration done in different PR's or some other way to make it reviewable.

readthedocs/vcs_support/utils.py
humitos: I explained this in https://github.com/readthedocs/readthedocs-corporate/pull/1336#issuecomment-977167702

PR #8706, opened 2021-11-23
oauth: sync remote repositories fix

humitos:
`.values_list` returns a list of id when accessing the related object. We use
that list of ids to query `Organization` model and get real organization objects instead.

Solves: https://sentry.io/organizations/read-the-docs/issues/2747536252/


stsewd (approved):
Looks like we need a test for this case.

PR #8727, opened 2021-11-29
structlog: migrate application code to better logging

agjohnson:
Initial work to migrate our application to structlog for better logging, making
them compatible with New Relic's integration.

There are more work that can be done here to make it better. For example, more
usage of `log.bind` to avoid repetition. Besides, rewriting some log message and
reduce them to have more Canonical Log
Lines (https://www.structlog.org/en/stable/logging-best-practices.html#canonical-log-lines)
instead.

### DIFF
--- a/readthedocs/analytics/utils.py
+++ b/readthedocs/analytics/utils.py
@@ -4,7 +4,7 @@
 
 import hashlib
 import ipaddress
-import logging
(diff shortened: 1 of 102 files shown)


humitos:
Closing this PR since the work is already merged.

PR #8736, opened 2021-12-01
Logging: tweaks and minor improvements

humitos:
The main purpose of this PR is to reduce the log entries sent to Sentry unless they are errors that we can fix only. We were sending "GitHub webhook creating failed", and the reason was the repository was deleted for example --which is not an error on our side.

Besides, there are some tweaks and improvements to have more structured logs where we can get analytics from.

### DIFF
--- a/readthedocs/oauth/services/github.py
+++ b/readthedocs/oauth/services/github.py
@@ -238,7 +238,12 @@ class GitHubService(Service):
 
         session = self.get_session()
         owner, repo = build_utils.get_github_username_repo(url=project.repo)
-        log.bind(project_slug=project.slug)
+        url = f'https://api.github.com/repos/{owner}/{repo}/hooks'
+        log.bind(
+            url=url,
+            project_slug=project.slug,
+            integration_id=integration.pk,
+        )
 
         rtd_webhook_url = 'https://{domain}{path}'.format(
             domain=settings.PRODUCTION_DOMAIN,
@@ -252,13 +257,7 @@ class GitHubService(Service):
         )
 
         try:
-            resp = session.get(
-                (
-                    'https://api.github.com/repos/{owner}/{repo}/hooks'
-                    .format(owner=owner, repo=repo)
-                ),
-            )
-
+            resp = session.get(url)
             if resp.status_code == 200:
                 recv_data = resp.json()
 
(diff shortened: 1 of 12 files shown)


readthedocs/oauth/services/github.py
```diff
@@ -325,31 +324,20 @@ def setup_webhook(self, project, integration=None):
...
+            log.exception('GitHub webhook creation failed for project.')
+        # TODO: I think this "else" clause here is executed when the `try` did
+        # not raise an exception. So we should probably delete it since it's
+        # calling log.error when there was no error.
         else:
-            log.error(
-                'GitHub webhook creation failed for project.',
-                project_slug=project.slug,
-            )
+            log.error('GitHub webhook creation failed for project.')
```
humitos, line 340: @stsewd does this TODO note sound accurate to you? Can we remove the `else` from here?

humitos: Example on GitLab that has the same logic: https://sentry.io/organizations/read-the-docs/issues/2827543888/?project=148442&query=is%3Aunresolved&statsPeriod=14d

Take a look at the reason: `404 - Not Found`.

readthedocs/projects/tasks.py
```diff
@@ -1326,9 +1326,6 @@ def fileify(version_pk, commit, build, search_ranking, search_ignore):
         return
     project = version.project
 
-    # TODO: remove this log once we find out what's causing OOM
```
ericholscher, line 1329: 😆

readthedocs/proxito/views/serve.py
```diff
@@ -389,9 +389,14 @@ def get(self, request, project):
         )
         path = build_media_storage.join(storage_path, 'robots.txt')
 
+        log.bind(
+            project_slug=project.slug,
+            version_slug=version.slug,
+        )
```
ericholscher, line 395: Does separating the `bind` and `info` give much value here? Are we just doing it to be consistent?

ericholscher (approved):
Looks like a good refactor & standardization.

readthedocs/proxito/views/serve.py
humitos: In this particular case, it does not, since we are calling `log.info` just once. However, if we add another line later, we don't need to repeat `project_slug` and `version_slug`.

seanfoltz (approved):

seanfoltz:
Thanks

PR #8766, opened 2021-12-14
Use ad client stickybox feature on RTD's own docs

davidfischer:
This shows off a new feature of our advertising client on our own docs. It basically works the same as before except the feature is built-in.

### DIFF
--- a/docs/_static/css/custom.css
+++ b/docs/_static/css/custom.css
@@ -5,7 +5,7 @@
 }
 
 
-/* RTD docs ad placement in the lower right */
+/* Center the RTD placement when not wide enough to float */
 #rtd-stickybox-container {
   text-align: center;
 }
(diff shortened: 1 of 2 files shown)


docs/_static/css/custom.css
```diff
@@ -6,42 +6,6 @@
...
-  top: 0.5rem;
-  right: 0.5rem;
-
-  background-color: #fefefe;
-  border: 1px solid #088cdb;
-  border-radius: 50%;
-  color: #088cdb;
-  font-size: 0.75rem;
+#rtd-stickybox-container {
   text-align: center;
```
ericholscher, line 10: Could perhaps update the comment here explaining why this is needed.

ericholscher (approved):

Issue #8775, opened 2021-12-20
ModuleNotFoundError: No module named 'setuptools._distutils'


knc6:
Hi,

We use readthedocs for the [jarvis-tools](https://jarvis-tools.readthedocs.io/) package. Recently, I am seeing an error due to setuptools:

Raw log here: https://readthedocs.org/api/v2/build/15584283.txt

I am wondering if somebody can help me with this.

---

📌 _Edit by @astrojuanlu_ 📌 This issue is triggered by [setuptools 60](https://setuptools.pypa.io/en/latest/history.html#v60-0-0), released on 2021-12-19. Current status:

- ~If your project _doesn't_ have a `pyproject.toml`, most likely you have the "use system packages" version enabled. You can (1) disable it, (2) cap setuptools yourself in your requirements, or (3) wait for us to deploy a fix (January 2022) . See [below](https://github.com/readthedocs/readthedocs.org/issues/8775#issuecomment-998366541) for more context.~ **Update: This has been fixed on RTD as of Dec 21.**
- If your project _does_ have a `pyproject.toml`, most likely pip is using its fallback mode and installing the problematic setuptools version in an isolated environment. In this case, there is nothing Read the Docs can do to fix your problem. You can do it yourself by adding a proper `[build-system]` PEP 518 table to your `pyproject.toml` file. See [below](https://github.com/readthedocs/readthedocs.org/issues/8775#issuecomment-998878270) for more information.
- You can temporarily roll back to the old setuptools behavior by exporting a `SETUPTOOLS_USE_DISTUTILS=stdlib` environment variable for your project. However, it is not guaranteed that this will work forever, and we encourage you to look into the solution above.
- If you were redirected here, it's because you opened a new issue for this exact problem. Please try the solution above first.
- If it didn't work, leave a comment below with a link to your failing build.

bhazelton:
We're seeing the same problem on PR builds for the [pyuvdata](https://pyuvdata.readthedocs.io/en/latest/index.html) package. 

This is the log:
https://readthedocs.org/projects/pyuvdata/builds/15585777/

I've done some digging, I think this is caused by setuptools version 60.0 first being installed and then downgraded to version 58.2.0. See these related issues:
https://github.com/pypa/setuptools/issues/2939
https://github.com/pypa/setuptools/issues/2353

[...]

PR #8774, opened 2021-12-20
Follow up to dev docs split

astrojuanlu:
Better to point the dev docs to `latest`. Continuation of #8751.

PR #8777, opened 2021-12-21
Cap setuptools even if installed packages are ignored

astrojuanlu:
Continuation of #8711, see discussion at #8775.

### DIFF
--- a/readthedocs/rtd_tests/tests/test_doc_building.py
+++ b/readthedocs/rtd_tests/tests/test_doc_building.py
@@ -5,6 +5,7 @@ Things to know:
 * the Command wrappers encapsulate the bytes and expose unicode
 """
 import hashlib
+from itertools import zip_longest
 import json
 import os
 import tempfile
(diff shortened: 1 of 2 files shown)


astrojuanlu:
The checks failure is fixed by #8774.

The fact that no tests failed makes me think that this code path is not being tested at all? I will try to add one.

astrojuanlu:
Tests passing :+1:

readthedocs/rtd_tests/tests/test_doc_building.py
```diff
@@ -1210,8 +1211,9 @@ def assertArgsStartsWith(self, args, call):
         with each element of args.
         """
         args_mock, _ = call
-        for arg, arg_mock in zip(args, args_mock):
+        for arg, arg_mock in zip_longest(args, args_mock):
```
astrojuanlu, line 1214: Otherwise, `zip` was discarding the elements of the longer list

humitos (approved):
Awesome! Thanks Juanlu for taking care of this!

I'd like to deprecate "use system package" eventually, tho. 😄

astrojuanlu:
> I'd like to deprecate "use system package" eventually, tho. 😄

Me too :)

PR #8786, opened 2021-12-23
Update redis version

mehrdad-khojastefar:
update redis version to resolve the conflict between redis and django-redis-cache

### DIFF
--- a/requirements/pip.txt
+++ b/requirements/pip.txt
@@ -31,7 +31,7 @@ pyyaml==5.4.1  # pyup: ignore
 Pygments==2.10.0
 
 # Basic tools
-redis==4.0.2
+redis
 kombu==5.2.2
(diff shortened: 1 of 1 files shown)


requirements/pip.txt
```diff
@@ -31,7 +31,7 @@ pyyaml==5.4.1  # pyup: ignore
 Pygments==2.10.0
 
 # Basic tools
-redis==4.0.2
+redis
```
humitos, line 34: You should pin to a specific version here and ignore it from pyup. Besides, add a small comment explaining why it is pinned to an old version.

humitos (changes requested):
Thanks for your PR. Note that we are pinning all our dependencies. Please, update the PR following my inline comment.

PR #8792, opened 2022-01-04
Spam: deny dashboard on spammy projects

humitos:
Quick/Initial implementation to deny showing a project's dashboard if spam score
is above the threshold.

- the old and non-used `ProjectSpamError` that checked for an invalid
  description was removed together with all its logic
- the mixin `ProjectSpamMixin` was re-purposed to check the spam score and deny
  serving the dashboard
- move "disable banned user to import projects" inside the `ImportWizardView`

Note that currently we are only denying the dashboard for the project's
detail ("Overview" in our UI) view. If we want to deny other pages like
"Downloads", "Search", "Builds", "Versions" and "Admin" we need to adapt other
views as well (e.g. migrate them from function-based view to class-based view to
be able to re-use the mixin). This can be implemented in a future version.
Together with this, we could only deny the dashboard to non-maintainers of the
project itself, allowing them to make some changes in case of a mistake.

### DIFF
--- a/readthedocs/projects/views/base.py
+++ b/readthedocs/projects/views/base.py
@@ -1,15 +1,12 @@
 """Mix-in classes for project views."""
 import structlog
-from datetime import timedelta
 from functools import lru_cache
 
 from django.conf import settings
 from django.http import HttpResponseRedirect
-from django.shortcuts import get_object_or_404
+from django.shortcuts import render, get_object_or_404
 from django.urls import reverse
-from django.utils import timezone
 
-from readthedocs.projects.exceptions import ProjectSpamError
 from readthedocs.projects.models import Project
 
 log = structlog.get_logger(__name__)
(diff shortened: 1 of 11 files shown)


requirements/pip.txt
```diff
@@ -78,12 +78,6 @@ django-crispy-forms==1.13.0
 
 docker==5.0.3
 
-django-textclassifier==1.0
-# django-textclassifier doesn't have pinned versions
-# if there is an update they could break our code
-nltk==3.6.6
-textblob==0.17.1
```
ericholscher, line 85: These are big dependencies, so 👍 on getting rid of them if we aren't using them.

readthedocs/projects/views/base.py
```diff
@@ -86,26 +83,17 @@ def get_form(self, data=None, files=None, **kwargs):
...
-            log.info('Rejecting project POST from shadowbanned user.')
-            return HttpResponseRedirect(self.get_failure_url())
-        try:
-            return super().post(request, *args, **kwargs)
-        except ProjectSpamError:
-            date_maturity = timezone.now() - timedelta(
-                days=settings.USER_MATURITY_DAYS
-            )
-            if request.user.date_joined > date_maturity:
-                request.user.profile.banned = True
```
ericholscher, line 103: So we will only have `banned` users manually now, correct? That seems fine.

ericholscher (approved):

ericholscher:
I would like to see us remove all the dashboard views, but the detail is definitely the big one for now. Let's see how this effects things.

readthedocs/projects/views/base.py
humitos: Yes. I didn't know about this code and I was surprised when I saw it yesterday. I think it wasn't working, tho. We can get back to "auto ban" in the future if we find a reliable way of doing it.

PR #8793, opened 2022-01-04
Cleanup: remove old/outdated code

humitos:
It seems wea re not using these views anymore, so I'm deleting them together
with their templates. Besides, I deleted some unused settings as well.


ericholscher (approved):
I still used the embed views on occasion to test something, but I haven't in a long while. Seems like hoverxref or some other place is a better place for a demo like this.

PR #8801, opened 2022-01-10
Bitbucket: update to match latest API changes

humitos:
* https://developer.atlassian.com/cloud/bitbucket/bitbucket-api-teams-deprecation/
* https://developer.atlassian.com/cloud/bitbucket/bitbucket-api-changes-workspaces/

Note that "Filter repositories" now will show "username" (`humitos`) and
"workspace name" (`Manuel Kaufmann`) which are duplicated. The first one,
`humitos`, will not show any repository when clicked because all repositories
belong to a workspace now (mapped as `RemoteOrganization` for us)

> Along those lines, repositories, projects and snippets will no longer have an owner. They will be located within a workspace

We may want to come back to this later and remove the "personal account filter"
from the "Filter repositories" list _only_ for the Bitbucket case.

Screenshot_2022-01-10_11-29-58

(in the image, `humitos` is my personal user and `Manuel Kaufmann` is _the workspace_ for my own account)

### DIFF
--- a/readthedocs/oauth/services/bitbucket.py
+++ b/readthedocs/oauth/services/bitbucket.py
@@ -82,17 +82,17 @@ class BitbucketService(Service):
         return remote_repositories
 
     def sync_organizations(self):
-        """Sync Bitbucket teams (our RemoteOrganization) and team repositories."""
+        """Sync Bitbucket workspaces(our RemoteOrganization) and workspace repositories."""
         remote_organizations = []
         remote_repositories = []
 
         try:
-            teams = self.paginate(
-                'https://api.bitbucket.org/2.0/teams/?role=member',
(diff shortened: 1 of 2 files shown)


humitos:
The `checks` error is not related to this PR:

```
projects/views/private.py
  Line: 287
    pylint: arguments-differ / Parameters differ from overridden 'post' method (col 4)
```

agjohnson (approved):
Looks great, though I also am not familiar with the team -> workspace change on the Bitbucket side. Seems simple enough though.

ericholscher:
Agreed, lets get this out 👍

ericholscher:
Tested are broken in `master`, so merging.

PR #8803, opened 2022-01-10
Remove explicit username from tutorial

astrojuanlu:
It turns out that lots of people following the tutorial
do not even want to think about the name of the project,
and they copy-paste whatever is in there.
I do not blame them!
But to avoid lots of tutorials with _my_ username,
let us write something generic.

### DIFF
--- a/docs/user/tutorial/index.rst
+++ b/docs/user/tutorial/index.rst
@@ -143,7 +143,7 @@ The next page will ask you to fill some details about your Read the Docs project
 Name
   The name of the project. It has to be unique across all the service,
   so it is better if you prepend your username,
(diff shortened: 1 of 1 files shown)


astrojuanlu:
With the old onboarding, before https://github.com/readthedocs/readthedocs.org/pull/8504, this happened automatically, by the way (that the username was prepended to the project name).

humitos (approved):

astrojuanlu:
I understand from https://github.com/readthedocs/readthedocs.org/pull/8801#issuecomment-1008744092 that the failure is unrelated and that it's failing on `master` already, so I'll go ahead and merge this.

PR #8823, opened 2022-01-19
Use stickybox ad placement on RTD themed projects

davidfischer:
During an initial rollout phase, we're going to use this hovering placement (the one we use on https://docs.readthedocs.io) 10% of the time. In the remaining 90%, there will be no changes.

For opting out of ads, users can become gold members. See https://docs.readthedocs.io/en/stable/advertising/ethical-advertising.html#opting-out


readthedocs/core/static-src/core/js/doc-embed/sponsorship.js
```diff
@@ -78,8 +86,9 @@ function create_ad_placement() {
         return $('<div />')
             .attr("id", "rtd-sidebar")
             .attr("data-ea-publisher", "readthedocs")
-            .attr("data-ea-type", "readthedocs-sidebar")
+            .attr("data-ea-type", ad_type)
             .attr("data-ea-manual", "true")
+            .attr("data-ea-style", style_name)
             .addClass(class_name)
```
humitos, line 92: Probably a dumb question but, does this work correctly when `ad_type` and `style_name` are undefined/none variables? Same question applies for `class_name`.

ericholscher (approved):
This looks like a good slow rollout. Let's aim to ship this next deploy? 👍

humitos:
I'm merging this so it goes out in today's deploy.

PR #8824, opened 2022-01-20
Docs: remove `USE_TESTING_BUILD_IMAGE`

humitos:
We don't really want people using this flag anymore. They should be using
`build.os: ubuntu-20.04` with `apt_packages` instead.


astrojuanlu (approved):
🚀

PR #8828, opened 2022-01-20
Canonical can point as stable, not just latest

simonw:
Clarifying note for the documentation about `<link rel="canonical">`.


astrojuanlu (approved):
Thanks a lot!

PR #8843, opened 2022-01-26
Docs: remove beta warning from config file's `build` key

humitos:
I think we can stop calling it beta since it has been working pretty well for
some months already and we are happy with it.


astrojuanlu (approved):
🚀

PR #8848, opened 2022-01-26
Celery: remove queue priority

humitos:
This feature never worked properly, so I'm removing this code.

### DIFF
--- a/readthedocs/core/utils/__init__.py
+++ b/readthedocs/core/utils/__init__.py
@@ -22,11 +22,6 @@ from readthedocs.doc_builder.exceptions import (
     BuildMaxConcurrencyError,
     DuplicatedBuildError,
 )
-from readthedocs.projects.constants import (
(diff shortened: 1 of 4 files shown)


agjohnson (approved):
Take my +1 with a grain of salt, I'm not sure I can answer if we missed anything here. The changes here look good though!

I'm sad this didn't work as we had hoped, it could have helped us with build concurrency and PR builds. However, build instances are cheap and running above capacity is a far easier solution.

humitos:
I went to check the PRs that @ericholscher opened for this and it seems that I reverted everything. See https://github.com/readthedocs/readthedocs.org/pull/7006/ and https://github.com/readthedocs/readthedocs.org/pull/6921/ so I think we should be good.

ericholscher:
😢  -- unfortunate but a good call to delete it 👍

PR #8856, opened 2022-01-27
Logging: reduce verbosity

humitos:
Make more usage of `log.debug` for things that are already well tested and we
don't usually require taking a look these logs to reduce the verbosity of them.

### DIFF
--- a/readthedocs/oauth/services/github.py
+++ b/readthedocs/oauth/services/github.py
@@ -325,7 +325,7 @@ class GitHubService(Service):
                 recv_data = resp.json()
                 integration.provider_data = recv_data
                 integration.save()
-                log.info('GitHub webhook creation successful for project.')
+                log.debug('GitHub webhook creation successful for project.')
                 return (True, resp)
 
             if resp.status_code in [401, 403, 404]:
@@ -471,7 +471,7 @@ class GitHubService(Service):
             )
             log.bind(http_status_code=resp.status_code)
             if resp.status_code == 201:
-                log.info("GitHub commit status created for project.")
+                log.debug("GitHub commit status created for project.")
                 return True
 
             if resp.status_code in [401, 403, 404]:
--- a/readthedocs/organizations/tasks.py
+++ b/readthedocs/organizations/tasks.py
@@ -15,7 +15,7 @@ def mark_organization_assets_not_cleaned(build_pk):
     try:
         build = Build.objects.get(pk=build_pk)
     except Build.DoesNotExist:
-        log.info("Build does not exist.", build_pk=build_pk)
+        log.debug("Build does not exist.", build_pk=build_pk)
         return
 
     organization = build.project.organizations.first()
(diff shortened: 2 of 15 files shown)


agjohnson (approved):
Looks good! :+1: on reducing some of our log noise

agjohnson:
Oops, I didn't realize the failing tests though.

Looks like we had tests on log lines :smile: 

https://app.circleci.com/pipelines/github/readthedocs/readthedocs.org/3858/workflows/5e45eded-9571-451a-9090-bca7ef2ee422/jobs/8592?invite=true#step-105-478

readthedocs/oauth/services/github.py
```diff
@@ -325,7 +325,7 @@ def setup_webhook(self, project, integration=None):
                 recv_data = resp.json()
                 integration.provider_data = recv_data
                 integration.save()
-                log.info('GitHub webhook creation successful for project.')
```
ericholscher, line 328: I do feel like I'm often trying to debug these things in production, and not having good logs will make that harder. I believe we aren't currently storing debug logs anywhere in prod, so for things that happen rarely and are important, I'm not sure I agree with downgrading them to `debug`.

readthedocs/organizations/tasks.py
```diff
@@ -15,7 +15,7 @@ def mark_organization_assets_not_cleaned(build_pk):
     try:
         build = Build.objects.get(pk=build_pk)
     except Build.DoesNotExist:
-        log.info("Build does not exist.", build_pk=build_pk)
+        log.debug("Build does not exist.", build_pk=build_pk)
```
ericholscher, line 18: This seems like it should be a `warning` not an `debug`? Is there a reason we'd expect this to fail?

ericholscher (approved):
Agreed with most of these downgrades, but I do worry about not seeing these logs anymore and making it harder to debug issues in production in some cases. We should probably figure out a better way to handle debug logs from our application, perhaps shipping them but only keeping them for a day or two?

readthedocs/oauth/services/github.py
humitos: With these ones, in particular, I downgraded the logs for the successful case. However, the ones that failed are still logging in info, I think.

readthedocs/organizations/tasks.py
humitos: Not really. There is no way to trigger a task with a build_pk and that task not finding the Build object. I would remove the try/except clause for this case, actually.

This protection scenario used here, is useful for other db objects, like Project or Version. However, for Build it does not make sense in the way we use it: "they are never deleted"

humitos:
> We should probably figure out a better way to handle debug logs from our application, perhaps shipping them but only keeping them for a day or two?

This could be a good approach. However, I'm a little worried about the _amount_ of debug log that we had. I assume that it will affect production performance if we dump this into a file, for example. I think we experienced this in proxito already when we deployed structlog changes and I made a mistake logging a lot of requests. So, that would definitely be something to test or consider at least.

humitos:
I'm merging this for now since I think the changes are good. We can come back to the discussion about how to increase log verbosity without affecting the performance.

PR #8877, opened 2022-02-02
Fix issue with build task routing and config argument

agjohnson:
Our config doesn't pass through a string, it passes through a dict that
matches:

```{"version": "1.2.3", "full_version": "1.2.3"}```


humitos (approved):

agjohnson:
Test fails are unrelated, force merging this one.

PR #8887, opened 2022-02-03
Docs: reduce visibility of Config File V1

humitos:
People is still arriving at Config File V1 and we want to avoid that as much as
possible. However, we can't yet remove this documentation page since it's still
working and there are people already using it.

We should find a way to migrate these projects off from Config File V1 and
finally remove this documentation page. However, that requires a much bigger
amount of work.

Closes #8857

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -19,7 +19,7 @@ search:
   ranking:
     # Deprecated content
     api/v1.html: -1
-    config-file/v1.html: -1
+    config-file/v1.html: -5
 
     # Useful content, but not something we want most users finding
     changelog.html: -6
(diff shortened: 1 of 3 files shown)


agjohnson:
Discussed in the underlying #8648, we can get more aggressive with this change and I think it would help doc clarity. The main points on that issue are 1) don't refer to our config file as v2 and 2) remove the nesting of the configuration file index, just get straight to our configuration file details.

humitos:
I'm moving this out to next's week deploy.

ericholscher (approved):
This is a good start, agreed on improving the overall structure as the next step (or as part of this PR if you feel like it).

humitos:
I will merge this small change as-is for now. We should come back to do a bigger refactor for our config docs. However, hiding this page will reduce the problem of people finding these docs and thinking they are up to date and they should use them.

PR #8899, opened 2022-02-08
Build: allow to not record commands on sync_repository_task

humitos:


readthedocs/doc_builder/environments.py
```diff
@@ -495,11 +495,13 @@ def __init__(
             build=None,
             config=None,
             environment=None,
+            record=True,
```
agjohnson, line 498: Note, come back to do docstring above

agjohnson (approved):

humitos:
I will merge this now for today's deploy, but _I have_ to come back and write that comment.

PR #8903, opened 2022-02-08
Build: define missing variable

humitos:


humitos:
We should come to this and add tests for `sync_reposity_task` because we found a couple on today's deploy.

agjohnson (approved):

Issue #8901, opened 2022-02-08
Build: verbose error in default errors


agjohnson:
Noticed this during deploy today:

![image](https://user-images.githubusercontent.com/1140183/153047081-b16b7f88-3042-4618-ac86-da5d053b26f1.png)

The error message is *a lot* and probably isn't more helpful than the code formatted section below. We should decide what to do in cases like these, and perhaps the answer is that if we can't give a clear error message, we show a generic error. Alternatively, we should at least be code formatting the error string in these cases, but I'm not sure that is more helpful than just the build output.

Long term, I'd like a helpful, clear message that links to the failed command for more context. This is probably a generic message in a lot of cases. But, for this case for example, we could give the user guidance like "There was an issue while installing your Python dependencies. Make sure you have the correct URLs and permissions".

agjohnson:
Actually, i guess I'd call this one a bit more of a bug after seeing a few examples here.

humitos:
This was solved already and we have other issues to track the long-term idea.

PR #8904, opened 2022-02-08
Reduce verbose logging on generic command failure

agjohnson:
We updated the copy used for generic errors with the celery refactor,
but also converted a generic command failure to `BuildUserError(msg)`.
This caused the command output to always be used as the error message
shown to the user.

More error classes is probably the direction we want to go here, perhaps
even going as far as to provide more helpful errors for each step that
commonly emits the generic user error class.

Refs #8901

### DIFF
--- a/readthedocs/doc_builder/environments.py
+++ b/readthedocs/doc_builder/environments.py
@@ -444,24 +444,21 @@ class BaseEnvironment:
             self.commands.append(build_cmd)
 
         if build_cmd.failed:
-            msg = 'Command {cmd} failed'.format(cmd=build_cmd.get_command())
-
-            # TODO: improve this error report. This is showing _the full_
-            # stdout to the user exposing it at the top of the Build Detail's
-            # page in red. It would be good to reduce the noise here and just
-            # point the user to take a look at its output from the command's
-            # output itself.
-            if build_cmd.output:
-                msg += ':\n{out}'.format(out=build_cmd.output)
-
             if warn_only:
+                msg = 'Command {cmd} failed'.format(cmd=build_cmd.get_command())
+                if build_cmd.output:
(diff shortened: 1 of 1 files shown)


agjohnson:
Untested locally, working on that now.

readthedocs/doc_builder/environments.py
```diff
@@ -444,24 +444,21 @@ def run_command_class(
...
                     project_slug=self.project.slug if self.project else '',
                     version_slug=self.version.slug if self.version else '',
                 )
             else:
-                raise BuildUserError(msg)
+                # TODO: for now, this still outputs a generic error message
+                # that is the same across all commands. We could improve this
+                # with more granular error messages that vary by the command
+                # being run.
+                raise BuildUserError()
```
humitos, line 461: I remembered that you noticed this in the original PR at https://github.com/readthedocs/readthedocs.org/pull/8815#discussion_r791120625

What about passing this instead to show a better message in this case? Still generic, but pointing to a problem in the project instead of the app
```suggestion
                raise BuildUserError(BuildUserError.GENERIC)
```

agjohnson: I should have noted, this is indeed the error I'm trying to raise. If you think explicit is better here, :+1: . I was keeping this logic centralized inside `on_failure`, letting this logic handle the missing `msg`:

https://github.com/readthedocs/readthedocs.org/blob/26e371b68f9369a7e87a1d2885d6d357a018e7fc/readthedocs/projects/tasks/builds.py#L380-L384

humitos: Oh, yeah! You are right 👍

I did a quick review while deploying, but your code is perfect 😎

PR #8917, opened 2022-02-14
Celery: use `on_retry` to handle `BuildMaxConcurrencyError`

humitos:
Instead of handling the retry of a task manually when the project hits max
concurrency error, we rely on Celey handlers to do it for us.

By raising the any of the exceptions defined in `autoretry_for`, Celery will
call `on_retry` automatically. There, inside `on_retry` we handle the particular
case for `BuildMaxConcurrencyError` by setting the error into the build object
and skipping the test to continue running.

See
[docs.celeryproject.org/en/master/userguide/tasks.html#automatic-retry-for-known-exceptions](https://docs.celeryproject.org/en/master/userguide/tasks.html#automatic-retry-for-known-exceptions)
and [docs.celeryproject.org/en/master/userguide/tasks.html#Task.autoretry_for](https://docs.celeryproject.org/en/master/userguide/tasks.html#Task.autoretry_for)

Continuation of https://github.com/readthedocs/readthedocs.org/pull/8905


readthedocs/core/utils/__init__.py
```diff
@@ -186,8 +186,8 @@ def prepare_build(
                 project_slug=project.slug,
                 version_slug=version.slug,
             )
-            options['countdown'] = 5 * 60
```
ericholscher, line 189: How does this code actually cancel the build? it seems it's just setting options on the task, and an `error`, but still triggering the build? 

Also seems like these aren't required, since it should take this from the task?

readthedocs/projects/tasks/builds.py
```diff
@@ -482,7 +471,16 @@ def on_success(self, retval, task_id, args, kwargs):
         self.data.build['success'] = True
 
     def on_retry(self, exc, task_id, args, kwargs, einfo):
```
ericholscher, line 473: This task could use a docstring. When in the build flow is this triggered?

ericholscher (commented):
This seems reasonable, but I don't really follow what is happening here as it's currently written. Definitely needs some comments or docstrings.

readthedocs/core/utils/__init__.py
humitos: This PR does not change the flow of this chunk of code. I just move these numbers into Django settings.

That said, this code is run in the webs when the build is triggered and it immediately sets a 5 minutes delay because we already know this project is concurrency limited at the moment. If we delete this code, builders will grab each of these triggered tasks and retry them immediately. So, we are just short-circuiting the process here.

Note that it _does not cancel_ the build, it just adds a countdown to be delayed 5 minutes. The error message is to communicate this delay to the user.

readthedocs/projects/tasks/builds.py
humitos: I added a small docstring explaining _when_ this celery handler is called mentioning both flows: raising a known exception to auto retry; and calling `self.retry`

readthedocs/projects/tasks/builds.py
```diff
@@ -263,25 +265,12 @@ def _check_concurrency_limit(self):
...
+                BuildMaxConcurrencyError.message.format(
                     limit=max_concurrent_builds,
-                ),
-                'builder': socket.gethostname(),
-            })
-            self.retry(
-                exc=BuildMaxConcurrencyError,
-                # We want to retry this build more times
-                max_retries=25,
+                )
```
agjohnson, line 273: Ooooh, I see what you were talking about now with `autoretry_for`. I wasn't familiar with this attribute when we were last talking about this. I like this syntax, it makes more sense :+1:

agjohnson (approved):

readthedocs/core/utils/__init__.py
ericholscher: Ah, interesting. That makes the comment above make a lot more sense. 👍

readthedocs/core/utils/__init__.py
```diff
@@ -186,8 +186,8 @@ def prepare_build(
...
-            options['max_retries'] = 25
+            options['countdown'] = settings.RTD_BUILDS_RETRY_DELAY
```
ericholscher, line 189:
```suggestion
            # Delay the start of the build for the build retry delay.
            # We're still triggering the task, but it won't run immediately,
            # and the user will be alerted in the UI from the Error below.
            options['countdown'] = settings.RTD_BUILDS_RETRY_DELAY
```

readthedocs/projects/tasks/builds.py
```diff
@@ -263,25 +265,12 @@ def _check_concurrency_limit(self):
...
-                project_slug=self.data.project.slug,
-                version_slug=self.data.version.slug,
-            )
-
-            # This is done automatically on the environment context, but
-            # we are executing this code before creating one
-            api_v2.build(self.data.build['id']).patch({
-                'error': BuildMaxConcurrencyError.message.format(
+            # By raising this exception and using ``autoretry_for``, Celery
+            # will handle this automatically calling ``on_retry``
```
ericholscher, line 269: This is a great comment!

readthedocs/settings/base.py
```diff
@@ -125,6 +125,8 @@ def SESSION_COOKIE_SAMESITE(self):
     RTD_STABLE_VERBOSE_NAME = 'stable'
     RTD_CLEAN_AFTER_BUILD = False
     RTD_MAX_CONCURRENT_BUILDS = 4
+    RTD_BUILDS_MAX_RETRIES = 25
```
ericholscher, line 128: Is this a reasonable default? This probably isn't the place to change it, but it feels like a lot.

ericholscher (approved):

readthedocs/settings/base.py
humitos: I thought about this as well. I think you are right that 25 as default for _all retry cases_ is not perfect and we should have a lower number for other cases. However, right now, the only situation where we perform a retry is the concurrency limit reached, as far as I can tell.

But we should expand this to be:

- `RTD_BUILDS_MAX_RETRIES=5`
- `RTD_BUILDS_CONCURRENT_LIMIT_MAX_RETRIES=25`

or similar in case we need the default value for other cases.

PR #8919, opened 2022-02-14
Sentry: ignore logging known exceptions

### LINKED ISSUE #8902: Logging: filter Sentry user build errors
Partial regression of #4590

During release of the task refactoring work, we are noticing a few exception classes filling up Sentry:

* BuildUserError
* RepositoryError

Just noting to come back to this to filter those out.

humitos:
By reading
https://docs.sentry.io/clients/python/integrations/#additional-settings it seems
that we should be able to ignore known exceptions to be sent to Sentry. Note
that we have this setting already defined at
https://github.com/readthedocs/readthedocs.org/blob/b5908b0fc65cb02fc840b23eeb4218b3115b6590/readthedocs/settings/base.py#L453

However, we need to add these known exceptions to
https://github.com/readthedocs/readthedocs.org/blob/b5908b0fc65cb02fc840b23eeb4218b3115b6590/readthedocs/projects/tasks/builds.py#L224-L229

This commit adds the known exceptions to be ignored by Celery and not sent to
Sentry either.

Closes  #8902


readthedocs/projects/tasks/builds.py
```diff
@@ -220,12 +220,21 @@ class UpdateDocsTask(SyncRepositoryMixin, Task):
     max_retries = 5  # 5 per normal builds, 25 per concurrency limited
     default_retry_delay = 7 * 60
 
-    # Expected exceptions that will be logged as info only and not retried
+    # Expected exceptions that will be logged as info only and not retried.
+    # These exceptions are not sent to Sentry either because we are using
+    # ``SENTRY_CELERY_IGNORE_EXPECTED=True``.
+    #
+    # All exceptions generated by a user miss-configuration should be listed
+    # here. Actually, every subclass of ``BuildUserError``.
```
agjohnson, line 228: Is this future proof too? Like, any subclasses of BuildUserError are already included here with the superclass in the `throws` list?

agjohnson (approved):

readthedocs/projects/tasks/builds.py
humitos: This is 100% clear to me. Sentry docs does not specify if indicating the superclass is enough to avoid sending their subclasses as an error as well. So, I'm specifying all of them for now.

humitos: Also, not sure if there is a good way to test it. However, we want _all_ `BuildUserError` to not be reported to Sentry because we know it's not our fault. The "problem" is that we need to specify _all_ the classes here, which may be a little tedious, but that's fine, I guess.

PR #8931, opened 2022-02-17
Tune build config migration

agjohnson:
Use the hacked up script that we ran in place of the migration. We hit
issues with a simple `.update()` backing up queries in production, so
ended up running this script instead. Mostly just archiving this logic
in case we ever want to come back to this.


agjohnson:
Oops, looks like I broke every single test :muscle: 

This should already be applied everywhere, except for our local envs, so might not need to be on next release project.

PR #8943, opened 2022-02-17
Build: reset build error before start building

humitos:
We need to clean up the build error previously shown to the user once the build
is retried for any reason. For example, this avoids the case that the build was
concurrency limited and shows "Concurrency limit reached (2), retrying in 5
minutes."

Then, in the next retry, even after the build is finished, the error is not
removed and we ended up with a successful build showing an error in the build's
detail page. Example: https://readthedocs.org/projects/test-builds/builds/16118772/


agjohnson (approved):
Good catch. I noticed this while the projects were building but didn't realize that the error persists once the build finished too.

PR #8948, opened 2022-02-19
Adding sphinx project

tanvimoharir:
Updating fixtures as per #8914


tanvimoharir:
> @tanvimoharir this looks good and since tests passed, I think that we shouldn't have any issues with test cases -- I'm going to try to include your work in #9319

Thanks @benjaoming :) Do let me know if there is anything which I can do (any changes required etc)

PR #8949, opened 2022-02-21
Test: remove verbose warnings

humitos:
Most of them are warnings we don't have control because they are from third
party applications.


agjohnson (approved):
Yes, thanks for squashing this! My fingers are tired from scrolling through test error output

PR #8950, opened 2022-02-21
Requirements: upgrade gitpython because of security issue

humitos:
- we are using py3.8 so we don't need to ignore gitpython anymore
- upgrade to latest version because it fixes a security issue reported at
  https://github.com/gitpython-developers/GitPython/issues/1409


stsewd (approved):
The issue was a ReDoS, we are executing all git commands inside a celery task and don't think we are using the affected code (directly at least).

PR #8951, opened 2022-02-21
Celery: use AppUserError exception to handle user errors properly

humitos:
- raise `AppUserError` on known exceptions that we want to communicate to users
- do not log these exceptions on Sentry
- re-use a base `ReadTheDocsBaseException` for all our own exceptions
- add note to refactor the usage of `__call__` to use `on_failure` instead


humitos:
Note that we should refactor `SyncServiceError` and use a similar approach as well.

PR #8959, opened 2022-02-22
Apply reverted reverts to rel

agjohnson:
It is 100% unclear why these commits were applied to master but are not applied
here.


agjohnson:
Noooope

PR #8960, opened 2022-02-22
Fix reverted reverts?

agjohnson:
Okay, so, no clue why this happened, but as far as I can tell git disregarded
these commits on `rel`.

`git diff rel master` was showing a diff, even though these commits were merged
into `master`, and `master` was merged into `rel`. My absolute guess is that git
got utterly confused because there was a revert hotfix, a revert PR merged in
master, and a revert-revert PR merged in master.

So, this PR is the cherry-picked reverted revert commits. It illustrates what I
did to `rel` and `relcorp` -- these commits are already in `rel`/`relcorp`.


agjohnson:
@humitos you should look these over to verify this is expected. I just made things match `master`

PR #8972, opened 2022-02-25
add note on setting locale_dirs

OriolAbril:
locale_dirs defaults to locales but the guide uses locale instead. If only the steps in the guide are followed, sphinx can't find the translations. Let me know if this looks good or you'd rather a different solution


ericholscher (approved):
Thanks, this looks like a useful addition 👍

PR #8974, opened 2022-02-28
Celery: increase timeout limit for `sync_remote_repositories` task

humitos:
Users belonging to huge GH organizations, like `conda-forge`, were never syncing
their repositories because the task timed out and none of the relationships were
created.

### DIFF
--- a/readthedocs/oauth/tasks.py
+++ b/readthedocs/oauth/tasks.py
@@ -26,7 +26,15 @@ log = structlog.get_logger(__name__)
 
 
 @PublicTask.permission_check(user_id_matches)
-@app.task(queue='web', base=PublicTask)
+@app.task(
+    queue='web',
+    base=PublicTask,
+    # We have experienced timeout problems on users having a lot of
+    # repositories to sync. This is usually due to users belonging to big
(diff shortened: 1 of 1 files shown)


humitos:
We are 100% sure we have a problem with `conda-forge`, so we could hardcode this name and skip it from the re-sync as next step if we want. Nobody is importing `conda-forge` repositories into Read the Docs and I don't think it makes sense at all.

Also, showing this repositories in the UI is currently distracting to users. This will be solved in the ext-theme because we will be allowing filtering, tho.

agjohnson:
> we could hardcode this name and skip it from the re-sync as next step if we want

This sounds like a great idea. If we felt it was necessary to show the `conda-forge` organization in the import page, we might just want to skip tracking repositories in the conda forge organization (still keeping the conda-forge organization relationship to the github connected user). We could at least show this organization in a disabled and skipped state.

agjohnson (approved):
This is a good fix as a temporary solution :+1: We should follow up to this and skip tracking conda-forge repos at all though -- agreed it doesn't make sense to show these.

PR #8981, opened 2022-03-01
Celery: trigger `archive_builds` frequently with a lower limit

humitos:
Using `LIMIT 5000` and also `LIMIT 2000` has been causing our database to use
too much CPU. This commit changes this limit to reduce it to 500.

Quickly checking CloudWatch logs, I see that we have around 10k builds per day.
Triggering the task every 15 minutes with a limit of 500 gives us 48k builds
archived per day. If my math is correct, I think we are fine.


stsewd (approved):
I think there is something else happening with our DB, since the problem started after the task was stable for like a week, but it makes sense to include this.

PR #8984, opened 2022-03-01
Build: cancel error message

humitos:
When canceling a build immediately after it's triggered the message was `None`
because we were raising a `RepositoryError` exception and Celery wasn't handling
it as it was a build cancelled.

By catching this exception on the VCS commands and re-raising `BuildCancelled`
again, `on_failure` method from Celery will handle it properly.

### DIFF
--- a/readthedocs/vcs_support/base.py
+++ b/readthedocs/vcs_support/base.py
@@ -3,7 +3,7 @@ import structlog
 import os
 import shutil
 
-from readthedocs.doc_builder.exceptions import BuildUserError
+from readthedocs.doc_builder.exceptions import BuildUserError, BuildCancelled
 from readthedocs.projects.exceptions import RepositoryError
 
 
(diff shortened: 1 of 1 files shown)


readthedocs/vcs_support/base.py
```diff
@@ -101,6 +101,11 @@ def run(self, *cmd, **kwargs):
 
         try:
             build_cmd = self.environment.run(*cmd, **kwargs)
+        except BuildCancelled as e:
+            # Catch ``BuildCancelled`` here and re raise it. Otherwise, if we
+            # raise a ``RepositoryError`` then the ``on_failure`` method from
+            # Celery won't treat this problem as a ``BuildCancelled`` issue.
+            raise BuildCancelled
```
agjohnson, line 108: Not sure if there is any context in `e`, but perhaps this should be `raise` or `raise(BuildCancelled(e))`? This is currently tossing out the context. If the exception context isn't required, `as e` above can be dropped

agjohnson (approved):
Looks good to hotfix, my note doesn't need to block release

PR #8987, opened 2022-03-02
Requirements: remove `django-permissions-policy`

humitos:
`interest-cohort` is not a thing anymore. See
https://github.com/readthedocs/readthedocs.org/pull/8954#discussion_r816962542


stsewd (approved):
There is also a nginx file that needs to be reverted https://github.com/readthedocs/readthedocs.org/pull/8145/files

Issue #8995, opened 2022-03-03
Build: incompatible Mercurial versions from VCS setup than build commands


ahal:
## Details

* Read the Docs project URL: https://readthedocs.org/projects/taskcluster-taskgraph/
* Build URL (if applicable): https://readthedocs.org/projects/taskcluster-taskgraph/builds/16259866/
* Read the Docs username (if applicable): ahal

## Expected Result

The builds should succeed :)

## Actual Result

Builds fail as seen [here](https://readthedocs.org/projects/taskcluster-taskgraph/builds/16259866/). I don't believe anything changed on the repo side, this was [working a week or two ago](https://readthedocs.org/projects/taskcluster-taskgraph/builds/).

```
hg identify --id
abort: repository requires features unknown to this Mercurial: sparserevlog!
(see https://mercurial-scm.org/wiki/MissingRequirement for more information) 
```

marmoute:
I only skimmed through the logs, but this is weird. It looks like the `hg id` call is suddently too old to be able to read the repository that just got cloned by that same hg.

So, might this is no longer the same hg and some of the virtualenv play in between downgraded it somehow ?

ahal:
I don't have any insight into reathedocs infrastructure, though a co-worker pointed out a [somewhat suspicious commit](https://www.mercurial-scm.org/repo/hg/rev/6d2ddea0721a2a70dc399bd94829ee36bf7597af) to Mercurial which would have been released a couple weeks before this started failing. But yeah, it's weird that the earlier `hg` commands aren't failing.

marmoute:
The "suspicious commit" (and the related work in this area) only affects the "streaming clone" feature. Streaming clone does not seems to be involved in this build at all. So I would be surprised if these changes play any part in the current failures.

mattip:
We are seeing this on the [PyPy doc builds](https://readthedocs.org/projects/pypy/builds/) as well. Is there a way to find all the projects that are using mercurial to find one that is succeeding?

mattip:
According to #8840 there are ~46 repos using mercurial. Are all their builds failing?

humitos:
@ahal can you try using a Read the Docs config file with build.os (https://docs.readthedocs.io/en/stable/config-file/v2.html#build-os) declared as Ubuntu 20.04?

humitos:
[...]

PR #8999, opened 2022-03-07
Build: `RepositoryError` message

humitos:
When a `RepositoryError` _without_ a message was raised we were setting the
message as `"None"` and reporting this to the user :(

This commit first checks if the exception already contains a message and use
it in that case. If the exception does not contains a message, it raises it
without a message so the generic one is added by the Celery
handler (`on_failure`).

Also, it improves the messages reported to the user to communicate a more
detailed error depending the case.

Related to the error reported in https://github.com/readthedocs/readthedocs.org/issues/8995

### DIFF
--- a/readthedocs/projects/exceptions.py
+++ b/readthedocs/projects/exceptions.py
@@ -52,7 +52,7 @@ class RepositoryError(BuildUserError):
     )
 
     @property
-    def CLONE_ERROR(self):
+    def CLONE_ERROR(self):  # noqa: N802
         if settings.ALLOW_PRIVATE_REPOS:
             return self.PRIVATE_ALLOWED
         return self.PRIVATE_NOT_ALLOWED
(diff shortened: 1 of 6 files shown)


readthedocs/projects/exceptions.py
```diff
@@ -47,7 +46,16 @@ class RepositoryError(BuildUserError):
 
     FAILED_TO_CHECKOUT = _('Failed to checkout revision: {}')
 
-    def get_default_message(self):
+    GENERIC_ERROR = _(
+        "There was a problem with your repository. "
+        "Please, take a look at the commands' output to find out the reason.",
```
agjohnson, line 51:
```suggestion
        "There was a problem cloning your repository. "
        "Please check the command output for more information.",
```

agjohnson:
There is a failing lint test that I'll try fixing for this PR to get it merged.

agjohnson (approved):

PR #9009, opened 2022-03-14
Build: allow users to use Ubuntu 22.04 LTS

humitos:
Ubuntu 22.04 LTS is not yet released, but it's already available to use. We can
start allowing people to use this image for their builds.

To make usage of this image in production, we need to:

- `docker build` and push to Docker Hub: https://github.com/readthedocs/readthedocs-docker-images/pull/178
- update the hash for this image on https://github.com/readthedocs/readthedocs-ops/pull/1144
- run `./scripts/compile_version_upload_s3.sh` for the new image with all the `build.tools` supported


agjohnson (approved):
Looks good :+1:

PR #9010, opened 2022-03-14
Update project to use Ubuntu 22.04 LTS

humitos:
Ubuntu 22.04 LTS is not yet released, but it's already available to use. We can
start allowing testing this image internally before upgrading production
servers.

Also, note that the default Python version was upgraded from 3.8 to 3.10.

Related to https://github.com/readthedocs/readthedocs.org/pull/9009


humitos:
This PR has to be merged after April 21, when Ubuntu 22.04 is released and it will be available on AWS as well.

humitos:
Hrm, it seems that tests are not running because it can install django-allauth since we are using an old version. We should upgrade it (see #9122) before upgrading Ubuntu/Python. There are other benefits mentioned in the issue as well.

humitos:
- ✅ All tests passed!
- ✅ Production AMIs were built successfully!
- ✅ Local Docker image was built correctly!

@ericholscher @agjohnson @stsewd we should be ready to merge this PR, test it for a week and use the new Ubuntu 22.04 LTS base on next deploy.

.circleci/config.yml
```diff
@@ -6,7 +6,7 @@ orbs:
 jobs:
   tests:
     docker:
-      - image: 'cimg/python:3.8'
+      - image: 'cimg/python:3.10'
```
ericholscher, line 9: Exciting. We'll finally be on a mostly up to date version of Python :)

ericholscher (approved):
This looks simple enough. I'm less worried about it locally as running in production, but overall we don't depend on system-level packages for much beyond Python.

PR #9015, opened 2022-03-15
Build: log when a build is reset

humitos:
If the build worker is killed for any reason, as we are using `ACKS_LATE`, the
build task will be picked by another worker and re-executed.

When that happens, we detect that the `Build` object already has some commands
saved on it and we delete those commands to start again from scratch.

This commit just adds a log line to be aware how often this situation happens in
production and be able to make other decisions.


ericholscher (approved):
This makes sense, and will make it easier to understand how often it happens.

PR #9029, opened 2022-03-18
Add bumpver configuration

agjohnson:
This is directly called from ops/deploy, so usage of this tool is
abstracted and not super important.

It's easy to use however, and has a string input method, instead of
incrementing each part individually, a la bump2version.

```
% bumpver update --dry --set-version "8.0.0"
INFO    - fetching tags from remote (to turn off use: -n / --no-fetch)
INFO    - Old Version: 7.4.2
INFO    - New Version: 8.0.0
--- docs/conf.py
+++ docs/conf.py
@@ -67,7 +67,7 @@
 copyright = '2010-{}, Read the Docs, Inc & contributors'.format(
     timezone.now().year
 )
-version = "7.4.2"
+version = "8.0.0"
 release = version
 exclude_patterns = ['_build']
 default_role = 'obj'
--- package.json
+++ package.json
@@ -1,6 +1,6 @@
 {
   "name": "readthedocs",
-  "version": "7.4.2",
+  "version": "8.0.0",
   "description": "Read the Docs build dependencies",
   "author": "Read the Docs, Inc <support@readthedocs.com>",
   "scripts": {
--- readthedocs/__init__.py
+++ readthedocs/__init__.py
@@ -2,5 +2,5 @@

 import os.path

-__version__ = "7.4.2"
+__version__ = "8.0.0"

--- setup.cfg
+++ setup.cfg
@@ -29,7 +29,7 @@
 github_repo = readthedocs.org

 [bumpver]
-current_version = "7.4.2"
+current_version = "8.0.0"
 version_pattern = "MAJOR.MINOR.PATCH[TAGNUM]"
 commit_message = "Bump version {old_version} -> {new_version}"
 commit = False
```

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -13,7 +13,6 @@ use f'{docset}/setting' as value on the setting, for example::
 
 import os
 import sys
-from configparser import RawConfigParser
 
 import sphinx_rtd_theme
 
@@ -45,13 +44,6 @@ for k, v in docsets[docset].items():
     locals()[k] = v
 
 
-def get_version():
-    """Return package version from setup.cfg."""
-    config = RawConfigParser()
-    config.read(os.path.join('..', 'setup.cfg'))
-    return config.get('metadata', 'version')
-
-
 sys.path.append(os.path.abspath('_ext'))
 extensions = [
     'sphinx.ext.autosectionlabel',
@@ -75,7 +67,7 @@ master_doc = 'index'
 copyright = '2010-{}, Read the Docs, Inc & contributors'.format(
     timezone.now().year
 )
-version = get_version()
+version = "7.4.2"
 release = version
 exclude_patterns = ['_build']
 default_role = 'obj'
--- a/package.json
+++ b/package.json
@@ -1,8 +1,8 @@
 {
   "name": "readthedocs",
-  "version": "7.4.1",
+  "version": "7.4.2",
   "description": "Read the Docs build dependencies",
-  "author": "Anthony Johnson <anthony@readthedocs.com>",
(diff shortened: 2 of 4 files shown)


docs/conf.py
```diff
@@ -13,19 +13,22 @@
...
 sys.path.append(os.path.dirname(__file__))
 os.environ.setdefault("DJANGO_SETTINGS_MODULE", "readthedocs.settings.dev")
 
-from django.utils import timezone
-
-import django
-django.setup()
+try:
+    # Nesting in try to avoid isort inflicting unwanted changes. It tries to
+    # move these imports above the very necessary sys.path manipulation.
```
ericholscher, line 25: Looks like we can use pramga comments for this as well: https://pycqa.github.io/isort/docs/configuration/action_comments.html

docs/conf.py
```diff
@@ -71,11 +67,9 @@ def get_version():
 
 templates_path = ['_templates']
 
-master_doc = 'index'
-copyright = '2010-{}, Read the Docs, Inc & contributors'.format(
-    timezone.now().year
-)
-version = get_version()
+master_doc = "index"
+copyright = "2010-{}, Read the Docs, Inc & contributors".format(timezone.now().year)
```
ericholscher, line 71:
```suggestion
copyright = "2010, Read the Docs, Inc & contributors"
```
I think we can just do this for even less dynamic magic :)

ericholscher (approved):
This seems like a cleaner approach. I'm 👍 assuming we integrate it in our deploy scripts.

docs/conf.py
agjohnson: Ah nice, `isort: split` was what I was looking for

PR #9048, opened 2022-03-29
Proxito: do not serve non-existent versions

humitos:
Avoid serving files that are still in the storage (S3) for some reason but its
associated version was deleted or de-activated.

Closes https://github.com/readthedocs/readthedocs-ops/issues/1139

### DIFF
--- a/readthedocs/proxito/views/serve.py
+++ b/readthedocs/proxito/views/serve.py
@@ -84,13 +84,19 @@ class ServeDocsBase(CachedView, ServeRedirectMixin, ServeDocsMixin, View):
             version_slug=version_slug,
             filename=filename,
         )
-
-        # All public versions can be cached.
         version = final_project.versions.filter(slug=version_slug).first()
+
+        # Skip serving versions that don't exist or are not active. This is to
+        # avoid serving files that we have in the storage, but its associated
+        # version does not exist anymore or it was de-activated
+        if not version or not version.active:
+            raise Http404("Version does not exist or is not active.")
+
         if (
             self._is_cache_enabled(final_project)
             and version and not version.is_private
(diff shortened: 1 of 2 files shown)


ericholscher (approved):

readthedocs/proxito/views/serve.py
```diff
@@ -84,23 +84,38 @@ def get(self,
...
+        # avoid serving files that we have in the storage, but its associated
+        # version does not exist anymore or it was de-activated.
+        #
+        # Note that we want to serve the page when `version is None` because it
+        # could be a valid URL, like `/` or `` (empty) that does not have a
+        # version associated to it.
+        #
+        # However, if there is a `version_slug` in the URL but there is no
+        # version on the database we want to return 404.
+        if version and not version.active or (not version and version_slug):
```
ericholscher, line 107:
```suggestion
        # However, if there is a `version_slug` in the URL but there is no
        # version on the database we want to return 404.
        if (version and not version.active) or (version_slug and not version):
```

I think we want this, to make it more explicit?

ericholscher (approved):
I think this looks good, but we definitely want to watch this closely. I feel like there's a non-zero chance of us accidently breaking single version projects or similar.

readthedocs/proxito/views/serve.py
humitos: Yeah, exactly. That's the problem this logic is trying to solve. I also agree it may be lead to some false positives, but I think we could cover them with test cases. If you have ideas of what other cases I'm missing here, I'm happy to write those test cases as well. Unfortunately, I didn't find a better / more explicit way to write this logic. Let me know.

Issue #8961, opened 2022-02-22
Build: cancel a build when a new one for the same branch/PR is triggered


humitos:
Now that we implemented a way to revoke Celery tasks we are able to cancel a running build. This is useful for people to manually cancel a build they are not interested in finishing for different reasons and free up the concurrency limit they have.

There are some scenarios where we can handle this automatically because we already know that the running build is useless and can be canceled:

- a new commit was pushed to a branch
- a build for the same branch was triggered (do not confuse with "Duplicated Build")
- ...

Related #8850

aiwalter:
that would be a really nice feature :)

jbusecke:
This would help us over at https://github.com/xgcm/xgcm quite a bit. +1 👍

astrojuanlu:
I think the SunPy and Astropy folks would love having this too :) @nabobalis @pllim

pllim:
> Astropy folks

I cannot speak for all of Astropy but personally I would love auto-cancel. Otherwise, sometimes when we get commit spam, RTD gives us a long queue.

Thanks for the ping, @astrojuanlu !

Zeitsperre:
Also chiming in that this would be a welcome addition for our PRs at https://github.com/Ouranosinc/xclim

humitos:
I went ahead and opened https://github.com/readthedocs/readthedocs.org/pull/9549 as a starting point. Let's see if we can have it reviewed and merged soon to start testing this new feature.

humitos:
I'm targeting this feature to go out tomorrow. Once deployed, I will enable this feature on the following projects:

- xclim
- sunpy
- astropy
- xgcm
- poliastro

Let me know if there are other people subscribed to this issue that want to try this feature at this beta phase and I will enable it for their projects as well. Hopefully, after some feedback from you 🙏🏼 , we can enable it by default 👍🏼 . Thank you all!

humitos:
This is currently deployed and all the projects listed in my previous comment have the feature enabled already. I quickly QAed this on production and it did work fine. Please, open an issue if you find something that's not working as you expected or contact support if you want to provide feedback to us! 💪🏼

Zeitsperre:
It seems to be doing the job:

![image](https://user-images.githubusercontent.com/10819524/187537282-d83af91b-1cd8-4654-a2a2-d2efcf7280dc.png)

Thanks again for this!

pllim:
[...]

PR #9049, opened 2022-03-29
Build: do not send VCS build status on specific exceptions

### LINKED ISSUE #9040: Duplicate build sends GH PR status, showing failure until the first build finishes
## Expected Result

I expect GH to show "build running" with a link pointing to the first build for that commit

## Actual Result

GH shows a failed build in the status result, pointing to the build that failed with `Duplicated build`. This is replaced with a ✔️ success after the first build finishes and overwrites the status.

humitos:
When the exception is an exception we know it does not require communicating the
VCS service about its state, we avoid sending the build status to the VCS
service.

This is currently used to avoid sending a "failure" (red cross) notification
when the build is marked as duplicated. In the future, it could also be used
when the build is auto-cancelled because a new commit on the same branch was
done (see https://github.com/readthedocs/readthedocs.org/issues/8961)

Closes #9040

### DIFF
--- a/readthedocs/projects/tasks/builds.py
+++ b/readthedocs/projects/tasks/builds.py
@@ -231,6 +231,9 @@ class UpdateDocsTask(SyncRepositoryMixin, Task):
         ProjectBuildsSkippedError,
     )
 
+    # Do not send external build status on failure builds for these exceptions.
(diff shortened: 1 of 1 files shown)


ericholscher (approved):
Looks great :+1:

PR #9055, opened 2022-03-30
Build: use same hack for VCS and build environments

### LINKED ISSUE #8995: Build: incompatible Mercurial versions from VCS setup than build commands

humitos:
`sync_versions` method executes some commands (e.g. `hg tags`) inside the Docker
container used for all the VCS commands. As it was being executed _outside_ the
`with` statement the container was being killed after executing `__exit__`.

This commit uses the same technique we used as a workaround for build
environment as well. However, this is not the most elegant solution and we
should find a way to make it better.

So far, it works and we can use it as a quick hotfix to make builds continue
working.

Closes #8995

### DIFF
--- a/readthedocs/doc_builder/director.py
+++ b/readthedocs/doc_builder/director.py
@@ -65,7 +65,30 @@ class BuildDirector:
                 ),
             )
 
-        environment = self.data.environment_class(
+        before_vcs.send(
+            sender=self.data.version,
+            environment=self.vcs_environment,
+        )
+
+        # Create the VCS repository where all the commands are going to be
+        # executed for a particular VCS type
+        self.vcs_repository = self.data.project.vcs_repo(
+            version=self.data.version.slug,
+            environment=self.vcs_environment,
+            verbose_name=self.data.version.verbose_name,
+            version_type=self.data.version.type,
+        )
+
+        self.pre_checkout()
+        self.checkout()
+        self.post_checkout()
+
+        commit = self.data.build_commit or self.vcs_repository.commit
+        if commit:
+            self.data.build["commit"] = commit
+
+    def create_vcs_environment(self):
+        self.vcs_environment = self.data.environment_class(
             project=self.data.project,
             version=self.data.version,
             build=self.data.build,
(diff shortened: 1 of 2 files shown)


ericholscher (approved):
Seems reasonable for a hotfix

readthedocs/doc_builder/director.py
```diff
@@ -65,7 +65,30 @@ def setup_vcs(self):
                 ),
             )
```
agjohnson, line 66: These two conditionals will have moved to executing in the context manager. It doesn't seem important that either are executed outside the environment in the first place. But in case anyone has any more information about why this is, these seem to be the only logic order changes.

agjohnson (approved):
I'm not caught up on the refactor we just did to this code, so I'm not confident my review here.

This seems safe enough, I understand the underlying issue and mostly understand this logic change I think, and I don't believe my note on the two conditionals is important.

So, take this as a +0.5, but only because this is code I'm not familiar with yet.

readthedocs/doc_builder/director.py
humitos: This code was being already executed inside the context manager. The difference now is that instead of the context manager being executed inside the `BuildDirector` is executed in the Celery task, so we can call an extra method there (`sync_versions`), which needs to be executed inside the context manager (Docker environment)

humitos:
> This commit uses the same technique we used as a workaround for build
> environment as well (see #9018). However, this is not the most elegant solution and we
> should find a way to make it better.

Another cleaner approach here would be to avoid using context manager (`with` statements) and use Celery handlers instead. This is, run the `__enter__` and `__exit__` code from the Celery task. Mainly, this is "creating the containers before they are needed", "removing the containers after they are used" and "performing a cleanup of the Docker containers in `after_return` in case there was an exception during the build process".

This will give us a better understanding of the flow without the complexity/magic/hiddenness that the context managers are currently providing.

humitos:
I'm merging this PR since it's already hotfixed and it's working. We can open an issue to track the refactor, tho.

PR #9056, opened 2022-03-30
Docs: `build.jobs` reference documentation

humitos:
Small reference documentation for `build.jobs` and all the pre/post jobs allowed.

Related #9016

### DIFF
--- a/docs/user/config-file/v2.rst
+++ b/docs/user/config-file/v2.rst
@@ -418,6 +418,42 @@ We don't currently support PPA's or other custom repositories.
    When possible avoid installing Python packages using apt (``python3-numpy`` for example),
    :ref:`use pip or Conda instead <guides/reproducible-builds:pinning dependencies>`.
 
+
+build.jobs
+``````````
+
+Commands to be run before or after a Read the Docs pre-defined job.
+This allows you to run custom commands at a particular moment in the build process.
+
+
+.. code-block:: yaml
+
+   version: 2
+
+   build:
+     os: ubuntu-22.04
+     tools:
+       python: "3.10"
+     jobs:
+       pre_create_environment:
+         - echo "Command run at 'pre_create_environment' step"
+       post_build:
+         - echo "Command run at 'post_build' step"
+         - echo `date`
+
+.. note::
+
+   Each key under ``build.jobs`` must be a list of strings.
+
+
+:Type: ``dict``
+:Allowed keys: ``post_checkout``, ``pre_system_dependencies``, ``post_system_dependencies``,
+   ``pre_create_environment``, ``post_create_environment``, ``pre_install``, ``post_install``,
+   ``pre_build``, ``post_build``
+:Required: ``false``
+:Default: ``{}``
+
+
 sphinx
 ~~~~~~
 


docs/user/config-file/v2.rst
```diff
@@ -418,6 +418,42 @@ We don't currently support PPA's or other custom repositories.
    When possible avoid installing Python packages using apt (``python3-numpy`` for example),
    :ref:`use pip or Conda instead <guides/reproducible-builds:pinning dependencies>`.
 
+
+build.jobs
+``````````
+
+Bash commands to be run before or after a Read the Docs pre-defined job.
```
agjohnson, line 425:
```suggestion
Shell commands to be run before or after a Read the Docs pre-defined job.
```

agjohnson: Or I suppose even just "Commands", "shell" is probably redundant.

humitos: Cool, I thought we were calling `/bin/bash -c` but I just saw we are doing `/bin/sh -c`.

https://github.com/readthedocs/readthedocs.org/blob/052f891dbab6b21a5bbdf4570fb149a0adef1581/readthedocs/doc_builder/environments.py#L364

docs/user/config-file/v2.rst
```diff
@@ -418,6 +418,42 @@ We don't currently support PPA's or other custom repositories.
    When possible avoid installing Python packages using apt (``python3-numpy`` for example),
    :ref:`use pip or Conda instead <guides/reproducible-builds:pinning dependencies>`.
 
+
+build.jobs
+``````````
+
+Commands to be run before or after a Read the Docs pre-defined job.
+This allows you to run custom commands at a particular moment in the build process.
```
ericholscher, line 426:
```suggestion
Commands to be run before or after a Read the Docs pre-defined build step.
This allows you to run custom commands at various times in the build process.

A few common use cases for this area:

* Modifying the state of the checkout of your repository
* Stopping a build by returning a non-zero exit code on certain conditions 
* Running a command to process generate markup files prior to building
```

Thinking more about this, we probably need to define this feature a lot more explicitly in the docs somewhere, particularly around exit codes.

docs/user/config-file/v2.rst
```diff
@@ -418,6 +418,42 @@ We don't currently support PPA's or other custom repositories.
...
+     jobs:
+       pre_create_environment:
+         - echo "Command run at 'pre_create_environment' step"
+       post_build:
+         - echo "Command run at 'post_build' step"
+         - echo `date`
+
+.. note::
+
+   Each key under ``build.jobs`` must be a list of strings.
```
ericholscher, line 446: I don't really know what this means. Can we communicate it more explicitly somehow? What else would users try to do here, that might be wrong?

ericholscher (approved):

docs/user/config-file/v2.rst
humitos: I'm trying to avoid users writing something like

```yaml
build:
  jobs:
    post_install: pip install mypackage
```

when it should be 

```yaml
build:
  jobs:
    post_install: 
      - pip install mypackage
```

docs/user/config-file/v2.rst
humitos: I didn't mention use cases here because none of the other config keys do that. Instead, I suggested writing a guide explaining what are the known use cases we have already and expanding on each of them clearly.

I see this page as reference documentation that only explains what are the possible/valid config keys and values of each of them, but it does not talk about use cases. I'd include the link to that guide page here once we write it, tho, as we do with `build.tools.python` and the Conda Support link (https://docs.readthedocs.io/en/stable/config-file/v2.html#build-tools-python)

What do you think?

docs/user/config-file/v2.rst
```diff
@@ -418,6 +418,42 @@ We don't currently support PPA's or other custom repositories.
...
+Commands to be run before or after a Read the Docs pre-defined job.
+This allows you to run custom commands at a particular moment in the build process.
+
+
+.. code-block:: yaml
+
+   version: 2
+
+   build:
+     os: ubuntu-22.04
```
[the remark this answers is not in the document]
humitos: This one, in particular, is not _required_ to be a string. However, we can declare it as it. We should change the other one as well where I copied it from https://docs.readthedocs.io/en/stable/config-file/v2.html#build

humitos:
I'm merging this so it goes out tomorrow together with the deploy and the `build.jobs` feature. We can come back to this comment later https://github.com/readthedocs/readthedocs.org/pull/9056#discussion_r840036949 and also continue the discussion about how the documentation should be written in https://github.com/readthedocs/readthedocs.org/issues/9067

PR #9066, opened 2022-03-31
Add gVisor runtime option for build containers

### LINKED ISSUE #9065: Build: add gvisor project feature flag for testing runsc runtime
We'll have some production operations changes to make to support this, but for now I'd like to start off as a feature flag that will only work in local development for now. I could also make this an application setting I suppose, but we probably want the feature flag for private beta testing this in production anyways.

This adds additional protection to arbitrary command execution that will be enabled with build.jobs and build.commands build configuration options.

agjohnson:
Also quick dev doc on setup.

Code change is in commit 1def3f85b
Linting is in commit f616b7f9f

Closes #9065

### DIFF
--- a/docs/dev/guides/gvisor.rst
+++ b/docs/dev/guides/gvisor.rst
@@ -21,7 +21,8 @@ daemon to avoid systemd cgroups.
             % sudo runsc install
 
         You do need to instruct Docker to avoid systemd cgroups. You will need
-        to make further changes to ``/etc/docker/daemon.json``:
+        to make further changes to ``/etc/docker/daemon.json`` and restart the
+        Docker service:
 
         .. code:: json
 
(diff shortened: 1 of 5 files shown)


docs/dev/guides/gvisor.rst
```diff
@@ -0,0 +1,52 @@
...
+                "runtimes": {
+                    "runsc": {
+                        "path": "/usr/bin/runsc"
+                    }
+                },
+                "exec-opts": ["native.cgroupdriver=cgroupfs"]
+            }
+
+Docker is correctly configured when you can run this command from the quick
+start guide:
```
humitos, line 38: Just note that you have to restart docker at this point. Otherwise, the runtime is unknown: `sudo systemctl restart docker`

humitos (approved):
Looks good to me and it worked fine! 🎉

docs/dev/guides/gvisor.rst
agjohnson: It's noted in the guide, but feels worth duplicating here

PR #9069, opened 2022-04-04
Github: use probot auto-assign bot

humitos:
This bot will auto-assign any PR created to it's own author. It will help us to
manage our new Roadmap using GitHub Beta Projects in a better way.

> We can easily disable if we find it does not work as we expected.


agjohnson (approved):
Seems like a good step. I also have not been assigning my PRs into the roadmap, as I find the duplicate issue and PR is noisy and confusing. I'm fine with just issues in the roadmap view, except for one-off cases where an issue isn't worth tracking.

PR #9083, opened 2022-04-06
Build: bugfix `RepositoryError.CLONE_ERROR` message

humitos:
This commit converts the `@property` used to select the correct message
depending on the platform (.org or .com) by a `@classmethod` to make it simpler
to realize that we need to call it as a regular method.

In Python 3.9, we will be able to combine these two decorators and we won't need
parenthesis anymore. For now, we need to add parenthesis to make the call of the
method.


ericholscher (approved):
Silly issue :( Good job tracking this down.

PR #9098, opened 2022-04-12
Build: do not upload `build.tool` to production S3

humitos:
Avoid uploading the .tar.gz to production S3 since we will be managing this via
CircleCI and an AWS orbs. See
https://github.com/readthedocs/readthedocs-docker-images/pull/179

We need to keep the upload part for development environment, tho.

### DIFF
--- a/scripts/compile_version_upload_s3.sh
+++ b/scripts/compile_version_upload_s3.sh
@@ -103,18 +103,21 @@ docker cp $CONTAINER_ID:/home/docs/$OS-$TOOL-$VERSION.tar.gz .
 # Kill the container
 docker container kill $CONTAINER_ID
 
-# Upload the .tar.gz to S3
-AWS_ENDPOINT_URL="${AWS_ENDPOINT_URL:-http://localhost:9000}"
-AWS_BUILD_TOOLS_BUCKET="${AWS_BUILD_TOOLS_BUCKET:-build-tools}"
-AWS_ACCESS_KEY_ID="${AWS_ACCESS_KEY_ID:-admin}"
-AWS_SECRET_ACCESS_KEY="${AWS_SECRET_ACCESS_KEY:-password}"
-
-if [[ -z $AWS_REGION ]]
+if [[ -z $CIRCLECI ]]
 then
-    # Development environment
+    # Upload the .tar.gz to S3 development environment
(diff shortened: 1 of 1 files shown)


scripts/compile_version_upload_s3.sh
```diff
@@ -81,17 +45,18 @@
 #
 # EXAMPLES
 #
-#  ./scripts/compile_version_upload.sh python 3.9.7
-#  ./scripts/compile_version_upload.sh nodejs 14.17.6
+#  ./scripts/compile_version_upload.sh python-3.9.7
+#  ./scripts/compile_version_upload.sh nodejs-14.17.6
```
humitos, line 49: Note to myself, this may not be required and we can do the same hack as in DOCKER_IMAGE :)

scripts/compile_version_upload_s3.sh
```diff
@@ -138,22 +116,21 @@ docker cp $CONTAINER_ID:/home/docs/$OS-$TOOL-$VERSION.tar.gz .
...
 docker container kill $CONTAINER_ID
 
-# Upload the .tar.gz to S3
-AWS_ENDPOINT_URL="${AWS_ENDPOINT_URL:-http://localhost:9000}"
-AWS_BUILD_TOOLS_BUCKET="${AWS_BUILD_TOOLS_BUCKET:-build-tools}"
-AWS_ACCESS_KEY_ID="${AWS_ACCESS_KEY_ID:-admin}"
-AWS_SECRET_ACCESS_KEY="${AWS_SECRET_ACCESS_KEY:-password}"
-
-if [[ -z $AWS_REGION ]]
+if [[ -z $CIRCLECI ]]
```
agjohnson, line 119: I noticed this on the shell scripting on the ops side of this change as well, and a few spots above, but just to note: when using variables in sh/bash, it's best to quote them:

https://tldp.org/LDP/abs/html/quotingvar.html

We don't expect are any problematic characters in these variables, so this isn't an issue.

I've definitely been bit by unquoted variables in the past, usually around string comparisons.

```suggestion
if [[ -z "$CIRCLECI" ]]
```

agjohnson (approved):
Looks great!

agjohnson:
Failure is docs, merging this.

PR #9104, opened 2022-04-13
External status: use admin users to hit VCS API

humitos:
We were using `AdminPermission.members`. However, that will return
`RemoteRepositoryRelation.admin=False`, so those users may not have permissions
to update the VCS status.

By using `AdminPermission.admins` we are sure the user selected will have
permissions to make this request.

Front logo [Front conversations](https://app.frontapp.com/open/top_409c7)


humitos:
Meh, this is wrong. This is inside a loop, so we pass to the next one if it fails.

PR #9124, opened 2022-04-19
Expierment with pip-tools for our docs.txt requirements

ericholscher:
This will then generate a full set of transitive dependencies,
so that our builds shouldn't break.


humitos (approved):
Sounds good to me! The next step is to do the same for our own app dependencies, maybe? 😄

ericholscher:
Yea, it was pretty darn straightforward here! Definitely something we should be recommending more. There's also a hash mode that adds some more security: https://pip-tools.readthedocs.io/en/latest/#using-hashes

humitos:
I introduced `pip-tools` + dependabot on https://github.com/readthedocs/sphinx-hoverxref/pull/182. We can use it as example to do to the same here and remove `pyup` that does not support `pip-tools` (I asked them via support).

The next feature is to open a PR for the dependencies that were released one week ago at least 😉

Issue #9100, opened 2022-04-12
Create better status for concurrency-limited & duplicated builds


ericholscher:
Currently when a build is concurrency limited it looks like this:

![Screen Shot 2022-04-12 at 9 54 58 AM](https://user-images.githubusercontent.com/25510/163014540-52abc24c-4b6e-4b66-8cc6-0386dae943ff.png)

And on the list screen:

![Screen Shot 2022-04-12 at 9 54 55 AM](https://user-images.githubusercontent.com/25510/163014559-89117672-5071-44b4-8dde-af83a69f58a0.png)

This is not communicating to the user the actual state of things. This build isn't waiting to run, it's delayed because of a concurrency limit. We should be communicating this more clearly to users.

Similarly, duplicated builds are marked as failed:

![Screen Shot 2022-04-12 at 9 55 55 AM](https://user-images.githubusercontent.com/25510/163014721-f225f9b4-e710-4f40-8d52-449647748780.png)

This makes the user think something went wrong, but the build was actually just cancelled. This should be communicated to the user in a nicer way as well. 

## Implementation

We could either add new statuses for these outcomes, or just change the display of the status based on the failure message. I don't feel strongly here, but it seems like an actual status is the most clear.

agjohnson:
So, I actually think "Triggered" isn't too bad. To the user, "Triggered" vs "Queued" is still effectively "Your build will be attempted soon". To me, this doesn't feel like an important designation as we've mostly reduced the time builds sit around before being accepted by a builder. So, "Triggered" feels like it's mostly gone away and "Queued" is the state we're talking about here. And it feels safe to lump both into one state.

I would agree that "Triggered" is not ideal though. I've advocated for "Queued" instead of "Triggered" for a few of the reasons here.

I am treating "Triggered" as "Queued" visually for now in fact, and will soon release something closer to:

![image](https://user-images.githubusercontent.com/1140183/163080136-f4e7e35c-4a06-46ce-8dad-84bbbf93319d.png)

I agree surfacing build metadata would be a UX improvement though, but I think that our previous conversations here were on the right track:

https://github.com/readthedocs/readthedocs.org/issues/3399

The issue that we have with errors now is that our build errors are **template logic** :grimacing:. These really should be some model state so that I can include basic conditional elements in multiple templates.

For example, if we wanted to show some error states on the build listing, I would inspect the build error objects attached to the build:

[...]

PR #9145, opened 2022-04-27
Build: add a new "Cancelled" final state

humitos:
This commit introduces a new build final state: "Cancelled".
This state is shown in the build listing page and also in the build details
page. The build will end up on this state when:

- The user cancels the build by clicking on the button
- The build is cancelled due to inactivity
- The build is cancelled because it was duplicated

Ther are a lot of changes required for this:

 - HTML template logic for build details and listing pages
 - API changes to add a new value
 - Knockout.js logic for build details page
 - `Build` model changes to support a new state
 - Update different places of core logic to change checking for `finished` state
   to `BUILD_FINAL_STATES` instead
 - Documentation changes to mention the new state
 - Celery build handler `after_return` to decide if it's `finished` or
   `cancelled` the ending state
 - Logic to decide if the build should be killed due to inactivity

### Build ending up in a cancelled state

Screenshot_2022-05-04_13-13-15

### Build cancelled by the user immediately

Screenshot_2022-05-04_13-13-42

### Build cancelled shown on listing page

Screenshot_2022-05-04_13-13-57

Reference https://github.com/readthedocs/readthedocs.org/issues/9100


ericholscher:
I think overall this makes sense. I feel like @agjohnson had a bit more context in the discussion thread, so happy to let him take primary reviewer here. Happy to do a technical review if we're 👍 on the direction, if that's useful.

readthedocs/templates/builds/build_detail.html
```diff
@@ -138,9 +138,14 @@
         {% trans "Build completed" %}
       </span>
       <span class="build-state-failed"
-            data-bind="visible: finished() && !success()"
+            data-bind="visible: finished() && !success() && !cancelled()"
```
agjohnson, line 141: This is the type of logic that I'll move to strictly JS and would like out of templates eventually. For now this is fine though.

readthedocs/templates/projects/project_dashboard_base.html
```diff
@@ -77,6 +77,7 @@ <h3>{% trans "Projects" %}</h3>
                               <span class="build-state build-state-passing">{% trans "passing" %}</span>
                             {% else %}
                               <span class="build-state build-state-failing">{% trans "failing" %}</span>
+                              {# TODO: What whould be show here when "not build.sucess and cancelled"? #}
```
agjohnson, line 80: I'm getting rid of this concept in the new templates. There is no singular status that makes sense here in any regard. That is, we are showing the status of the project based on the `get_latest_build` -- a failure on some minor version can make the entire project look failing.

Instead, the new templates don't show this status at all, but show individual version statuses instead. Even in that case, I don't feel we show the version as "cancelled" though.

agjohnson (approved):
This does highlight why I want to centralize this logic. I don't have a great grasp on what could have been missed here, but this looks thorough enough to test.

PR #9153, opened 2022-04-28
Mamba: remove CONDA_USES_MAMBA feature flag

humitos:
Mamba can be used via `build.tools.python: mambaforge` now.

This PR removes the `CONDA_USES_MAMBA` feature flag since we already contact
most of the users and opened a PR on their repositories with the config file
required for this change and we also sent an in-site notification to all those
projects still using it.

The message said this feature flag will be removed on May 1st. So, we are all
good to merge and deploy next week (May 3rd).

### DIFF
--- a/readthedocs/doc_builder/python_environments.py
+++ b/readthedocs/doc_builder/python_environments.py
@@ -447,10 +447,6 @@ class Conda(PythonEnvironment):
         """
         Decide whether use ``mamba`` or ``conda`` to create the environment.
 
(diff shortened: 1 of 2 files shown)


ericholscher (approved):
This looks good. I do wish there was a better way to note this removal, but I guess the CHANGELOG is good enough?

humitos:
Hrm, yeah. I don't think we have a better way than that, unfortunately.

Issue #9039, opened 2022-03-25
Update Reproducible Builds docs with pip-tools


ericholscher:
Currently we have a great doc on specifying requirements, but it doesn't show how to pin requirements. We should recommend an approach like pip-tools to this problem. 

## Input

https://github.com/pallets/jinja/blob/e542e100027aaa131df73ad06b71ba9a4ee348c8/requirements/docs.in#L1-L4

## Output

https://github.com/pallets/jinja/blob/e542e100027aaa131df73ad06b71ba9a4ee348c8/requirements/docs.txt#

PR #9158, opened 2022-05-02
Add docs showing pip-tools usage on dependencies

ericholscher:
This will help users not have random jinja or docutils updates break their builds.

Fixes https://github.com/readthedocs/readthedocs.org/issues/9039


docs/user/guides/reproducible-builds.rst
```diff
@@ -182,3 +182,89 @@ or our Conda docs about :ref:`environment files <guides/conda:creating the \`\`e
...
+
+✅ Good:
+    All your transitive dependencies will stay defined,
+    not allowing a package upgrade to break your docs.
+
+   .. code-block::
+
+      # File: docs/requirements.in
+
+      sphinx==4.2.0
```
agjohnson, line 208: Captions show up a bit nicer, and outside the code example.

```suggestion
   .. code-block::
      :caption: docs/requirements.in

      sphinx==4.2.0
```

docs/user/guides/reproducible-builds.rst
```diff
@@ -182,3 +182,89 @@ or our Conda docs about :ref:`environment files <guides/conda:creating the \`\`e
...
+
+   .. code-block:: yaml
+
+      # File: docs/requirements.txt
+
+      # This file is autogenerated by pip-compile with python 3.7
+      # To update, run:
```
agjohnson, line 215:
```suggestion
   .. code-block:: yaml
      :caption: docs/requirements.txt

      # This file is autogenerated by pip-compile with python 3.7
      # To update, run:
```

docs/user/guides/reproducible-builds.rst
```diff
@@ -182,3 +182,89 @@ or our Conda docs about :ref:`environment files <guides/conda:creating the \`\`e
...
+
+We recommend `pip-tools`_ to help address this problem.
+It allows you to specify a ``requirements.in`` file with your first-level dependencies,
+and it generates a ``requirementes.txt`` file with the full set of transitive dependencies.
+
+.. _pip-tools: https://pip-tools.readthedocs.io/en/latest/
+
+✅ Good:
+    All your transitive dependencies will stay defined,
+    not allowing a package upgrade to break your docs.
```
agjohnson, line 202: A few more clarity updates

```suggestion
    All your transitive dependencies will stay defined,
    which ensures new package releases will not break your docs.
```

docs/user/guides/reproducible-builds.rst
```diff
@@ -160,7 +160,7 @@ Some examples:
       readthedocs-sphinx-search
 
    .. code-block:: yaml
-      
+
       # File: docs/environment.yaml
```
agjohnson, line 164: Also `:caption:` here as well.

agjohnson (commented):
It's probably also worth noting to users that there is specific tooling that does this job much better, like Poetry and friends with lock files. Our suggestion is just the easiest to migrate to.

humitos (commented):
This is a good recommendation, I'm 👍🏼 on merging it.

However, I think the "Pinning transitive dependencies" section that you are adding is very similar to the previous one "Pinning dependencies" that does not pin transitive dependencies. So, I'd say that we should either:

* refer to the previous section communicating that pinning all the versions, including transitive dependencies is better than the previous section but it implies an extra step
* merge the two sections into one and only recommend pinning with pip-tools

Besides, this new section does not have a " Bad" subsection as the other have.

ericholscher:
Yea, there's definitely some additional things to work on here, this is just a basic start.

PR #9166, opened 2022-05-04
Docs: mention `ubuntu-22.04` as a valid option

humitos:
Add `ubuntu-22.04` as a valid option when describing `build.os` in the
documentation reference.


stsewd (commented):
The spec should be updated https://github.com/readthedocs/readthedocs.org/blob/c4376600613fe8658acadf4e369665e2e065a431/readthedocs/rtd_tests/fixtures/spec/v2/schema.json#L87-L89

PR #9169, opened 2022-05-05
Remove our old out-dated architecture diagram

ericholscher:
We should do a blog post perhaps with a new one.


stsewd (approved):

agjohnson (approved):
At least this wasn't horribly incorrect, but I'd agree this would make a better blog post. Perhaps we could cover more of our providers and sponsors in a bit more detail too.

PR #9182, opened 2022-05-10
Development: only pull the images required

humitos:
We are not using `readthedocs/build:latest` anymore for the clone step. We are
now using `readthedocs/build:ubuntu-20.04` for it, so that image is the
required one.


stsewd (approved):

ericholscher:
Seems you generally need both though, if you're building a users docs who specify a build image..

humitos:
@ericholscher yeah, the required one is `ubuntu-20.04` because it's _always_ used. The same image can be used to build the docs as well if the user specifies it. 

However, if the user specify any other image, you need that image as well. You can download them all without using the `--only-required` but that's it like `5 images * 8Gb`.

We can add in the documentation that with `--only-required` you can build projects that are using `ubuntu-20.04` only and if not, you need to download the other images as well.

What image(s) do you think is best to download as _required_ initially?

ericholscher:
I think this is fine for now, just wanted to note it since I hit an issue building docs locally the other day. Not much we can do 🤷

humitos:
The note is valid, tho. Maybe we should add a troubleshooting section in this guide explaining how to solve the problem when "Image is not found" appears in the log when building.

humitos:
I updated the docs. Let me know if that works for you.

ericholscher (approved):

PR #9185, opened 2022-05-10
Note sub-folders with a single domain.

choldgraf:
This is a useful feature of RTD that I didn't know existed. This is a small PR to try and boost the visibility of sub-projects by using a keyword that I think others might search for "sub-folders" and adding a note to the custom domain section.

### DIFF
--- a/docs/user/custom_domains.rst
+++ b/docs/user/custom_domains.rst
@@ -85,6 +85,14 @@ for our own domains including ``*.readthedocs.io``, ``*.readthedocs-hosted.com``
 
 .. _Strict Transport Security header: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Strict-Transport-Security
 
+Multiple documentation sites as sub-folders of a domain
++++++++++++++++++++++++++++++++++++++++++++++++++++++++
+
+You may host multiple documentation repositories as **sub-folders of a single domain**.
+For example, ``docs.foo.org/repo1`` and ``docs.foo.org/repo2``.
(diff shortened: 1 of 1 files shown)


docs/user/custom_domains.rst
```diff
@@ -85,6 +85,14 @@ for our own domains including ``*.readthedocs.io``, ``*.readthedocs-hosted.com``
 
 .. _Strict Transport Security header: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Strict-Transport-Security
 
+Multiple documentation sites as sub-folders of a domain
++++++++++++++++++++++++++++++++++++++++++++++++++++++++
+
+You may host multiple documentation repositories as **sub-folders of a single domain**.
+For example, ``docs.foo.org/repo1`` and ``docs.foo.org/repo2``.
```
ericholscher, line 92:
```suggestion
For example, ``docs.example.org/projects/repo1`` and ``docs.example.org/projects/repo2``.
```

Unfortunately this is still a limitation of this feature, so we don't end up with weird URL overlap. We can and should probably remove this at some point.

ericholscher (approved):

docs/user/custom_domains.rst
choldgraf: ah so it **must** have `projects/` in it?

ericholscher:
Thanks!

docs/user/custom_domains.rst
humitos: Yes. This is a current limitation 😢

choldgraf: ah bummer, that is indeed a big limitation

PR #9186, opened 2022-05-10
Disable codecov annotations

agjohnson:
Not a fan of these, review is really hard with these enabled. The docs
aren't super clear, but I'm going off an example from:

https://docs.codecov.com/docs/github-checks#disabling-github-checks-patch-annotations-via-yaml

These were enabled in #9119

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -9,3 +9,5 @@ coverage:
         informational: true
 codecov:
   require_ci_to_pass: false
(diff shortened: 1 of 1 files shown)


ericholscher (approved):

ericholscher:
Yes, this was annoying me as well.

humitos (approved):

PR #9241, opened 2022-05-18
Docs: jsdoc example using `build.jobs` and `build.tools`

humitos:
Expand our build customization page with an example for jsdoc.

See https://test-builds.readthedocs.io/en/jsdoc/ as example of usage.

### DIFF
--- a/docs/user/build-customization.rst
+++ b/docs/user/build-customization.rst
@@ -196,3 +196,22 @@ It's possible to use ``post_checkout`` user-defined job for this.
          - ./git-lfs fetch
          # Make local files to have the real content on them
          - ./git-lfs checkout
+
+
+Document your code with ``jsdoc``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+It's possible to make usage of ``jsdoc`` to document your Javascript code directly from your source code.
+To setup it, you need to first define the version of ``nodejs`` you want to use and also install ``jsdoc`` as shown below:
+
+.. code:: yaml
+
+   version: 2
+   build:
+     os: "ubuntu-22.04"
+     tools:
+       python: "3.9"
+       nodejs: "16"
+     jobs:
+       post_install:
+         - npm install -g jsdoc
--- a/docs/user/build-customization.rst
+++ b/docs/user/build-customization.rst
@@ -198,11 +198,11 @@ It's possible to use ``post_checkout`` user-defined job for this.
          - ./git-lfs checkout
 
 
-Document your code with ``jsdoc``
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Install Node.js dependencies
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-It's possible to make usage of ``jsdoc`` to document your Javascript code directly from your source code.
-To setup it, you need to first define the version of ``nodejs`` you want to use and also install ``jsdoc`` as shown below:
+It's possible to install Node.js together with the required dependencies by using :term:`user-defined build jobs`.
+To setup it, you need to define the version of Node.js to use and install the dependencies by using ``build.jobs.post_install``:
 
 .. code:: yaml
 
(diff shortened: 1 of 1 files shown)


docs/user/build-customization.rst
```diff
@@ -196,3 +196,22 @@ It's possible to use ``post_checkout`` user-defined job for this.
...
+
+   version: 2
+   build:
+     os: "ubuntu-22.04"
+     tools:
+       python: "3.9"
+       nodejs: "16"
+     jobs:
+       post_install:
+         - npm install -g jsdoc
```
ericholscher, line 217: Seems like this is missing the part where you actually implement it in Sphinx? Could be neat to include to make it more copy-pastable.

ericholscher (approved):

docs/user/build-customization.rst
humitos: Ah, yeah. I wasn't sure if this page is the right place for that. I felt that I was contaminating it with a lot of code that it's outside the scope of `build.jobs` / build customization. Probably, the full example fits better on a blog post or a guide explaining the full process to document Javascript documentation. We could link to that post/guide from here.

docs/user/build-customization.rst
```diff
@@ -196,3 +196,22 @@ It's possible to use ``post_checkout`` user-defined job for this.
          - ./git-lfs fetch
          # Make local files to have the real content on them
          - ./git-lfs checkout
+
+
+Document your code with ``jsdoc``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
agjohnson, line 202: This might be better off generalized, rather than specific to jsdoc. There is a lot more to say about jsdoc usage, and duplicating that here will probably guide users to the wrong guide content.

```suggestion
Install Node.js dependencies
~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

docs/user/build-customization.rst
```diff
@@ -196,3 +196,22 @@ It's possible to use ``post_checkout`` user-defined job for this.
...
+
+It's possible to make usage of ``jsdoc`` to document your Javascript code directly from your source code.
+To setup it, you need to first define the version of ``nodejs`` you want to use and also install ``jsdoc`` as shown below:
```
agjohnson, line 205: If generalizing this, the copy here would need to be tweaked. But generally, noting to use the official terminology:

```suggestion
To setup it, you need to first define the version of Node.js you want to use and also install ``jsdoc`` as shown below:
```

docs/user/build-customization.rst
agjohnson: Also, if we can generalize this example, maybe it's wise to point out `npm ci` as well. This is probably what the user wants ultimately, for reproducible builds, so would be a good suggestion for us to give.

```suggestion
         - npm ci
         - npm install -g jsdoc
```

agjohnson (commented):
Perhaps this is better off just as a generalized example. There is a lot to say about actually using sphinx-js and jsdoc together, especially once you get past a basic example and on to documenting a full JS library.

I've wanted to write up a more complete guide around this. Perhaps this guide should be my doc addition focus one of the next sprints. I found the sphinx-js docs not super helpful, and struggled with a few configuration pieces to make everything work.

docs/user/build-customization.rst
humitos: I like this idea, :100:

humitos:
Thanks for the feedback on this. I will update the documentation to make it more generalized as Anthony suggested. I think that's the right step. After that, we can write a guide / post for a specific use case of this feature together with `jsdoc` 👍🏼

humitos:
I updated the PR to make it more generalized, and I like it more. Let me know if it's ready to merge.

I also opened #9247 to track the writing of a blog post / guide for `jsdoc`.

agjohnson (approved):
:100:

ericholscher:
Will be curious to see if we get any SEO traffic for this page 🤔

PR #9254, opened 2022-05-21
Update german

Justman10000:


agjohnson:
Hi! Thanks for the contribution. We actually manage translations with Transifex, do you have an account there that you could submit these edits with?

Justman10000:
> Hi! Thanks for the contribution. We actually manage translations with Transifex, do you have an account there that you could submit these edits with?

Yes, only this would not do anything as I would have to re-submit the translations and this PR would have wasted my time! For this i bought extra myself [PoEdit](https://poedit.net) for [40€](https://poedit.net/pro).

agjohnson:
These edits would have to be updated at Transifex either way. You should be able to submit your changes by importing your PO file to Transifex however.

We have some notes in our contributing documentation about our localization process, which is all done through Transifex. We do have a number of folks contributing to the German localization, so it's important to work in the same place as other translators.

Justman10000:
> These edits would have to be updated at Transifex either way. You should be able to submit your changes by importing your PO file to Transifex however.
> 
> We have some notes in our contributing documentation about our localization process, which is all done through Transifex. We do have a number of folks contributing to the German localization, so it's important to work in the same place as other translators.

Ah ok... Imported it, or at least I think it didn't overwrite strings that were already approved (actually logical)! Would therefore like to be a proofreader for german...

agjohnson:
Awesome, thanks! I didn't realize you were already on the team too, I've bumped you up to reviewer to get some of those strings reviewed.

I've also pushed up a PR to update all translations and push translation sources to Transifex. Unfortunately our source files were extremely out of date. We are talking about making this part of our weekly release process however.

agjohnson:
Yeah, sorry about that :disappointed:

This is a known issue, and core team does have a plan for increasing consistency here. I'll be continuing to push for translation updates on releases (weekly) so we can avoid out of date source strings and translations.

As far as this PR is concerned, shall we close this? I believe your edits might already be in our translations, as I pulled from Transifex after you uploaded. If they aren't they just need reviewed and should be included in next sync of translations (which will be next week during release!)

Justman10000:
> As far as this PR is concerned, shall we close this? I believe your edits might already be in our translations, as I pulled from Transifex after you uploaded. If they aren't they just need reviewed and should be included in next sync of translations (which will be next week during release!)

Yes it can be closed...

humitos:
BTW, thanks @Justman10000 for all your contributions on translations 👍🏼

PR #9259, opened 2022-05-23
Update translations, pull from Transifex

agjohnson:
This process has happened in a long time, looks like most files haven't been
updated in 3 years. We've talked about automated options on release here, which
is wise.

Besides all of the translations that were updated, this adds the
following translations:

- Turkish `tr`
- Hebrew (100% translated!) `he`
- Lithuanian `lt`
- Hungarian `hu`
- Dutch `nl` and Netherlands Dutch `nl_NL`
- Indonesian `id_ID`
- Korean `ko_KR`

Not sure there is much to review here, this is almost entirely just pulling
files down from Transifex.


stsewd (approved):

humitos:
> We've talked about automated options on release here, which
> is wise.


I think the low-effort and high-benefit here is to add this as part of our release process as suggested in https://github.com/readthedocs/readthedocs.org/issues/4463#issuecomment-1122467742 instead of creating a super-automated-and-complex thing. If we follow that idea, it should be a couple of extra commands that will be run in the `./deploy.sh` script.

agjohnson:
Yup, agreed. It's only a little bit of extra set up for the deploy tasks ultimately.

PR #9277, opened 2022-05-31
Localization updates for 8.1.1

agjohnson:
Lots of removals here as our last push to Transifex changed many of the
previously translated strings. This is mostly expected, and so this
following update will be pulling back in all of the removals.


humitos (approved):

agjohnson:
Test failures are unrelated, merging!

PR #9297, opened 2022-06-06
Flyout: Use raw SVG instead of FontAwesome webfont

mgeier:
WARNING: This is untested!

This is a proof of concept for using inline SVG icons instead of the webfont for the two icons in the flyout.

Something like this could avoid having to load the webfont in the case that the RTD theme is not used, see https://github.com/readthedocs/sphinx_rtd_theme/issues/1298.

### DIFF
--- a/readthedocs/api/v2/templates/restapi/footer.html
+++ b/readthedocs/api/v2/templates/restapi/footer.html
@@ -5,9 +5,13 @@
   {% if not new_theme %}
   <div class="rst-versions rst-badge" data-toggle="rst-versions">
     <span class="rst-current-version" data-toggle="rst-current-version">
-      <span class="fa fa-book">&nbsp;</span>
+      <span style="width: 0.9em; height: 0.9em; fill: #fcfcfc; float: left;">
+        <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 448 512"><!-- Font Awesome Free 5.15.4 by @fontawesome - https://fontawesome.com License - https://fontawesome.com/license/free (Icons: CC BY 4.0, Fonts: SIL OFL 1.1, Code: MIT License) --><path d="M448 360V24c0-13.3-10.7-24-24-24H96C43 0 0 43 0 96v320c0 53 43 96 96 96h328c13.3 0 24-10.7 24-24v-16c0-7.5-3.5-14.3-8.9-18.7-4.2-15.4-4.2-59.3 0-74.7 5.4-4.3 8.9-11.1 8.9-18.6zM128 134c0-3.3 2.7-6 6-6h212c3.3 0 6 2.7 6 6v20c0 3.3-2.7 6-6 6H134c-3.3 0-6-2.7-6-6v-20zm0 64c0-3.3 2.7-6 6-6h212c3.3 0 6 2.7 6 6v20c0 3.3-2.7 6-6 6H134c-3.3 0-6-2.7-6-6v-20zm253.4 250H96c-17.7 0-32-14.3-32-32 0-17.6 14.4-32 32-32h285.4c-1.9 17.1-1.9 46.9 0 64z"></path></svg>
+      </span>
       v: {{ current_version.explicit_name }}
-      <span class="fa fa-caret-down"></span>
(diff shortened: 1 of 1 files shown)


mgeier:
I don't know anything about CSP, I've just used inline CSS to avoid coming up with CSS class names and CSS files where to put the definitions.

It seems that the CSS link (`badge_only.css`) is included at build time and part of the initially served HTML, while the `footer.html` is inserted dynamically via JavaScript, which makes the synchronization between the two a bit hard, doesn't it?

agjohnson:
@mgeier yeah, syncing the two up is awkward. The inject HTML needs to be supported by all versions of the Sphinx theme, historical releases included to a large degree.

> If we implement a copy of the fa CSS style in badge_only.css, we could have class="fa" that works well for both new and old documentation builds.

I'd agree this seems like a great approach. We can remove 99% of font awesome from the theme build, and retain backward compatibility.

mgeier:
> The inject HTML needs to be supported by all versions of the Sphinx theme

I'm not sure if I fully understand, so let me paraphrase:

"Any new version of the injected HTML has to work with any old version of `badge_only.css`."

Is that correct?

If yes, it doesn't help if we add CSS stuff to `badge_only.css`, because it won't be available in older versions.

So we actually *have to* add inline CSS to the injected HTML?

And just to make sure I understand: we are allowed to remove some CSS from (future versions of) `badge_only.css`?

To paraphrase again:

"New versions of `badge_only.css` can be changed (including removals) as long as new versions of the injected HTML keep working (but old versions of the injected HTML are irrelevant)"

This isn't part of this PR (since `badge_only.css` isn't even in this repo), but the main idea is to remove the FontAwesome stuff from there.

agjohnson:
> "Any new version of the injected HTML has to work with any old version of badge_only.css."

Indeed, this sounds correct to me.

> So we actually have to add inline CSS to the injected HTML?

Well, inline CSS is something I always advocate for avoiding. The option that I would opt for is to disregard old releases of the sphinx theme and only put this feature on the latest sphinx theme release, likely 1.2 as we're preparing to do a push for a bugfix only 1.1 release.

That gets around trying to manage a backwards compatible version of this from our HTML injection. The styles ultimately still come from the Sphinx theme anyways.

Additionally, an technical solution that is worth discussing would be subsetting the Font Awesome fonts files/SVG sprite with only the glyph(s) we care about. This would avoid inline styles and inline SVGs, and is minimally invasive because SASS/CSS doesn't change. FA has a tool for subsetting which could be used as well, which we have access to:

https://fontawesome.com/docs/web/dig-deeper/subsetter

agjohnson:
I tested out the font subsetter -- actually I tested this out https://www.npmjs.com/package/fontawesome-subset -- and it seems to work great. I got the ttf/woff files both down to less than 2KB.

This seems like the easiest approach, however the compromise will have to be that this operation is performed manually, and with a newer Node release. Currently, Wyrm has caused us to pin very specific dependencies at Node, Sass, Neat, and Bourbon, because Wyrm has been unmaintained for a long time. We don't change these files often though, so this seems fine. The eventual goal would be to plug this into Webpack instead.

mgeier:
> I tested out the font subsetter [...] and it seems to work great.

Would this mean that the injected HTML can stay the same and only `badge_only.css` has to be changed?

That sounds like it might be simpler than switching to SVG icons.

agjohnson:
Basically, yeah. More specifically, even the CSS stays the same, it's just the bundled web font files that would change.

Currently, FA4 web fonts are only ~75kB, though FA5/6 are far larger if we do upgrade. We'd be able to save the user ~73kB on requests now, so not a huge performance gain.

mgeier:
> We'd be able to save the user ~73kB on requests now, so not a huge performance gain.

Still, it's more than nothing!

humitos:
We are refactoring the flyout completely in our addons implementation: https://github.com/readthedocs/addons/issues/70

The new flyout should not have this issue anymore. We plan to start slowly rolling this out in the next few weeks 🚀 .

I'm closing this PR for now, but feel free to open any issue you may have related to the flyout in the addons repository 🙏🏼

PR #9302, opened 2022-06-07
Improve scoping of two potentially overlapping Triage sections

benjaoming:
* [x] Improve the scope of the two Triaging sections
* [x] Fixes wrong ``:doc:`team` `` referencing Write the Docs via intersphinx
* [x] ~This is a bit meta-level: Since I'm going over dev docs right now, I wonder how exactly this PR lands: It's a PR without an issue, so it's not actually triaged. Some projects prefer that an issue is always opened for discussion, some prefer that contributors have a shortcut by simply opening a PR.~ I believe that this is resolved, we keep things as they are, adding a bit more incentives on the triaging process.
* [x] Do not allow any `:doc:` or `:ref:` to resolve to intersphinx mapping without `:external:` prefix
* [x] Ignore `:external:` role in rstcheck

### DIFF
--- a/docs/dev/contribute.rst
+++ b/docs/dev/contribute.rst
@@ -87,15 +87,22 @@ projects, feel free to suggest the language on Transifex to start the process.
 
 .. _our projects on Transifex: https://www.transifex.com/projects/p/readthedocs/
 
-Triaging tickets
-----------------
+Triaging issues
+---------------
 
-Core team uses the following guidelines for issue triage on all of our projects.
-These guidelines describe the issue lifecycle, and what is required at each step
-of this lifecycle.
+Everyone is encouraged to help improving, refining, verifying and prioritizing
+issues on Github. The Read the Docs core :external+rtd:doc:`team` uses the following
+guidelines for issue triage on all of our projects. These guidelines describe
+the issue lifecycle step-by-step.
 
 .. note:: You will need Triage permission on the project in order to do this.
-          You can ask one of the members of the :doc:`team` to give you access.
+          You can ask one of the members of the :external+rtd:doc:`team` to give you access.
+
+.. tip:: Triaging helps identify problems and solutions and ultimately what
+         issues that are ready to be worked on. The core
+         :external+rtd:doc:`team` maintains a separate :doc:`/roadmap`
+         of prioritized issues - issues will only end up on that Roadmap after
+         they have been triaged.
 
 Initial triage
 ~~~~~~~~~~~~~~
--- a/docs/dev/roadmap.rst
+++ b/docs/dev/roadmap.rst
@@ -12,7 +12,7 @@ Here, you will find several views into our roadmap:
 
 `Backlog <https://github.com/orgs/readthedocs/projects/156/views/4>`_
     Work that we have planned for future sprints. Items with an assigned
-    timeframe have generally been discussed already by the team. Iteams that do
+    timeframe have generally been discussed already by the team. Items that do
     not yet have a timeframe assigned are not yet a priority of the core team.
 
 The focus of the core team will be on roadmap and sprint items. These items are
@@ -20,11 +20,13 @@ promoted from our backlog before each sprint begins.
 
 .. _GitHub Roadmap: https://github.com/orgs/readthedocs/projects/156/views/1
 
-Triaging issues
-~~~~~~~~~~~~~~~
+Triaging issues for the Roadmap
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-Much of this is already covered in our guide on :doc:`contribute`, however to
-summarize the important pieces:
+Issues are triaged before they are worked on, involving a number of steps that
+are covered in :doc:`/contribute`. Everyone can take part in helping to
+triage issues, read more in :ref:`contribute:Triaging issues`. Additionally, issues are
+considered for the Roadmap according to the following process:
 
 * New issues coming in will be triaged, but won't yet be considered part of our
   roadmap.
--- a/docs/.rstcheck.cfg
+++ b/docs/.rstcheck.cfg
@@ -1,6 +1,6 @@
 [rstcheck]
 ignore_directives=automodule,http:get,tabs,tab,prompt,http:patch,http:post,http:put,http:delete
-ignore_roles=djangosetting,setting,featureflags,buildpyversions
+ignore_roles=djangosetting,setting,featureflags,buildpyversions,external+rtd:doc,external,external+rtd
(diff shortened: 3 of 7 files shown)


docs/dev/contribute.rst
```diff
@@ -87,12 +87,15 @@ projects, feel free to suggest the language on Transifex to start the process.
...
+.. _contribute_triage:
 
-Core team uses the following guidelines for issue triage on all of our projects.
-These guidelines describe the issue lifecycle, and what is required at each step
-of this lifecycle.
+Triaging issues
+---------------
+
+Everyone is encouraged to help improving, refining, verifying and prioritizing
+issues on Github. The Read the Docs core :doc:`team` uses the following
```
stsewd, line 96: We prefer to use absolute paths (in case pages are moved)

```suggestion
issues on Github. The Read the Docs core :doc:`/team` uses the following
```

benjaoming: ``:doc:`/team` `` breaks correctly (because there is no such document)

There is actually no document `/team` and Intersphinx was automatically converting ``:doc:`team` `` to https://www.writethedocs.org/team/ -- I copy-pasted without looking at where it was pointing. I'll remove this reference from the original `note::`.

It seems like a good idea to use absolute paths. In case `/dev/teams.rst` has existed (I don't see it in the git repo history though), removing the file would also break the build, which is what we want.

humitos: We should probably use this config https://www.sphinx-doc.org/en/master/usage/extensions/intersphinx.html#confval-intersphinx_disabled_reftypes to avoid this happening again.

docs/dev/contribute.rst
```diff
@@ -87,12 +87,15 @@ projects, feel free to suggest the language on Transifex to start the process.
 
 .. _our projects on Transifex: https://www.transifex.com/projects/p/readthedocs/
 
-Triaging tickets
-----------------
+.. _contribute_triage:
```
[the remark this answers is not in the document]
benjaoming: This is fixed, thanks! I think I understand how to use it -- when changing a section title, we rely on the documentation build to fail - so it's all the same as if explicit section labels were used?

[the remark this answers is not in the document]
benjaoming: I guess the only catch is that we cannot rename a section without also changing all the references?

docs/dev/contribute.rst
```diff
@@ -87,15 +87,24 @@ projects, feel free to suggest the language on Transifex to start the process.
...
 
-Triaging tickets
-----------------
+Triaging issues
+---------------
 
-Core team uses the following guidelines for issue triage on all of our projects.
-These guidelines describe the issue lifecycle, and what is required at each step
-of this lifecycle.
+.. rstcheck: ignore-roles="external+rtd:doc"
```
humitos, line 93: Is this needed here? I saw it in the `.rstcheck.cfg` file as well.

docs/dev/roadmap.rst
```diff
@@ -20,11 +20,15 @@ promoted from our backlog before each sprint begins.
...
-~~~~~~~~~~~~~~~
+.. roadmap_triage:
 
-Much of this is already covered in our guide on :doc:`contribute`, however to
-summarize the important pieces:
+Triaging issues for the Roadmap
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Issues are triaged before they are worked on, involving a number of steps that
+are covered in :doc:`contribute`. Everyone can take part in helping to
```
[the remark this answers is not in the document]
humitos: It seems this is not applied, right?

docs/dev/contribute.rst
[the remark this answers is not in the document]
humitos: Yes, that's kind of a downside, but keeps things more explicit.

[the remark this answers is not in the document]
stsewd: We have a small guide about this at https://docs.readthedocs.io/en/stable/guides/cross-referencing-with-sphinx.html#automatically-label-sections if it's useful :)

docs/dev/contribute.rst
benjaoming: Sorry no, this is an artifact from a previous attempt :)

dockerfiles/Dockerfile
```diff
@@ -31,6 +31,13 @@ RUN apt-get -y install \
         telnet \
         lsb-release
 
+# Gets the MinIO mc client used to add buckets upon initialization
+# If this client should have issues running inside this image, it is also
+# fine to defer it to a separate image.
+# The current minio/mc Docker image could be a lot smaller
+RUN curl -s -q https://dl.min.io/client/mc/release/linux-amd64/archive/mc.RELEASE.2022-06-11T21-10-36Z -o /usr/bin/mc && \
+    chmod +x /usr/bin/mc
```
humitos, line 39: It seems you included this change in this PR by mistake.

benjaoming: Oh yeah, that wasn't supposed to have happened... CTRL-C'ed the git push process too late :)

humitos (approved):
Looks good to me 👍🏼 

One comment. Please, do not force push new commits if there is a review already done in the PR because that forces the reviewers to review all the files again since GitHub interprets as all the content is new. If there is a way to solve this problem involving force push I'm happy to know how, but I didn't find it. If you want to rebase/change history/etc you can do that immediately before merging the PR

benjaoming:
> Please, do not force push new commits if there is a review already done in the PR because

This is done so that - as a PR Author - can ensure that the commit history is nice and clean before a merge. There are several alternatives to the same goal:

1) A maintainer asks the PR Author to rewrite the commit history
2) A maintainer chooses squash/rebase and the branch is rewritten before a merge -- but without any manual choices or interventions
3) Maintainer uses the `!fixup` notation from the beginning -- I think in this case, it will be quite explicit for the maintainer that a squash is expected before a final merge. Github.com doesn't support anything out of the box for this yet, so it's the same process as 1) AFAIK.

Of course, there is also the option of keeping all the commits and merging the old way.

> that forces the reviewers to review all the files again since GitHub interprets as all the content is new.

This is not really true, you can see all comments and what segments of the code they belong to. All comments are also outdated.

Personally, I like to take a pragmatic approach and not see any of these as a mandatory option, as it raises the bar for contribution and makes it too easy for developers to have discussions :smile:

humitos:
> > that forces the reviewers to review all the files again since GitHub interprets as all the content is new.
> 
> This is not really true, you can see all comments and what segments of the code they belong to. All comments in this PR were outdated.

I'm not talking about the comments, but the changes themselves. If you force push, when I go to "File changed" tab, GitHub shows me _all the files_ again, instead of only the last changes since I reviewed this PR last time. So, I lost all the files I marked as "Viewed" and I'm not sure what changes from the last time I review the PR.

PR #9303, opened 2022-06-07
Integrations: handle `ping` event on GitHub

humitos:
It feels _weird_ to tell the users: "If the webhook fails it means that it's
configured correctly". So, I removed that paragraph from the documentation and I
implemented an expected response for when GitHub sends us a `ping` event.


stsewd (approved):

agjohnson (approved):
Great improvement, this is much clearer UX.

PR #9304, opened 2022-06-07
Docs: improve "Features" page

humitos:


ericholscher:
@humitos is this ready for review? What are the next steps here?

humitos:
@ericholscher no, it's not ready for review. I pushed this as a draft because I started working on this page but I wasn't able to finish it yet. It was demanding me a lot of time to focus and do it right, and I swapped to other tasks while waiting for the inspiration 😄

ericholscher:
@humitos I think we can call this done, and spin up new ideas on site-community?

agjohnson:
Also, from what I was mentioning in our roadmap call today, there is some balance here for content. Our documentation could take more of a technical approach to discussing features, and our website should promote features from a marketing approach. That is, documentation is the *how* to use features, our website is the *why* to use features. That would help keep content duplication down, and also give us a clearer focus on our documentation content.

These changes do seem more like marketing content, but I strongly agree we need better documentation of features in our documentation as well. We do have some gaps we need to fill in our documented feature set.

I'll move this off the current sprint for now, as it's been draft for a while.

humitos:
I got distracted with other stuffs and I was not able to come back to this PR. We are working on a better approach now using Diátaxis. I proceed and close this one.

PR #9305, opened 2022-06-07
Build: avoid overwriting a variable

humitos:
Use `job_commands` and `commands` to differentiate them.

This bug was producing the last "job command" to be saved as `build.commands`, unfortunately.


ericholscher (approved):
Probably should call this something totally different to avoid confusion, but this will work for now.

Issue #9318, opened 2022-06-09
COMPOSE_PROJECT_NAME should not be required for development setup


benjaoming:
## Details

If we setup a fresh environment, following Development Install, these steps will not work because `COMPOSE_PROJECT_NAME` isn't specified. Adding `COMPOSE_PROJECT_NAME=community` fixes the issue.

## Expected Result

We can run `inv docker.build` and `inv.docker.up` without setting `COMPOSE_PROJECT_NAME`.

## Actual Result

This is necessary:

`COMPOSE_PROJECT_NAME=community inv docker.build`
`COMPOSE_PROJECT_NAME=community inv docker.up`

## Tested platform

Latest `main`, Ubuntu 22.04 with Docker 20.10.4

benjaoming:
I'm not sure why it doesn't work for me and works for others. I am running the commands from the same directory as the `.env` file is located in. According to documentation, this seems to have been a feature since `1.28`, so then it should be because I am using a Docker Compose that's too new. But I think we are on the same. See: https://docs.docker.com/compose/environment-variables/

Another thing that I wonder about: Aren't .env files supposed to be about setting the environment for the containers, and `COMPOSE_PROJECT_NAME` is actually a variable meant for the host system?

benjaoming:
What happens during `inv docker.up` is this:

```
ERROR: pull access denied for community_server, repository does not exist or may require 'docker login': denied: requested access to the resource is denied
```

...

> I did cross-check with my setup and it seems that I ran inv docker.build which succeeded correctly without building `community_server`

Additionally:

> it's not reading it for some reason and using `dockerfiles_` to prefix the docker image

So this is actually normal Docker behavior. From [Docker docs](https://docs.docker.com/compose/reference/envvars/#compose_project_name)

> If you do not set this, the `COMPOSE_PROJECT_NAME` defaults to the basename of the project directory.

benjaoming:
@humitos I have had this working since Wednesday - I wanted to add the above for clarity of the issue :)

Since docker-compose is invoked from a base directory called `dockerfiles`, that's why I have ended up with server and image names prefixed `dockerfiles_`.

I understand that everyone wants the [Development Install](https://dev.readthedocs.io/en/latest/install.html) to work without specifying `COMPOSE_PROJECT_NAME` and having it default to `community` in one way or another.

benjaoming:
[...]

PR #9319, opened 2022-06-09
Updates and fixes to Development Install guide

benjaoming:
* [x] Adds a bit more guidance and requirements for a development setup.
* [x] ~Work-around for #9318~ Fixed in `common`
* [x] Fixes MinIO instructions - **NB! DEPENDS ON** https://github.com/readthedocs/common/pull/144
* [x] Mentions private/incognito mode several places
* [x] Fix project fixtures - a new setup was building readthedocs `master` branch instead of `main`
* [x] Add test-builds to project fixtures


docs/dev/install.rst
```diff
@@ -4,79 +4,114 @@ Development Installation
...
+Installing Docker
+-----------------
+
+You need `Docker <https://www.docker.com/>`_ to run a development setup of Read the Docs. The easiest place to start is `Docker's installation guide <https://docs.docker.com/install/>`_, however there might be other more specific ways to get Docker for your exact system.
+
+If you run Windows, you can run Docker on `Windows Subsystem for Linux (WSL) <https://docs.microsoft.com/en-us/windows/wsl/tutorials/wsl-containers>`_ or `Docker Desktop for Windows <https://docs.docker.com/desktop/windows/install/>`_.
+
+   .. tip::
+
+      We always support Docker CE for the latest Ubuntu LTS. On Ubuntu 22.04, we are therefore supporting Docker 20.x, the version of the ``docker.io`` package.
```
humitos, line 33: We use `docker-ce` and `containerd.io` packages. In fact, the following ones are removed on purpose:

- docker
- docker-engine
- docker.io
- containerd
- runc

docs/dev/install.rst
```diff
@@ -4,79 +4,114 @@ Development Installation
...
+#. Build the Docker image for the servers:
 
    .. warning::
 
       This command could take a while to finish since it will download several Docker images.
 
    .. prompt:: bash
 
-      inv docker.build
+      COMPOSE_PROJECT_NAME=community inv docker.build
```
humitos, line 68: We should fix the environment variable to not be required on these commands instead of telling users to prepend them every time they need to run `inv`.

I'd guess there is something different in your environment that this is not working. We've seen this problem in the past when people were using a different version of Docker Compose that the one installed in the previous step. Currently, 1.29.2. Are you using this specific version?

docs/dev/install.rst
```diff
@@ -4,79 +4,114 @@ Development Installation
...
-#. login as ``admin`` /  ``admin`` and verify that the project list appears
+#. Login as ``admin`` /  ``admin`` and verify that the project list appears.
 
-#. go to the "Read the Docs" project, click on the "Build version" button to build ``latest``,
-   and wait until it finishes
+#. Go to the "Read the Docs" project, under section "Build a version", click on the "Build version" button selecting "latest",
+   and wait until it finishes (this can take several minutes).
 
-#. click on the "View docs" button to browse the documentation, and verify that it works
+#. Click on the "View docs" button to browse the documentation, and verify that it works.
```
humitos, line 114:
```suggestion
#. Click on the "View docs" button to browse the documentation, and verify that it shows the Read the Docs documentation page.
```

docs/dev/install.rst
benjaoming: Is there a specification of this somewhere that I've missed? You mentioned earlier that the version for Ubuntu 22.04 should be supported, but I was unsure which, so I guessed `docker.io` since it's very easy to install as it's maintained by Ubuntu maintainers?

docs/dev/install.rst
```diff
@@ -4,79 +4,114 @@ Development Installation
...
+
+* Is Unix-like system (Linux, BSD, Mac OSX) which **supports Docker**. Windows systems should have WSL+Docker or Docker Desktop.
+* Has **10 GB or more of free disk space** on the drive where Docker's cache and volumes are stored. If you want to experiment with customizing Docker containers, you'll likely need more.
+* Can spare *2 GB of system memory* for running Read the Docs, this typically means that a development laptop should have **8 GB or more of memory** in total.
 
 .. note::
 
-   We do not recommend to follow this guide to deploy an instance of Read the Docs for production usage.
-   Take into account that this setup is only useful for developing purposes.
+   Take into account that this setup is intended for development purposes. We do not recommend to follow this guide to deploy an instance of Read the Docs for production.
```
stsewd, line 21: We break long lines on semantic breaks https://dev.readthedocs.io/en/latest/docs.html#content

```suggestion
   Take into account that this setup is intended for development purposes.
   We do not recommend to follow this guide to deploy an instance of Read the Docs for production.
```

docs/dev/install.rst
stsewd: Also, the version of docker compose comes from `common/dockerfiles/requirements.txt`

docs/dev/install.rst
```diff
@@ -4,79 +4,114 @@ Development Installation
...
 
-#. start all the containers:
+#. Start all the containers:
 
    .. prompt:: bash
 
-      inv docker.up  --init  # --init is only needed the first time
+      COMPOSE_PROJECT_NAME=community inv docker.up  --init  # --init is only needed the first time
+
+#. Go to http://community.dev.readthedocs.io to access your local instance of Read the Docs. On newer versions of Firefox and Chrome, use a fresh Private/Incognito session.
```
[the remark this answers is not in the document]
stsewd: I think this happens if you have firefox HTTPS-Only mode enabled (I use a private window on FF bc of that), on chromium it works in a normal window for me (I just use chromium for testing)

readthedocs/projects/fixtures/test_data.json
```diff
@@ -15,6 +15,22 @@
...
+            "modified_date": "2010-08-15T14:36:46+00:00",
+            "description": "",
+            "project_url": "",
+            "repo": "https://github.com/readthedocs/test-builds",
+            "users": [
+        1
+        ],
+            "pub_date": "2010-08-14T01:37:58+00:00",
+            "slug": "test-builds",
+            "name": "Test Builds"
```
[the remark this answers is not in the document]
humitos: We should probably want here `default_version="main"` because otherwise, it will fail when building.

docs/dev/install.rst
```diff
@@ -4,79 +4,114 @@ Development Installation
...
+* Can spare ``2 GB`` of **system memory** for running Read the Docs, this typically means that a development laptop should have ``>=8 GB`` of memory in total.
 
 .. note::
 
-   We do not recommend to follow this guide to deploy an instance of Read the Docs for production usage.
-   Take into account that this setup is only useful for developing purposes.
+   Take into account that this setup is intended for development purposes. We do not recommend to follow this guide to deploy an instance of Read the Docs for production.
+
+
+Installing Docker
```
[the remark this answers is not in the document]
ericholscher: 👍 on having less content that other people have already written. I'm generally -1 on adding more content to this page unless it's strongly required, less is definitely better here.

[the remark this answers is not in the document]
benjaoming: Reading the steps below ("Set up your environment"), they are flat instructions that can read like a tutorial.. having "install Docker" there doesn't accomplish that on the same level, makes the list unbalanced, that's why I wanted to move it somewhere else.

We shouldn't underestimate that getting setup with Docker is easily the most time-consuming step here :)

The separate section is meant to make it easier to skip to the actual instructions. Someone that already has Docker can go "fine fine" and skip to the next section, but we also have space to elaborate what we mean by "Docker". To eliminate the proposed "Installing Docker", I would rather move it into requirements (some prefer "prerequisites").

readthedocs/projects/fixtures/test_data.json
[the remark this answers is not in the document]
benjaoming: I tested it and it works - I'm not sure why, but the single `"builds.version"` item makes it default to have `latest` pointing to `main`. This also works for the `readthedocs.org` project.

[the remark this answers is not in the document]
benjaoming: Will test again and revert :hourglass_flowing_sand:

docs/dev/install.rst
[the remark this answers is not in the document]
saadmk11: I faced this issue when I was setting up local environment. I think this happens because of browser HSTS settings (I think Read the Docs uses HSTS on production and it gets set on `*.readthedocs.io` domain). It works after you clear HSTS cache from the browser.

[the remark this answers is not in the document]
humitos: Yeah. I've been there as well. I'd like to avoid mentioning incognito at this point and put a note about using `http://` instead. Then in the troubleshooting section, we can add an entry that explains how to clear the cache browser in case this problem is hit.

readthedocs/projects/fixtures/test_data.json
```diff
@@ -15,6 +15,22 @@
...
+        }
+    },
     {
         "pk": 22,
         "model": "projects.project",
```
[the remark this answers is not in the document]
benjaoming: Alright, let me get to that :+1:

docs/dev/install.rst
benjaoming: This is actually going to become important.

docker-compose on PyPi was stopped in V1 and rewritten in Go for V2.

https://pypi.org/project/docker-compose/
vs.
https://github.com/docker/compose/

> Docker Compose V2 is a major version bump release of Docker Compose. It has been completely rewritten from scratch in Golang (V1 was in Python). The installation instructions for Compose V2 differ from V1. V2 is not a standalone binary anymore, and installation scripts will have to be adjusted. Some commands are different.

docs/dev/install.rst
benjaoming: Thanks for the reference! I see that we are using Docker's distribution of Docker Engine for Ubuntu (docker-ce)

I'm not sure if our production is on Ubuntu 22.04 so I've just written that it's a good idea to match the current production system to not say too much. I also added that django-compose is the 1.x series.

Moved all this to the requirements section.

docs/dev/install.rst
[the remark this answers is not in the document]
benjaoming: I've elaborated the issue now. Let's get to #9310 after this documentation update so it doesn't get stuck? :)

docs/dev/install.rst
benjaoming: Issue with environment variable fixed here: https://github.com/readthedocs/common/pull/145

benjaoming: Once https://github.com/readthedocs/common/pull/145 is merged, we can remove this part of the changeset.

humitos: @benjaoming 

> docker-compose on PyPi was stopped in V1 and rewritten in Go for V2.

This is unfortunate 😞 . I think it will complicate a little more our initial development setup. We should probably delegate this step to the official documentation as well. (e.g. in Arch, this is `pacman -S docker-compose` and it seems that for Ubuntu it comes when installing the Docker Engine server?)

humitos: @stsewd 

> Also, the version of docker compose comes from common/dockerfiles/requirements.txt

It seems we upgraded the version from that file without double checking the `.env` file was working properly. The latest version of docker-compose that read that file properly is `1.27.4`. So, we should downgrade that requirement instead of adding `COMPOSE_PROJECT_NAME` everywhere.

> Docker Compose versions earlier than 1.28, load the .env file from the current working directory, where the command is executed, or from the project directory if this is explicitly set with the --project-directory option

(from https://docs.docker.com/compose/env-file/)

---

@benjaoming 

We need to change the order of the `-f` passed to use `-f docker-compose.override.yml` as the first one. That way, it will load the `.env` file automatically. See

> The project directory is specified by the order of precedence:
>
>     --project-directory flag
>     Folder of the first --file flag
>     Current directory

I've tested this with Docker Compose v1.29.2 _and_ Docker Compose v2 and it keeps working fine with `.env` file in the repository. In the future, when we upgrade to Docker Compose v2 we won't need touch/change anything else.

readthedocs/projects/fixtures/test_data.json
```diff
@@ -1,346 +1,1161 @@
 [
```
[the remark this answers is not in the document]
benjaoming: The JSON file is not _maintained_, it's an automated artifact. I will respond with your own instructions:

https://github.com/readthedocs/readthedocs.org/issues/8914#issuecomment-1046916832

I wouldn't say that _I_ upgraded the whole file, I re-used the work of a contributor who had already done it, then I verified that it works which it did :+1: 

My immediate reaction: Creating a script like this is moving the goal line for the second time: Initially, I wanted to change `master` to `main` on the readthedocs.org fixture, I added `test-builds` while doing that -- you suggested to fix another issue and I did include the work already submitted as a PR by another contributor. With a suggestion to create a utility script, I think we are too far off the radar for what this PR should do.

I think that the work here is good to go:

1) It fixes an immediate bug
2) It's an overall improvement of previous situation
3) It's possible to defer a utility script to a separate issue.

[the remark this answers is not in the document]
humitos: Yeah. I'm not saying the creation of the script like this has to be done in this PR. It's a separate issue 👍

However, I'd like to do that eventually because even if the JSON is generated automatically by Django, we have to update it from a fresh DB. Otherwise, a lot of extra data is added. In the end, it's harder to maintain than just a Python file with exactly what we want.

docs/dev/install.rst
stsewd: If we change the order it won't override the values anymore

> You can supply multiple -f configuration files. When you supply multiple files, Compose combines them into a single configuration. Compose builds the configuration in the order you supply the files. Subsequent files override and add to their predecessors.

https://docs.docker.com/compose/reference/#specifying-multiple-compose-files

stsewd: Adding `--project-directory=.` seems to work for me

docs/dev/install.rst
```diff
@@ -4,79 +4,98 @@ Development Installation
...
+Requirements
+------------
+
+A development setup can be hosted by your laptop, in a VM, on a separate server etc. Any such scenario should work fine, as long as it can satisfy the following:
+
+* Is Unix-like system (Linux, BSD, Mac OSX) which **supports Docker**. Windows systems should have WSL+Docker or Docker Desktop.
+* Has **10 GB or more of free disk space** on the drive where Docker's cache and volumes are stored. If you want to experiment with customizing Docker containers, you'll likely need more.
+* Can spare *2 GB of system memory* for running Read the Docs, this typically means that a development laptop should have **8 GB or more of memory** in total.
+* Your system should *ideally* match the production system which uses the **latest official+stable Docker** distribution for `Ubuntu <https://docs.docker.com/engine/install/ubuntu/>`_ (the ``docker-ce`` package). If you are on Windows or Mac, you may also want to try `Docker Desktop <https://docs.docker.com/desktop/>`_.
+* We use the latest stable `docker-compose 1.x <https://pypi.org/project/docker-compose/>`_ (this can be installed and run through a virtual environment).
```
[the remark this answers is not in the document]
benjaoming: Yeah, it's strictly speaking not a requirement anymore but more a note in case issues pop up. Removing :+1:

docs/dev/install.rst
```diff
@@ -4,79 +4,98 @@ Development Installation
...
+----------------
+
+This manual step is intended to be made redundant.
+
+#. MinIO buckets need to be configured. Visit MinIO: http://127.0.0.1:9000
+
+#. Login as ``admin`` /  ``password``. Here you should see a list of "buckets".
 
-#. go to http://community.dev.readthedocs.io to access your local instance of Read the Docs.
+#. Click "Manage" on each bucket and change the policy from Private to Public.
```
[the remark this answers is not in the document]
benjaoming: This section is now removed

dockerfiles/Dockerfile
```diff
@@ -31,13 +31,6 @@ RUN apt-get -y install \
         telnet \
         lsb-release
 
-# Gets the MinIO mc client used to add buckets upon initialization
-# If this client should have issues running inside this image, it is also
-# fine to defer it to a separate image.
-# The current minio/mc Docker image could be a lot smaller
-RUN curl -s -q https://dl.min.io/client/mc/release/linux-amd64/archive/mc.RELEASE.2022-06-11T21-10-36Z -o /usr/bin/mc && \
-    chmod +x /usr/bin/mc
-
```
stsewd, line 40: Think we need this?

benjaoming: Yes!

humitos (approved):
Looks good to me! 👍🏼

PR #9321, opened 2022-06-09
EmbedAPIv3: make usage of CDN

humitos:
* Move `CDNCacheTagsMixin` one level up in the directories, since it's shared between API v2 and v3
* Do not CDN cache requests that hit external pages because we can purge them properly (we don't have control over the external page)
* Refactor EmbedAPI v2 and v3 to share some code via a new mixin class
* Update tests:
  * Do not run tests on Sphinx 1.8 and 2.0 anymore
  * Skip a bibtex that's failing because the HTML changed on docutils 0.18
  * Fix a glossary test to work on docutils 0.17 and 0.18

### DIFF
--- a/readthedocs/api/mixins.py
+++ b/readthedocs/api/mixins.py
@@ -1,6 +1,12 @@
-import structlog
+import functools
 
+import structlog
+from django.shortcuts import get_object_or_404
+from django.utils.functional import cached_property
+
+from readthedocs.core.unresolver import unresolve
 from readthedocs.core.utils import get_cache_tag
+from readthedocs.projects.models import Project
 
 log = structlog.get_logger(__name__)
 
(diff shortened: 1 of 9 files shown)


readthedocs/api/mixins.py
```diff
@@ -55,3 +61,35 @@ def _get_cache_tags(self):
         if project and self.project_cache_tag:
             tags.append(get_cache_tag(project.slug, self.project_cache_tag))
         return tags
+
+
+class EmbedAPIMixin:
```
ericholscher, line 66: This class should have a docstring, probably explaining why we're caching these various things?

readthedocs/embed/views.py
```diff
@@ -68,29 +63,10 @@ class EmbedAPIBase(CDNCacheTagsMixin, APIView):
...
-
-    @cached_property
-    def unresolved_url(self):
-        url = self.request.GET.get('url')
-        if not url:
-            return None
-        return unresolve(url)
+    @property
+    def external(self):
+        # Always return False because APIv2 does not support external domains
```
ericholscher, line 68: Is this documented somewhere? Seems like a random detail that would be confusing for users.

ericholscher (approved):
This mostly just looks like a refactor and upgrade to Sphinx 5.0. Seems fine to me, but definitely fiddly with docutils & sphinx changes.

readthedocs/embed/views.py
humitos: Hrm, I think it's not explicitly documented.

* v2: https://docs.readthedocs.io/en/stable/api/v2.html#embed
* v3: https://docs.readthedocs.io/en/stable/api/v3.html#embed

We are not mentioned that external domains are allowed in v3 either --we only mention it in the response. I think this was on purpose because we originally didn't want to expose these external domains to people due to concerns about misuse, etc.

But yeah, we should probably re-think this, update the docs and communicate better how we are using external domains and how people can propose extra domains maybe.

PR #9272, opened 2022-05-28
Remote repository: Add user admin action for syncing remote repositories

saadmk11:
This PR implements a User Admin Action to run `sync_remote_repositories()` task for selected users.

Closes https://github.com/readthedocs/readthedocs.org/issues/9269

PR #9325, opened 2022-06-12
Remote repository: Add organization admin action for syncing remote repositories

saadmk11:
This PR implements a Organization Admin Action to run sync_remote_repositories() task for all members of the selected organizations.

Ref: https://github.com/readthedocs/readthedocs.org/pull/9272#pullrequestreview-988932451

### DIFF
--- a/readthedocs/organizations/admin.py
+++ b/readthedocs/organizations/admin.py
@@ -1,0 +1,85 @@
+"""Django admin interface for organization models."""
+
+from django.contrib import admin
+from django.contrib.auth.models import User
+from django.db.models import Q
+from django.urls import reverse
+from django.utils.html import format_html
+from django.utils.safestring import mark_safe
+
+from readthedocs.oauth.tasks import sync_remote_repositories
+from readthedocs.organizations.models import Organization
+
+
+class OrganizationAdmin(admin.ModelAdmin):
+
+    """Admin configuration for Organization."""
+
+    readonly_fields = (
+        "pub_date",
+        "modified_date",
+        "slug",
+    )
+    list_display = (
+        "id",
+        "name",
+        "max_concurrent_builds",
+        "disabled",
+        "artifacts_cleaned",
+        "modified_date",
+        "pub_date",
+    )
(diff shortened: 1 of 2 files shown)


readthedocs/organizations/admin.py
```diff
@@ -0,0 +1,85 @@
...
+from django.db.models import Q
+from django.urls import reverse
+from django.utils.html import format_html
+from django.utils.safestring import mark_safe
+
+from readthedocs.oauth.tasks import sync_remote_repositories
+from readthedocs.organizations.models import Organization
+
+
+class OrganizationAdmin(admin.ModelAdmin):
```
humitos, line 14: We will need to combine this code with our "Read the Docs for Business" repository because we already have a `OrganizationAdmin` class there.

readthedocs/organizations/admin.py
```diff
@@ -0,0 +1,85 @@
...
+        formatted_task_urls = []
+
+        users = (
+            User.objects.filter(
+                Q(teams__organization__in=queryset)
+                | Q(owner_organizations__in=queryset),
+            )
+            .values_list("id", "username")
+            .distinct()
+        )
```
humitos, line 63: This query will need some adjustments to support SSO on "Read the Docs for Business"

humitos (changes requested):
This PR looks good to me and it's a pretty useful action that we have wanted previously!

However, I think we can't merge this PR as is because there is code from "Read the Docs for Business" that it's not yet migrated into "Read the Docs Community". I thought we had already done that when I mentioned you to do this action. My fault!

That said, this PR will need to be merged with that code first. We could either:

1. move the code from this PR into our private repository adding the action with the required changes
1. move the code from `readthedocs-corporate` to here and combine it with this PR's code

2) makes more sense to me since we want to have organizations in Community in the near future.

humitos:
It seems we haven't needed this. We updated our code to automatically re-sync remote repositories on login and it seems that was enough. I will close this PR for now but we can come back if we consider.

PR #9326, opened 2022-06-13
Translations: integrate Transifex into our Docker tasks

humitos:
I tried to integrate these commands into the release/deploy process but it was complicated. The main problem is that we need to install all the Python dependencies since we need to run `django-admin` commands as part of the process. So, instead of dealing with a new Python environment for the release process, we could use the Docker image where all the other Django commands are executed to update the translations.

Once this PR is merged, the process will be:

```
inv docker.translations --pull
inv docker.translations --push
```

I'm going to update the release/deploy script to call these commands _before_ running the other commands.

Requires: https://github.com/readthedocs/common/pull/147
Reference: https://github.com/readthedocs/readthedocs-ops/issues/1195
Reference: https://github.com/readthedocs/readthedocs.org/issues/4463

### DIFF
--- a/.tx/config
+++ b/.tx/config
@@ -1,9 +1,7 @@
-[readthedocs.readthedocs]
+[main]
+host = https://www.transifex.com
+
+[o:readthedocs:p:readthedocs:r:readthedocs]
 file_filter = readthedocs/locale/<lang>/LC_MESSAGES/django.po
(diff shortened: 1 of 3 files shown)


agjohnson (approved):
This seems fine, I don't think we need this as an invoke task any direction we go with this work.

PR #9327, opened 2022-06-13
Translations: migrate `.tx/config` to new client's version format

humitos:
Use `tx migrate` (v1.1.0) to update our `.tx/config` file.

Related #9326


readthedocs/locale/ar/LC_MESSAGES/django.po
```diff
@@ -24,757 +24,843 @@ msgstr ""
...
-#: analytics/models.py:64 builds/models.py:616 projects/models.py:1380
-#: search/models.py:28 sphinx_domains/models.py:33 templates/base.html:255
-#: templates/search/elastic_search.html:54
-msgid "Version"
-msgstr "الإصدار"
+#: bower_components/sphinx-rtd-theme/sphinx_rtd_theme/breadcrumbs.html:31
+#, fuzzy
+#| msgid "No Docs"
+msgid "Docs"
+msgstr "لا ملفات تعليمات"
```
agjohnson, line 31: This doesn't look right. Why do we have bower_components files in our translation files now? :disappointed: 

I'm guessing that the old client was Django aware and the new client is just pulling in anything it can get its hands on.

agjohnson (changes requested):
I think the client needs some more tuning, it's pulling in a lot of extra files into our translation sources.

readthedocs/locale/ar/LC_MESSAGES/django.po
humitos: I think I know why... https://github.com/readthedocs/readthedocs.org/pull/9326/files#diff-e071df9d84c80d98b07a7efe464ea1b4e92ab493b5c52f5d0f37ca69266ebc3fL8

humitos: Nope, that's not the problem. I'm running the command from the root directory, and it has to be run from inside `readthedocs/` folder.

agjohnson: That looks to have indeed fixed the issue :+1:

agjohnson (approved):
I just eyeballed a few files, looks good!

PR #9333, opened 2022-06-14
EmbedAPI: handle special case for Sphinx manual references

humitos:
When using `.. _my-reference:` above a title, Sphinx adds an empty `<span
id="my-reference"></span>` to the resulting HTML.

Then, our backend is queried for `my-reference` fragment and as it finds it
immediately, it returns the corresponding HTML tag. The problem is that tag is
empty, so the tooltip shows no content at all.

This commit handles this special case when the API is called with
`doctool=sphinx` and returns the parent node of this `span` tag which is the
section itself.

See https://github.com/readthedocs/sphinx-hoverxref/issues/199 for a longer
description of the problem.


ericholscher (approved):
This is super fiddly :( I really wish Sphinx had more semantic HTML for these kinds of things, but I imagine there's always going to be weird edge cases like this.

humitos:
Yeah, we knew this when designing the API, that's why we added the `doctool=`. At least with that, we can take different flows since we know a little more about the HTML structure (_in theory_).

It's also hard to deal with the tests, but I'm happy we can immediately detect changes in the resulting HTML and it's easy to test different combinations of sphinx/docutils. Thanks to that, probably, this is the first time we touched this code after releasing it 🎉

PR #9337, opened 2022-06-15
 EmbedAPI: clean images (`src`) properly from inside a tooltip 

humitos:
When returning the content to be injected in a tooltip, besides cleaning the
`a.href` we are also cleaning the `img.src` to make them absolute and render the
images properly.

### Working example

Screenshot_2022-06-15_16-16-39

See https://github.com/readthedocs/sphinx-hoverxref/issues/200

### DIFF
--- a/readthedocs/embed/utils.py
+++ b/readthedocs/embed/utils.py
@@ -18,11 +18,11 @@ def recurse_while_none(element):
 
 def clean_references(obj, url, html_raw_response=False):
     """
-    Rewrite (internal) links to make them absolute.
+    Rewrite (internal) links (href) and images (src) to make them absolute.
 
-    1. external links are not changed
+    1. external links/images are not changed
     2. prepend URL to links that are just fragments (e.g. #section)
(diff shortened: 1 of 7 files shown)


ericholscher (approved):

readthedocs/embed/utils.py
```diff
@@ -15,13 +16,13 @@ def recurse_while_none(element):
     return {element.text: href}
 
 
-def clean_links(obj, url, html_raw_response=False):
+def clean_references(obj, url, html_raw_response=False):
     """
-    Rewrite (internal) links to make them absolute.
+    Rewrite (internal) links (href) and images (src) to make them absolute.
```
ericholscher, line 21: I wonder if there's other external things we want to be rewriting here as well?

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/source is one example, perhaps.

humitos: I created #9344  to track this and improve the API with other cases that may require rewriting the HTML content.

Issue #8802, opened 2022-01-10
Local development: celery/build container don't restart on files changes


humitos:
We are using `watchmedo` to listen to file changes and restart `celery` and `build` containers. However, this has stopped working for some reason. If you change any Python file you will notice that these containers are not restarted and you have to manually execute `inv docker.restart "celery build"` to load the changes, which takes more time to restart than with `watchmedo`.

The `watchmedo` command executed is at https://github.com/readthedocs/common/blob/4d31f030bde728e2a86ccafbaceadca17122ae15/dockerfiles/entrypoints/build.sh#L12-L20 and https://github.com/readthedocs/common/blob/4d31f030bde728e2a86ccafbaceadca17122ae15/dockerfiles/entrypoints/celery.sh#L14-L23

alella:
This is due to a [bug in watchdog](https://github.com/gorakhargosh/watchdog/issues/870) library. A quick fix would be to remove the `--kill-after=5` flag and let it default to 10 seconds. 

Traceback from watchmedo
```
  File "/usr/local/bin/watchmedo", line 8, in <module>
    sys.exit(main())
  File "/usr/local/lib/python3.8/dist-packages/watchdog/watchmedo.py", line 645, in main
    args.func(args)
  File "/usr/local/lib/python3.8/dist-packages/watchdog/watchmedo.py", line 636, in auto_restart
    handler.stop()
  File "/usr/local/lib/python3.8/dist-packages/watchdog/tricks/__init__.py", line 192, in stop
    kill_time = time.time() + self.kill_after
TypeError: unsupported operand type(s) for +: 'float' and 'str'
```

humitos:
@alella thanks a lot for sharing this! I opened a PR at https://github.com/readthedocs/common/pull/106 to fix this issue. I tested it locally and it worked fine.

humitos:
I'm hitting these issues now, and I'm not sure how to exclude `.tox` directory for example, and others

- https://github.com/gorakhargosh/watchdog/issues/335
- https://github.com/gorakhargosh/watchdog/issues/212

We could replace watchmedo by watchman, but that will require some extra effort: https://facebook.github.io/watchman/docs/watchman-make.html

PR #9338, opened 2022-06-15
Local development: use `nodemon` to watch files instead of `watchmedo`

humitos:
We have been dealing with a problem with `watchmedo` that restart the process
when it shouldn't be restarted and it doesn't restart it when it should.

I got tired debugging `watchmedo` and I was suggested to give it a try to
`nodemon` because we are already using it in other projects. I'm not super happy
adding a node dependency to the Dockerfile, but I didn't find a better way to do it.

I did some tests with the configuration proposed and it seemed to work pretty
well. I'm sure we will find some edge cases, but we always can tune it a little
more later.

Requires: https://github.com/readthedocs/common/pull/150
Closes https://github.com/readthedocs/readthedocs.org/issues/8802

### DIFF
--- a/dockerfiles/Dockerfile
+++ b/dockerfiles/Dockerfile
@@ -29,7 +29,8 @@ RUN apt-get -y install \
         sqlite \
         netcat \
         telnet \
-        lsb-release
+        lsb-release \
+        npm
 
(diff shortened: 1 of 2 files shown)


agjohnson (approved):
This looks easy enough, it's what I would have added :+1:

PR #9350, opened 2022-06-17
Security: avoid requests with NULL characters (0x00) on GET

humitos:
Block requests that contain "\x00" characters on their GET attributes. Raise
`SuspiciousOperation` on these cases.

Screenshot_2022-06-17_14-28-15

----

Related to https://github.com/readthedocs/readthedocs.org/pull/9349/
Sentry issues: https://sentry.io/organizations/read-the-docs/issues/2941721921/?project=148442

### DIFF
--- a/readthedocs/core/middleware.py
+++ b/readthedocs/core/middleware.py
@@ -194,6 +194,7 @@ class ReferrerPolicyMiddleware:
 
 
 class NullCharactersMiddleware:
+
     """
     Block all requests that contains NULL characters (0x00) on their GET attributes.
 
@@ -211,6 +212,6 @@ class NullCharactersMiddleware:
         for key, value in request.GET.items():
             if "\x00" in value:
                 raise SuspiciousOperation(
-                    f"There are NULL (0x00) characters in at least one of the parameters ({key}) passed to the requests."
+                    f"There are NULL (0x00) characters in at least one of the parameters ({key}) passed to the request."  # noqa
                 )
         return self.get_response(request)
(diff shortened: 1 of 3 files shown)


readthedocs/core/middleware.py
```diff
@@ -192,3 +191,20 @@ def __call__(self, request):
...
+class NullCharactersMiddleware:
+    """
+    Block all requests that contains NULL characters (0x00) on their GET attributes.
+    """
+
+    def __init__(self, get_response):
+        self.get_response = get_response
+
+    def __call__(self, request):
+        for key, value in request.GET.items():
```
ericholscher, line 205: Do we only want to check `GET`? Should we also check `POST`, or the `path`?

readthedocs/core/middleware.py
```diff
@@ -192,3 +191,20 @@ def __call__(self, request):
         response = self.get_response(request)
         response['Referrer-Policy'] = settings.SECURE_REFERRER_POLICY
         return response
+
+
+class NullCharactersMiddleware:
+    """
+    Block all requests that contains NULL characters (0x00) on their GET attributes.
```
ericholscher, line 198: Why are we doing this? Should explain a bit more in this docstring. 
```suggestion
    Block all requests that contains NULL characters (0x00) on their GET attributes.
    
    This is because these requests cause issues in our form handling code,
    and lead to 500s showing up in Sentry.
```

readthedocs/core/middleware.py
ericholscher: Makes sense. 👍

PR #9357, opened 2022-06-20
Add an explicit flyout placement option

ericholscher:
This PR does a few things:

* Lets themes define a `#readthedocs-embed-placement` element, where we will inject the footer
* Documents the flyout menu under Advanced Features

Refs https://github.com/pydata/pydata-sphinx-theme/issues/705


docs/user/flyout-menu.rst
```diff
@@ -0,0 +1,121 @@
...
+* If a page is an old version of the Read the Docs Sphinx theme, update the place where the theme expected it to be injected.
+* All other pages have the flyout appended to the ``<body>``
+
+Styling the flyout
+~~~~~~~~~~~~~~~~~~
+
+HTML themes can style the flyout to make it match the overall style of the HTML.
+By default the flyout has it's `own CSS file <https://github.com/readthedocs/sphinx_rtd_theme/blob/master/src/sass/_theme_badge.sass>`_,
+which you can look at to see the basic CSS class names.
+It also uses fontawesome to define some basic UI elements,
```
agjohnson, line 56:
```suggestion
It also uses Font Awesome to define some basic UI elements,
```

readthedocs/core/static-src/core/js/doc-embed/footer.js
```diff
@@ -1,14 +1,21 @@
...
+    // * The page looks like our Sphinx theme, updated the existing div
+    // * All other pages just get it appended to the <body>
 
-    // If the theme looks like ours, update the existing badge
-    // otherwise throw a a full one into the page.
-    // Do not inject for mkdocs even for the RTD theme
-    if (config.is_sphinx_builder() && config.is_rtd_like_theme()) {
+    var config = rtddata.get();
+    if ($(EXPLICIT_FLYOUT_PLACEMENT_SELECTOR).length > 0) {
+        $(EXPLICIT_FLYOUT_PLACEMENT_SELECTOR).html(data['html']);
```
agjohnson, line 16: This will execute the query for `EXPLICIT_FLYOUT_PLACEMENT_SELECTOR` twice. You could store this to a variable or use a more functional jQuery pattern here to avoid two lookups.

agjohnson (commented):
Great docs addition :+1: Something we've needed for a very long time

docs/user/flyout-menu.rst
```diff
@@ -0,0 +1,121 @@
...
+People who are making custom documentation themes often want to specify where the flyout is injected,
+and also what it looks like.
+We support both of these use cases for themes.
+
+Defining where the flyout menu is injected
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The flyout menu injection follows this logic:
+
+* When a ``<div id="readthedocs-embed-placement">`` is defined, inject the flyout menu there.
```
humitos, line 46: I read this and I thought that "embed" may be too generic and we may want to use something more specific, just in case in the future we need more placements. Why not use something like `readthedocs-flyout-placement`?

docs/user/flyout-menu.rst
```diff
@@ -0,0 +1,121 @@
...
+* If a page is an old version of the Read the Docs Sphinx theme, update the place where the theme expected it to be injected.
```
humitos, line 47: This bullet is not clear to me. What does "update the place where the theme expected it to be injected" mean?

docs/user/flyout-menu.rst
humitos: There is a PR opened to remove this and hardcode only the icons used. So, this may conflict with that. See https://github.com/readthedocs/readthedocs.org/pull/9297

humitos (commented):
Looks 💯

docs/user/flyout-menu.rst
ericholscher: Yea, I did that originally, but in the issue Pradyun suggested using the same prefix as the JS/CSS files in the docs, so people know where it came from. We need to unify all these names at some point..

docs/user/flyout-menu.rst
ericholscher: We could probably just remove this. It's an internal detail most users don't need to care about.

docs/user/flyout-menu.rst
ericholscher: I'll go ahead and remove this mention for now, so we don't have to remember to remove it. That does seem like a better approach.

docs/user/flyout-menu.rst
ericholscher: Went ahead with `readthedocs-embed-flyout`. I could also do `readthedocs-embed-flyout-placement`, but that seems overly verbose.

pradyunsg:
This isn't a blocker, but a future feature request -- would it be possible to make the CSS styling of the flyout not get injected, if a certain attribute is set on the object? This could look like adding an additional class on the element injected in that case, and scoping the styles to a `:not(.that-class)`.

ericholscher:
@pradyunsg Yea, I just had a conversation with @agjohnson, and he suggested using `data-` attributes. This is what we've done with the ad client, and it could enable something like `<div data-rtd-version="full"` data-rtd-style="false">` or similar. That would give us more customizability and options over time that can be configured in the HTML injection.

docs/user/flyout-menu.rst
agjohnson: Note: Font Awesome isn't being removed, just the need for font files. It's still a Font Awesome licensed graphic.

humitos (approved):
Looks good to me! The code is simple 👍🏼  I haven't QAed this locally, tho 😬

pradyunsg:
What can I do to help this get landed and deployed, so that the various active downstream themes can switch to this? :)

pradyunsg:
FYI: The PR description needs to be updated to reflect the implementation (which uses `readthedocs-embed-flyout`)

ericholscher:
@pradyunsg Going to merge this and deploy it tomorrow 👍

ericholscher:
@pradyunsg This should now be live, but I haven't heavily tested it other than ensuring it injects properly. Please let me know if it's working for you!

humitos:
@ericholscher I just realized that the example used in the documentation does not have translations. I think it would be good to improve the example by adding translations so theme authors have a full flyout example and can style it properly. What do you think?

ericholscher:
@humitos Sounds good to me 👍 I just copied an existing style, but we should definitely be documenting (and probably have an example) of the full possible content.

PR #9363, opened 2022-06-21
Development: remove silent and use long attribute name

humitos:
- remove `-s` (silent) because it takes some time to download the file and we
  don't know what it's happening; I cancelled the process multiple times due to
  this

- use `--disable` instead `-d` since it's more expressive


benjaoming (approved):
LGTM, thanks for fixing - and oh - haha - I thought it was some coincidence/exception when I thought it was funny how MinIO's CDN was slow... it is actually slow? :cold_face:

PR #9373, opened 2022-06-23
CSP header: enforce mode

humitos:
We have been in report mode for some time (years?) already and we haven't
detected any issues with our application. Most of the reports we received in
Sentry are valid and should be blocked.

Because of that, we are making `Content-Security-Policy` HTTP header to be
enfoced instead of keep in report only mode.

Related: https://github.com/readthedocs/readthedocs-ops/pull/1204


davidfischer (approved):
I think this should go smoothly.

ericholscher (approved):
Looks good. This only applies to the .org/.com, and not docs sites, right?

PR #9375, opened 2022-06-23
Properly log build exceptions in Celery

ericholscher:
This code wasn't in an exception,
so we have to pass the exc_info explicitly.

### DIFF
--- a/readthedocs/projects/tasks/builds.py
+++ b/readthedocs/projects/tasks/builds.py
@@ -425,9 +425,9 @@ class UpdateDocsTask(SyncRepositoryMixin, Task):
         else:
             # We don't know what happened in the build. Log the exception and
             # report a generic message to the user.
-            log.exception('Build failed with unhandled exception.')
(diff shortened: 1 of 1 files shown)


readthedocs/projects/tasks/builds.py
```diff
@@ -146,7 +146,9 @@ def on_failure(self, exc, task_id, args, kwargs, einfo):
             )
         else:
             # Catch unhandled errors when syncing
-            log.exception('An unhandled exception was raised during VCS syncing.')
+            log.exception(
+                "An unhandled exception was raised during VCS syncing.", exc_info=exc
+            )
```
humitos, line 151:
```suggestion
            # Note we are using `log.error(exc_info=...)` instead of `log.exception` 
            # because this is not executed inside a try/except block.
            log.error(
                "An unhandled exception was raised during VCS syncing.", exc_info=exc
            )
```

humitos (approved):
Good catch! Makes sense to me!

PR #9377, opened 2022-06-27
Docs: remove old feature flags

### LINKED ISSUE #9290: Migrate user-facing feature flags into the product
I'm creating an issue from an old Trello card that we have, so we can prioritize this work easily.

Taking a look at https://docs.readthedocs.io/en/stable/feature-flags.html#feature-flags, there are still some feature flags that we can migrate/remove:

- `PIP_ALWAYS_UPGRADE`: this can be replaced by `build.jobs.post_create_environment: pip install -U pip`
- `UPDATE_CONDA_STARTUP`: replaced by `build.jobs.pre_create_environment: conda install conda`
- `CONDA_APPEND_CORE_REQUIREMENTS`: I'm not sure there is an _easy_ way to replace this, but we could suggest a bashy solution maybe?
- `DONT_OVERWRITE_SPHINX_CONTEXT`: this one does not seem to be replaceable either
- `LIST_PACKAGES_INSTALLED_ENV`: this can be done now via `build.jobs.pre_build: pip freeze`

I think we could add these examples to FAQs (or even the "Build customization" page) and remove them from the page linked.

humitos:
I'm removing these because we don't want users ask for them to use in their
projects. All of the ones removed can be achieved by using `build.jobs.*` config.

Closes #9290

### DIFF
--- a/docs/user/feature-flags.rst
+++ b/docs/user/feature-flags.rst
@@ -11,22 +11,12 @@ or reaching out to the administrator of your service.
 Available Flags
 ---------------
 
-``PIP_ALWAYS_UPGRADE``: :featureflags:`PIP_ALWAYS_UPGRADE`
-
-``UPDATE_CONDA_STARTUP``: :featureflags:`UPDATE_CONDA_STARTUP`
-
-The version of ``conda`` used in the build process could not be the latest one.
-This is because we use Miniconda, which its release process is a little more slow than ``conda`` itself.
-In case you prefer to use the latest ``conda`` version available, this is the flag you need.
-
 ``CONDA_APPEND_CORE_REQUIREMENTS``: :featureflags:`CONDA_APPEND_CORE_REQUIREMENTS`
 
 Makes Read the Docs to install all the requirements at once on ``conda create`` step.
 This helps users to pin dependencies on conda and to improve build time.
 
 ``DONT_OVERWRITE_SPHINX_CONTEXT``: :featureflags:`DONT_OVERWRITE_SPHINX_CONTEXT`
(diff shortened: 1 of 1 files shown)


humitos:
We don't have a plan. However, I'm not sure if it's a big deal:

```
In [3]: Feature.objects.get(feature_id=Feature.LIST_PACKAGES_INSTALLED_ENV).projects.count()
Out[3]: 9

In [4]: Feature.objects.get(feature_id=Feature.UPDATE_CONDA_STARTUP).projects.count()
Out[4]: 12

In [5]: Feature.objects.get(feature_id=Feature.PIP_ALWAYS_UPGRADE).projects.count()
Out[5]: 5
```

None of these features are "breaking changes". Even if we remove these projects from these features, they should continue building. Maybe the conda one break --we should probably tell them to migrate to mamba if we haven't already (some of them are in the Google Sheet with a PR open/merged).

I'm happy to send them an email saying that in 30 days this feature will be removed completely, tho

ericholscher:
> I'm happy to send them an email saying that in 30 days this feature will be removed completely, tho

Yea, I'd like to see us doing this proactively. I really liked the approach we took previously with opening PR's on their repos as well. I think getting a really nice workflow for proactively migrating users will definitely help us do it more frequently. Perhaps something that @benjaoming could help with.

benjaoming (approved):
LGTM - would be nice to have this removed from the documentation.

benjaoming:
@ericholscher sounds like a good idea! Kind of like a "feature flag deprecation bounty". Added to our roadmap.

PR #9387, opened 2022-06-29
Build: handle Celery timeout events

humitos:
Solve the problem reported by Sentry when the Celery task times out: https://sentry.io/organizations/read-the-docs/issues/3038754375/

> Each commit has a detailed explanation of the changes

Note to myself: the latest commit tries to handle soft timeout cancelled properly. It's close but it needs a little more of QA because for some reason `SIGUSR1` is triggered/catched as well.

### DIFF
--- a/readthedocs/builds/views.py
+++ b/readthedocs/builds/views.py
@@ -193,7 +193,7 @@ class BuildDetail(BuildBase, DetailView):
             build_task_id=build.task_id,
             terminate=terminate,
         )
-        app.control.revoke(build.task_id, signal=signal.SIGINT, terminate=terminate)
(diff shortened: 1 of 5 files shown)


humitos:
Note that we are not hitting this too frequently because we have an _pretty high_ timeout due to this: https://github.com/readthedocs/readthedocs.org/blob/634f85bf00ffcab7bc91926cc9bf9cc394d375fb/readthedocs/projects/tasks/utils.py#L95-L100

humitos:
I'm closing this PR for now. We can come back in the future if we consider. The idea is good, but I wasn't able to properly QA on production to be sure it works as I expect.

PR #9410, opened 2022-07-07
OAuth: resync `RemoteRepository` weekly for active users

humitos:
Trigger a daily task to compare user's last login isoweekday with today's
isoweekday for all active users. If they matches, we resync the
`RemoteRepository` for this user.

This logic is the same as "resync `RemoteRepository` once a week per each user".

We consider active users those that have logged in at least once in the last 90
days.

Related: https://github.com/readthedocs/readthedocs.org/issues/8229
Related: https://github.com/readthedocs/readthedocs.org/pull/9409


humitos:
Note that this is about 2k users to resync per day, which I don't think it's terrible, but it's definitely a lot more than what we are doing currently:

```
In [75]: Counter(
    ...:     User.objects.annotate(weekday=ExtractIsoWeekDay("last_login"))
    ...:     .filter(last_login__gt=three_months_ago, socialaccount__isnull=False)
    ...:     .order_by("weekday")
    ...:     .values_list("weekday", flat=True)
    ...: )
Out[75]: Counter({1: 1749, 2: 1873, 3: 1927, 4: 1749, 5: 1603, 6: 1025, 7: 942})
```

This may stress our `web-celery` ASG a little. We could split those 2ks into 10 and trigger ~200 tasks every hour instead of 2k at the same time. Does it worth trying this as-is and splitting it later if we found that's too many tasks at once?

There are some improvements to do here too:

- ignore GH organization with plenty of repositories (see https://github.com/readthedocs/readthedocs.org/issues/8979)
- mark users with revoked permissions for our app to not resync until they reconnect
- migrate to GH Application that allows us to receive webhooks instead (this migration is not trivial, as we found out)

humitos:
There is an unrelated problem with the docs,

```
/home/circleci/project/docs/user/guides/jupyter.rst:306: WARNING: unknown document: poliastro:gallery
```

PR #9413, opened 2022-07-07
Docs: improve the flyout page to include a full example

humitos:
- add the Languages section
- add the Downloads section
- include some HTML comments explaining when the sections are present

Note that this is _not a real example_ since Flask does not have translations
nor downloads, but I don't think that's super important for the purpose of the
example. The most important is the structure of the HTML than the URLs included
on it.

We could use a real example if we find a good one, tho.

> Based on the content of the `restapi/footer.html` file: https://github.com/readthedocs/readthedocs.org/blob/d77cc3a70d32d7621f252ae78bf763d1ee2a34cb/readthedocs/api/v2/templates/restapi/footer.html

Related: https://github.com/readthedocs/readthedocs.org/pull/9357#issuecomment-1171729910

### DIFF
--- a/docs/user/flyout-menu.rst
+++ b/docs/user/flyout-menu.rst
@@ -21,15 +21,17 @@ Closed
 ~~~~~~
 
 .. figure:: /_static/images/flyout-closed.png
+   :align: center
 
-    The flyout when it's closed
+   The flyout when it's closed
 
 Open
 ~~~~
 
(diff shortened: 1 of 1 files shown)


agjohnson (approved):
I'm looking forward to providing users with a cleaner implementation of the flyout, but until we have that, we might as well be thorough :+1:

PR #9433, opened 2022-07-14
Requirements: update them with `pur`

humitos:
Use `pur` and run it manually because `pyup` has been broken lately. See
https://github.com/readthedocs/readthedocs.org/issues/9273

So, I went ahead and use `pur` which does exactly the same but it requires a
manual run. I updated all the requirements we are tracking as `txt` but I
haven't touched the `in` since that is using a different tool.


humitos:
Superseded by the technique used in https://github.com/readthedocs/sphinx-hoverxref/pull/221

Issue #9438, opened 2022-07-19
Environment variables not available when overriding build


agoose77:
I'm working on documenting a project that will require long build times (exceeding available time on RTD). We already use GitHub Actions to build the necessary assets, and I'd hoped to build the documentation on the CI as well. In order to connect to RTD, my plan was to trigger a build on RTD via webhook, and have RTD pull the assets from the GitHub Actions Workflow. This requires a custom build process (`build.commands`).

Unfortunately, I observed that the access token I'd added to the environment variables in the project settings was not exposed to the builder logic. Here's a build that demonstrates the problem:

(`SPECIAL_TOKEN` is not visible in the output of `env` in this build)

https://readthedocs.org/projects/test-ccache-gha/builds/17481074/

Apologies that this is so terse, and thanks for all of the hard work!

agjohnson:
I hit this today as well. I made a test-build test case to test it at https://github.com/readthedocs/test-builds/pull/2091

agoose77:
@agjohnson what are the odds! Thanks for making a cleaner test case.

humitos:
Hi! The fix for this issue was already deployed. I did a quick test in https://readthedocs.org/projects/test-builds/builds/17604894/ and I can see my variable there. Let me know if this works for you!

PR #9443, opened 2022-07-20
Update env var docs

agjohnson:
A user asked about pull request builds and env vars, so figured it's a
good page to update.

* Mention value READTHEDOCS_VERSION for pull request builds
* Convert table view to definition list view
* Use envvar directive to give linkable env vars and env var roles that
  resolve from other docs
* Update "slug" definition

### DIFF
--- a/docs/user/environment-variables.rst
+++ b/docs/user/environment-variables.rst
@@ -6,7 +6,10 @@ Read the Docs supports two types of environment variables in project builds:
 * `Default environment variables`_
 * `User-defined environment variables`_
 
-Both are merged together during the build process and are exposed to all of the executed commands.
+Both are merged together during the build process and are exposed to all of the executed commands. There are two exceptions for user-defined environment variables however:
+
+* User-defined variables are not available during the checkout step of the :doc:`build process </builds>`
+* User-defined variables that are not marked as public will not be available in :docs:`pull request builds </pull-requests>`
 
 Default environment variables
 -----------------------------
--- a/docs/user/environment-variables.rst
+++ b/docs/user/environment-variables.rst
@@ -9,7 +9,7 @@ Read the Docs supports two types of environment variables in project builds:
 Both are merged together during the build process and are exposed to all of the executed commands. There are two exceptions for user-defined environment variables however:
 
 * User-defined variables are not available during the checkout step of the :doc:`build process </builds>`
-* User-defined variables that are not marked as public will not be available in :docs:`pull request builds </pull-requests>`
+* User-defined variables that are not marked as public will not be available in :doc:`pull request builds </pull-requests>`
 
 Default environment variables
 -----------------------------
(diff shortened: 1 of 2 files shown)


docs/user/environment-variables.rst
```diff
@@ -1,36 +1,51 @@
...
+Read the Docs supports two types of environment variables in project builds:
 
-* default environment variables
-* user-defined environment variables
-
-Both are merged together during the build process and exposed to all the commands executed when building the documentation.
+* `Default environment variables`_
+* `User-defined environment variables`_
 
+Both are merged together during the build process and are exposed to all of the executed commands.
```
ericholscher, line 9: @stsewd is this right? I thought we didn't expose the env vars in some cases like checking out the code, but I'm not 100% sure.

ericholscher (approved):

docs/user/environment-variables.rst
humitos: Yeah, "VCS environment" does not expose environment variables. Only in the "Build environment" they are exposed.

I documented this the other day when a user hit https://github.com/readthedocs/readthedocs.org/issues/9438#issuecomment-1189288612

docs/user/environment-variables.rst
```diff
@@ -1,36 +1,51 @@
...
-   ``READTHEDOCS_LANGUAGE``, The Read the Docs' language slug of the project which is being built, ``en``
+    Whether the build is running inside Read the Docs.
 
-.. note::
+    :Default: ``True``
+
+.. envvar:: READTHEDOCS_VERSION
+
+    The :term:`slug` of the version being built, such as ``latest``, ``stable``,
+    or a branch name like ``feature-1234``. For `pull request builds </pull-requests>`,
```
humitos, line 25:
```suggestion
    or a branch name like ``feature-1234``. For `pull request builds </pull-requests>`_,
```

humitos:
```suggestion
    or a branch name like ``feature-1234``. For :doc:`pull request builds </pull-requests>`,
```

humitos: Not sure which one is the one wanted here.

agjohnson: Oops, yeah it was supposed to be `:doc:...` if I had copy pasted correctly.

docs/user/environment-variables.rst
agjohnson: So, #9438 bug aside, the VCS env is only used for the checkout step, correct?

Is this accurate then?

```suggestion
Both are merged together during the build process and are exposed to all of the executed commands. There are two exceptions for user-defined environment variables however:

* User-defined variables are not available during the checkout step of the :doc:`build process </builds>`
* User-defined variables that are not marked as public will not be available in :docs:`pull request builds </pull-requests>`
```

ericholscher: That sounds correct 👍

PR #9445, opened 2022-07-21
Docs: `poetry` example on `build.jobs` section

humitos:
There are multiple ways to install Poetry on Read the Docs and all of them have, at least,
one thing that I don't like. So, I'm recommending using the official one
described as the first option in their own page --which is not perfect, but it's
the official one :)

### DIFF
--- a/docs/user/build-customization.rst
+++ b/docs/user/build-customization.rst
@@ -244,8 +244,8 @@ To setup it, you need to define the version of Node.js to use and install the de
 Install dependencies with Poetry
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-In case the project is managing Python dependencies with `Poetry <https://python-poetry.org/>`__,
-it can use ``post_create_environment`` user-defined job to install them while building the documentation.
+Projects managed with `Poetry <https://python-poetry.org/>`__,
+can use the ``post_create_environment`` user-defined job to use Poetry for installing Python dependencies.
 Take a look at the following example:
 
 
--- a/docs/user/build-customization.rst
+++ b/docs/user/build-customization.rst
@@ -241,6 +241,37 @@ To setup it, you need to define the version of Node.js to use and install the de
          - npm install -g jsdoc
 
 
+Install dependencies with Poetry
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+In case the project is managing Python dependencies with `Poetry <https://python-poetry.org/>`__,
+it can use ``post_create_environment`` user-defined job to install them while building the documentation.
+Take a look at the following example:
+
+
+.. code-block:: yaml
+   :caption: .readthedocs.yaml
+
+   version: 2
+
+   build:
+     os: "ubuntu-22.04"
+     tools:
+       python: "3.10"
+     jobs:
+       post_create_environment:
+         # Install poetry
+         # https://python-poetry.org/docs/#osx--linux--bashonwindows-install-instructions
+         - curl -sSL https://raw.githubusercontent.com/python-poetry/poetry/master/get-poetry.py | python -
+         # Tell poetry to not use a virtual environment
+         - $HOME/.poetry/bin/poetry config virtualenvs.create false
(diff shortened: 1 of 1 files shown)


ericholscher (approved):
This looks reasonable 👍 It's not super clean, but works well enough.

agjohnson:
I think they suggest this version, but more for local development purposes. For our use case, where we are building a fresh environment every build and Poetry and dependencies will not have a chance to conflict with anything, the pip install method seems totally reasonable.

docs/user/build-customization.rst
```diff
@@ -241,6 +241,37 @@ To setup it, you need to define the version of Node.js to use and install the de
          - npm install -g jsdoc
 
 
+Install dependencies with Poetry
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+In case the project is managing Python dependencies with `Poetry <https://python-poetry.org/>`__,
+it can use ``post_create_environment`` user-defined job to install them while building the documentation.
```
agjohnson, line 248: Just making language a bit clearer.

```suggestion
Projects managed with `Poetry <https://python-poetry.org/>`__,
can use the ``post_create_environment`` user-defined job to use Poetry for installing Python dependencies.
```

humitos:
Mrm... another option could be to install Poetry on `pre_create_environment` which will use the Python from the system (not the virtualenv) and then use `post_create_environment` to install the dependencies? How do we tell Poetry to install the dependencies inside the virtualenv _it's not running from_ then? 🤔

agjohnson:
Ah good point. I don't have any strong objections here, I've just found that the pip install method works fine for me. I'm sure we'll get some conflicting advice from Poetry users here soon either way -- seemed that there were some opinions on this pattern.

> How do we tell Poetry to install the dependencies inside the virtualenv it's not running from then?

I think I get what you're saying. Wouldn't this be the same case for the curl installed method, which is just packaged up away from the system/venvs? I don't know the answer to your question, but your suggestion sounds reasonable.

humitos:
> Mrm... another option could be to install Poetry on `pre_create_environment` which will use the Python from the system (not the virtualenv) and then use `post_create_environment` to install the dependencies?

I tested this and it didn't work: https://readthedocs.org/projects/test-builds/builds/17527514/. I tried to do some other tests locally and I'm not exactly sure how poetry internals work, but also I'm not sure it's worth going too deep into this either. I prefer to recommend the official way to install it than the not recommended by themselves because it looks nicer but introduce our users to random issues that we are not aware right now.

humitos:
We can always update this example if we find there is a better and supported way of doing this.

davidorme:
@humitos - apologies for resurrecting this PR, it seemed like a natural place to post this and I couldn't see any more recent issues about `poetry` and installing.

I'm trying to build a site for a package managed using `poetry` and following the officially documented approach in the `build.jobs` docs [section](https://docs.readthedocs.io/en/stable/build-customization.html#install-dependencies-with-poetry). That approach now appears to be failing:

https://readthedocs.org/projects/virtual-rainforest/builds/18663328/

The TLDR on that build log error is that the old `get-poetry.py` install _will not_ install recent versions (`>1.2.0a1`) and that the new `https://install.python-poetry.org` approach must be used instead. I tried using that approach:

```yaml
build:
  os: ubuntu-20.04
  tools:
    python: "3.9"
  jobs:
    post_create_environment:
      # Install poetry
      - curl -sSL https://install.python-poetry.org | python3 -
      # Tell poetry to not use a virtual environment
      - $HOME/.local/bin/poetry config virtualenvs.create false
      # Install project's dependencies
      - $HOME/.local/bin/poetry install
```

but that doesn't seem to be working: https://readthedocs.org/projects/virtual-rainforest/builds/18663177/

I know that is an apparent error in my package, but it doesn't occur with local `sphinx` builds and seems like a problem with the installed environment?

agjohnson:
Hi @davidorme! Thanks for the update on the install method, we should probably note that in our docs. Care to open this up as a new issue instead?

PR #9446, opened 2022-07-22
Docs: Adds Jupyter Book to examples table

benjaoming:
Not sure how this current example table should look wrt. headline.. proposing to rename it to "List of current example projects", supposing that we might have a list of "outdated" examples or "unmaintained examples".

Before:

image

After:

image

### DIFF
--- a/docs/user/examples.rst
+++ b/docs/user/examples.rst
@@ -31,7 +31,7 @@ Sphinx and MkDocs examples
    * - Jupyter Book
      - Jupyter Book and Sphinx
      - `[Git] <https://github.com/readthedocs-examples/example-jupyter-book/>`__ `[Rendered] <https://example-jupyter-book.readthedocs.io/>`__
-     - Jupyter Book w/ popular integrations configured for RTD
+     - Jupyter Book with popular integrations configured
 
 
 Real-life examples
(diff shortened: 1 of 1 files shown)


humitos:
It looks like before/after images are the same? 🤔

benjaoming:
Thanks, updated! The screenshots were definitely also overdoing it :)

docs/user/examples.rst
```diff
@@ -28,6 +28,10 @@ Sphinx and MkDocs examples
      - MkDocs
      - `[Git] <https://github.com/readthedocs-examples/example-mkdocs-basic/>`__ `[Rendered] <https://example-mkdocs-basic.readthedocs.io/en/latest/>`__
      - Basic example of using MkDocs
+   * - Jupyter Book
+     - Jupyter Book and Sphinx
+     - `[Git] <https://github.com/readthedocs-examples/example-jupyter-book/>`__ `[Rendered] <https://example-jupyter-book.readthedocs.io/>`__
+     - Jupyter Book w/ popular integrations configured for RTD
```
agjohnson, line 34: I like to expand contractions or short hand english, as you never know what translation processes, screen readers, etc do with short hand like "w/"

"for RTD" seems redundant here, as the context is already for example projects on RTD

```suggestion
     - Jupyter Book with popular integrations configured
```

agjohnson (approved):
:tada:

Unrelated to this PR, but this table would probably flow better as a definition list. The table version feels really dense, given it's maybe the focus of this page.

benjaoming:
@agjohnson I guess that we could embed these columns in a definition list. I am, though, inclined to thinking that it's actually the table functionality that can benefit from improvements. For instance, I have this antique PR :smile: https://github.com/readthedocs/sphinx_rtd_theme/pull/432

agjohnson:
That's a valid, separate topic, but I think this content would look better without the table entirely.

PR #9447, opened 2022-07-22
Docs: Read the Docs for Science

benjaoming:
Opening for viewing Pull Request build: https://docs--9447.org.readthedocs.build/en/9447/science.html

Reviews: This is still draft'ish - Please comment on general contents, we can tweak language and fix typos once ready.

PR #9460, opened 2022-07-29
Docs: Read the Docs for Science - new alternative with sphinx-design

### LINKED ISSUE #9467: Data Science landing page in docs
This is an initial start of a single page to link people when we talk about our grant work and data science outreach.

This was converted from a note, I'm sure there is more context somewhere though.

benjaoming:
This is opened to have an alternative PR build to the term list in #9447

The idea of having dropdowns is that we can easily mange more contents in the same page without presenting readers to a wall of text or a lot of scrolling.

Each drop-down can contain links, examples, screenshots etc. An incremental process following this will be to enrich these drop-downs. This PR lays out the path but doesn't conclude the destination :)

Rendered link: https://docs--9460.org.readthedocs.build/en/9460/science.html

Fixes #9467

----
:books: Documentation preview :books:: https://docs--9460.org.readthedocs.build/en/9460/


humitos:
I like how it looks. However, I'm not 100% convinced about how these drop-downs are used in this particular example, mainly because each of the drop-downs only have one sentence inside once you click them. So, the reader's effort is "too much" compared with what they get when they click on it. On the other hand, if the content of the drop-down grows too much, I'd say that's better to have a specific section (or page) for that topic in particular. That said, at this stage, I think it _looks good_ but it does not increase the UX for readers, IMHO.

Direct link to the rendered page: https://docs--9460.org.readthedocs.build/en/9460/science.html

benjaoming:
@humitos 
> However, I'm not 100% convinced about how these drop-downs are used in this particular example, mainly because each of the drop-downs only have one sentence inside once you click them. 

See description of PR:

> The idea of having dropdowns is that we can easily mange more contents in the same page without presenting readers to a wall of text or a lot of scrolling.

benjaoming:
The PR is in draft mode, so before I go ahead and add a lot of content, I wanted to see how it played out :+1: There's no intention to entertain the reader with meaningless clicks to open up 1 line of text :)

benjaoming:
@agjohnson this PR is now ready for a second review :) It has a lot of consideration. I'm still especially interested in any obvious gaps.

I was actually helped to identify one such gaping mistake in the "🪄 Easy to use" section which had some incredibly dumb and not-so-easy sentences (they have since then been fixed).

docs/user/science.rst
```diff
@@ -0,0 +1,142 @@
...
+* Course and research websites
+* Interactive classroom activities
+* Data science software documentation
+
+
+.. button-ref:: executablebook:gallery
+    :ref-type: doc
+    :color: primary
+
+    Visit the gallery of solutions built with Jupyter Book
```
humitos, line 106: I'd not use a button for this since it just a link to an external page. I think it's more accurate to use just a sentence with a link for this.

docs/user/science.rst
```diff
@@ -0,0 +1,142 @@
...
+    .. TODO: Where should this video live?
+
+    .. video:: https://overtag.dk/files/enable-pull-request-builders.mp4
+       :width: 500
+       :height: 300
+
+.. dropdown:: 💬 Collaboration and community
+
+    .. figure:: /img/screenshot_edit_on_github.png
+        :align: right
```
humitos, line 69: This image could be cropped to remove the top part of the Firefox window: x button, space and "Search" input.

docs/user/science.rst
```diff
@@ -0,0 +1,142 @@
...
+
+.. dropdown:: ⚙️ Automatic builds
+
+    Build and publish your project for every change made through Git (GitHub, GitLab, BitBucket etc). Preview changes via Pull Requests. Receive notifications when something is wrong. How does this work? Have a look at this video:
+
+    .. TODO: Where should this video live?
+
+    .. video:: https://overtag.dk/files/enable-pull-request-builders.mp4
+       :width: 500
+       :height: 300
```
humitos, line 64: This could be centered inside the dropdown.

docs/user/science.rst
```diff
@@ -0,0 +1,142 @@
...
+""""""""""""""""""
+
+.. TODO: get the correct link for
+.. :external+jupyter:ref:`the many sub-projects of Jupyter <index.md#sub-project-documentation>`
+
+Read the Docs community for science is already big and keeps growing. The :external+jupyter:doc:`Jupyter Project <index>` itself and `the many sub-projects of Jupyter <https://docs.jupyter.org/en/latest/#sub-project-documentation>`__ are built and published with Read the Docs.
+
+.. grid:: 3
+    :gutter: 2
+    :padding: 0
```
humitos, line 128: I loved this grid! 💯

docs/user/science.rst
```diff
@@ -0,0 +1,142 @@
...
+    .. grid-item-card:: Jupyter Project Documentation
+      :img-top: img/logo_jupyter.png
+      :link: https://docs.jupyter.org/
+
+    .. grid-item-card:: Chainladder - Property and Casualty Loss Reserving in Python
+      :img-top: img/logo_chain_ladder.png
+      :link: https://chainladder-python.readthedocs.io/
+
+    .. grid-item-card:: Feature-engine - A Python library for Feature Engineering and Selection
+      :img-top: img/logo_feature_engine.png
```
humitos, line 139: These images could be resized to save some bandwidth when loading the page.

humitos (commented):
I really like this page! 👍🏼 

I'm still not sure about the dropdowns (it's not a -1, tho, just "meh" 😄 ). I like how they look, but, as a reader I will end up by clicking on all of them to see its content, so I'd prefer them to have been expanded originally 😄 

However, I like the structure of the page: introduction, dropdowns, getting started (gallery, grid with examples). I think the usage of a button for the gallery is _too much_, tho.

I didn't review the content itself just yet.

docs/user/science.rst
```diff
@@ -0,0 +1,142 @@
...
+
+    Science and academia have a big kinship with software developers: We ❤️ community. Our solutions and projects become better when we foster inclusivity and active participation. Read the Docs features easy access for readers to suggest changes via GitHub, GitLab, BitBucket etc. But not just any unqualified feedback. Instead, the *code* and all the tools are available for your community to forge qualified contributions.
+
+    Your readers can become your co-authors!
+
+    Discuss changes via Pull Request and track all changes in your project's version history.
+
+    Does this sound chaotic? Don't worry, the full ownership and permission handling remains in your hands: Project and organization owners govern what is released and who has access to approve and build changes.
+
+.. dropdown:: 🔎 Search Engine Optimization
```
agjohnson, line 79: This section feels a little out of place, perhaps it could be generalized to cover more search features instead?

Neither RTD or Sphinx have really strong features to help SEO, but we do have strong search features otherwise. Maybe what we're focusing on here is that readers can easily find what they are looking for, and enumerate search features instead.

Unless we have input from projects that list SEO as a strong concern for their project, though search just in general is usually a higher priority item for projects.

docs/user/science.rst
```diff
@@ -0,0 +1,142 @@
...
+
+    Don't reinvent SEO: Have SEO handled jointly by the combined best-practices from Sphinx, its themes and Read the Docs' platform.
+
+.. dropdown:: 🌱 Grow your own solutions
+
+    The eco-system is Open Source and makes it accessible for anyone with Python skills to build their own extensions.
```
agjohnson, line 85: Open source doesn't need proper capitalization here. Microsoft's styleguide has some guidance on the term:

https://docs.microsoft.com/en-us/style-guide/a-z-word-list-term-collections/o/open-source


```suggestion
    The eco-system is open source and makes it accessible for anyone with Python skills to build their own extensions.
```

docs/user/science.rst
```diff
@@ -0,0 +1,142 @@
...
+
+    Use the latest state-of-the-art themes for web and have PDFs and e-book formats automatically generated.
+
+    New themes are improving every day, and when you write documentation based on Jupyter Book and Sphinx, you will separate your contents and semantics from your presentation logic. This way, you can keep up with the latest theme updates or try new themes.
+
+    Another example of the benefits from separating content and presentation logic: Your documentation also transforms into printable books and eBooks.
+
+.. dropdown:: 📐 Widgets, widgets and more widgets
+
+    Design your science project's layout and components with widgets from a rich eco-system of Open Source extensions built for many purposes. Special widgets help users display and interact with graphs, maps and more. :external+jupyterbook:doc:`Several <content/components>` `extensions <https://sphinx-gallery.github.io/>`__ are built and invented by the science community.
```
agjohnson, line 54:
```suggestion
    Design your science project's layout and components with widgets from a rich eco-system of open-source extensions built for many purposes. Special widgets help users display and interact with graphs, maps and more. :external+jupyterbook:doc:`Several <content/components>` `extensions <https://sphinx-gallery.github.io/>`__ are built and invented by the science community.
```

docs/user/science.rst
agjohnson: It's also wise to be careful with buttons. This doesn't seem like the place we want to send most readers, but a button will be really prominent.

If you want to keep this as a button, it is a very wordy button. It might be better to break up the button into a paragraph with details and a shorter button.

docs/user/science.rst
```diff
@@ -0,0 +1,142 @@
...
+    Visit the gallery of solutions built with Jupyter Book
+
+Ready to get started?
+"""""""""""""""""""""
+
+.. Note that this is a deliberate repitition of a previous segment. Should it repeat? Maybe not, but for now it's nice to be sure that people see it.
+
+* All new to this? Take the official :external+jupyterbook:doc:`Jupyter Book Tutorial <start/your-first-book>`
+* Curious for practical code? See :doc:`/examples`
+* Familiar with Sphinx? See :doc:`/guides/jupyter`
```
agjohnson, line 115: This might be a better section for a prominent button instead, if we have a singular place we want to send readers to continue.

docs/user/science.rst
```diff
@@ -0,0 +1,142 @@
...
+
+    Design your science project's layout and components with widgets from a rich eco-system of Open Source extensions built for many purposes. Special widgets help users display and interact with graphs, maps and more. :external+jupyterbook:doc:`Several <content/components>` `extensions <https://sphinx-gallery.github.io/>`__ are built and invented by the science community.
+
+.. dropdown:: ⚙️ Automatic builds
+
+    Build and publish your project for every change made through Git (GitHub, GitLab, BitBucket etc). Preview changes via Pull Requests. Receive notifications when something is wrong. How does this work? Have a look at this video:
+
+    .. TODO: Where should this video live?
+
+    .. video:: https://overtag.dk/files/enable-pull-request-builders.mp4
```
agjohnson, line 62: Noted in chat, but this is better as an embed from a streaming service. A static embed can't be sampled down to a lower bitrate and takes more JS work to add features that an embed get us.

agjohnson (commented):
The layout of the page is quite nice, though it feels like there are some tweaks needed. The content is going to be rather hidden behind the dropdown sections. Readers can't easily scan the page, and so it's hard to build up engagement with a reader without more of the content showing. I agree it's a rather dense page if we expand all the sections however. For now, I might echo Manuel's suggestion and open all the sections by default, so the user can minimize the sections if they want.

There are some longer term options for improving reader flow, but let's see what use this page gets before we optimize things too much.

humitos:
> Readers can't easily scan the page, and so it's hard to build up engagement with a reader without more of the content showing. I agree it's a rather dense page if we expand all the sections however

This was one of my main concerns the first time I saw these dropdowns in action. Depending on the content inside them, it could be preferable to have regular sections (h2 or h3). Note that we can't link to a particular dropdown or do Ctrl+F to search for a term while the dropdown is closed. These are some things that make me prefer regular sections.

docs/user/science.rst
benjaoming: It's turned into a normal link now. Not sure I'm liking how it's dangling on its own now, so I think it needs to be fixed with some other means.

I'm not too worried about people exiting to the Jupyter Book Gallery. If someone in science/academia sees this page, I feel quite confident that they will come back in order to get started. They will have already scrolled passed our feature list to reach this point.

docs/user/science.rst
benjaoming: The directive doesn't support centering as an option, however using `100%` width does the work

image

docs/user/science.rst
benjaoming: Yes, maybe that isn't achieving what I wanted. Just wanted the screenshot to be easy to decode by giving some context. But I guess for someone on a different OS with a different browser, it's not much help :)

benjaoming: Fixed!

docs/user/science.rst
benjaoming: It was possible to reduce the Jupyter logo from 117K to 33K.

The other logos aren't possible to reduce in file size by downsampling the image.

docs/user/science.rst
benjaoming: With several options to choose from, I'm not sure which one is the right one to go with.

Using [sphinxcontrib-video](https://github.com/sphinx-contrib/video) and:

* Hosting our video tutorials as mp4 somewhere

Using [sphinxembeddedvideos](https://github.com/Peque/sphinxembeddedvideos) and:

* Hosting on archive.org (not sure which embed features it has)
* Hosting on a Peertube service (we need to choose or host one)
* Hosting on Youtube (tracking, ads etc)
* Paid Vimeo?

docs/user/science.rst
```diff
@@ -0,0 +1,142 @@
...
+    .. figure:: /img/screenshot_edit_on_github.png
+        :align: right
+
+    Science and academia have a big kinship with software developers: We ❤️ community. Our solutions and projects become better when we foster inclusivity and active participation. Read the Docs features easy access for readers to suggest changes via GitHub, GitLab, BitBucket etc. But not just any unqualified feedback. Instead, the *code* and all the tools are available for your community to forge qualified contributions.
+
+    Your readers can become your co-authors!
+
+    Discuss changes via Pull Request and track all changes in your project's version history.
+
+    Does this sound chaotic? Don't worry, the full ownership and permission handling remains in your hands: Project and organization owners govern what is released and who has access to approve and build changes.
```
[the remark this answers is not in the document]
benjaoming: I agree and adding to this: The previous sentence (now fixed) also had an assumption that the reader might be confused here. I'm quite okay with having a warm and friendly text, but it shouldn't unnecessarily derogatory/disparaging.

Just quickly noting that especially terms like "VCS" should not be in this text. I know that you didn't mean to suggest that, but it reminds me to mention that the entire text needs to be verified by someone who is not "internal", or at the very least, we have to train our own reading to the POV of an academic worker :)

docs/user/science.rst
benjaoming: Since these are 3 alternative exit points, I don't want to give any of them more attention than the next. I don't think that the buttons will be all that nice in the list, and I think we might want to add some more "Getting started" resources here over time, as the audience is very non-homogeneous.

benjaoming: I've tested with buttons and I can't get it to work. The sphinx-design buttons are block level elements and won't look good inline, even if we CSS'ed them.

There isn't really a design element with buttons that fits what we want here: Multiple resources listed with clearly visible links.

I've adapted the page slightly now, I think we should work with how anonymous the text is becoming in a future version :)

image

benjaoming:
I'm quite happy with the dropdowns for now :)

The page has a typical design paradox: We can't give everything priority and making one choice for the layout often contradicts another. I think @agjohnson said it well:

> The content is going to be rather hidden behind the dropdown sections. Readers can't easily scan the page

So while readers may get an easy way to scan the page if we hide some of the contents initially, they might have to click a bit more to unhide them - we need to choose, I don't see a compromise. Of course, if we were building a full website, we could use slideshows, animations and take more advantage of the entire screen... but this is nothing more than a documentation article with a few widgets available :)

I'll mention a few things as to why I like this layout:

1. We don't know **who** the readers are, and I'm more afraid of losing them because of in-decipherable content, foreign words and high LIX number. I prefer to hide most contents inside dropdowns to be **cautious** by having **easy and friendly headlines**. You can say that the reader might read all the feature headlines and happily know that they are substantiated without de-collapsing and reading the actual contents of each box. That's actually fine!

2. Once we know more about the reader, we may well find out that they're not interested in the full feature list. My point here is: No matter if we can adapt our text to the knowledge of the reader, we might still bore them with the wrong list of features. That's my reason to believe that the dropdowns or a similar type of widget will be useful in the future (as long as this page is nested inside the frames of our documentation project).

3. Some outcomes that I would look for here:

   * A reader whom otherwise did not know what Sphinx and Read the Docs was and what software documentation is about is convinced that there is something worth looking into
   * An existing user of Jupyter Book gets interested in using Read the Docs (there are MANY that don't, so huge potential here)
   * Once published, we embark on a path to appeal to a highly resourceful and productive community (science and academia) that would otherwise not think of themselves as writing "documentation". Our messages might be untested, and hopefully we will get some feedback from the existing documentation community who observe it, and we might want to have some more feedback and debate around the vision. Especially the Executable Book community. It's really really exciting :)

docs/user/science.rst
```diff
@@ -111,17 +111,17 @@ Here are some popular activities that are well-supported by Jupyter Book:
...
 """""""""""""""""""""
 
 .. Note that this is a deliberate repitition of a previous segment. Should it repeat? Maybe not, but for now it's nice to be sure that people see it.
 
-* All new to this? Take the official :external+jupyterbook:doc:`Jupyter Book Tutorial <start/your-first-book>`
-* Curious for practical code? See :doc:`/examples`
-* Familiar with Sphinx? See :doc:`/guides/jupyter`
+* All new to this? Take the official :external+jupyterbook:doc:`Jupyter Book Tutorial » <start/your-first-book>`
+* Curious for practical code? See the list of :doc:`Example Projects » </examples>`
+* Familiar with Sphinx? Read :doc:`How to use Jupyter notebooks in Sphinx » </guides/jupyter>`
```
humitos, line 124: Why are you using a `»` at the end of the links? If it's to denote that it's an external link, I think there is some consensus about using the "square with the arrow". Like this one: https://fontawesome.com/icons/arrow-up-right-from-square?s=solid

docs/user/science.rst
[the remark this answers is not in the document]
agjohnson: Agreed! I'm all for being careful about terminology. We have a few terms that we use commonly that have significance internally, but might not be clear language outside internal discussions.

agjohnson:
@benjaoming +1, I think this is worth a test, and you're right that we might not know the reader's behaviors yet. We can always come back and toggle the dropdowns opened as a test, and see what that does for engagement with the page.

> The page has a typical design paradox

One pattern that I like to help avoid this paradox is to use introductory content at the top of the page to guide the user. For documentation pages, this might be the `contents` directive, or more helpful, a definition list with a brief description of each heading and why it matters to the reader. Like here:

https://docs.readthedocs.io/en/stable/build-customization.html

Then we can get as wordy as we want in the individual sections, as the user has both a way to scan what they are looking for and scroll sequentially.

I'm just speaking generally here, I'm not suggesting any more changes to the page.

ericholscher (approved):
I didn't read the full backlog of comments around structure, but it seems like folks are mostly happy with the content. I think this page is a great improvement over what we have. I also have some thoughts and ideas for improving the layout over time, but I'd say we should just go ahead and merge this and work on some future improvements in another PR. 

Hopefully we can ship some additions to our new marketing pages for this work, since I think some of the complexity here is around trying to do too much with a documentation page. Great pages are written with multiple rounds of iteration and feedback, and I think we'll start to get some feedback when we ship it 👍

docs/user/science.rst
benjaoming: THANKS!! :heart: I was actually contemplating exactly what convention to use here. Will see if this can be used.

humitos:
🚢

PR #9487, opened 2022-08-10
Use Read the Docs action v1

humitos:
Read the Docs repository was renamed from `readthedocs/readthedocs-preview` to `readthedocs/actions/`. Now, the `preview` action is under `readthedocs/actions/preview` and is tagged as `v1`

----
:books: Documentation preview :books:: https://docs--9487.org.readthedocs.build/en/9487/


stsewd (approved):
We should add links for the dev docs too now that it's possible.

PR #9509, opened 2022-08-17
Preview: escape characters

humitos:
Avoid the GH Action to break when running due to unescaped characters.


stsewd (approved):
We would need to revert this once this has been fixed in the action.

PR #9513, opened 2022-08-17
API v3: added support for tags in API

SyedMa3:
## Details
Closes issue #7354 

## Changes

Added support for tags to use in API. 
Added `'default_version', 'default_branch'` fields in `ProjectCreateSerializerBase` for user's convenience.

---

- User's documentation (`docs`): https://docs--9513.org.readthedocs.build/en/9513/

- Developer's documentation (`dev`): https://dev--9513.org.readthedocs.build/en/9513/


SyedMa3:
@humitos  I can't understand the failed tests. Can you please elaborate them? Thanks.

SyedMa3:
@humitos 
The test `FAILED readthedocs/api/v3/tests/test_projects.py::ProjectsEndpointTests::test_import_project_with_extra_fields` is not compatible with adding `default_version` field in project creation API. Can you please look into this?

And, can you also please explain the failed checks.
Thanks.

SyedMa3:
@humitos @stsewd Gentle reminder ping!

readthedocs/api/v3/serializers.py
```diff
@@ -472,6 +474,11 @@ class Meta:
             'programming_language',
             'repository',
             'homepage',
+            'tags',
+
+            # Advanced Settings -> General Settings
+            'default_version',
+            'default_branch',
```
humitos, line 481: Is this related to this PR? It seems that it's adding support to change other fields besides the `tags`

humitos (changes requested):
Thanks! I think this PR is good. Before merging it, I think there are some things that would be good to include on it:

- [ ] Update the API documentation including the `tags` fields where required
- [ ] Add or update test cases to match these changes
- [ ] Return the `tags` for the project on get/listing endpoints in case they are not already returning them (probably need to use the `TaggitSerializerField` there as well)
- [x] Remove new fields added that are not related to tags (eg. `default_`*)
- [x] Run the pre-commit linter on this PR to blackify the code

Again, thanks for your contribution and hoping to move this PR forward 💪🏼

readthedocs/api/v3/serializers.py
SyedMa3: It was done because you said ` I'm fine adding support to change most (or all) the fields at this point.`  And adding other fields while creating project didn't make sense.

humitos: Yes. I think that's fine, but we can take that work in a separate PR 👍

SyedMa3:
@humitos  
Can you please help me in the documentation part? I am not able to set it up.

When I do 
```bash
 $ cd docs
 $ make livehtml
```
it gives an error:
```
sphinx-autobuild --port 4444 -b html -d _build/user/doctrees   . _build/user/html
make: sphinx-autobuild: No such file or directory
make: *** [Makefile:67: livehtml] Error 127
```

What am I doing wrong here?
I have followed the steps from here (https://dev.readthedocs.io/en/latest/docs.html).

humitos:
@SyedMa3 It seems that you don't have the virtualenv enabled and that's why it does not find the `sphinx-autobuild` command.

humitos:
The PR looks good! The only thing that seems missing to me is to update/add a test case for the feature you are adding here. Would you like to take a look at that?

SyedMa3:
> The PR looks good! The only thing that seems missing to me is to update/add a test case for the feature you are adding here. Would you like to take a look at that?

My bad! I am new to "testing". But i'll look into it and do it.

SyedMa3:
@humitos 
I am encountering a problem in asserting the tags. The solution I implemented is, converting the `project.tags.names()` into a list. But the order of elements is not always same, since `project.tags.names()` is a QuerySet. So the fix for that is sorting. But is this fix optimal?

humitos:
@SyedMa3 can you use there the same queryset returned by the serializer? I'd assume they are using `order_by=` so it's always the same response.

I took a look at the code and it seems it supports an `order_by` attribute when defining the field: https://github.com/jazzband/django-taggit/blob/master/taggit/serializers.py#L48. We could use that everywhere and keep consistency all over across our code and responses as well.

benjaoming:
ping @SyedMa3 - are you still available for this PR? It seems a lot of good work has gone in to it and it's really close!

SyedMa3:
> ping @SyedMa3 - are you still available for this PR? It seems a lot of good work has gone in to it and it's really close!

Hi @benjaoming , sorry I had uni exams, so I was focusing on that and forgot about this. I will restart working on this. Thanks for the reminder!

benjaoming:
@SyedMa3 hope they went well! welcome back :)

SyedMa3:
@humitos @benjaoming I have made changes, and the tests have passed. Can you look into it and ping me if there should be any changes.

docs/user/api/v3.rst
```diff
@@ -413,6 +417,10 @@ Project update
             "language": "ja",
             "programming_language": "py",
             "homepage": "https://readthedocs.org/",
+            "tags" : [
+                "new tag",
+                "updated tag"
+            ]
```
benjaoming, line 423: How does "updated tag" work? It sounds like it updates an existing tag? But isn't it more that this set of tags entirely overwrites any existing tags?

benjaoming (approved):
It seems that test cases were added and everything is looking great :+1: 

I have just a question about the example in the documentation. Maybe it would be nice to clarify somewhere what it means to add the `tags` field, and if omitting it will delete all tags or just not update them?

docs/user/api/v3.rst
SyedMa3: Oh my bad! The nomenclature I used was ambiguous.
For `PUT`,  omitting them will not change them, but sending a list of tags will replace the old tags.
Should I change the names in the test files?

benjaoming: Thanks for clarifying!

For the tests, I think it seems okay. As long as the tests verifies that the behavior is what is intended.

But for the documentation, it would be nice if it's easy to understand what's going on. You could add it in a slightly creative way:

```
            "tags" : [
                "adding a list here",
                "will replace",
                "all existing tags",
                "with these tags"
            ]
```

But otherwise, I think it's also good if you can find a place in the documentation text to explicitly state this behavior.

SyedMa3: I have made a few changes.

And the docs looks like this.
image

benjaoming: This looks good, yes. It seems more clear now how this part works.

docs/user/api/v3.rst
```diff
@@ -422,6 +432,9 @@ Project update
 
         }
 
+    Adding ``tags`` will replace existing tags with the new list,
+    and if omitted won't change the tags.
```
humitos, line 436:
```suggestion
    .. note::

        Adding ``tags`` will replace existing tags with the new list,
        and if omitted won't change the tags.
```

humitos (approved):

benjaoming (approved):
Thanks for addressing the reviews. This looks good :+1:

SyedMa3:
@humitos @benjaoming Thank you for your guidance. I am just getting started in open source contribution and this helped me a lot.

benjaoming:
@SyedMa3 what a great start!! Really appreciate that you took the time to get around all the angles with respect to testing and documentation updates :+1:

PR #9514, opened 2022-08-17
Telemetry: collect user's requirements

humitos:
We are saving `packages.pip.user` that are "top level packages" --which does not
mean anything too useful for our purposes.

The original idea was to have `packages.pip.all` that list all the packages
installed in the virtual environment and `packages.pip.user` that list all the
packages installed because the user specified them in the requirements file.

This PR parses all requirements files specified by the user and saves packages'
names and versions into `packages.pip.user`. These ones are pretty useful to
know what projects are specifically defining/pinning a package.

---

- User's documentation (`docs`): https://docs--9514.org.readthedocs.build/en/9514/

- Developer's documentation (`dev`): https://dev--9514.org.readthedocs.build/en/9514/

### DIFF
--- a/readthedocs/telemetry/collectors.py
+++ b/readthedocs/telemetry/collectors.py
@@ -155,14 +155,16 @@ class BuildDataCollector:
                     ).serialize()
                     dependencies = df.get("dependencies", [])
                     for requirement in dependencies:
-                        name = requirement.get("name").lower()
+                        name = requirement.get("name", "").lower()
+                        if not name:
+                            continue
 
                         # If the user defines a specific version in the
                         # requirements file, we save it Otherwise, we don't
                         # because we don't know which version will be
                         # installed.
                         version = "undefined"
-                        specs = str(requirement.get("specs"))
+                        specs = str(requirement.get("specs", ""))
                         if specs:
                             if specs.startswith("=="):
                                 version = specs.replace("==", "", 1)
--- a/readthedocs/telemetry/collectors.py
+++ b/readthedocs/telemetry/collectors.py
@@ -58,6 +58,12 @@ class BuildDataCollector:
         Data that can be extracted from the database (project/organization)
         isn't collected here.
         """
+
+        # NOTE: we could run each command inside a try/except block to have a
+        # more granular protection and be able to save data from those commands
+        # that didn't fail. Otherwise, if one command fails, all the data for
+        # this Build is lost.
+
         data = {}
         data["config"] = {"user": self.config.source_config}
         data["os"] = self._get_operating_system()
--- a/readthedocs/telemetry/collectors.py
+++ b/readthedocs/telemetry/collectors.py
@@ -129,7 +129,19 @@ class BuildDataCollector:
         return []
 
     def _get_user_pip_packages(self):
-        """Get all the packages to be installed defined by the user."""
+        """
+        Get all the packages to be installed defined by the user.
+
+        It parses all the requirements files specified in the config file by
+        the user (python.install.requirements) using ``dparse`` --a 3rd party
+        package.
+
+        If the version of the package is explicit (==) it saves that particular
+        version. Otherwise, if it's not defined, it saves ``undefined`` and if
+        it's a non deterministic operation (like >=, <= or ~=) it saves
+        ``unknown`` in the version.
+
+        """
         results = []
         # pylint: disable=too-many-nested-blocks
         for install in self.config.python.install:
--- a/readthedocs/telemetry/collectors.py
+++ b/readthedocs/telemetry/collectors.py
@@ -2,7 +2,10 @@
 
 import json
 
+import dparse
 import structlog
+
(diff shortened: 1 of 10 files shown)


readthedocs/telemetry/collectors.py
```diff
@@ -151,11 +193,21 @@ def _get_pip_packages(self, include_all=False):
...
+            "python",
+            "-m",
+            "pip",
+            "list",
+            "--pre",
+            "--local",
+            "--format",
+            "json",
+            "--not-required",
+        ]
```
ericholscher, line 206: Seems like it would be nicer to just do a single string split on spaces?

readthedocs/telemetry/collectors.py
```diff
@@ -121,9 +128,44 @@ def _get_all_conda_packages(self):
...
+        Get all the packages to be installed defined by the user.
         """
-        Get all the packages installed by the user using pip.
+        results = []
+        for install in self.config.python.install:
+            if isinstance(install, PythonInstallRequirements):
+                if install.requirements:
+                    cmd = ["cat", install.requirements]
+                    code, stdout, _ = self.run(*cmd, cwd=self.checkout_path)
+                    df = dparse.parse(
```
ericholscher, line 141: I've never heard of dparse. Probably would be useful to add a comment explaining what this is doing, or make the docstring more complete?

readthedocs/telemetry/collectors.py
```diff
@@ -62,8 +69,8 @@ def collect(self):
         )
         data["packages"] = {
             "pip": {
-                "user": self._get_pip_packages(include_all=False),
-                "all": self._get_pip_packages(include_all=True),
+                "user": self._get_user_pip_packages(),
+                "all": self._get_all_pip_packages(),
```
ericholscher, line 73: This feels really brittle. I'd love to have a bit more protection here against exceptions, given that we're running a ton of code here, interacting with third-party modules and the filesystem, which feels highly likely to break or time out..

ericholscher (approved):
This looks like more useful data, but I do worry about how much logic we're adding here, and not having any protection against bugs, timeouts, or exceptions.

readthedocs/telemetry/collectors.py
humitos: The whole code is ran inside a try/except block from outside because of this concern specifically: https://github.com/readthedocs/readthedocs.org/blob/100bf6c08e16c2917272b8b1e4ff816c38616dda/readthedocs/projects/tasks/builds.py#L694-L706

So, I think we are already protected about this problem.

readthedocs/telemetry/collectors.py
humitos: I don't have a strong preference on this. This chunk of code was just black-ified in this PR. However, we are using a list for all the other commands in multiple files, so I think I will keep it like that just for consistency.

readthedocs/telemetry/collectors.py
humitos: I updated the docstring explaining a little more what's the content that's saved and where it comes from.

readthedocs/telemetry/collectors.py
ericholscher: Yea, it just means that if one thing breaks, all the data is lost, but I guess that's probably fine?

humitos: Oh, I see. Yes, this is how it has been working so far. We could wrap each command on a try except block to always save as much data as we can. However, we haven't seen this problem so far. I will add a note in the code.

readthedocs/telemetry/collectors.py
```diff
@@ -121,9 +134,60 @@ def _get_all_conda_packages(self):
...
+        Get all the packages to be installed defined by the user.
+
+        It parses all the requirements files specified in the config file by
+        the user (python.install.requirements) using ``dparse`` --a 3rd party
+        package.
+
+        If the version of the package is explicit (==) it saves that particular
+        version. Otherwise, if it's not defined, it saves ``undefined`` and if
+        it's a non deterministic operation (like >=, <= or ~=) it saves
+        ``unknown`` in the version.
```
benjaoming, line 148: Can we have another status like "private" for private repositories? I don't fully understand the intention of collecting the data so I'll leave the suggestion here :)

humitos: > Can we have another status like "private" for private repositories? 

I'm not sure to understand what "private" would mean here and how we would detect that. Can you expand?

> I don't fully understand the intention of collecting the data so I'll leave the suggestion here :)

This data is useful to answer questions like: "How many projects are using Sphinx 5.0.0?" and be able to differentiate that from "How many projects are _pinning_ Sphinx 5.0.0?"

benjaoming: Using readthedocs.com one can build private git repos. I'm wondering if we can mark certain build data as private, rather than collecting it.

PR #9537, opened 2022-08-23
Dependencies: pin deploy dependencies

humitos:
We should not use the `requirements.txt` file from `readthedocs-ops` and use
`requirements/deploy.in` from this repository instead. This way, we can keep
track of all dependencies via pip-tools and avoid creating conflicting
environments.

---

- User's documentation (`docs`): https://docs--9537.org.readthedocs.build/en/9537/

- Developer's documentation (`dev`): https://dev--9537.org.readthedocs.build/en/9537/

### DIFF
--- a/requirements/deploy.in
+++ b/requirements/deploy.in
@@ -10,3 +10,9 @@ django-redis-cache
 pillow
 
 structlog-sentry
+
+# We should unpin this dependency.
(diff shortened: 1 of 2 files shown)


humitos:
Merging this as it's already deployed and it's working.

PR #9549, opened 2022-08-25
Build: cancel old builds 

humitos:
> NOTE: I split the work in different commits to make it easier to review.

This PR implements a simple logic to cancel old running builds when a new
build for the same project/version arrives:

1. look for running builds for the same project/version
2. if there are any, it cancels them all one by one via Celery's revoke method
3. trigger a new build for the current commit received

Note that this new feature is behind a feature flag (`CANCEL_OLD_BUILDS`) for
now so we can start testing it on some projects that have shown their interest
on this feature.

The current behavior for `DEDUPLICATE_BUILDS` will be replaced by this new logic
in the future. However, it was not removed in this commit since it's still
useful for projects that won't be using the new feature flag yet.

Closes https://github.com/readthedocs/readthedocs.org/issues/8961

---

- User's documentation (`docs`): https://docs--9549.org.readthedocs.build/en/9549/

- Developer's documentation (`dev`): https://dev--9549.org.readthedocs.build/en/9549/

### DIFF
--- a/readthedocs/builds/tests/test_trigger_build.py
+++ b/readthedocs/builds/tests/test_trigger_build.py
@@ -1,0 +1,95 @@
+from unittest import mock
+
+import django_dynamic_fixture as fixture
+import pytest
+
+from readthedocs.builds.constants import (
+    BUILD_FINAL_STATES,
+    BUILD_STATE_BUILDING,
+    BUILD_STATE_CANCELLED,
+    BUILD_STATE_CLONING,
+    BUILD_STATE_FINISHED,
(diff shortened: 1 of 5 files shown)


humitos:
I've been experiencing with this locally and the experience as a user is great 💯 .

- I trigger a build for `latest`
- it starts running
- I do some changes on my code
- I trigger another build for `latest`
- the previous build got cancelled
- the new build with the new code starts building 🥳 

I don't have to wait for the old build to finish and show me outdated documentation 💪🏼

stsewd (approved):

Issue #9562, opened 2022-08-30
Build: "automagic" builder is broken for some Python versions


humitos:
Read the Docs executes some code in the build process on behalf the user to facilitate the on boarding process. This means that you can just put some `.rst` or `.md` files in your repository and Read the Docs will create a nice documentation site for you without requiring any type of configuration.

This is a nice feature for new users without too much knowledge on Sphinx or MkDocs. However, it has bitten _us_ and also _users_ due that sometimes there are unexpected errors that are hard to debug due to a disparity between Read the Docs and a local build for the same exact commit.

That said, **this "automagic" builder is currently broken** 😞 . If you follow the example that I mentioned (just put some `.rst` files in your repository without a `conf.py` file), Read the Docs will create one for you. Well, that `conf.py` generated by Read the Docs contains a problematic line:

```python
from recommonmark.parser import CommonMarkParser
```

that fails with the following error:

```
Traceback (most recent call last):
  File "/home/docs/checkouts/readthedocs.org/user_builds/test-builds/envs/no-conf-py/lib/python3.9/site-packages/sphinx/config.py", line 347, in eval_config_file
    exec(code, namespace)
  File "/home/docs/checkouts/readthedocs.org/user_builds/test-builds/checkouts/no-conf-py/docs/conf.py", line 7, in <module>
    from recommonmark.parser import CommonMarkParser
  File "/home/docs/checkouts/readthedocs.org/user_builds/test-builds/envs/no-conf-py/lib/python3.9/site-packages/recommonmark/parser.py", line 9, in <module>
    from commonmark import Parser
  File "/home/docs/checkouts/readthedocs.org/user_builds/test-builds/envs/no-conf-py/lib/python3.9/site-packages/commonmark/__init__.py", line 4, in <module>
    from commonmark.main import commonmark
  File "/home/docs/checkouts/readthedocs.org/user_builds/test-builds/envs/no-conf-py/lib/python3.9/site-packages/commonmark/main.py", line 14, in <module>
    from commonmark.blocks import Parser
  File "/home/docs/checkouts/readthedocs.org/user_builds/test-builds/envs/no-conf-py/lib/python3.9/site-packages/commonmark/blocks.py", line 5, in <module>
    from commonmark import common
  File "/home/docs/checkouts/readthedocs.org/user_builds/test-builds/envs/no-conf-py/lib/python3.9/site-packages/commonmark/common.py", line 14, in <module>
[...]

PR #9563, opened 2022-08-30
Build: upgrade `commonmark` to 0.9.1

humitos:
Currently, we are installing an older version of `commonmark` that has some
problems with newer versions of Python. Upgrading it to 0.9.1 seems to solve
this problem.

See #9562 for a detailed explanation of the problem

---

- User's documentation (`docs`): https://docs--9563.org.readthedocs.build/en/9563/

- Developer's documentation (`dev`): https://dev--9563.org.readthedocs.build/en/9563/

### DIFF
--- a/readthedocs/doc_builder/python_environments.py
+++ b/readthedocs/doc_builder/python_environments.py
@@ -182,7 +182,7 @@ class Virtualenv(PythonEnvironment):
             [
                 "mock==1.0.1",
                 "alabaster>=0.7,<0.8,!=0.7.5",
-                "commonmark==0.8.1",
+                "commonmark==0.9.1",
                 "recommonmark==0.5.0",
             ]
         )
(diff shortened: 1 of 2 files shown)


stsewd (commented):
We should update https://github.com/readthedocs/readthedocs.org/blob/fix-embed-api-cors-check/docs/user/build-default-versions.rst?plain=true#L67-L68

humitos:
What's the workflow for adding commits on a PR that it was already hotfixed into `rel`/`relcorp` now that we are using squash & merge? I'd assume that we will get conflicts when doing a new release next week? cc @stsewd @agjohnson

agjohnson:
We could also hotfix the new commits on this PR, or alternatively (and probably my suggestion) move on to a new PR -- that is, treat this PR as hotfix, merge it, and follow up with a PR that is not hotfix for the new commits.

PR #9620, opened 2022-09-24
Made some small changes to the MyST migration how-to

evildmp:
---

- User's documentation (`docs`): https://docs--9620.org.readthedocs.build/en/9620/

- Developer's documentation (`dev`): https://dev--9620.org.readthedocs.build/en/9620/

### DIFF
--- a/docs/user/guides/authors.rst
+++ b/docs/user/guides/authors.rst
@@ -16,4 +16,4 @@ and :doc:`/intro/getting-started-with-mkdocs`.
    cross-referencing-with-sphinx
    intersphinx
    jupyter
-   migrate-rest-myst
+   Migrate from rST to MyST <migrate-rest-myst>
--- a/docs/user/guides/migrate-rest-myst.md
+++ b/docs/user/guides/migrate-rest-myst.md
@@ -1,4 +1,8 @@
-# Migrating from reStructuredText to MyST Markdown
+# How to migrate from reStructuredText to MyST Markdown
+
+In this guide, you will find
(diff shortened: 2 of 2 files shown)


humitos (approved):
💯. Thanks 🙏 🧉

benjaoming:
Thanks so much for the wonderful approach given today, @evildmp :heart: 

### Notes and context to the change

@evildmp had a look at our documentation and made a local environment. He recommends that we start with random walks like this one to find the articles that are easiest to refactor towards Diátaxis.

The PR exemplifies:

* Giving a clear title (if it's a "How To" then make that clear in the title)
* Starting with a paragraph scoping out what the article is about (for the reader and author)
* Making a shorter title for the menu

Step-wise what we can then do:

1) We keep doing changes like this for easy, low-hanging fruits
2) We simply skip articles that aren't easy to process
3) At some point, it will _certainly_ be much more obvious and simple to make larger transformations

A side-note: We should likely very soon add sections in the menu for Reference and Explanation in order to make articles easier to move here. Once these sections exist, we can use the above approach. We shouldn't worry much that they may seem empty to begin with, if we have a quick processing of small changesets following the above recipe, they will quickly fill up.

benjaoming (approved):
I love this change, it's a small, simple and positive improvement that takes us in the right direction. If we keep doing this, the documentation will quickly be transformed to a better structure without heaps of planning, overwhelming reviews and merge conflicts.

:100: :+1: :rocket:

PR #9637, opened 2022-10-03
Dashboard: right-floating elements should go before inline text

benjaoming:
Wrapping and floating gets messed up with long branch names. And yes, we probably want to rewrite all this, use flexbox etc :)

Before:

image

After:

image

---

- User's documentation (`docs`): https://docs--9637.org.readthedocs.build/en/9637/

- Developer's documentation (`dev`): https://dev--9637.org.readthedocs.build/en/9637/


humitos:
This page has changed a lot in the new templates.

Screenshot_2023-03-07_11-08-54

If there is anything that should be done for this, it has to be done there: https://github.com/readthedocs/ext-theme

PR #9654, opened 2022-10-11
Do not assign html_theme_path

benjaoming:
I would like to have a second :brain: and :eyes: on this:

* Confirmation that defining this value will skip calls to `sphinx_rtd_theme.setup(app)` and thus the logic there is lost
* Confirmation that removing this configuration value doesn't break anything?

Longer story: https://github.com/readthedocs/sphinx_rtd_theme/issues/1355

For reference, here's the full `sphinx_rtd_theme.setup(app)` which we are not calling AFAICT.

https://github.com/readthedocs/sphinx_rtd_theme/blob/20607e85259f5a1885505807f932fc16cdad9449/sphinx_rtd_theme/__init__.py#L41-L72

---

- User's documentation (`docs`): https://docs--9654.org.readthedocs.build/en/9654/

- Developer's documentation (`dev`): https://dev--9654.org.readthedocs.build/en/9654/


benjaoming:
I asked on IRC #sphinx-doc and got this crystal-clear response:

> because `html_theme` only calls the entry point if the theme is not yet in `html_theme_path`

humitos (commented):
I want to mention some things here:

* I don't have a full understanding of this template `conf.py` file
* This file is created for all the projects that don't have a `conf.py` file in their repository. We usually refer to this as "Magic Builder"
* Touching this file will have an impact on _all those projects_

That said, is this change _required_? What are the benefits of it? I'd not jump into this deep research that involves some risk of breaking people's builds if it's not strongly required for a particular reason 😄

benjaoming:
Touching that file can certainly have an impact if assumptions are incorrect. At the same time, it would also be weird if we didn't fix things in that file that are already there because of wrong assumptions. I'm waiting for someone else to analyze this... as mentioned in the description, the entire `sphinx_rtd_theme.setup(app)` is _most likely_ not called.

I think it's a bit of a messy situation to not know if `setup(app)` is called or skipped. For instance, it means that we cannot rely on the theme's jQuery injection being launched from its own `setup(app)`.

For the theme's `setup(app)` to be called, it's required that current users of RTD would explicitly unset the variable, `html_theme_path=[]`.

benjaoming:
Another note on this: There seems to be an instruction for `sphinx_rtd_theme` to add it as an extension and also import the module in `conf.py`.

See: https://sphinx-rtd-theme.readthedocs.io/en/1.0.0/installing.html#installation

The instructions mention translations as a reason to include it in `extensions`. Let's investigate if that's redundant as well.

benjaoming:
On a local build, I tested that the theme is correctly enabled with a configuration that sets `html_theme="sphinx_rtd_theme"` on Sphinx 1.6.7 with sphinx-rtd-theme 0.5.1

The configuration did not add `sphinx_rtd_theme` to `extensions` in `conf.py` (the workaround).

benjaoming:
Tested on a local RTD server that it builds [tutorial-template](https://github.com/readthedocs/tutorial-template) -- this project has neither .readthedocs.yaml nor requirements.txt. Its conf.py is minimal.

* Python 3: Gets Sphinx 5.3.0 (because it's Python 3.7 on the default image). Build works :heavy_check_mark: 
* Python 2: Gets Sphinx 1.8.6 (the last version to support Python 2.7. Build works (needed to adjust the template project itself) and the result looks fine, all template assets are there :heavy_check_mark:

humitos:
@benjaoming I understand that we need this PR for Sphinx 6.x to work properly with our own theme because we depend on calling `sphinx_rtd_theme.setup(app)` to inject jQuery, right?

My concerns raised in https://github.com/readthedocs/readthedocs.org/pull/9654#pullrequestreview-1143967818 are still valid. If we want to move forward with these changes, we should make them in a progressive way, starting by making them under a feature flag.

Also, in case this change _only affect people with Sphinx >6.x_, we can add that restriction as well. My point here is to reduce the risk/impact of something that _looks like a simple change_ but affect all the builds for all the projects.

benjaoming:
> @benjaoming I understand that we need this PR for Sphinx 6.x to work properly with our own theme because we depend on calling sphinx_rtd_theme.setup(app) to inject jQuery, right?

That is the most urgent part for sure!

We might have unknown or unreported issues with other themes that depend on calling `setup(app)`. There have been suggestions to use `setup(app)` to add Jinja context variables which broke things. This originally happened here: https://github.com/readthedocs/sphinx_rtd_theme/issues/1355

> Also, in case this change only affect people with Sphinx >6.x, we can add that restriction as well. My point here is to reduce the risk/impact of something that looks like a simple change but affect all the builds for all the projects.

Nothing about this is simple, in fact it has already taken a lot of time :) It'd be great to remove this bit-by-bit with a feature flag. I'd like to end up having it completely removed, though.

Could launch with a `DONT_OVERWRITE_HTML_THEME_PATH` feature flag:

* Enabled on all Sphinx 6+ builds
* Enabled on all future projects
* Deprecation notice*)


## Deprecation notice *)

In order to figure out what is being deprecated and give users the right notice, we need to figure out why `html_theme_path` was necessary. What environment needs it? Why was it added?

agjohnson:
@benjaoming and I chatted briefly on this one. The big thing is that this issue probably should block a sphinx_rtd_theme 1.2.0 release, as anyone using Sphinx 6 will immediately hit this issue. The effect is JS silently breaking, so our flyout will just stop working -- very easy to miss.

A couple more notes:

> I don't have a full understanding of this template conf.py file

Me neither :grimacing:

+1 on being careful here. I think this logic was added/removed because Sphinx included rtd_theme as a dependency for a few releases -- without looking, 1.3-2.0?

Gating this logic change on Sphinx version seems wise :+1: That might be enough for me, but I'm not opposed to a feature flag too.

I'll add: we might want to just ensure `sphinx_rtd_theme`  is in `extensions` instead. I don't know if it's enough to test if `sphinx_rtd_theme` can be imported, but that could be an option. I'm guessing we'd have to inspect `extensions` after the configuration is loaded and ensure `sphinx_rtd_theme` was loaded as an extension.

Would that make us feel better about this change?

I'm not sure exactly what this looks like, but it would be more than relying on Sphinx to do the right thing. Without some additional logic, my guess is there would still be some Sphinx versions that can't find `sphinx_rtd_theme` without the `html_theme_path` addition.

agjohnson:
> In order to figure out what is being deprecated and give users the right notice, we need to figure out why html_theme_path was necessary

My memory is fuzzy here, but for Sphinx releases where sphinx_rtd_theme was not a dependency, Sphinx did not know how to find the theme path. This was likely before we suggested executing the theme as an extension, and so the logic in our theme extension code wouldn't have been used:

https://github.com/readthedocs/sphinx_rtd_theme/blob/master/sphinx_rtd_theme/__init__.py#L62-L63

humitos:
@agjohnson 

> I'll add: we might want to just ensure `sphinx_rtd_theme` is in `extensions` instead. I don't know if it's enough to test if `sphinx_rtd_theme` can be imported, but that could be an option. I'm guessing we'd have to inspect `extensions` after the configuration is loaded and ensure `sphinx_rtd_theme` was loaded as an extension.

This confuses me a little. We are not currently doing this check, so this will be a new approach, right?

I'm just proposing to keep the same code we have now for all the projects _except_ they have a feature flag enabled or they are using Sphinx >= 6:

```diff
diff --git a/readthedocs/doc_builder/templates/doc_builder/conf.py.tmpl b/readthedocs/doc_builder/templates/doc_builder/conf.py.tmpl
index 05bd3f0e0..fe3ba6734 100644
--- a/readthedocs/doc_builder/templates/doc_builder/conf.py.tmpl
+++ b/readthedocs/doc_builder/templates/doc_builder/conf.py.tmpl
@@ -58,10 +58,10 @@ using_rtd_theme = (
     ) or 'html_theme' not in globals()
 )
 if using_rtd_theme:
-    theme = importlib.import_module('{{ html_theme_import }}')
     html_theme = '{{ html_theme }}'
     html_style = None
     html_theme_options = {}
+    if ({{ project.has_feature(Feature.OVERWRITE_HTML_THEME_PATH) }} and not sphinx.version_info >= (6, 0, 0)) or sphinx.version_info < (6, 0, 0):
     if 'html_theme_path' in globals():
         html_theme_path.append(theme.get_html_theme_path())
     else:
```

`OVERWRITE_HTML_THEME_PATH` will be default `True` for all past projects. This way, I think we are not changing any current behavior _unless_ the project is using Sphinx >= 6; which is what we want.

benjaoming:
> This confuses me a little. We are not currently doing this check, so this will be a new approach, right?

Yes - while it might work, I also find it hacky and potentially having other side-effects. Ideally, we take an approach that can give us _less_ headache in the future.

I would prefer what you suggest @humitos , essentially gating the Sphinx version -- it's much more clear what is happening in this case. It also acts as a parameter -- as our knowledge of the issue is more firm, we can lower it to include Sphinx 5.

agjohnson:
> This confuses me a little. We are not currently doing this check, so this will be a new approach, right?

Correct. Instead of implicitly addressing the Spinx 6 behavior change, we'd explicitly address it. That is, removing `html_theme_path` tricks Sphinx into executing our theme as an extension, I'm suggesting instructing Sphinx to do this directly instead of relying on side effects. We can still gate on Sphinx version or whatever we're comfortable with.

Removing this logic when using Sphinx 6 is definitely fine for now, but I do think we'll want a longer term solution here. I'm doubtful we can get away with removing this logic for all Sphinx versions.

So, I'd lean towards using Sphinx version as the conditional here I suppose. The problem is that we likely can't remove `html_theme_path` for older releases of Sphinx, or we can't expect the same implicit behavior from older releases. So we will likely will have long-lived conditionals against the Sphinx version in this code. This is the scenario where we've been trying to get away from using feature flags.

benjaoming:
> Instead of implicitly addressing the Spinx 6 behavior change, we'd explicitly address it. That is, removing `html_theme_path` tricks Sphinx into executing our theme as an extension

This is not a trick, though, this is the documented and expected behavior. See: https://www.sphinx-doc.org/en/master/development/theming.html

All themes have their `setup(app)` called, except in cases when `html_theme_path` is defined. The trick of adding the theme in `extension` is a workaround for this behavior but in no way conventional.

agjohnson:
> This is not a trick, though, this is the documented and expected behavior. See: https://www.sphinx-doc.org/en/master/development/theming.html

What exactly are you referring to in that doc? The fact that Sphinx does not load a theme as an extension if `html_theme_path` is defined does not seem communicated in the Sphinx docs.

> I think we can remove html_theme_path for a lot of Sphinx versions

This is what I'm very doubtful of. We're still only guessing why this option was added in the first place, so this sounds optimistic.

Also, if I'm understanding your issue correctly, if _the project_ sets `html_theme_path`, our theme will still not be loaded as an extension -- so JS will silently break still. Is that correct?

benjaoming:
> What exactly are you referring to in that doc? The fact that Sphinx does not load a theme as an extension if `html_theme_path` is defined does not seem communicated in the Sphinx docs.

[No, it's not.](https://github.com/readthedocs/readthedocs.org/pull/9654#issuecomment-1275275561) But why even do that in the first place?

benjaoming:
> Also, if I'm understanding your issue correctly, if the project sets `html_theme_path`, our theme will still not be loaded as an extension -- so JS will silently break still. Is that correct?

Yes, but since that is a configuration that isn't found in any docs, I wonder why anyone would do that :)

benjaoming:
> This is what I'm very doubtful of. We're still only guessing why this option was added in the first place, so this sounds optimistic.

[See my comment here.](https://github.com/readthedocs/readthedocs.org/pull/9654#issuecomment-1396795016) - I'm not talking about eliminating this at once, but to be on the right path.

agjohnson:
> Yes, but since that is a configuration that isn't found in any docs, I wonder why anyone would do that :)

Well, because we told them to :smile: Using `html_theme_path` was the prescribed method of using the theme for quite some time, in both our docs and the theme docs. I don't know if we have this in telemetry or if the figure there would be polluted by our overrides setting this for users, but any long standing project would have potential to use this old configuration style.

benjaoming:
[I remember simplifying those instructions slightly](https://github.com/readthedocs/sphinx_rtd_theme/pull/1365), they actually still contain adding the theme in `extensions` so in some way, we have been working actively around our self-created `html_theme_path` issue :smile: [Your comment here](https://github.com/readthedocs/sphinx_rtd_theme/pull/1365/files#r1004869239) was really spot on @agjohnson :100:

benjaoming:
But I think we can save these "risk assessments" for later changes. Our changes for now can be considered safe and then we'll figure out how to "expand" them.

agjohnson:
Correct, we've had a few conflicting versions of this. Either way, we can expect users to be using all of them. It's quite likely that a project started on Sphinx 1.8 with old installation directions is still using that same code. There isn't any reason to update this once it's working.

Because this involves a silent failure of our flyout, and not an error with Sphinx, user docs will break in a very unfortunate way that users/we won't detect. It's very unfortunate that Sphinx has conditional behavior here. We likely should either ensure the theme is executed as an extension for the user, or throw a hard error when the theme is not executed correctly so the user can fix their configuration.

I still agree we can proceed with this hack for now, but this approach is not going to solve the problem for us.

humitos:
@agjohnson 

> Because this involves a silent failure of our flyout, and not an error with Sphinx, user docs will break in a very unfortunate way that users/we won't detect

I'd suggest checking for the combination of _incorrect_ config setup and raise and exception in that case, making the build to hard fail. It's not perfect, but it's better than silently fail. This idea together with a feature flag will help us to deploy these changes gradually. What do you think?

agjohnson:
> I'd suggest checking for the combination of incorrect config setup and raise and exception in that case

Yup, I'm leaning that way now too :+1:

I've lost patience with trying to fix the jQuery removal at the Sphinx level. We'll address this more thoroughly by 1) removing jQuery from our flyout and 2) loading all flyout dependencies we do need from the flyout code, not Sphinx/theme code. We still need something immediately though.

> This idea together with a feature flag will help us to deploy these changes gradually. What do you think?

Yup, that's what I'm describing too :+1:  We can proceed with the hack here for now, but I wanted to note that we'll be back on this bug rather quickly, as projects upgrade to Sphinx 6 with old instantiation methods.

benjaoming:
Oh yeah one other thing I keep forgetting.... There is a rather deceptive variable name `use_rtd_theme` in the mix. If a theme defines `html_theme = "sphinx_rtd_theme"`, `use_rtd_theme` remains False. So this whole behavior should already be isolated to projects that do not even define a theme.

humitos:
@agjohnson 

> We can proceed with the hack here for now, but I wanted to note that we'll be back on this bug rather quickly, as projects upgrade to Sphinx 6 with old instantiation methods.

If you can describe what are those installation methods, we can probably use Telemetry to track them to eventually communicate with the authors and start deprecating old patterns. We've hit this problem about "avoid moving forward" or "moving forward is _so complex_" because supporting pretty old stuffs.

_(note that I don't have a full understanding of the whole situation here --theme, conf.py, installation methods, etc-- but I think we've bitten by this situation multiple times)_

benjaoming:
@agjohnson in case you haven't noticed, this PR has been updated.

docs/user/feature-flags.rst
```diff
@@ -18,6 +18,15 @@ This helps users to pin dependencies on conda and to improve build time.
 
 ``DONT_OVERWRITE_SPHINX_CONTEXT``: :featureflags:`DONT_OVERWRITE_SPHINX_CONTEXT`
 
+``SKIP_SPHINX_HTML_THEME_PATH``: :featureflags:`SKIP_SPHINX_HTML_THEME_PATH`
```
agjohnson, line 21: We stopped advertising feature flags like this. If a feature flag is in this list, it should be a configuration option instead, something the user can opt into without requiring support time.

readthedocs/doc_builder/templates/doc_builder/conf.py.tmpl
```diff
@@ -58,15 +58,37 @@ using_rtd_theme = (
     ) or 'html_theme' not in globals()
 )
 if using_rtd_theme:
-    theme = importlib.import_module('{{ html_theme_import }}')
     html_theme = '{{ html_theme }}'
     html_style = None
     html_theme_options = {}
+
+# Legacy behavior for html_theme_path:
```
agjohnson, line 65: This is more internal commentary that user commentary. It's probably better as a template comment block so we're not outputting debug output to users. This will be output in builds on the cat conf.py step.

readthedocs/projects/models.py
```diff
@@ -1859,6 +1860,12 @@ def add_features(sender, **kwargs):
                 'Do not overwrite context vars in conf.py with Read the Docs context',
             ),
         ),
+        (
+            SKIP_SPHINX_HTML_THEME_PATH,
+            _(
+                "On Sphinx < 6, do not define html_theme_path",
```
agjohnson, line 1866: Less passive voice here makes items a little easier to scan. The features list is a bit unwieldy when you need to interact with it. Also, typo here I assume.

```suggestion
                "Do not define html_theme_path on Sphinx >= 6.0",
```

docs/user/feature-flags.rst
benjaoming: That sounds good. By "configuration option", do you mean a variable that's defined in `conf.py` and then we check `globals()` ?

agjohnson:
> If you can describe what are those installation methods, we can probably use Telemetry to track them 

The case is anyone setting the `html_theme_path` option. It would be great to have a list of these values from projects. I suspect the ratio will be quiet high, and we'll have to face a decision to hack in a fix or raise errors on a fair amount of builds or something similar.

readthedocs/projects/models.py
benjaoming: In [the other comment](https://github.com/readthedocs/readthedocs.org/pull/9654#discussion_r1085629807), you indicate that a user-defined configuration value is better than a feature flag?

benjaoming:
> The case is anyone setting the `html_theme_path` option.

+ Using Sphinx unpinned in Python 3.8+ or Sphinx>=6.

agjohnson:
> Using Sphinx unpinned in Python 3.8+ or Sphinx>=6

Why limit the amount of telemetry data we're collecting to this?

docs/user/feature-flags.rst
agjohnson: Not `conf.py`, but in our own configuration file, `.readthedocs.yaml`. We have a number of issues open about deprecation of feature flags, but this particular doc was a focus of most of our removals.

I'm not suggesting this fix is implemented as a configuration option though. Just that feature flags should be reserved for internal use only.

readthedocs/projects/models.py
agjohnson: Sorry -- to clarify: this fix falls solidly into feature flag usage. That means that at some point core team will expand our testing sample in include more projects, in order to get more feedback. This won't be an option that we expect users can ask us to enable for their project.

benjaoming:
> Why limit the amount of telemetry data we're collecting to this?

Not meant to limit it - it just seems to be the most important segment, since those are the intersections where we know things will be broken.

humitos:
@agjohnson 

> With the addition of tracking `html_theme_path` values in our telemetry, we can correlate Python/Sphinx versions if it's really necessary.

I'm not sure how to track this, since we are already defining it. So, I'm not sure how we can differentiate if **it's the user who declared the variable** or **if it's _us_**. If you have an idea for this, I'm happy to expand our telemetry to save it somehow.

agjohnson:
I don't want to continue sidetracking this issue, we can discuss metrics later. This bug will be one we need to monitor anyways. The remaining conversations on metrics is a later discussion either way, not a today thing. 

> I'm not sure how to track this, since we are already defining it

I wasn't sure either. If we're getting similar values after the configuration is loaded in Sphinx, we could store a copy of the original value of `html_theme_path`, before we alter the option in our conf.py template. I'm not familiar with the telemetry code here though, so :shrug: 

> Not meant to limit it - it just seems to be the most important segment

We can just query this segment in Metabase, we have the data to correlate versions. We're better off with more data collection around builds in general.

docs/user/feature-flags.rst
benjaoming: Right, got it, yeah `.readthedocs.yaml` is overkill, I don't expect anyone to really need the feature flag, it's more like a safety net so behavior remains fully configurable. Hopefully this feature flag will be unnecessary soon, as the number of builds for _very old_ is in decline.

readthedocs/doc_builder/templates/doc_builder/conf.py.tmpl
```diff
@@ -58,15 +58,39 @@ using_rtd_theme = (
...
+# Discussion of issue:
+# https://github.com/readthedocs/readthedocs.org/pull/9654
+#
+# We need to isolate the old legacy behavior because it causes Sphinx
+# to skip calling the theme's setup(app) so its initialization never happens.
+{% endcomment %}
+# This following legacy behavior will gradually be sliced out until its deprecated and removed.
+# Skipped for Sphinx 6+
+# Skipped by internal Feature flag SKIP_SPHINX_HTML_THEME_PATH
+# Skipped by all new projects since SKIP_SPHINX_HTML_THEME_PATH's introduction (jan 2023)
```
benjaoming, line 81: @agjohnson I believe we should also do this, just want to make sure that you have seen it. New projects won't follow strange old install guides, and they are likely on later versions of Sphinx.

benjaoming:
Aligned with @agjohnson about merging this now :+1:

agjohnson (approved):
Updates look good to go! We have next steps planned already too, we should get an issue up and in our roadmap for these.

PR #9656, opened 2022-10-11
Adds more basic info to the default 404 page

benjaoming:
Addresses #2551 with a very simple improvement. The 404 page context is still unaware of what project it belongs to, that should follow in a different update.

TODO:

* [x] Verify that this is covered by a test

image

---

- User's documentation (`docs`): https://docs--9656.org.readthedocs.build/en/9656/

- Developer's documentation (`dev`): https://dev--9656.org.readthedocs.build/en/9656/

### DIFF
--- a/readthedocs/templates/404.html
+++ b/readthedocs/templates/404.html
@@ -16,7 +16,7 @@
 {% block language-select-form %}{% endblock %}
 
 {% block content %}
-<p>{% blocktrans context "404 page" %}This is a 404 error page. The link you have followed or the URL that you entered did not work. It might be that the documentation has been updated.{% endblocktrans %}</p>
+<p>{% blocktrans context "404 page" %}This is a 404 error page. The link you have followed or the URL that you entered does not exist. It might be that the documentation has been updated, or this page has been removed.{% endblocktrans %}</p>
 
 <a href="/">{% blocktrans with site_name=request.get_host|default:_("this site") context "404 page" %}Back to the front page of {{ site_name }}{% endblocktrans %}</a>
 
--- a/readthedocs/templates/404.html
+++ b/readthedocs/templates/404.html
@@ -15,7 +15,11 @@
 {# Hide the language select form so we don't set a CSRF cookie #}
 {% block language-select-form %}{% endblock %}
 
+<p>{% blocktrans context "404 page" %}This is a 404 page. The link you have followed or the URL that you entered did not work. It might be that the documentation has been updated.{% endblocktrans %}</p>
+
 {% block content %}
+<a href="/">{% blocktrans context "404 page" %}Back to the main page{% endblocktrans %}</a>
+
 <pre style="line-height: 1.25; white-space: pre;">
 
         \          SORRY            /
--- a/readthedocs/templates/404.html
+++ b/readthedocs/templates/404.html
@@ -15,7 +15,7 @@
 {# Hide the language select form so we don't set a CSRF cookie #}
 {% block language-select-form %}{% endblock %}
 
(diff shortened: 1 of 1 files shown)


humitos:
I think the message should clearly communicate this page belong to a documentation project that it's hosted by Read the Docs but it's not a 404 page on Read the Docs itself. To me, the "Back to the main page" is confusing in that regard. I would expect that I'd arrive at the main page of the documentation that I'm looking for. Why not include the information about the project on this PR as well? I think that will work to remove the ambiguity in the message.

benjaoming:
I'm writing a separate PR about adding context to the 404 page.

See the PR description:

> The 404 page context is still unaware of what project it belongs to, that should follow in a different update.

ericholscher:
I do agree with Manuel that the `Back to main page` could use some work, and could be confusing. Perhaps `Go back to the top of this site` or something, but any copy we write is going to be weird, because we don't know 100% where the link will go.

benjaoming:
Text updated! I added `request.get_host`.

readthedocs/templates/404.html
```diff
@@ -16,6 +16,10 @@
 {% block language-select-form %}{% endblock %}
 
 {% block content %}
+<p>{% blocktrans context "404 page" %}This is a 404 error page. The link you have followed or the URL that you entered did not work. It might be that the documentation has been updated.{% endblocktrans %}</p>
+
+<a href="/">{% blocktrans with site_name=request.get_host|default:_("this site") context "404 page" %}Back to the front page of {{ site_name }}{% endblocktrans %}</a>
```
ericholscher, line 21: That's a nice addition 👍

readthedocs/templates/404.html
```diff
@@ -16,6 +16,10 @@
 {% block language-select-form %}{% endblock %}
 
 {% block content %}
+<p>{% blocktrans context "404 page" %}This is a 404 error page. The link you have followed or the URL that you entered did not work. It might be that the documentation has been updated.{% endblocktrans %}</p>
```
ericholscher, line 19:
```suggestion
<p>{% blocktrans context "404 page" %}This is a 404 error page. The link you have followed or the URL that you entered does not exist. It might be that the documentation has been updated, or this page has been removed.{% endblocktrans %}</p>
```

We should be explicit that 404 means doesn't exist, not "did not work".

ericholscher (approved):
This is a good, simple addition.

readthedocs/templates/404.html
benjaoming: Good change!

humitos (approved):
I don't see too much value on this PR, but I'm fine merging it if you consider. I think that #9657 is the way to go and will replace this PR.

Another extra step related to "improve 404 page" work, IMO, would be to have _ completely different_ 404 pages for "application 404 page" and "documentation 404 page". That way, it will be a lot more clear if that 404 page the user reached is related to Read the Docs itself or if it's part of the project they wanted to read documentation about.

benjaoming:
@humitos I do see some value here: Instead of having an almost arrogant 404 page that doesn't tell the user anything, we now have a friendlier version that on a more fundamental level says what might have gone wrong.

I agree that it can be improved through more context, this PR was meant as the fastest way to improve the current situation + a bit of exercise before a larger changeset. Contextualizing the current 404 page or making 2 versions definitely makes sense, but I'm sure it's also more complex and requires a bit more ideas to be tossed around :)

benjaoming:
I like the ascii art, but the word "yet" at the very end is misleading, perhaps we should remove that in future updates?

PR #9657, opened 2022-10-12
Contextualize 404 page

benjaoming:
Update: This PR has gone through manual testing and is now ready for review. Would like to sync up on this work with you @stsewd - I really really think you are right about wanting to refactor this :100: 

Work flow:

 * Base branch was changed to https://github.com/readthedocs/readthedocs.org/pull/10074
 * This PR is extends work from #9656

I don't think that we should go to extreme extends to give suggestions -- but we could given a `project` context give **basic** suggestions (drafted below and also in the implementation)

### TODO

* [x] Version not found
* [x] Subproject page not found
* [x] Handle exception types with `instanceof` or similar
* [x] Remove too much debug logging
* [x] For "project page not found", we could try to suggest searching because `{% url "elastic_project_search" project.slug %}` affords us this option across Documentation project types?
* [x] Remember to remove `errors/404/include_search.html` in cases where searching isn't possible
* [x] Display correct `request.path`, not the one with `_proxito_404_` in the path...
* [x] Add exception for "language not found"

### Suggestions for "project page not found"

* Inform people that project "xyz" does not contain this page, possibly because it got updated.
* Look at the URL to understand if it's the `/latest/` branch and tell the user that projects are updated over time, so a link to "latest" may change and that's a good thing. They can look for the page in older versions if the latest version doesn't contain a similar page.
* ~Give people a search form for the project? That should be easy, right? `<form method="GET" action="/en/{{ default_branch }}/search.html?>`~ This needs a lot more context, i.e. "Is this a Sphinx project?" and "Which language are we looking at?". I don't find it feasible to implement.
* Update: Instead of in-doc search, we can give people our ES powered search using the project scope :100: :tada: 
* A suggestion box for documentation developers
  * Use custom 404 pages: https://docs.readthedocs.io/en/latest/hosting.html#custom-not-found-404-pages
  * Use redirects: https://docs.readthedocs.io/en/stable/user-defined-redirects.html

### For "project not found"

* Tell people that the URL is wrong
* Project might have been deleted

### For "subproject page not found"

* Tell people that the URL is wrong
* Tell people that the page in the subproject did not exist
* That the page may have failed to build
* That a translation may not be available (not ready)
* A suggestion box for documentation developers

### For "readthedocs.org page not found"

* The link you followed is broken
* We will look into this if it's a technical error

References #2551

---

- User's documentation (`docs`): https://docs--9657.org.readthedocs.build/en/9657/

- Developer's documentation (`dev`): https://dev--9657.org.readthedocs.build/en/9657/


benjaoming:
Note on the current state of the PR:

I'm still not looking for a code review :) But a general comment on the approach of contextualizing the `Http404` exception is most welcome. I had the idea today and once I started implementing it, I saw a really nice pattern emerging.

readthedocs/templates/404.html
```diff
@@ -16,6 +16,46 @@
 {% block language-select-form %}{% endblock %}
 
 {% block content %}
+
+<h1>{% trans "404 - page not found" %}</h1>
+
+{% if project %}
```
humitos, line 22: Since we are raising custom exceptions now, I'd suggest to raise a _specific exception for each particular case_ (e.g. `ProxitoHttp404ProjectNotFound`, `ProxitoHttp404VersionNotFound`, etc) instead one generic `ProxitoHttp404` and remove all this logic from the template.

That way, we have all the logic in Python and we can render different 404 templates with only the content they need. It also simplifies the reading of this code, since we don't have to follow Python and Django template logic to understand what's the final response.

benjaoming: I agree - that will mean the contextualization can also scale quite a lot if needed... in terms of both user-facing messages and the view logic such as notifications, automated redirects, error logging etc.

I'll have to _not_ respect the `template_name` keyword argument for the 404 view. But I think that's fine, we can still use template inheritance from one main `404.html` base template.

benjaoming: Very cool suggestion, I've added a note to also handle "version not found". The templates are now being split up and new exception classes are being added.

readthedocs/proxito/exceptions.py
```diff
@@ -1,7 +1,54 @@
...
+Because an Http404 can be raised anywhere in the code,
+we want to add more context for targeted error handling and user-facing communication
+"""
 from django.http.response import Http404
 
 
 class ProxitoHttp404(Http404):
+    """
+    General error class for proxity 404s
+    """
```
humitos, line 18: I think these will fail the linter:

* no final dot
* it keeps in one line

readthedocs/proxito/exceptions.py
```diff
@@ -1,7 +1,54 @@
...
+        self.project = project
+        self.subproject_slug = subproject_slug
+        super().__init__(message)
+
+
+class ProxitoProjectPageHttp404(ProxitoHttp404):
+    """
+    Raised if a page inside an existing project was not found
+
+    Note: The containing project can be both a project or a subproject inside another project
```
humitos, line 49: I'm not sure to understand this note. In this case, I'd assume that the project is "the parent project" and the subproject is the one where the page was not found.

readthedocs/templates/404.html
```diff
@@ -1,84 +1,2 @@
...
-           ]         NNN         [
-           ]         N/"         [
-           ]         N H         [
-          /          N            \
-         /           q,            \
-        /                           \
-</pre>
-{% endblock %}
+{% extends "errors/404/base.html" %}
+{# this is a deprecated template, will be removed. Please use one of the targeted 404s in templates/404/ #}
```
humitos, line 2: Why we don't just remove it?

readthedocs/templates/errors/404/base.html
```diff
@@ -0,0 +1,50 @@
...
+{% endblock %}
+
+{% if project %}
+
+{% elif "_proxito_404_" in request.path %}
+
+{% else %}
+
+{% endif %}
+
```
humitos, line 24:
```suggestion
```

readthedocs/templates/errors/404/community.html
```diff
@@ -0,0 +1,12 @@
+{% extends "errors/404/base.html" %}
+{% load i18n %}
+
+{# something on readthedocs.org does not exist - this 404 is not intended for projects #}
```
humitos, line 4:
```suggestion
{# This page on the application site does not exist. This 404 is not intended for user hosted projects. #}
```

readthedocs/templates/errors/404/include_tips.html
```diff
@@ -0,0 +1,11 @@
+{% load i18n %}
+
+<h3>{% trans "Are you the project owner?" %}</h3>
+
+<p>{% trans "Here are some tips to address 404 errors:" %}</p>
+
+<ul>
+  <li>{% trans "Use custom 404 pages:" %} <a href="https://docs.readthedocs.io/en/latest/hosting.html#custom-not-found-404-pages">{% trans "Read more" %}»</a></li>
+  <li>{% trans "Create redirects when you move contents:" %} <a href="https://docs.readthedocs.io/en/stable/user-defined-redirects.html">{% trans "Read more" %}»</a></li>
+  <li>{% trans "Use rediraffe to manage redirects in Sphinx:" %} <a href="https://pypi.org/project/sphinxext-rediraffe/">{% trans "Read more" %}»</a></li>
```
humitos, line 10: I'm not sure we want to recommend rediraffe here since it uses Javascript/meta tags for redirects.

readthedocs/templates/errors/404/no_subproject.html
```diff
@@ -0,0 +1,22 @@
+{% extends "errors/404/base.html" %}
```
humitos, line 1: Isn't this template the same as `no_project_page.html`? If so, we should extend that one. Otherwise, change the content to reflect that this is a subproject; in case it gives the user more context, but I don't think so...

humitos (commented):
I love where are we going with this. 404 pages will be way better with these changes 💯

readthedocs/core/views/__init__.py
```diff
@@ -48,17 +48,25 @@ def get_context_data(self, **kwargs):
         return context
 
 
-def server_error_404(request, template_name="404.html", exception=None):
+def server_error_404(
+    request, template_name="errors/404/community.html", exception=None
```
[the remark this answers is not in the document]
benjaoming: I didn't get far with this thought, but I had the idea that this would be configured in Read the Docs for Business through it's URL conf. I'll get back to it :+1:

readthedocs/settings/base.py
```diff
@@ -73,7 +73,7 @@ def SHOW_DEBUG_TOOLBAR(self):
     PRODUCTION_DOMAIN = 'readthedocs.org'
     PUBLIC_DOMAIN = None
     PUBLIC_DOMAIN_USES_HTTPS = False
-    USE_SUBDOMAIN = False
+    USE_SUBDOMAIN = True
```
benjaoming, line 76: Note to self: This should not be checked in

humitos: Why do you need this for testing? If you are using Docker, this is already set as `True`

https://github.com/readthedocs/readthedocs.org/blob/502b9b35f19197c452b4e73d70f9b78c129d4098/readthedocs/settings/docker_compose.py#L16

readthedocs/templates/404.html
benjaoming: I'll have a look and see. But I prefer removing it in a later change set, simply to reduce the potential impact of this change.

readthedocs/templates/errors/404/include_tips.html
benjaoming: Yes, I wanted to discuss precisely these suggestions earlier :100: 

In your experience, what are some commonly missed 404 options?

humitos: I'm not sure to understand the question 😅 

BTW, the other 2 suggestions are 💯 to me: custom 404 and our own redirects.

This was a project we will want to recommend in the future once it gets stable: https://github.com/nextstrain/readthedocs-cli. It allows you to sync redirects on RTD with a YAML file.

benjaoming: Totally agree that Rediraffe is too hacky a solution.

Exciting to see where nexstrain's readthedocs-cli is going :+1:

readthedocs/templates/404.html
benjaoming: Two gotchas of why I don't feel fully comfortable on removing this:

* `404.html` is the default 404 template for Django's default 404 view.
* I'm not sure about other projects that might somehow anticipate this file to exist, for instance I haven't looked in readthedocs-corporate.

It seems rather harmless to leave it like this.

readthedocs/core/views/__init__.py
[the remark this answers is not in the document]
benjaoming: It's gone :+1: I don't think we need to work with any alternative templates or settings here.

benjaoming:
404 pages now contain a search!

Project not found:
image

Page not found:
image

stsewd:
I'm big +1 on implementing this, and with our new proxito implementation, this should be easier to support. So, I'd +1 on implementing this on the new implementation instead of the old one. With the new implementation, we probably won't be needing `_get_project_data_from_request`, and instead all the handling would be done at / similar at:

https://github.com/readthedocs/readthedocs.org/blob/db3b96822b4b64eff89d236bc3f7ea14648df4e2/readthedocs/proxito/views/serve.py#L299-L309

And here:
https://github.com/readthedocs/readthedocs.org/blob/c0c3f02606ff0f2100444ed1791c30338cd062f8/readthedocs/proxito/middleware.py#L216-L218

which you are already using, probably we could implement it just in this place to start? That doesn't require waiting for the new proxito implementation.

benjaoming:
I can definitely do that @stsewd! Thanks for looking. I'll try to exclude logic from `_get_project_data_from_request` and do the handling outside.

Just noticed that I've been using `request.path` where I shouldn't so will fix that, too...

image

agjohnson:
Are these pages going to be used on commercial too, and if so, has that been tested?

benjaoming:
> Are these pages going to be used on commercial too, and if so, has that been tested?

No, I would want to do that, too. Thanks for the reminder. I initially was wondering if the 404s needed to be adapted for RTD4B, and I guess that adding the search functionality definitely won't work.

benjaoming:
Overriding the include `errors/404/include_search.html` with an empty file should be enough though.

agjohnson:
> and I guess that adding the search functionality definitely won't work

Aye, this is my main concern. The best option here would probably would have been to point back to the in-doc search, as we don't need to worry about dashboard access or removing readers from the documentation experience and placing them in an unfamiliar UI. We can override the dashboard search on commercial for now though.

humitos:
I just wanted to mention that I'd not block this PR/work on https://github.com/readthedocs/readthedocs.org/issues/10061 because that re-structure of Proxito is a lot bigger than this PR and it may take some time to take off.

benjaoming:
@humitos this PR has two outstanding small tasks and I'm syncing up with @stsewd today on finalizing  [re:his comment](https://github.com/readthedocs/readthedocs.org/pull/9657#issuecomment-1442252067):

> probably we could implement it just in this place to start? That doesn't require waiting for the new proxito implementation.

readthedocs/settings/base.py
```diff
@@ -46,7 +46,7 @@ class CommunityBaseSettings(Settings):
     DEFAULT_AUTO_FIELD = 'django.db.models.AutoField'
 
     # Debug settings
-    DEBUG = True
```
humitos, line 49: I think you commit this here by mistake 😄

benjaoming:
I'm changing this PR to target `use-proxito-v2-in-404-handler` instead. The idea is to avoid doing the same Q&A twice, or to have to adapt this code for some compromise. By targeting `use-proxito-v2-in-404-handler`, I will also be able to test the refactoring Proxito refactoring work and help to review it.

benjaoming:
@stsewd it'd be great if you can have a look here, as this is definitely something that you have been working with recently :)

benjaoming:
I'm yet to test it with readthedocs-corporate, I would like to know if that's something you've been doing @stsewd with other proxito work and if maybe we could have a combined look at https://github.com/readthedocs/readthedocs.org/pull/10074 on readthedocs-corporate?

stsewd:
Hmm, looks like GitHub didn't change the base branch to main and just closed this PR, can't re-open this PR :/

benjaoming:
No worries, I tried this before :) Will fix!

benjaoming:
When a base branch is deleted, the PR is closed and you cannot change the base branch anymore. I'm not sure why it's happening automatically in some cases, maybe it's like @humitos says because it's depending on the PR's source repo.

In any case, it has an easy workaround:

1) Recreate the old base branch (just click "restore" in the PR where it got deleted)
2) Navigate back to the closed PR, reopen it
3) Change to the new base branch
4) Go back and delete the old base branch again

benjaoming:
Description of this PR has been updated.

readthedocs/templates/errors/404/include_error_message.html
```diff
@@ -0,0 +1,16 @@
+{% load i18n %}
+<section>
+  <p>
+    {% blocktrans trimmed %}
+      You are browsing the documentation of <code>{{ project_slug }}</code>.
```
benjaoming, line 5:
```suggestion
      You are browsing the documentation of the project <code>{{ project_slug }}</code>.
```

benjaoming:
@stsewd pinging you in case you hadn't seen that this is ready for review :+1:

readthedocs/core/views/__init__.py
```diff
@@ -53,6 +61,52 @@ def get_context_data(self, **kwargs):
...
+        # These attributes are not guaranteed.
+        context.update(
+            {
+                "project": getattr(exception, "project", None),
+                "project_slug": getattr(exception, "project_slug", None),
+                "subproject_slug": getattr(exception, "subproject_slug", None),
+                "version_slug": getattr(exception, "version_slug", None),
+                "language_slug": getattr(exception, "language_slug", None),
+                "path_not_found": getattr(exception, "proxito_path", None),
+            }
```
stsewd, line 85: We could probably fill the context in the code below where you check for each type of exception, that woul avoid using `getattr`.

readthedocs/proxito/exceptions.py
```diff
@@ -0,0 +1,110 @@
...
+    """General error class for proxity 404s."""
+
+
+class ProxitoProjectHttp404(ProxitoHttp404):
+
+    """Raised if a project was not found."""
+
+    def __init__(self, message, project_slug=None, proxito_path=None):
+        self.project_slug = project_slug
+        self.proxito_path = proxito_path
```
stsewd, line 28: I think it should be fine to use just `path` for all these

readthedocs/proxito/exceptions.py
```diff
@@ -0,0 +1,110 @@
...
+
+    """Raised if a page inside an existing project was not found."""
+
+    def __init__(
+        self,
+        message,
+        project_slug=None,
+        proxito_path=None,
+        project=None,
+        subproject_slug=None,
```
stsewd, line 59: We can probably just pass the `project` and `subproject` as objects instead of passing their slugs separately

readthedocs/proxito/views/serve.py
```diff
@@ -46,12 +53,20 @@ def get(self, request, subproject_slug=None, filename=""):
...
 
         # Use the project from the domain, or use the subproject slug.
         if subproject_slug:
-            project = get_object_or_404(
-                project.subprojects, alias=subproject_slug
-            ).child
+            try:
+                project = project.subprojects.get(alias=subproject_slug).child
+            except ProjectRelationship.DoesNotExist:
+                raise ProxitoSubProjectHttp404(
```
stsewd, line 63: This 404 will go directly to the internal nginx redirect (`ServeError404Base`), so it won't work as expected.

stsewd (commented):
A couple of notes:

- We should implement this on the new proxito implemenation, no the old one
  
https://github.com/readthedocs/readthedocs.org/blob/830b899c91ab9f0ad8f8448947461d1eeafa23a1/readthedocs/proxito/views/serve.py#L327

https://github.com/readthedocs/readthedocs.org/blob/830b899c91ab9f0ad8f8448947461d1eeafa23a1/readthedocs/proxito/views/serve.py#L722

- 404 that are generated outside the `ServeError404Base` view won't work as expected, as all those responses will go through the internal nginx redirect, and back to the `ServeError404Base` view, where it will evaluate the type of 404 again. Raising a 404 directly from outside the `ServeError404Base` can be achieved with https://github.com/readthedocs/readthedocs.org/issues/10061. 404 from the middleware work too.
- We are missing lots of tests :)

readthedocs/proxito/exceptions.py
benjaoming: Can you clarify this? I'm not sure what objects you mean to pass instead?

benjaoming:
Thanks for the review @stsewd :heart_eyes: 

Regarding tests, I'm not writing them before after-the-fact.. as you can see, there's a number of design/architecture questions open, so it's a bit hard to add tests into the mix.

readthedocs/core/views/__init__.py
benjaoming: It would. But it will also cause a lot of extra code lines.

readthedocs/proxito/exceptions.py
benjaoming: I didn't like the ambiguity around this since we for instance distinguish between a resolved proxito path and the path of the URL. Don't want these two mixed up, i.e. `_proxito_not_found_/path/to/file`

readthedocs/proxito/views/serve.py
benjaoming: It replaces a call to `get_object_or_404`, so wouldn't this be exactly the behavior that we want? I.e. replace a general 404 exception with a contextualized one?

readthedocs/settings/base.py
```diff
@@ -46,7 +46,7 @@ class CommunityBaseSettings(Settings):
...
+    DEBUG = False
```
benjaoming, line 49: LOL, I keep doing this. But I also said to open a PR to make this a general setting.

readthedocs/templates/errors/404/no_language.html
```diff
@@ -0,0 +1,10 @@
+{% extends "errors/404/base.html" %}
+{% load i18n %}
+
+{# project exists but not in the requested language #}
+
+{% block 404_error_message %}
+  {% include "errors/404/include_error_message.html" with not_found_subject="language" %}
```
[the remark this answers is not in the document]
benjaoming: Totally agree! The `not_found_subject` itself is also not translatable. The approach was nice to avoid duplicating the same text in each template while developing the solution. But ultimately, it's not nice.

readthedocs/urls.py
```diff
@@ -13,12 +13,14 @@
...
     SupportView,
     do_not_track,
+    server_error_404,
     server_error_500,
 )
 from readthedocs.search.views import GlobalSearchView
 
 admin.autodiscover()
 
+handler404 = server_error_404
```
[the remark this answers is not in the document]
benjaoming: I have actually been confused about what to do wrt `ProjectDetailView` and `project.urls.public`

https://github.com/readthedocs/readthedocs.org/blob/7128a001f900b0236d441053543bf324c2c452af/readthedocs/projects/urls/public.py#L32-L36

I haven't read the implementation or pondered anything here, thinking that it has very little or no impact to add a nice 404 here.

The intention was that we can raise the exceptions from anywhere in the codebase. I also wondered about moving them out of proxito, but they are inspired by proxito internals, so I think they belong there :)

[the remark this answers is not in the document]
benjaoming: Curious if you have thoughts on this?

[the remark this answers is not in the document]
stsewd: I'm +1 on having nicer 404 pages on the main application, but if we do so, we probably want its own implementation separate from the proxito one, since proxito has restrictions that the main app doesn't, specially: URLs from the dashboard aren't present in proxito, we have access to the current user session in 404 from the main app. So I think it makes sense to have them separate.

readthedocs/core/views/__init__.py
stsewd: with getattr it is easy to misspell an attribute, and makes it harder to refactor attribute names, having it on each condition is also more clear what exception has what attributes.

readthedocs/proxito/exceptions.py
stsewd: Is there a case where we want the non-proxito path? I think we should be fine following the sama names from https://github.com/readthedocs/readthedocs.org/blob/main/readthedocs/proxito/README.rst#url-parts.

readthedocs/proxito/exceptions.py
stsewd: Here you are passing the `project` as an object, and `project_slug` separately, when you can just pass `project` and get the slug from there.

readthedocs/proxito/views/serve.py
stsewd: Yes, but the 404 will go to the internal nginx redirect, it won't hit your custom 404 handler that serves the specific page. And the internal redirect will hit the `ServeError404Base` view, so you won't see this specific exception there.

readthedocs/urls.py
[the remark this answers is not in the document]
benjaoming: Great! It sounds like this modular scoping will be much more clarified when moving 404 templates from the main app to proxito, as you already suggested!

Once all "proxito-scoped" code and template logic lives in proxito, the main application should be able to use or derive from it as is feasible?

readthedocs/proxito/exceptions.py
benjaoming: Maybe `doc_path` would be a better term than just `path`? I wanted to avoid mixing  up the resolved path with the HTTP path (I did that in the first versions of this implementation). In the very nice ascii diagram, things would get confusing if we introduced a URL prefix, i.e. `/en/version/page/` became `/docs/en/version/page/` ?

readthedocs/proxito/exceptions.py
benjaoming: Ah okay yeah, it was to have consistency with the other exception classes so they could have the same kwarg pattern. You can always have `project_slug`, but `project` is depending on whether a Project was resolved.

I'll have a look at if it's easy to minimize the kwargs for this one.

ericholscher:
Deprioritizing this because we have some higher-level goals. If we can scope this work down into something we can complete in ~4 hours, I'm fine moving forward on it, but don't want to spend much more than that to get something usable here.

stsewd:
+1 on scoping this down, since with our current architecture we can't cover all 404s from all views https://github.com/readthedocs/readthedocs.org/pull/9657#pullrequestreview-1340445046. I don't remember where I mentioned this, but the easiest would be to start with the 404 that happen at the middleware level

https://github.com/readthedocs/readthedocs.org/blob/f141a8f931ffe78d6aad90a4710b7ec566994dcd/readthedocs/proxito/middleware.py#L213-L213

But as always, we need to take into consideration that 404s execute our middleware twice.

readthedocs/proxito/views/decorators.py
```diff
@@ -20,14 +23,17 @@ def map_subproject_slug(view_func):
 
     @wraps(view_func)
     def inner_view(  # noqa
-            request, subproject=None, subproject_slug=None, *args, **kwargs
+        request, subproject=None, subproject_slug=None, *args, **kwargs
     ):
+        # Something not entirely clear is happening when unpacking args and kwargs
+        # so the project is fetched from kwarg dictionary.
+        project = kwargs["project"]
```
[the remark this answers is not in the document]
benjaoming: Removed :+1:

benjaoming:
Just need to revert a couple of changes on the old proxito implementation and then test everything on the new implementation.

In order to scope things down, I'm going to hold off with adding contextualized exceptions on cases that aren't easy to resolve.

readthedocs/templates/errors/404/include_search.html
```diff
@@ -0,0 +1,13 @@
+{% load i18n %}
+
+<h2>{% trans "Try searching?" %}</h2>
+
+<section>
+  {% if search_project %}
+    {% with "project:"|add:search_project as search_term %}
+      {% include "core/widesearchbar.html" with form_action="https://readthedocs.org/search/" term=search_term %}
```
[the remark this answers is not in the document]
benjaoming: Done here and below :+1:

readthedocs/proxito/views/serve.py
benjaoming: This was much easier to understand after reading the new implementation. It has a lot more clarity than the old implementation. Well done :+1: :+1:

readthedocs/proxito/exceptions.py
benjaoming: Fixed in the new exception classes :+1:

readthedocs/proxito/exceptions.py
benjaoming: This has also been simplified, it's now an argument that can be added to all contextualized Http404 exceptions, so it's just done once in the new unresolver method in `ServeError404Base`.

readthedocs/core/views/__init__.py
benjaoming: This is now handled with a pattern that just assumes all additional kwargs for a contextualized Http404 exception as a context variable for its template.

readthedocs/proxito/middleware.py
```diff
@@ -218,9 +218,16 @@ def process_request(self, request):  # noqa
...
                 status=400,
             )
-        except (InvalidSubdomainError, InvalidExternalDomainError):
+        except (InvalidSubdomainError, InvalidExternalDomainError) as exc:
             log.debug("Invalid project set on the subdomain.")
-            raise Http404
+            # We pass e.domain as 'project_slug' because we cannot unpack a slug from the
+            # domain at this stage. The domain can be a custom domain docs.foobar.tld,
+            # in which case 'docs' would be an incorrect slug.
+            raise ProxitoProjectHttp404(
```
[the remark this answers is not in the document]
benjaoming: I guess because we try to limit the scope, I'm not gonna look at it this time around, but will be easy to add if anyone bumps into a good use case :+1:

readthedocs/core/views/__init__.py
```diff
@@ -53,6 +61,52 @@ def get_context_data(self, **kwargs):
         return context
 
 
+def server_error_404(request, template_name="errors/404/base.html", exception=None):
```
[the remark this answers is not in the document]
benjaoming: Done :+1:

readthedocs/proxito/exceptions.py
```diff
@@ -0,0 +1,110 @@
...
+    ):
+        super().__init__(message, project_slug=project_slug, proxito_path=proxito_path)
+        self.project = project
+        self.subproject_slug = subproject_slug
+
+
+class ProxitoProjectPageHttp404(ProxitoProjectHttp404):
```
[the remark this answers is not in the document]
benjaoming: Done :+1:

benjaoming:
Nice - the Django test client seems to have an option `raise_view_exception` that can test exception handling: https://code.djangoproject.com/ticket/18707#comment:14

benjaoming:
@stsewd ready for review - I added a POC of using `raise_view_exception` without adding a new bunch of test cases, but I think it's nice.

readthedocs/templates/errors/404/include_search.html
```diff
@@ -0,0 +1,13 @@
+{% load i18n %}
+
+<h2>{% trans "Try searching?" %}</h2>
```
stsewd, line 3: This is including two titles

Screenshot 2023-04-19 at 11-59-19 404 Not Found Read the Docs

I think we should remove the "search the docs"

stsewd (commented):
Much simpler implementation :+1:

Left some general suggestions

readthedocs/proxito/views/serve.py
```diff
@@ -30,6 +30,12 @@
 from readthedocs.redirects.exceptions import InfiniteRedirectException
 from readthedocs.storage import build_media_storage
 
+from ..exceptions import (
```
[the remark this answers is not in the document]
benjaoming: Oh but not right below this change :see_no_evil:

readthedocs/proxito/exceptions.py
```diff
@@ -0,0 +1,100 @@
...
+        c.update(self.kwargs)
+        return c
+
+
+class DomainDNSHttp404(ContextualizedHttp404):
+
+    """Raised if a DNS record points to us but we do not know the domain."""
+
+    template_name = "errors/404/dns.html"
+    not_found_subject = _("domain matching DNS record")
```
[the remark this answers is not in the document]
benjaoming: it's a subject, though.. it's used _in a sentence_.

But I can use https://docs.djangoproject.com/en/4.2/ref/utils/#django.utils.translation.pgettext instead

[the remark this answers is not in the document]
benjaoming: Fixed!

readthedocs/proxito/middleware.py
```diff
@@ -213,20 +214,23 @@ def process_request(self, request):  # noqa
...
         except SuspiciousHostnameError as exc:
             log.warning("Weird variation on our hostname.", domain=exc.domain)
-            return render(
-                request,
-                "core/dns-404.html",
-                context={"host": exc.domain},
-                status=400,
+            # Raise a contextualized 404 that will be handled by proxito's 404 handler
+            raise DomainDNSHttp404(
+                http_status=400,
```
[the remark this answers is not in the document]
benjaoming: It's used with a 404 below.

readthedocs/proxito/middleware.py
```diff
@@ -213,20 +214,23 @@ def process_request(self, request):  # noqa
...
+                host=exc.domain,
```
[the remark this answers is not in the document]
benjaoming: Fixed in 344a845ecca4daa6a57f65ab68bb1dd6d93dbd97

readthedocs/proxito/views/serve.py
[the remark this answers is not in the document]
benjaoming: Fixed both

readthedocs/proxito/views/serve.py
```diff
@@ -761,21 +786,26 @@ def get_using_unresolver(self, request, path):
             filename = unresolved.filename
             lang_slug = project.language
             version_slug = version.slug
+            contextualized_404_class = ProxitoProjectFilenameHttp404
```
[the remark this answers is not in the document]
benjaoming: It's simpler to keep the current pattern for now, maybe this change can happen in a later refactor?

readthedocs/templates/errors/404/include_search.html
benjaoming: Fixed

benjaoming:
Thanks for the great and detailed review @stsewd :100: 

Will do some manual testing and get back :hourglass_flowing_sand:

readthedocs/proxito/exceptions.py
```diff
@@ -0,0 +1,100 @@
...
+    def get_context(self):
+        c = {
+            "not_found_subject": self.not_found_subject,
+            "path_not_found": self.path_not_found,
+        }
+        c.update(self.kwargs)
+        return c
+
+
+class DomainDNSHttp404(ContextualizedHttp404):
```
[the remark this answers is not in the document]
stsewd: It's easy to introduce bugs here, making a typo while passing the kwargs since they don't have any validation. I'd prefer if we are explicit about what the template needs.

benjaoming:
@stsewd changes work locally, the failure in "checks" is unrelated.

readthedocs/proxito/exceptions.py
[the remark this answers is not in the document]
benjaoming: I've added positional arguments where they are required :+1:

readthedocs/proxito/exceptions.py
```diff
@@ -0,0 +1,174 @@
...
+    This is currently used very broadly.
+    It indicates a number of reasons for the user.
+    """
+
+    template_name = "errors/404/no_project.html"
+    not_found_subject = pgettext_lazy(
+        "Names an object not found in a 404 error", "project"
+    )
+
+    def __init__(self, domain=None, **kwargs):
```
stsewd, line 77: Domain seems to always be passed here. This also doesn't require us to use `no domain` in the template, we should always have access to the domain.

```suggestion
    def __init__(self, domain, **kwargs):
```

readthedocs/templates/errors/404/include_error_message.html
```diff
@@ -0,0 +1,16 @@
+{% load i18n %}
+<section>
+  <p>
+    {% blocktrans trimmed with project_slug=project.slug|default:_("unknown slug") %}
```
stsewd, line 4: We should always assume to have a project/project slug here (all the templates that include this file should have it available)

readthedocs/proxito/exceptions.py
```diff
@@ -0,0 +1,174 @@
...
+class DomainDNSHttp404(ContextualizedHttp404):
+
+    """Raised if a DNS record points to us and we don't know the domain."""
+
+    template_name = "errors/404/dns.html"
+    not_found_subject = pgettext_lazy(
+        "Names an object not found in a 404 error", "matching DNS record"
+    )
+
+    def __init__(self, domain, **kwargs):
```
stsewd, line 52: Having named parameters is more clear, but allowing the kwargs still lets room for some errors (like passing `http_statu` instead of `http_status`).

But I'm okay moving forward with this for now.

readthedocs/proxito/exceptions.py
```diff
@@ -0,0 +1,174 @@
...
+    Subclasses may define their own template name,
+    HTTP status and object that was not found.
+
+    The contextualized exception is handled by proxito's 404 handler
+    """
+
+    template_name = "errors/404/base.html"
+    not_found_subject = pgettext_lazy(
+        "Names an object not found in a 404 error", "page"
+    )
```
stsewd, line 19: This is being used in the templates as

```
The {{ not_found_subject }} you are looking for at <code>{{ path_not_found }}</code> was not found.
```

Not sure if other languages require the word to be translated as a whole with the sentence, @agjohnson may have some input here.

stsewd (approved):
There is some information that will (should) always be available to templates, other than that I'm fine moving forward with this.

readthedocs/proxito/exceptions.py
benjaoming: The required arguments are now explicit through the `__init__` constructor, meaning that the pattern is hopefully clear for anyone that wants to add more context to exceptions and templates in the future :+1:

readthedocs/templates/errors/404/include_error_message.html
benjaoming: I removed the `|default` filter if that is what you meant here.

readthedocs/proxito/exceptions.py
benjaoming: Added better translator context guidelines in 5e9017e69781214296ad807e2f6aae4433e2a693

I can imagine several ways to improve things, also for translators - but I don't think this will lead to inaccuracy, now that there's also a pretty solid context.

I don't think it's proportional to put more efforts on the translation of this particular part, given that there hasn't been much effort in designing the error messages (yet).

benjaoming:
Super happy to move forwards on this, thanks for all the feedback @stsewd :+1: 

I hope that anyone encountering a 404 in the future will quickly pitch in with improvements to texts and suggestions.

The new contents haven't been worked much with in the PR (although we did add some significant helpful texts and functions, as mentioned in the PR description). But it's great that we have now introduced a pattern that enables better 404 error handling with the option of using a targeted context, rather than _unhelpfully_ broad messages.

PR #9660, opened 2022-10-13
Docs: Use current year instead of hard-coded 2010

benjaoming:
Writing "2010" in the Copyright notice gives a wrong impression that whatever documentation article is displayed hasn't been updated since. And many of them aren't even written in 2010, so it's also factually very wrong :)

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -13,6 +13,7 @@ for more information read https://sphinx-multiproject.readthedocs.io/.
 
 import os
 import sys
+from datetime import datetime
 
 import sphinx_rtd_theme
 from multiproject.utils import get_project
@@ -68,7 +69,7 @@ docset = get_project(multiproject_projects)
 templates_path = ["_templates"]
 
 master_doc = "index"
-copyright = "2010, Read the Docs, Inc & contributors"
+copyright = "{}, Read the Docs, Inc & contributors".format(datetime.now().year)
 version = "8.7.0"
 release = version
 exclude_patterns = ["_build"]
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -13,7 +13,6 @@ for more information read https://sphinx-multiproject.readthedocs.io/.
 
 import os
 import sys
-from datetime import datetime
 
 import sphinx_rtd_theme
 from multiproject.utils import get_project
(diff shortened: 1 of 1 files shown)


docs/conf.py
```diff
@@ -68,7 +69,7 @@
 templates_path = ["_templates"]
 
 master_doc = "index"
-copyright = "2010, Read the Docs, Inc & contributors"
+copyright = "{}, Read the Docs, Inc & contributors".format(datetime.now().year)
```
ericholscher, line 72: We should just not have a date here, instead.

benjaoming: It seems to be a Sphinx default to always hard-code a year, which I think goes wrong for most projects :) So it's great to get rid of in any case.

I like putting the current year there because it becomes the only indicator that what you (the reader) are looking at is "fresh". An alternative would be to put an exact date in the footer "this page was built on 2022-10-14", but that would add too much detail since it's strictly speaking not when the article was actually updated. Which is why I like just the year.

ericholscher: That is a different job than the copyright. I don't think we need a copyright at all in the footer, if that's the goal.

benjaoming: Yes, what I'm saying is that in lieu of a build timestamp, I would just put the current year here and keep the Copyright so the change becomes minimal. I'm sure there are more interesting stuff to put here... a copy-left statement or a fun sign-off. But then because the setting is called `copyright`, it feels a bit wrong to put that there :)

benjaoming: I'll empty the copyright statement and have another look to see if it makes sense to replace it with something now, or it's fine to leave it empty and find another addition for the footer in a later change.

humitos: I think the copyright note is fine as it is currently. If we want to introduce the "Last time the page was updated" we could use something like https://github.com/mgeier/sphinx-last-updated-by-git/ but that's different than the copyright, as Eric mentioned before.

benjaoming: I've removed the year now. I don't think that the copyright notice is fine saying "2010" because it risks making pages look out-dated.

But if there's too much friction around this hot fix, I'm happy to withdraw it so a better fix can be contributed at a later time.

ericholscher (approved):

PR #9665, opened 2022-10-17
Proposal for sphinxcontrib-jquery

benjaoming:
This is a proposal for a new Sphinx extension, as elaborated in the proposal itself.

Direct comments and suggestions are welcome through GitHub's code review.

A more general discussion or questions are welcome as comments to the Pull Request

**Rendered version**: https://dev--9665.org.readthedocs.build/en/9665/design/sphinx-jquery.html

Discussions about jQuery removal in Sphinx and extensions:

* https://github.com/sphinx-doc/sphinx/issues/7405
* https://github.com/sphinx-doc/sphinx/issues/10070
* https://github.com/readthedocs/sphinx-hoverxref/issues/160
* https://github.com/readthedocs/sphinx_rtd_theme/issues/1253
* https://github.com/pydata/pydata-sphinx-theme/issues/764

---

- User's documentation (`docs`): https://docs--9665.org.readthedocs.build/en/9665/

- Developer's documentation (`dev`): https://dev--9665.org.readthedocs.build/en/9665/

### DIFF
--- a/docs/dev/design/sphinx-jquery.rst
+++ b/docs/dev/design/sphinx-jquery.rst
@@ -1,13 +1,15 @@
-Design proposal for sphinxcontrib-query
-=======================================
+sphinxcontrib-jquery
+====================
 
-JQuery is being removed from Sphinx 6.0.0. We can expect 6.0.0 to ship in late 2023. See also:
+JQuery will be removed from Sphinx 6.0.0. We can expect 6.0.0 to ship in late 2023. See also:
 
 * https://github.com/sphinx-doc/sphinx/issues/7405
 * https://github.com/sphinx-doc/sphinx/issues/10070
 
-This is a "request for comments" for initiating a community-owned Sphinx extension
-that bundels jQuery.
+This is a "request for comments" for a community-owned Sphinx extension that bundles jQuery.
+
+Comment deadline: **November 1st, 2022**
+
 
 Overview
 --------
@@ -19,4 +21,88 @@ python package:
   ``sphinxcontrib.jquery``
 
 ownership:
-  community-owned, lives in https://github.com/sphinx-contrib/jquery
+  Read the Docs core team will implement the initial releases of an otherwise community-owned package that lives in https://github.com/sphinx-contrib/jquery
+
+functionality:
+  sphinxcontrib-jquery is a Sphinx extension that provides a simple mechanism for other Sphinx extensions and themes to ensure that jQuery is included into the HTML build outputs and loaded in the HTML DOM itself.
+  More specifically, the extension ensures that jQuery is loaded exactly once no matter how many themes and extensions that request to include jQuery nor the version of Sphinx.
+
+scope:
+  This extension assumes that it's enough to provide a single version of jQuery for all of its dependent extensions and themes.
+  As the name implies, this extension is built to handle jQuery only.
+  It's not a general asset manager and it's not looking to do dependency resolution of jQuery versions.
+
+Usage
+-----
+
+The primary users of this package are
+**a) Sphinx theme and extension developers** and
+**b) documentation project owners**.
+
+Theme and extension developers:
+
+  The following 2 steps need to be completed:
+
+  #. A Sphinx theme or extension should depend on the python package ``sphinxcontrib-jquery``.
+  #. Calling either ``sphinxcontrib.jquery.add_jquery`` or ``sphinxcontrib.jquery.add_jquery_if_sphinx_lt_6``.
+
+  In addition to this, we recommend extension and theme developers to log to ``console.error`` in case jQuery isn't found. The log message could for instance say:
+
+    "<package-name> depends on sphinxcontrib-jquery. Please ensure that <package-name>.setup(app) is called or add 'sphinxcontrib-jquery' to your conf.py extensions setting."
+
+
+Documentation project owners:
+
+  If you are depending on a theme or extension that did not itself address the removal of jQuery from Sphinx, you can patch up your project like this:
+
+  #. Add ``sphinxcontrib-jquery`` to your ``requirements.txt``.
+  #. Add ``sphinxcontrib.jquery`` to your ``extensions`` setting in ``conf.py``.
+
+
+``sphinxcontrib.jquery.add_jquery(app)``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Adds jQuery no matter what, but at most once.
+It's useful for themes and extensions that want jQuery added unconditionally or if they want to handle the conditions themselves.
+
+``sphinxcontrib.jquery.add_jquery_if_sphinx_lt_6(app)``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Adds jQuery if Sphinx is older than version 6, at most once.
+This is recommended for extensions and themes that maintain compatibility of Sphinx before and after version 6.
+
+
+Release
+-------
+
+All releases follow semantic versioning: ``<major>.<minor>.<patch>``
+
+* ``major``: Changes that break existing themes and extensions - we'll aim to only ever have ``1.x.y`` releases!
+* ``minor``: Changes that contain jQuery updates or substantial feature additions.
+* ``patch``: Bug fixes and added compatibility.
+
+For distributed packages with a longer lifecycle, we propose a loose pinning of the patch and minor versions to allow for future updates, which are likely to happen when jQuery receives security updates.
+Example: ``sphinxcontrib-jquery>=1.0.0,<2.0.0`` will include all compatible future releases, including jQuery upgrades.
+
+
+jQuery version and inclusion
+----------------------------
+
+Sphinx has kept relatively up to date with jQuery, and this package intends to follow.
+The most recently bundled jQuery version was v3.5.1 and only two releases have happened since: 3.6.0 and 3.6.1.
+The 3.6.0 release had a very small backwards incompatibility which illustrates how harmless these upgrades are for the general purpose Sphinx package.
+
+Therefore, we propose to start 1.0.0 at 3.5.1 (the currently shipped version) and subsequently release 3.6.1 as the first update of jQuery as 1.1.0.
+
+The bundled jQuery version will be NPM pre-minified and distributed together with the PyPI package.
+
+The minified jQuery JS file is ultimately included by calling `app.add_js_file <https://www.sphinx-doc.org/en/master/extdev/appapi.html#sphinx.application.Sphinx.add_js_file>`_, which is passed the following arguments:
+
+.. code:: python
+
+  app.add_js_file(
+      get_jquery_url_path(),
+      loading_method="defer",
+      priority=200,
+      integrity="sha256-{}".format(get_jquery_sha256_checksum())
+  )
--- a/docs/dev/design/sphinx-jquery.rst
+++ b/docs/dev/design/sphinx-jquery.rst
@@ -59,7 +59,7 @@ The following 2 steps need to be completed:
 
 In addition to this, we recommend extension and theme developers to log to the browser's ``console.error`` in case jQuery isn't found. The log message could for instance say::
 
-  if ($ == "undefined") console.error("<package-name> depends on sphinxcontrib-jquery. Please ensure that <package-name>.setup(app) is called or add 'sphinxcontrib-jquery' to your conf.py extensions setting.")
+  if (typeof $ == "undefined") console.error("<package-name> depends on sphinxcontrib-jquery. Please ensure that <package-name>.setup(app) is called or add 'sphinxcontrib-jquery' to your conf.py extensions setting.")
 
 
 Documentation project owners
@@ -67,7 +67,7 @@ Documentation project owners
 
 If you are depending on a theme or extension that did not itself address the removal of jQuery from Sphinx 6, you can patch up your project like this:
 
-#. Add ``sphinxcontrib-jquery`` to your ``requirements.txt``.
+#. Add ``sphinxcontrib-jquery`` to your installed dependencies.
 #. Add ``sphinxcontrib.jquery`` to your ``extensions`` setting in ``conf.py``.
 
 
--- a/docs/dev/design/sphinx-jquery.rst
+++ b/docs/dev/design/sphinx-jquery.rst
@@ -52,7 +52,7 @@ Theme and extension developers
 The following 2 steps need to be completed:
 
 #. A Sphinx theme or extension should depend on the python package ``sphinxcontrib-jquery``.
-#. Calling either ``sphinxcontrib.jquery.add_jquery`` or ``sphinxcontrib.jquery.add_jquery_if_sphinx_gte_6``.
+#. Calling either ``sphinxcontrib.jquery.add_jquery`` in the theme's/extension's ``setup(app)``.
 
 In addition to this, we recommend extension and theme developers to log to the browser's ``console.error`` in case jQuery isn't found. The log message could for instance say:
 
@@ -68,24 +68,18 @@ If you are depending on a theme or extension that did not itself address the rem
 #. Add ``sphinxcontrib.jquery`` to your ``extensions`` setting in ``conf.py``.
 
 
-``sphinxcontrib.jquery.add_jquery(app)``
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+``sphinxcontrib.jquery.add_jquery(app, force=False)``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-Adds jQuery no matter what, but at most once.
-It's useful for themes and extensions that want jQuery added unconditionally or if they want to handle the conditions themselves.
+Adds jQuery if Sphinx is from version 6 and up. Adds jQuery at most once.
+
+When setting ``force=True``, adds jQuery no matter what, but at most once. This is useful if you want to handle alternative conditions for adding jQuery.
+
+Adding ``sphinxcontrib.jquery`` to a documentation project's ``conf.extensions`` will call this function with ``force=False``.
 
 .. warning::
 
-  If you call this function, you most likely should also add ``Sphinx>=6`` to your theme's/extension's dependencies since versions before this already bundles jQuery!
-
-
-``sphinxcontrib.jquery.add_jquery_if_sphinx_gte_6(app)``
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-
-Adds jQuery if Sphinx is from version 6 and up. Adds jQuery at most once.
-This is recommended for extensions and themes that maintain compatibility of Sphinx before and after version 6.
-
-Adding ``sphinxcontrib.jquery`` to a documentation project's ``conf.extensions`` will call this function.
+  If you call this function with ``force=False``, you most likely should also add ``Sphinx>=6`` to your theme's/extension's dependencies since versions before this already bundles jQuery!
 
 
 Release
--- a/docs/dev/design/sphinx-jquery.rst
+++ b/docs/dev/design/sphinx-jquery.rst
@@ -89,19 +89,6 @@ When setting ``jquery_force_enable=True``, jQuery is added no matter the Sphinx 
   If you set ``jquery_force_enable=True``, you most likely should also add ``Sphinx>=6`` to your theme's/extension's dependencies since versions before this already bundles jQuery!
 
 
-Release
--------
-
-All releases follow semantic versioning: ``<major>.<minor>.<patch>``
-
-* ``major``: Changes that break existing themes and extensions - we'll aim to only ever have ``1.x.y`` releases!
-* ``minor``: Changes that contain jQuery updates or substantial feature additions.
-* ``patch``: Bug fixes and added compatibility.
-
-For distributed packages with a longer lifecycle, we propose a loose pinning of the patch and minor versions to allow for future updates, which are likely to happen when jQuery receives security updates.
-Example: ``sphinxcontrib-jquery>=1.0.0,<2.0.0`` will include all compatible future releases, including jQuery upgrades.
-
-
 jQuery version and inclusion
 ----------------------------
 
--- a/docs/dev/design/sphinx-jquery.rst
+++ b/docs/dev/design/sphinx-jquery.rst
@@ -1,7 +1,7 @@
 sphinxcontrib-jquery
 ====================
 
-JQuery will be removed from Sphinx 6.0.0. We can expect 6.0.0 to ship in late 2023. See also:
+JQuery will be removed from Sphinx 6.0.0. We can expect 6.0.0 to ship in late 2022. See also:
 
 * https://github.com/sphinx-doc/sphinx/issues/7405
 * https://github.com/sphinx-doc/sphinx/issues/10070
--- a/docs/dev/design/sphinx-jquery.rst
+++ b/docs/dev/design/sphinx-jquery.rst
@@ -26,6 +26,9 @@ Package-name:
 Python package:
   ``sphinxcontrib.jquery``
 
+Dependencies:
+  Python 2.7+, Sphinx 1.8+
+
 Ownership:
   Read the Docs core team will implement the initial releases of an otherwise community-owned package that lives in https://github.com/sphinx-contrib/jquery
 
--- a/docs/dev/design/sphinx-jquery.rst
+++ b/docs/dev/design/sphinx-jquery.rst
@@ -5,6 +5,9 @@ JQuery will be removed from Sphinx 6.0.0. We can expect 6.0.0 to ship in late 20
 
 * https://github.com/sphinx-doc/sphinx/issues/7405
 * https://github.com/sphinx-doc/sphinx/issues/10070
+* https://github.com/readthedocs/sphinx-hoverxref/issues/160
+* https://github.com/readthedocs/sphinx_rtd_theme/issues/1253
+* https://github.com/pydata/pydata-sphinx-theme/issues/764
 
 This is a "request for comments" for a community-owned Sphinx extension that bundles jQuery.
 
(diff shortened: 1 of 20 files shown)


docs/dev/design/sphinx-jquery.rst
```diff
@@ -0,0 +1,110 @@
...
+Adds jQuery no matter what, but at most once.
+It's useful for themes and extensions that want jQuery added unconditionally or if they want to handle the conditions themselves.
+
+``sphinxcontrib.jquery.add_jquery_if_sphinx_lt_6(app)``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Adds jQuery if Sphinx is older than version 6, at most once.
+This is recommended for extensions and themes that maintain compatibility of Sphinx before and after version 6.
+
+Adding ``sphinxcontrib.jquery`` to a documentation project's ``conf.extensions`` will call this function.
```
humitos, line 75: I wouldn't add this method. I think it's better if theme/extension maintainers deal with this by themselves that will know more about their package than us.

humitos (commented):
I think this is simple enough and it's a good start. I would reduce the scope a little bit and and remove one of the method proposed to avoid handling `Sphinx<6` since it's not the target of this extension.

docs/dev/design/sphinx-jquery.rst
```diff
@@ -0,0 +1,110 @@
...
+
+  In addition to this, we recommend extension and theme developers to log to the browser's ``console.error`` in case jQuery isn't found. The log message could for instance say:
+
+    "<package-name> depends on sphinxcontrib-jquery. Please ensure that <package-name>.setup(app) is called or add 'sphinxcontrib-jquery' to your conf.py extensions setting."
+
+Documentation project owners:
+  If you are depending on a theme or extension that did not itself address the removal of jQuery from Sphinx, you can patch up your project like this:
+
+  #. Add ``sphinxcontrib-jquery`` to your ``requirements.txt``.
+  #. Add ``sphinxcontrib.jquery`` to your ``extensions`` setting in ``conf.py``.
```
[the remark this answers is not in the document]
benjaoming: If you are using a theme or extension that didn't address the Sphinx 6 release, you can fix it yourself as a project owner.

A project owner doesn't call the function but simply needs to enable the extension. It's the theme or extension that can call these methods directly to save the project owner from having to update their `conf.py`.

docs/dev/design/sphinx-jquery.rst
```diff
@@ -0,0 +1,110 @@
...
+
+The primary users of this package are
+**a) theme and extension developers** and
+**b) documentation project owners**.
+
+Theme and extension developers:
+  The following 2 steps need to be completed:
+
+  #. A Sphinx theme or extension should depend on the python package ``sphinxcontrib-jquery``.
+  #. Calling either ``sphinxcontrib.jquery.add_jquery`` or ``sphinxcontrib.jquery.add_jquery_if_sphinx_lt_6``.
```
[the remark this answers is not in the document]
benjaoming: Ah yes `gt` :see_no_evil:

docs/dev/design/sphinx-jquery.rst
```diff
@@ -0,0 +1,110 @@
+sphinxcontrib-jquery
+====================
+
+JQuery will be removed from Sphinx 6.0.0. We can expect 6.0.0 to ship in late 2022. See also:
```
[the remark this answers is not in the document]
benjaoming: Good idea!

docs/dev/design/sphinx-jquery.rst
```diff
@@ -0,0 +1,110 @@
...
+Overview
+--------
+
+Comment deadline:
+  November 1st, 2022
+
+package-name:
+  ``sphinxcontrib-jquery``
+
+python package:
```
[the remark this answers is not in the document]
benjaoming: Yes, definitely, fixed

docs/dev/design/sphinx-jquery.rst
benjaoming: Let's just say that Sphinx 7 reintroduces jQuery :smile: or we need to detect the existence of another popular extension shipping jQuery.

docs/dev/design/sphinx-jquery.rst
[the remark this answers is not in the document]
benjaoming: That's a good idea! Will update it!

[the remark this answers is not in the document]
benjaoming: This is now added

AA-Turner:
I've been directed to this discussion by Ben.

As opposed to the method, why not suggest `app.setup_extension("sphinxcontrib.jquery")`?

A

benjaoming:
@AA-Turner 

That's interesting, I haven't considered that at all. However, one thing that would require an extra configuration value then would be `force=True` in case an extension needs more control. We consider this case for the "unexpected", i.e. an extension or theme that we haven't considered needs for and that want to load jQuery based on some other set of conditions. I'm not sure how to do that, I'm also not sure how realistic this case is. Would it be possible to register a config value `jquery_force_enable` and `jquery_force_disable` so that an additional layer of control is yielded to both other extensions and conf.py?

Maybe having config values is a better approach. It could ultimately mean that someone who wants to load jQuery by themselves (from their own CDN etc) is free to do so.

AA-Turner:
Dear Ben,

I am not sure I understand the rationale for the `force` keyword, please may you briefly explain?

A

benjaoming:
We introduce sphinxcontrib-jquery for other themes and extensions to be able to provide jQuery through the extension. The general case is clear: Themes and extensions should have an easy way to continue using jQuery.

However, do we want an additional layer of control for themes, extensions and projects to specify exactly when they want jQuery loaded?

That's the ultimate question. I think that the proposal might have chosen a wrong mechanism for that.

* If I write a theme or extension that has its own jQuery and I know that another extension might also load it, should I be able to switch off sphinxcontrib-jquery with a config value or some other mechanism?
* If I have a project with its own CDN, should I be able to switch off sphinxcontrib-jquery? (config value seems obvious here)
* If I have a theme or extension that doesn't even want Sphinx 5.x's jQuery but is still compatible with Sphinx 5.x? (not sure about this one, but it also seems unlikely)

AA-Turner:
Dear Ben,

Thank you for your explanation; perhaps my understanding was too simple.

I had envisioned the extension as providing a way to pretend Sphinx had never removed jQuery, such that the jQuery library would be included *as if* it was provided by core Sphinx.

This would mean that extensions providing jQuery themselves would deal with the duplication similar to how they have in the past.

I do see some value in a global disable flag set somehow.

A

docs/dev/design/sphinx-jquery.rst
```diff
@@ -68,24 +68,18 @@ If you are depending on a theme or extension that did not itself address the rem
...
-``sphinxcontrib.jquery.add_jquery_if_sphinx_gte_6(app)``
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Adding ``sphinxcontrib.jquery`` to a documentation project's ``conf.extensions`` will call this function with ``force=False``.
 
-Adds jQuery if Sphinx is from version 6 and up. Adds jQuery at most once.
-This is recommended for extensions and themes that maintain compatibility of Sphinx before and after version 6.
+.. warning::
 
-Adding ``sphinxcontrib.jquery`` to a documentation project's ``conf.extensions`` will call this function.
+  If you call this function with ``force=False``, you most likely should also add ``Sphinx>=6`` to your theme's/extension's dependencies since versions before this already bundles jQuery!
```
humitos, line 82: I think this is not what we want. The idea of the `force=False` argument was to do not have this condition.

* Using `force=False` will only add jQuery if Sphinx>=6
* When `force=True` will add jQuery _always_, not matter what

That way, themes calling this function with `force=False` and using Sphinx>6 will get the jQuery added. Also, themes using Sphinx<6 won't get jQuery added.

benjaoming: Totally agree, it should say:

```
If you call this function with ``force=True``
```

benjaoming:
> I had envisioned the extension as providing a way to pretend Sphinx had never removed jQuery, such that the jQuery library would be included as if it was provided by core Sphinx.

That's also the main objective. That should be the default behavior.

I think we should throw other scenarios over board as long as no one is weighing in on them... or just do a design that can accommodate for them _later_. So it's all just pure speculation with potential for over-engineering.

@humitos what do you think about using config values instead of the `force=True/False` kwarg? I like how it pushes the configuration into a global layer, since it makes very little sense that Extension A wants jQuery and Extension B doesn't and the project owner maybe wants to include it from somewhere else or put a debugging version in place or whatevs... at the end of the day, there should be just 1 jQuery version included, hence the global configuration layer seems the right place, even though we may not know how/if it's going to used.

I'd like to change the proposal to use config values and then I think it's starting to look good :+1:

benjaoming:
> I'd use only one config value.

That sounds good, I'll get it updated :+1:

benjaoming:
@AA-Turner I noticed that you already released a 1.0.0, great work. I noticed the Python requirement set to 3.7, and I wanted to add that we aim to have this package perhaps work for 3.6.. I can't tell what is out there building "in the wild", nor how much we are willing to break with an update of sphinx-rtd-theme. At this stage, when there is nothing technically in the way of supporting older Python and Sphinx releases, I think we shouldn't impose minimal version requirements until we know exactly what the aim is.

I don't know exactly what Sphinx and Python versions that we'll remove from sphinx-rtd-theme 2.0. I can imagine that we will remove Python 2.7 and Sphinx<4.

benjaoming:
Hi AA-Turner

I see our comments have crossed each other. Your question should be answered already in the former comment.

Thanks for clarifying the release!

We need to do this if we want "zero downtime" for users that upgrade to Sphinx 6:

1) sphinxcontrib-jquery is ready and tested with Sphinx 6 beta releases
2) sphinx-rtd-theme 2.0 is ready with Sphinx 6 beta releases and sphinxcontrib-jquery
3) other important extensions with jquery needs are ready with sphinxcontrib-jquery
4) Sphinx 6.0.0 is released.

docs/dev/design/sphinx-jquery.rst
benjaoming: Alright, we have this solved with a single function and a parameter (which later turned into a configuration value)

docs/dev/design/sphinx-jquery.rst
```diff
@@ -0,0 +1,125 @@
...
+
+When a Sphinx theme or extension calls `setup_extension() <https://www.sphinx-doc.org/en/master/extdev/appapi.html#sphinx.application.Sphinx.setup_extension>`_, a call to ``sphinxcontrib.jquery.setup(app)`` will happen. Adding ``sphinxcontrib.jquery`` to a documentation project's ``conf.extensions`` will also call ``sphinxcontrib.jquery.setup(app)`` (at most once).
+
+In ``sphinxcontrib.jquery.setup(app)``, jQuery is added. The default behaviour is to detect the Sphinx version and include jQuery via `app.add_js_file <https://www.sphinx-doc.org/en/master/extdev/appapi.html#sphinx.application.Sphinx.add_js_file>`__ when Sphinx is from version 6 and up. jQuery is added at most once.
+
+
+Config value: ``jquery_force_enable``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+When setting ``jquery_force_enable=True``, jQuery is added no matter the Sphinx version, but at most once. This is useful if you want to handle alternative conditions for adding jQuery.
```
humitos, line 85: I would call it `jquery_force`. If we want to go verbose here, I'd say `jquery_force_include`.

AA-Turner:
Dear Ben,

It is too painful to test on pre-3.7 (I tried for aroudn 2 hours to get it to work), but I have declared suppport for 2.7 and later.

Thanks,
Adam

AA-Turner:
I have invited Ben and Manuel to the `sphinxcontrib-jquery` repo.

Thanks,
Adam

benjaoming:
AA-Turner, thanks for adding us.

> It is too painful to test on pre-3.7

I think that the pain of the developer is considerable and that we should not spend too much time on old stacks. My suggestion was to get rid of it if it comes for free.. but when it doesn't, then get rid of it :smile: :+1:

benjaoming:
@AA-Turner 

Since you are already starting to implement and release stuff, do you mind if we do work to have the full proposal implemented and would you release that or grant access for us to release updates?

AA-Turner:
Dear Ben,

Sounds good, hopefully I'll have time to review PRs etc.

A

docs/dev/design/sphinx-jquery.rst
```diff
@@ -0,0 +1,125 @@
...
+If you are depending on a theme or extension that did not itself address the removal of jQuery from Sphinx 6, you can patch up your project like this:
+
+#. Add ``sphinxcontrib-jquery`` to your ``requirements.txt``.
+#. Add ``sphinxcontrib.jquery`` to your ``extensions`` setting in ``conf.py``.
+
+
+Calling ``app.setup_extension("sphinxcontrib.jquery")``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+When a Sphinx theme or extension calls `setup_extension() <https://www.sphinx-doc.org/en/master/extdev/appapi.html#sphinx.application.Sphinx.setup_extension>`_, a call to ``sphinxcontrib.jquery.setup(app)`` will happen. Adding ``sphinxcontrib.jquery`` to a documentation project's ``conf.extensions`` will also call ``sphinxcontrib.jquery.setup(app)`` (at most once).
```
humitos, line 77: This paragraph is confuses me a little. It seems that even doing both, calling `setup_extension()` _and_ adding it to `conf.extensions` will call the function _just once_ which I understand it's not how it works.

I would rephrase it as:

```suggestion
Either calling `setup_extension() <https://www.sphinx-doc.org/en/master/extdev/appapi.html#sphinx.application.Sphinx.setup_extension>`_ from the theme/extension or adding ``sphinxcontrib.jquery`` to a documentation project's ``conf.extensions``  will produce a call to ``sphinxcontrib.jquery.setup(app)`` (at most once).
```

docs/dev/design/sphinx-jquery.rst
```diff
@@ -0,0 +1,125 @@
...
+-------
+
+All releases follow semantic versioning: ``<major>.<minor>.<patch>``
+
+* ``major``: Changes that break existing themes and extensions - we'll aim to only ever have ``1.x.y`` releases!
+* ``minor``: Changes that contain jQuery updates or substantial feature additions.
+* ``patch``: Bug fixes and added compatibility.
+
+For distributed packages with a longer lifecycle, we propose a loose pinning of the patch and minor versions to allow for future updates, which are likely to happen when jQuery receives security updates.
+Example: ``sphinxcontrib-jquery>=1.0.0,<2.0.0`` will include all compatible future releases, including jQuery upgrades.
```
humitos, line 102: I think this section is outdated and we should probably remove it, since we won't be following. The 1.x and 2.x already happened.

docs/dev/design/sphinx-jquery.rst
```diff
@@ -0,0 +1,125 @@
...
+
+
+jQuery version and inclusion
+----------------------------
+
+Sphinx has kept relatively up to date with jQuery, and this package intends to follow.
+The most recently bundled jQuery version was v3.5.1 and only two releases have happened since: 3.6.0 and 3.6.1.
+The 3.6.0 release had a very small backwards incompatibility which illustrates how harmless these upgrades are for the general purpose Sphinx package.
+
+Therefore, we propose to start 1.0.0 at 3.5.1 (the currently shipped version) and subsequently release 3.6.1 as the first update of jQuery as 1.1.0.
```
humitos, line 112: This also makes references to 1.x, but it should be updated.

docs/dev/design/sphinx-jquery.rst
```diff
@@ -0,0 +1,125 @@
...
+The bundled jQuery version will be NPM pre-minified and distributed together with the PyPI package.
+
+The minified jQuery JS file is ultimately included by calling `app.add_js_file <https://www.sphinx-doc.org/en/master/extdev/appapi.html#sphinx.application.Sphinx.add_js_file>`__, which is passed the following arguments:
+
+.. code:: python
+
+  app.add_js_file(
+      get_jquery_url_path(),
+      loading_method="defer",
+      priority=200,
```
humitos, line 123: Where this 200 as priority comes from? Is it 200 for some specific reason?

humitos (approved):
The proposal looks good to me. Considering that the repository and code is already done and the package already released, we need to make some adjustments here in the docs. Also, we need to update the code to reflect this document.

I'll approve these changes and ping @ericholscher and @agjohnson who were the other folks discussing this to do a final review, but I think we are good moving forward here.

docs/dev/design/sphinx-jquery.rst
benjaoming: See https://www.sphinx-doc.org/en/master/extdev/appapi.html#sphinx.application.Sphinx.add_js_file

docs/dev/design/sphinx-jquery.rst
```diff
@@ -0,0 +1,125 @@
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The following 2 steps need to be completed:
+
+#. A Sphinx theme or extension should depend on the python package ``sphinxcontrib-jquery``.
+#. In your extension's or theme's ``setup(app)``, call ``app.setup_extension("sphinxcontrib.jquery")``.
+
+In addition to this, we recommend extension and theme developers to log to the browser's ``console.error`` in case jQuery isn't found. The log message could for instance say::
+
+  if ($ == "undefined") console.error("<package-name> depends on sphinxcontrib-jquery. Please ensure that <package-name>.setup(app) is called or add 'sphinxcontrib-jquery' to your conf.py extensions setting.")
```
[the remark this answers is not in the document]
benjaoming: I think the original suggestion was to have a final safety net? If the extension is responsible for this, then the extension is required for it to fire. But the tip is to install the extension :)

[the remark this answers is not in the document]
benjaoming: So I would say: No, it doesn't make sense. But it makes sense to put this in the extensions README.md as a small tip for extension owners.

docs/dev/design/sphinx-jquery.rst
benjaoming: These are different patterns.

One pattern is for extension owners (that call `setup_extension`), the other pattern is for project owners that write a `conf.py`.

> which I understand it's not how it works.

That's how it works though :) It's even implemented here:

https://github.com/sphinx-contrib/jquery/blob/master/sphinxcontrib/jquery.py#L8

docs/dev/design/sphinx-jquery.rst
benjaoming: Yes, I don't know what we are following. @AA-Turner please observe the thoughts behind using a versioning scheme :arrow_up:

docs/dev/design/sphinx-jquery.rst
benjaoming: Also, I think that 200 is the correct value.

docs/dev/design/sphinx-jquery.rst
AA-Turner: Counter: https://hynek.me/articles/semver-will-not-save-you/

Pinning doesn't make much sense in the Python ecosystem as I see it, given the culture of longer deprecations. Given the intentionally limited scope there should be no breaking changes, but breaking is different from one man to the next.

A

benjaoming: I don't think the blog post is what you think it is.

That aside, if you - as the maintainer - does not want to use SemVer, then the proposal cannot mention anything about it.

So I'll remove it. I disagree, but I will remove it :+1:

AA-Turner: I linked the post (one of many on both sides of the debate) as my view is that how maintainers choose to implement SemVer doesn't necessarily align with users' expectations of SemVer, and I concluded that it isn't worth the hassle. That was the read I got from Hyneks's post, though perhaps I did misinterpret it.

A

docs/dev/design/sphinx-jquery.rst
```diff
@@ -0,0 +1,125 @@
...
```
agjohnson, line 62: Accidentally an operator here:

```suggestion
  if (typeof $ == "undefined") console.error("<package-name> depends on sphinxcontrib-jquery. Please ensure that <package-name>.setup(app) is called or add 'sphinxcontrib-jquery' to your conf.py extensions setting.")
```

docs/dev/design/sphinx-jquery.rst
```diff
@@ -0,0 +1,125 @@
...
+
+
+Documentation project owners
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+If you are depending on a theme or extension that did not itself address the removal of jQuery from Sphinx 6, you can patch up your project like this:
+
+#. Add ``sphinxcontrib-jquery`` to your ``requirements.txt``.
```
agjohnson, line 70: To cover other supported installation methods besides requirements files:

```suggestion
#. Add ``sphinxcontrib-jquery`` to your installed dependencies.
```

docs/dev/design/sphinx-jquery.rst
```diff
@@ -0,0 +1,125 @@
...
+The minified jQuery JS file is ultimately included by calling `app.add_js_file <https://www.sphinx-doc.org/en/master/extdev/appapi.html#sphinx.application.Sphinx.add_js_file>`__, which is passed the following arguments:
+
+.. code:: python
+
+  app.add_js_file(
+      get_jquery_url_path(),
+      loading_method="defer",
+      priority=200,
+      integrity="sha256-{}".format(get_jquery_sha256_checksum())
+  )
```
agjohnson, line 125: This feature is still on my wishlist :+1:

I'd like to see the default behavior installing from a local file. Google doesn't need our help tracking users

docs/dev/design/sphinx-jquery.rst
benjaoming: I did also notice that you've already shipped two releases as "1.0.0" and "2.0.0", so the user-directed warnings of the blog post are spot on :smile: 

It's no biggy, and I think that we - as users of Python libraries - are rather used to projects that look like SemVer but aren't (yet). When I read a version string, I interpret it together with its release history and community... the longer the history, the more likely that it has a settled understanding of SemVer.

If we have a complicated understanding of version (as the proposal goes), then there is also a risk that we forget our own guidelines or someone else takes over the project and never reads them.

So I have no problem seeing this from another perspective, I'll remove this from the proposal.

docs/dev/design/sphinx-jquery.rst
humitos: Awesome! Thanks 👍🏼

docs/dev/design/sphinx-jquery.rst
benjaoming: And definitely just want to repeat what I sad about this... no biggy! I'm sure the extension will do its job well without this and I thank you for the efforts :pray:

agjohnson: As far as this section is concerned, I agree with @humitos and don't think we need to define versioning too strictly. It seems we're on a different path from the plan noted here.

I'm still :+1: on semver though. It at least gives users and maintainers *some* common language to describe backwards compatibility. There will be edge cases and bugs, but designing around these is always a trap.

humitos: I'm fine by "not following semver" on some projects. However, if that's the decision made, I usually try to use a different version string that does not look like semver to avoid misunderstanding in communication. That is, for example, instead of using `x.y.z` and _not_ following semver, I'd prefer to use `2022.10.24` (calver, or something similar).

That said, I don't have a strong opinion on this and either is fine to me 👍🏼 

My 2 cents.

agjohnson (commented):
:+1: This all seems reasonable, it's still pretty close to what we discussed previously.

docs/dev/design/sphinx-jquery.rst
humitos: From that link, I'd say that we should use `500: default priority for extensions` since `200: default priority for built-in JavaScript files` and this is not a built-in javascript (it _was_, tho).

docs/dev/design/sphinx-jquery.rst
benjaoming: To complicate things @humitos ... let's say _for some given project_ we are NOT using SemVer now, but we might use it in the future... then having a version scheme that _looks like_ SemVer but _isn't_ would be necessary :exploding_head: 

I agree that using a scheme that definitely doesn't look like SemVer is good if you decide to not use SemVer.

docs/dev/design/sphinx-jquery.rst
benjaoming: Yes, the understanding with @AA-Turner is that we will contribute the remaining changes from the proposal. So both the local file and the `integrity` check will be added :+1:

benjaoming:
> +1 This all seems reasonable, it's still pretty close to what we discussed previously.

@agjohnson yes definitely - the intention was to move an internal effort into a public scope here. Hopefully everything from the previous discussions got reflected and duplicated :smile:

docs/dev/design/sphinx-jquery.rst
benjaoming: I removed everything about versioning now, let's take it as it comes :+1:

benjaoming:
Thanks for all the comments! I'd suggest to keep the PR open for any further commenting, but that we move forwards with implementing the remaining part of the proposal. I'll open issues in https://github.com/sphinx-contrib/jquery

benjaoming:
@AA-Turner I'm slightly "under the weather", but I'll try to get to the freshly added issues in the repo and create tests where meaningful.

If anyone wants to jump in, please feel free to self-assign the issue :+1:

PRO-BROS:
Very well. Facing the same issue but now resolved on my [site](https://prhocoding.com/)

humitos:
I think we can just merge this PR as-is since it already worked as a canonical place to discuss the proposal implementation and the work is already happening. If anything, you can just add a small paragraph mentioning the repository where this work is happening and then merge it.

benjaoming:
Yes, I think we are also out of time for processing more feedback :+1:

Issue #9502, opened 2022-08-16
Add blacken-docs to pre-commit hooks (and blacken docs)


benjaoming:
While waiting for pre-commit hooks and Black to be added (and indeed run for the entire code base to be blackened), we can test out this pre-commit hook:

https://github.com/asottile/blacken-docs

To fix the issue:

* Add the hook to the pre-commit config over in https://github.com/readthedocs/common/blob/main/pre-commit-config.yaml
* Run blacken-docs on all files (but please don't run all hooks on all files)

ArhyaSaha:
Hi I am starting my open source journey!
Would love to be able to contribute on this.

Can you assign me this issue?

ArhyaSaha:
`-   repo: https://github.com/asottile/blacken-docs
    rev: v1.12.1
    hooks:
    -   id: blacken-docs
        additional_dependencies: [black==22.6.0]`

 Also, will adding this over in https://github.com/readthedocs/common/blob/main/pre-commit-config.yaml solve the issue?

benjaoming:
Hi @ArhyaSaha - this issue should ideally be solved in the two steps outlined above. I would recommend opening up a PR on the `common` repo and then link it to a PR on the `readthedocs.org` repo (this one) that shows the results of running `--all-files` (should only affect `docs/` of course).

One more thing: The `common` module is shared between several repositories (hence the name). I think that you can assume that documentation always lives in a `docs/` folder.

ArhyaSaha:
Hi @benjaoming I am unable to run the pre-commit hook in my local repo. It gives me an error. 

An error has occurred: InvalidConfigError: 
==> File .pre-commit-config.yaml
=====> Expected a Config map but got a str

I cloned the `readthedocs.org` repo and also cloned the `common` repo and moved it in place of the `common` folder in the `readthedocs.org` repo. After that I also made sure to install pre-commit. Then I added the code i suggested above(previous comment) in https://github.com/readthedocs/common/blob/main/pre-commit-config.yaml . When I try to run the pre-commit hooks it gives me the mentioned error.

How should i move forward with this? Please guide me.

ArhyaSaha:
Hi @benjaoming. I tried setting up my environment as mentioned in the docs. I am not able to build docker image for the servers as it gives me an error when I run the given command. 
The error might be because I am on a windows platform. 
Is there any other way i can set up my local environment for this issue?

benjaoming:
[...]

PR #9672, opened 2022-10-18
Run `blacken-docs` precommit hook on all files

dojutsu-user:
Related Issue: https://github.com/readthedocs/readthedocs.org/issues/9502
Related PR: https://github.com/readthedocs/common/pull/156

---

- User's documentation (`docs`): https://docs--9672.org.readthedocs.build/en/9672/

- Developer's documentation (`dev`): https://dev--9672.org.readthedocs.build/en/9672/

### DIFF
--- a/docs/dev/install.rst
+++ b/docs/dev/install.rst
@@ -175,7 +175,9 @@ you can use ``celery.contrib.rdb``:
 
 .. code-block:: python
 
-    from celery.contrib import rdb; rdb.set_trace()
+    from celery.contrib import rdb
+
+    rdb.set_trace()
 
 When the breakpoint is hit, the Celery worker will pause on the breakpoint and
 will alert you on STDOUT of a port to connect to. You can open a shell into the container
--- a/docs/dev/design/better-doc-urls-handling.rst
+++ b/docs/dev/design/better-doc-urls-handling.rst
@@ -418,17 +418,18 @@ the parts from the URL.
 
    LANGUAGES = {"es", "en"}
 
+
    def pop_parts(path, n):
-       if path[0] == '/':
-          path  = path[1:]
-       parts = path.split('/', maxsplit=n)
+       if path[0] == "/":
+           path = path[1:]
+       parts = path.split("/", maxsplit=n)
        start, end = parts[:n], parts[n:]
-       end = end[0] if end else ''
+       end = end[0] if end else ""
        return start, end
 
 
    def resolve(canonical_project: Project, path: str, check_subprojects=True):
-       prefix = '/'
+       prefix = "/"
        if canonical_project.prefix:
            prefix = canonical_project.prefix
        subproject_prefix = "/projects"
--- a/docs/dev/design/theme-context.rst
+++ b/docs/dev/design/theme-context.rst
@@ -33,72 +33,70 @@ Note that this dictionary is injected under the main key `readthedocs`:
 .. code:: python
 
    {
-       'readthedocs': {
-           'v1': {
-               'version': {
-                   'id': int,
-                   'slug': str,
-                   'verbose_name': str,
-                   'identifier': str,
-                   'type': str,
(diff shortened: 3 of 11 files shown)


benjaoming (approved):
Yay! This has worked extremely well, really glad to see all these code-snippets tidied up :tada: 

Great work!

I'll await to see if anyone else on the team has comments on this, but a big "LGTM" from me :)

dojutsu-user:
@benjaoming Possible to merge these PRs?

docs/dev/install.rst
```diff
@@ -175,7 +175,9 @@ you can use ``celery.contrib.rdb``:
 
 .. code-block:: python
 
-    from celery.contrib import rdb; rdb.set_trace()
+    from celery.contrib import rdb
+
+    rdb.set_trace()
```
ericholscher, line 180: This feels like somewhere that we want the existing syntax, as it's standard for using PDB-like things. Is there a way to disable the auto-formatting here?

ericholscher (commented):
This looks good, but we should figure out a pattern for excluding things. Often documentation has a different goal with a code sample than code, so we should have good ways to preserve custom formatting as needed.

docs/dev/install.rst
benjaoming: It's possible to use the new example, arguing like this:

1. Example code is blackened, so someone copy-pasting won't have to bother with linting errors from copy-pasting it.
2. Importing `rdb` in itself is harmless, but if we only comment out `set_trace`, we'll get an "unused import" linting error anyways
3. We can start the trace at any point in our code, the example of using the module for instance doesn't have the line above... https://github.com/celery/celery/blob/master/celery/contrib/rdb.py

razor-vg: Also, `blacken-docs` is just a wrapper around `black`, so we can use all the features of `black`, like disabling auto-formatting.
I tried this example and it worked:
```
.. code-block:: python

    # fmt: off
    from celery.contrib import rdb; rdb.set_trace()
```

Let me know if we need to update the PR.

cc: @benjaoming @ericholscher

dojutsu-user: Also, `blacken-docs` is just a wrapper around `black`, so we can use all the features of `black`, like disabling auto-formatting.
I tried this example and it worked:
```
.. code-block:: python

    # fmt: off
    from celery.contrib import rdb; rdb.set_trace()
```

Let me know if we need to update the PR.

cc: @benjaoming @ericholscher

benjaoming: Great work @dojutsu-user - glad to see that the possibility exists.

I think that it's great to have documentation that lints with black and a little odd/distracting to have comments in documentation examples for disabling linting rules. My concern would be that the reader is confused.

What do you think, @ericholscher ?

ericholscher: Hrm, I'd really like to avoid having the comment in the code block, as that will confuse users as well.

This code should never be checked into a repo, since it will block and break if it's ever checked in. I'm not really sure what you're trying to say @benjaoming?

I guess we can just move forward with this change, but I do think it's overall worse in the docs because of the lack of flexibility here.

benjaoming: The new linting caught a wrong indentation and made it easier to copy-paste without re-formatting, so I think that the benefits can definitely outweigh the disadvantages :+1: 

A possible approach is to say that when/if there's a code example that does not conform to black, then there must be a better alternative with black formatting.

> This code should never be checked into a repo, since it will block and break if it's ever checked in. I'm not really sure what you're trying to say @benjaoming?

Well, you could deploy it to non-prod in order to debug a development environment. Yes, you are right that it would be great and flexible if the linting could be disabled for a segment that intentionally doesn't need it. In this case, it seems possible to accept, unless there's a way to disable it without adding comments in the code itself.

ericholscher (approved):

PR #9696, opened 2022-10-26
GHA: only run the preview links action on `docs/` path

humitos:
Instead of adding a preview link for every single PR opened, the action will only be called if there is a modification in the `docs/` direction. This is only to avoid the a little the noise when there is no documentation changes.

---

- User's documentation (`docs`): https://docs--9696.org.readthedocs.build/en/9696/

- Developer's documentation (`dev`): https://dev--9696.org.readthedocs.build/en/9696/

### DIFF
--- a/.github/workflows/pr-preview-links.yaml
+++ b/.github/workflows/pr-preview-links.yaml
@@ -3,6 +3,8 @@ on:
   pull_request_target:
     types:
       - opened
+    paths:
+      - "docs/**"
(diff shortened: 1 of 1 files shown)


stsewd (approved):
there may be cases where changing things outside this directory will change the docs, but this is only to generate the link preview, so.

humitos:
Yeap, exactly.

PR #9717, opened 2022-11-08
Docs: refactor "skipping a build" section

humitos:
- Move the explanation about the "Cancelling builds" feature to the "Builds" page
- Keep the examples (How-To) for "Cancel a build" into the "Build customization" page

---

- User's documentation (`docs`): https://docs--9717.org.readthedocs.build/en/9717/

- Developer's documentation (`dev`): https://dev--9717.org.readthedocs.build/en/9717/

### DIFF
--- a/docs/user/build-customization.rst
+++ b/docs/user/build-customization.rst
@@ -73,7 +73,7 @@ There are some caveats to knowing when using user-defined jobs:
 * Environment variables are expanded in the commands (see :doc:`environment-variables`)
 * Each command is executed in a new shell process, so modifications done to the shell environment do not persist between commands
 * Any command returning non-zero exit code will cause the build to fail immediately
-  (note there is a special exit code to `skip the build <skip-build-based-on-a-condition>`_)
+  (note there is a special exit code to `cancel the build <cancel-build-based-on-a-condition>`_)
 * ``build.os`` and ``build.tools`` are required when using ``build.jobs``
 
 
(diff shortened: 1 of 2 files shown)


docs/user/build-customization.rst
```diff
@@ -105,24 +105,14 @@ To avoid this, it's possible to unshallow the clone done by Read the Docs:
...
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Cancel build based on a condition
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-There may be situations where you want to skip a build that was automatically triggered when someone on your team pushed to the repository.
-Skipping builds will allow you to speed up review times and also help us reduce server costs and ultimately our environmental footprint.
-Consider the following scenarios:
+When a command exits with code ``439``,
+Read the Docs will cancel the build immediately.
+We can use this approach to cancel builds we don't want to complete based on a condition.
```
ericholscher, line 113: Who is `We`? I think we want to talk to the user right?

```suggestion
You can use this approach to cancel builds that you don't want to complete based on some conditional logic.
```

docs/user/build-customization.rst
```diff
@@ -105,24 +105,14 @@ To avoid this, it's possible to unshallow the clone done by Read the Docs:
...
 
-Skip build based on a condition
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Cancel build based on a condition
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-There may be situations where you want to skip a build that was automatically triggered when someone on your team pushed to the repository.
-Skipping builds will allow you to speed up review times and also help us reduce server costs and ultimately our environmental footprint.
-Consider the following scenarios:
+When a command exits with code ``439``,
```
ericholscher, line 111: This error code isn't a valid exit code (https://tldp.org/LDP/abs/html/exitcodes.html), we shouldn't document a code above 255 otherwise users will get confused.

docs/user/builds.rst
```diff
@@ -68,6 +68,46 @@ The following are the pre-defined jobs executed by Read the Docs:
...
+Cancelling builds
+-----------------
+
+There may be situations where you want to cancel a particular running build.
+Cancelling running builds will allow your team to speed up review times and also help us reduce server costs and ultimately,
+our environmental footprint.
+
+Consider the following scenarios:
+
+* the build depends on an external situation that's not met yet
```
ericholscher, line 80:
```suggestion
* the build has an external dependency that hasn't been updated
```

docs/user/builds.rst
```diff
@@ -68,6 +68,46 @@ The following are the pre-defined jobs executed by Read the Docs:
...
+* there were no changes on the documentation files
+* the user wants it for any particular reason
```
ericholscher, line 82: This is not clear to me? Not 100% sure what it's trying to say, but we can be more explicit here, or not include it if it's just "other"

docs/user/builds.rst
```diff
@@ -68,6 +68,46 @@ The following are the pre-defined jobs executed by Read the Docs:
     it's possible to run user-defined commands and :doc:`customize the build process <build-customization>`.
 
 
+Cancelling builds
```
ericholscher, line 71: This heading can definitely be improved to be explicit about what this is about. It's definitely [explanation](https://diataxis.fr/explanation/#the-language-of-explanation):

```suggestion
When to cancel builds
```

ericholscher (approved):
This is a great addition to the docs.

docs/user/build-customization.rst
humitos: Yes, "we" is incorrect here.

docs/user/builds.rst
humitos: My idea here was to communicate the developer could have "any reason" and know upfront they want to skip the build for this particular push/commit. Because of that "any reason" the user will add `skip ci` in the commit message.

I'm not sure how to explain that situation in a better way (without mentioning the commit message, since that's the implementation, not the scenario)

ericholscher:
```suggestion
* many other use cases that can be solved with custom logic
```

PR #9734, opened 2022-11-15
Docs: cancel PR builds if there is no documentation changes

humitos:
Use the new feature to cancel builds if there are not documentation changes

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -25,7 +25,7 @@ build:
       # If there are no changes (exit 0) we force the command to return with 183.
       # This is a special exit code on Read the Docs that will cancel the build immediately.
       - |
-        if [ $READTHEDOCS_VERSION_TYPE = "external" ];
+        if [ "$READTHEDOCS_VERSION_TYPE" = "external" ];
         then
           ! git diff --quiet origin/main -- docs/ && exit 183;
         fi
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -14,6 +14,21 @@ build:
   os: ubuntu-22.04
   tools:
     python: "3.10"
+  jobs:
+    post_checkout:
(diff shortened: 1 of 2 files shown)


benjaoming (approved):
LGTM! We should also be running this ASAP in production to make sure that we gather maximum hands-on experience with the approach :)

benjaoming:
@humitos we'll have to investigate why the latest build fails with exit code 1: https://readthedocs.org/projects/dev/builds/18636879/

humitos:
Meh, this doesn't work as we expected. When `git diff` returns 1 it exists immediately and break the build. We need a way to execute the `git diff` command and save its exit code into a variable _without_ exiting the whole command.

humitos:
I don't think that works either. Have you tested it?

humitos:
Cool! I figured it out. My last command change worked! 💪🏼

benjaoming:
> I don't think that works either. Have you tested it?

Yes, worked fine locally :+1:

.readthedocs.yml
```diff
@@ -14,6 +14,19 @@ build:
...
+      # https://docs.readthedocs.io/en/stable/build-customization.html#cancel-build-based-on-a-condition
+      #
+      # Cancel building pull requests when there aren't changed in the docs directory.
+      #
+      # If there are no changes (git diff exits with 0) we force the command to return with 183.
+      # This is a special exit code on Read the Docs that will cancel the build immediately.
+      - |
+        if [ "$READTHEDOCS_VERSION_TYPE" = "external" ] && git diff --quiet origin/main -- docs/;
+        then
+          exit 183;
```
benjaoming, line 28:
```suggestion
          echo "No changes to docs/ - exiting the build.";
          exit 183;
```

benjaoming:
Since this _does_ change `docs/`, we should remember to check that it actually "skips" builds after it's been merged :)

PR #9743, opened 2022-11-20
Docs: update instructions to install deps with Poetry

### LINKED ISSUE #9740: Docs `build.job` recipe for `poetry` does not work with newer versions
I'm trying to build a site for a package managed using `poetry` and following the officially documented approach in the `build.jobs` docs [section](https://docs.readthedocs.io/en/stable/build-customization.html#install-dependencies-with-poetry). That approach now appears to be failing:

https://readthedocs.org/projects/virtual-rainforest/builds/18663328/

The TLDR on that build log error is that the old `get-poetry.py` install _will not_ install recent versions (`>1.2.0a1`) and that the new `https://install.python-poetry.org` approach _must_ be used instead. I tried using that approach:

```yaml
build:
  os: ubuntu-20.04
  tools:
    python: "3.9"
  jobs:
    post_create_environment:
      # Install poetry
      - curl -sSL https://install.python-poetry.org | python3 -
      # Tell poetry to not use a virtual environment
      - $HOME/.local/bin/poetry config virtualenvs.create false
      # Install project's dependencies
      - $HOME/.local/bin/poetry install
```

but that doesn't seem to be working: https://readthedocs.org/projects/virtual-rainforest/builds/18663177/

I know that is an apparent error in my package, but it doesn't occur with local `sphinx` builds and seems like a problem with the installed environment? I think @humitos tried this approach and also had problems with the resulting build environment : 

https://github.com/readthedocs/readthedocs.org/pull/9445#issuecomment-1193736700

_Originally posted by @davidorme in https://github.com/readthedocs/readthedocs.org/issues/9445#issuecomment-1318911972_

browniebroke:
Updating instructions with the [latest from the offiical docs](https://python-poetry.org/docs/#installing-with-the-official-installer).

> The previous get-poetry.py and install-poetry.py installers are deprecated.

Fix #9740 

---

- User's documentation (`docs`): https://docs--9743.org.readthedocs.build/en/9743/

- Developer's documentation (`dev`): https://dev--9743.org.readthedocs.build/en/9743/

### DIFF
--- a/docs/user/build-customization.rst
+++ b/docs/user/build-customization.rst
@@ -310,7 +310,7 @@ Install dependencies with Poetry
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 Projects managed with `Poetry <https://python-poetry.org/>`__,
-can use the ``post_create_environment`` user-defined job to use Poetry for installing Python dependencies.
+can use the ``post_create_environment`` user-defined job to install Poetry and ``post_install`` to install Python dependencies.
 Take a look at the following example:
 
 
@@ -326,24 +326,16 @@ Take a look at the following example:
      jobs:
        post_create_environment:
          # Install poetry
-         # https://python-poetry.org/docs/#installing-with-the-official-installer
-         - curl -sSL https://install.python-poetry.org | python3 -
+         - pip install poetry
          # Tell poetry to not use a virtual environment
-         - $HOME/.local/bin/poetry config virtualenvs.create false
-       pre_install:
-         # Export project dependencies to requirements.txt
-         - $HOME/.local/bin/poetry export --with docs > docs/requirements.txt
+         - poetry config virtualenvs.create false
+       post_install:
+         # Install dependencies with 'docs' dependency group
+         # https://python-poetry.org/docs/managing-dependencies/#dependency-groups
+         - poetry install --with docs
 
    sphinx:
      configuration: docs/conf.py
-   
-   python:
-     install:
-       # Install exported dependencies
-       - requirements: docs/requirements.txt
-       # Optional: install current package if imported from `docs/conf.py`
-       - method: pip
-         path: .
 
 
 Override the build process
--- a/docs/user/build-customization.rst
+++ b/docs/user/build-customization.rst
@@ -310,7 +310,7 @@ Install dependencies with Poetry
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 Projects managed with `Poetry <https://python-poetry.org/>`__,
-can use the ``post_create_environment`` user-defined job to install Poetry and ``post_install`` to install Python dependencies.
+can use the ``post_create_environment`` user-defined job to use Poetry for installing Python dependencies.
 Take a look at the following example:
 
 
@@ -326,10 +326,10 @@ Take a look at the following example:
      jobs:
        post_create_environment:
          # Install poetry
+         # https://python-poetry.org/docs/#installing-manually
          - pip install poetry
          # Tell poetry to not use a virtual environment
          - poetry config virtualenvs.create false
-       post_install:
          # Install dependencies with 'docs' dependency group
          # https://python-poetry.org/docs/managing-dependencies/#dependency-groups
          - poetry install --with docs
--- a/docs/user/build-customization.rst
+++ b/docs/user/build-customization.rst
@@ -326,12 +326,12 @@ Take a look at the following example:
      jobs:
        post_create_environment:
          # Install poetry
-         # https://python-poetry.org/docs/#osx--linux--bashonwindows-install-instructions
-         - curl -sSL https://raw.githubusercontent.com/python-poetry/poetry/master/get-poetry.py | python -
+         # https://python-poetry.org/docs/#installing-with-the-official-installer
+         - curl -sSL https://install.python-poetry.org | python3 -
          # Tell poetry to not use a virtual environment
-         - $HOME/.poetry/bin/poetry config virtualenvs.create false
+         - $HOME/.local/bin/poetry config virtualenvs.create false
          # Install project's dependencies
-         - $HOME/.poetry/bin/poetry install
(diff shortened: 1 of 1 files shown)


humitos (changes requested):
It would be good to research a little more to try to figure it out how to install poetry without required `python.install` as it was doing before.

`get-poetry.py` seems to be deprecated right now and builds using it are failing: https://readthedocs.org/projects/test-builds/builds/18691179/

However, using the new installer seems to not install the dependencies in the right virtualenv: https://readthedocs.org/projects/test-builds/builds/18691225/

I'm more tempted to recommend using `pipx` or `pip` instead (the other installation methods shown in the official docs), where the instructions will be simpler that the required YAML shown on this PR. What do you think?

humitos:
> I'm using a recent version of Sphinx, and the RTD install was installing `sphinx<2` by default, which uninstalled my version (4) because it ran after. I'm a bit surprised that it wasn't an issue before, to be honest.

This is because your project is "old" and we haven't touched those project to avoid breaking them. See https://docs.readthedocs.io/en/stable/build-default-versions.html#python. If you want me to disable this feature flag in your project, please use the from in https://readthedocs.org/support/

browniebroke:
> This is because your project is "old" and we haven't touched those project to avoid breaking them.

Yes, I understand: better to remain stable. I can't imagine how many docs pages haven't touched in years and are stuck on the old build system and RTD have to keep this. The great thing is that the config file is flexible enough to let more advanced users customise their build.

I was more sharing my finding to explain why I changed it, compared to how it was documented before.

docs/user/build-customization.rst
```diff
@@ -326,12 +326,13 @@ Take a look at the following example:
...
          # Install poetry
-         # https://python-poetry.org/docs/#osx--linux--bashonwindows-install-instructions
-         - curl -sSL https://raw.githubusercontent.com/python-poetry/poetry/master/get-poetry.py | python -
+         - pip install poetry
          # Tell poetry to not use a virtual environment
-         - $HOME/.poetry/bin/poetry config virtualenvs.create false
-         # Install project's dependencies
-         - $HOME/.poetry/bin/poetry install
+         - poetry config virtualenvs.create false
+       post_install:
```
humitos, line 332: We should remove this `post_install` here to simplify the config a little since it's not required for the most common use case. I tried this in `test-builds` project and it worked properly https://readthedocs.org/projects/test-builds/builds/18704586/

docs/user/build-customization.rst
```diff
@@ -326,12 +326,13 @@ Take a look at the following example:
      jobs:
        post_create_environment:
          # Install poetry
-         # https://python-poetry.org/docs/#osx--linux--bashonwindows-install-instructions
```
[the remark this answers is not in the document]
browniebroke: Right, of course!

docs/user/build-customization.rst
```diff
@@ -326,12 +326,13 @@ Take a look at the following example:
...
+         - pip install poetry
          # Tell poetry to not use a virtual environment
-         - $HOME/.poetry/bin/poetry config virtualenvs.create false
-         # Install project's dependencies
-         - $HOME/.poetry/bin/poetry install
+         - poetry config virtualenvs.create false
+       post_install:
+         # Install dependencies with 'docs' dependency group
+         # https://python-poetry.org/docs/managing-dependencies/#dependency-groups
+         - poetry install --with docs
```
browniebroke, line 335:
```suggestion
         # Install dependencies with 'docs' dependency group
         # https://python-poetry.org/docs/managing-dependencies/#dependency-groups
         - poetry install --with docs
```

docs/user/build-customization.rst
browniebroke: I thought it was best to cover more edge cases, in the hope that it would avoid potential issues for old projects and therefore support requests/GitHub issues.

That being said, I don't mind either way & I'll remove this section 👍🏻

davidorme:
Just to cross reference that this recipe closes #9740.

humitos:
@davidorme thanks for confirming this works for your case! 💪🏼

humitos (approved):
Awesome! Thanks a lot 💯

docs/user/build-customization.rst
```diff
@@ -310,7 +310,7 @@ Install dependencies with Poetry
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 Projects managed with `Poetry <https://python-poetry.org/>`__,
-can use the ``post_create_environment`` user-defined job to use Poetry for installing Python dependencies.
+can use the ``post_create_environment`` user-defined job to install Poetry and ``post_install`` to install Python dependencies.
```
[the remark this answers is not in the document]
browniebroke: Updated 😄

juantocamidokura:
Hi @browniebroke. Thank you for your contribution. It was really useful for our team. I was planning to open a similar PR regarding this documentation page https://docs.readthedocs.io/en/stable/guides/poetry.html.

Do you think your PR deprecates it? In our investigation on migrating to Poetry it mislead us on how to integrate it with ReadTheDocs.

Thanks in advance :smile:

browniebroke:
Good question. I think I myself stumbled upon that page a while ago before finding the dedicated example. It might predate the more advance build hooks.
 
I'm not a maintainer, but my guess is that it would be better to open a new issue for it. Closed issues and pull requests are generally not the best place to have discussions.

benjaoming:
It would be great if you can open an issue @juantocamidokura !

A lot of our documentation is being updated these days, and it would be great to have some help from Poetry users to update this particular How-to Guide :pray:

humitos:
> Hi @browniebroke. Thank you for your contribution. It was really useful for our team. I was planning to open a similar PR regarding this documentation page https://docs.readthedocs.io/en/stable/guides/poetry.html.

This guide was written when `build.jobs` and `build.commands` did not exist yet. So, at that time, the only of using poetry is the one described in that guide. I'm not a poetry user, so I'm not sure which approach is better currently.

Issue #9746, opened 2022-11-21
Docs: Diátaxis refactor - iteration 1 high-level tracking


benjaoming:
Successor of #9745 

This is a high-level issue moved from the [initial planning document](https://docs.google.com/document/d/1LQXMUOAKBZ4RhAew2F4TCDp-QW7czH5Dq6BPxlyA_dw/edit#). Further issues may be opened and linked from here.

**Acceptance criteria:** All documentation content is correctly categorized in _Explanation_, _How-to_ and _Reference_, according to Diátaxis. This means moving pages, splitting up pages and writing new content. This also merges all _Business Features_ into _Explanation_ and _How-to_. Removes everything from _Features_ that isn't a feature, and combines _Features_ with _Advanced Features_.

**Outcome:** _Features_ is a list of old feature files without any purely Explanation/Reference/How-To left. _RTD for Business_ and _Advanced Features_ are gone from the sidebar. _About RTD_ is mostly untouched. To avoid scope creep, we are not changing sub-levels of How-to guides but keeping the old ones - new How-to guides can be added casually in those existing levels since it will change in the next iteration.

### Relabel
Existing contents are moved with light changes to title and content. Intentionally not called “move” because we don’t change URLs and break stuff, we just relabel it.

* [x] Relabel "Choosing Between Our Two Platforms" to Explanation https://github.com/readthedocs/readthedocs.org/pull/9784
* [x] Relabel "Subprojects" in Explanation (Title suggestion: "Combining many projects in one: Subprojects") https://github.com/readthedocs/readthedocs.org/pull/9785
* [x] Relabel "Localization of Documentation" in Explanation (Title suggestion: "Localization (Translation): Considerations and workflows") https://github.com/readthedocs/readthedocs.org/pull/9790
* [x] Relabel "Read the Docs for Science" in Explanation (Title suggestion: "Using documentation in Science and Academic Publishing") https://github.com/readthedocs/readthedocs.org/pull/9832
* [ ] Relabel "Build process" in Explanation (Title suggestion: "Build process overview") (@ericholscher)
  * Implement inputs from @agjohnson https://github.com/readthedocs/readthedocs.org/issues/9575
* [x] Relabel "Single Version Documentation" to Explanation (same title?) https://github.com/readthedocs/readthedocs.org/pull/9835
[...]

PR #9677, opened 2022-10-19
Docs: Split Traffic Analytics to a How-to guide and a Feature entry (Diátaxis)

benjaoming:
Refs: #9746 

---
:books: Documentation previews :books:

- User's documentation (`docs`): https://docs--9677.org.readthedocs.build/en/9677/

- Developer's documentation (`dev`): https://dev--9677.org.readthedocs.build/en/9677/

PR #9755, opened 2022-11-23
Docs: hosting integrations for `build.commands`

humitos:
Initial file to start documenting hosting integrations.

This is not ready to review or merge. I'm just pushing this changes to ask for an early quick review to know if this is the direction we want to follow with this document. 

Document at https://docs--9755.org.readthedocs.build/en/9755/guides/hosting-integrations.html

Related #9036
Related https://github.com/readthedocs/meta/issues/71
Related https://github.com/readthedocs/meta/discussions/35

---

- User's documentation (`docs`): https://docs--9755.org.readthedocs.build/en/9755/

- Developer's documentation (`dev`): https://dev--9755.org.readthedocs.build/en/9755/

### DIFF
--- a/docs/user/guides/hosting-integrations.rst
+++ b/docs/user/guides/hosting-integrations.rst
@@ -1,6 +1,7 @@
 Hosting Integrations
 ====================
 
+
 .. This document is not ready to be presented to our users and tell them to integrate with us.
    It has been more useful as an internal exercise to understand how our integrations work,
    and be able to standardize and migrate them out from the build process to a JavaScript library.
--- a/docs/user/guides/hosting-integrations.rst
+++ b/docs/user/guides/hosting-integrations.rst
@@ -42,7 +42,6 @@ These are all the hosting features that Read the Docs provides by default:
    Read more in our blog post
    `Knowing more about how people use our service <https://blog.readthedocs.com/knowing-more-about-ourselves/>`_
 
-
 `Advertisement`_
    Ad shown on documentation pages that helps Read the Docs to be sustainable.
    Read more at :doc:`advertising`.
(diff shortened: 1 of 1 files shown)


docs/user/guides/hosting-integrations.rst
```diff
@@ -0,0 +1,97 @@
...
+`Search index`_
+   Smart search index powered by Elasticsearch able to differentiate between regular text,
+   function names in code, and search in all subprojects, among others.
+   Read more at :doc:`server-side-search`.
+
+`Search as you type`_
+   Level up your search by providing immediate feedback while the user is typing.
+   Read more at :doc:`advanced-search.html#enable-search-as-you-type-in-your-documentation`.
+
+`Version warning`_
```
ericholscher, line 26: Hrm, I thought we had an extension for this?

docs/user/guides/hosting-integrations.rst
```diff
@@ -0,0 +1,97 @@
...
+   Banner displayed on old versions when there is a newer version available,
+   communicating readers they may be careful since they are reading an older version of the documentation.
+   Read more at :doc:`versions#version-warning`.
+
+`External (pull request) version warning`_
+   Benner displayed on versions of the documentation built from an external version (pull request),
+   warning readers this is not the final version of this documentation.
+   Read more at :doc:`pull-requests`.
+
+.. TODO: write a new page explaining this and link it from here.
```
ericholscher, line 36: I'm pretty sure users won't care about integrating this, but the page explaining that we do it with some public metabase charts could be cool.

ericholscher (commented):
This looks like a great start. I think we can probably document this, but then have the goal to have the next step be to wrap all this up so you just need a single JS file that pulls the rest of that stuff in. 👍

docs/user/guides/hosting-integrations.rst
humitos: We have a core feature that handles the common use case _and_ a Sphinx extension that's more configurable. However, IIRC it broke some time ago due to CORS changes.

docs/user/guides/hosting-integrations.rst
humitos: Yeah, I'm not expecting final users to integrate this, but probably doctools/themes/etc. I think it's fine to have it documented and let people decide whether or not they want to support us by dumping this data. The more data they dump, the better support we can offer to them. So, it's a win-win in the end.

docs/user/guides/hosting-integrations.rst
```diff
@@ -0,0 +1,97 @@
+Hosting Integrations
+====================
+
+Read the Docs performs some actions by default when building documentation for its users to.
```
benjaoming, line 4: It's like something is off with this sentence?

But a good way to start it, could be like writing to the reader and the author and reviewers: "In this section, you can learn about <...>".

benjaoming (commented):
I know this is a draft, so I want to just leave a few general inputs :)

* I think that this is _explanation_? In this regard, we should consider moving it out of the `guides/` folder because that folder is _mostly_ How-to. It would make a lot of sense if you already now suggest a folder for explanation, perhaps it's as easy as `explanation/`? (none of the current iterations have any proposals on where to put explanation)

* Start the section by writing _what it is about_. This might be shortened later, but it's really helpful in the beginning.

* Flyout menu: The word "Flyout" is super weird. I've put it in actual quotes in some page titles. Even what it refers to is confusing to me :D

>  A flyout occurs when a batter hits the ball in the air (not including balls designated as line drives) and an opposing defender catches it before it hits the ground or fence. 

 :question:  :confused: 

If you happen to think of something, please say so.. already in Iteration 1 https://github.com/readthedocs/readthedocs.org/issues/9746 we are looking at the pages about "Flyout" menus.

You've started explaining the "Flyout" menu already in a subsection, so that might also be a reason to coordinate things closely. Otherwise, this work might be easier in Iteration 3 or 4 when we have wrestled with the existing contents and have started to focus more on developing new content.

docs/user/guides/hosting-integrations.rst
humitos: I updated this a little to mention this as the first sentence. I think it's better now 👍🏼

humitos:
@ericholscher I updated this page a little more adding more explanations for some of the other features. I added some notes to consider/discuss/talk about because telling people to "include this chunk of HTML" won't work for some features (e.g. version warning).

I think it is still fine to communicate "what we do", but I don't think it works as "integrate these features in your doctool" yet. IMO, we are not there yet and we should decouple them from being Sphinx extensions and migrate them into a Javascript library that does all this work (for the version warning example, adds the HTML, put the CSS to make it sticky at the top, generate the correct links, etc).

Writing this document is being a good exercise to realize what is required to make it fully compatible and generic enough, 💯

docs/user/guides/hosting-integrations.rst
```diff
@@ -0,0 +1,211 @@
...
+   Read more at :doc:`advertising`.
+
+
+.. TODO: explain "Visual diff" when we have it available for our users.
+
+
+Flyout menu
+-----------
+
+Displying the flyout menu requires adding some Javascript and CSS files in each of the documentation pages.
```
benjaoming, line 56:
```suggestion
Displaying the flyout menu requires adding some JavaScript and CSS files in each of the documentation pages.
```

docs/user/guides/hosting-integrations.rst
benjaoming: I still didn't entirely understand it, maybe it's just a redundant "to" at the end.

Here's what I mean by introducing the article in a way that tells both the author and the reader what the scope and intention of the article is... I think it's nice to start with, it can maybe be trimmed later.

```suggestion
Read the Docs performs a set of default actions to build and publish your documentation. In this article, we explain more about these actions, how they work, what they produce, how they relate. Ultimately we give you pointers on how to configure them.
```

ericholscher:
@benjaoming Flyout is a standard term for what we're doing.. https://elementor.com/resources/glossary/what-is-flyout-navigation/

benjaoming:
Good reference to know. It'd still be interesting to figure out a better name. For now, I think we should resort to be careful to always mention it with a definition or reference.

To me as a reader, the word "flyout" is kind of alienating if it doesn't come with a definition or reference.

ericholscher:
@benjaoming Yea, I think in general we try to `:term:` it.

docs/user/guides/hosting-integrations.rst
benjaoming: Great change :+1: 

I see also that "explain" is used, which is really good. At the end of content reorganizing, the idea of having a real reference for features will come up (so let's hope we are there in 1 month :crossed_fingers: ). In that regards, this overview and explanation will be valuable because it means that the feature reference can simply link to the explanation article.

humitos:
📓 Note to myself:

> If you require more detailed analytics, Read the Docs has native support for Google Analytics.

(from https://github.com/readthedocs/readthedocs.org/pull/9677/files#diff-98cc398741c4326533ef8f26b360e24f914d4ef3998e69474d10a163e42eaa4dR87)

We need to explain here how this works and _where_ this is injected? How do we plan this to work on non-Sphinx projects?

humitos:
This is another potentially good integration feature: https://github.com/readthedocs/readthedocs.org/issues/9854

humitos:
I'm closing this PR for now since is superseded by #10127. The important conversation will be happening there.

PR #9758, opened 2022-11-24
Docs: 4 diátaxis categories at the top of the navigation sidebar (Diátaxis iteration 0)

benjaoming:
Notice that this PR targets `diataxis/main`

Refs #9745

I opted to having all 4 categories at the top. That makes it easy to observe the end-result as well as grasping how much work is left (i.e. all the sections that are still left in the old categories below).


ericholscher (approved):
Simple enough 👍

PR #9759, opened 2022-11-25
Docker Compose v2 compatibility

benjaoming:
This PR adds an important line in our `.env`.

The issue with Docker Compose v2 is that they decided to use `-` as a separator instead if `_`. So for instance, images are then tagged as `community-server:latest` and not `community_server:latest` etc. `COMPOSE_COMPATIBILITY=true` fixes this.

~I wrote something in the documentation about this, and I also added a troubleshooting section on Docker networking based on fresh experience.~ Removed again due to push-back.

In the future, once everyone on the team is using Docker Compose v2, we could remove the option and replace underscores with dashes.

---

- User's documentation (`docs`): https://docs--9759.org.readthedocs.build/en/9759/

- Developer's documentation (`dev`): https://dev--9759.org.readthedocs.build/en/9759/

### DIFF
--- a/.env
+++ b/.env
@@ -1,1 +1,4 @@
 COMPOSE_PROJECT_NAME=community
+# Current setup assumes Docker Compose v1. In v2, underscores are replaced by dashes in image and network
+# names, which breaks the assumption.
+COMPOSE_COMPATIBILITY=true
--- a/.env
+++ b/.env
@@ -1,4 +1,5 @@
 COMPOSE_PROJECT_NAME=community
-# Current setup assumes Docker Compose v1. In v2, underscores are replaced by dashes in image and network
-# names, which breaks the assumption.
+# Current setup assumes Docker Compose v1.
+# In v2, underscores are replaced by dashes in image and network names,
+# which breaks the assumption.
 COMPOSE_COMPATIBILITY=true
--- a/docs/dev/install.rst
+++ b/docs/dev/install.rst
@@ -279,58 +279,6 @@ you have to follow these steps:
 Once this is done, you should be able to trigger a new build on that project and it should succeed.
 
 
-Network connection failures
-~~~~~~~~~~~~~~~~~~~~~~~~~~~
-
-After starting up your services with ``inv docker.up``, you might not be able to connect to the ``web`` instance on ``http://devthedocs.org:80``.
-Firstly, check all the outputs from Docker Compose to see that one of the instances hasn't crashed for a particular reason.
-If everything is up and running, it's likely related to a Docker network issue.
-
-In many cases, it can help to bring down any existing containers and docker networks.
-
-.. code-block:: console
-
-    # Give Docker Compose a chance to clean up what it can
-    inv docker.down
-    # After this step, you can try to start Docker Compose again
-    inv docker.up
-
-
-If it still doesn't work, we need to gather some more information to understand the issue.
-Have a look at how your networking traffic is routed:
-
-.. code-block:: console
-
-    # Run the command "route" to see how network traffic is routed.
-    route
-
-This should contain a network interface ``br-<docker-id>`` handling traffic for the subnet ``10.10.0.0``.
-If you have several networks handling traffic for that subnet, that's an issue you need to solve.
-This can happen because of Docker but you might also have other services occupying the subnet.
-
-It's a good idea to run ``docker network prune`` to remove Docker's unused network devices.
-If the problem persists, you may want to reboot or manually remove the network device,
-for instance with ``sudo ifconfig <device> down``.
-
-You may also want to have a look at how the containers are connected to the network:
-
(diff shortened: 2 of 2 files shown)


.env
```diff
@@ -1 +1,4 @@
 COMPOSE_PROJECT_NAME=community
+# Current setup assumes Docker Compose v1. In v2, underscores are replaced by dashes in image and network
+# names, which breaks the assumption.
```
humitos, line 3:
```suggestion
# Current setup assumes Docker Compose v1.
# In v2, underscores are replaced by dashes in image and network names,
# which breaks the assumption.
```

Just use semantic line break

docs/dev/install.rst
```diff
@@ -267,6 +278,50 @@ you have to follow these steps:
 
 Once this is done, you should be able to trigger a new build on that project and it should succeed.
 
+
+Network connection failures
```
humitos, line 282: I think this section goes too deep into pretty technical details that may not be required, making the section more confusing.

I've had similar problems and the solution that I applied was simpler: `inv docker.down` which deletes all the containers, networks and everything created by our own application.

IMO, telling people to modify `/etc/hosts` and debug using `route`, `docker network inspect` or similar is out of the scope and will only confuse users.

docs/dev/install.rst
```diff
@@ -267,6 +278,50 @@ you have to follow these steps:
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+After starting up your services with ``inv docker.up``, you might not be able to connect to the ``web`` instance on ``http://devthedocs.org:80``. Firstly, check all the outputs from Docker Compose to see that one of the instances hasn't crashed for a particular reason. If everything is up and running, it's likely related to a Docker network issue.
```
humitos, line 285: Remember to use semantic break lines when writing documentation / comments. There are some paragraph that should be reviewed in this file.

humitos (commented):
I think it's fine to document Docker Compose v2 for now. However, I think the real solution here is to upgrade the instructions/code to use Docker Compose v2 instead of maintaining compatibility with both of them.

Also, I'd remove the section about networking issue completely, or reduce it to just recommend using `inv.docker. down` instead that deep debugging process.

docs/dev/install.rst
benjaoming: It's a troubleshooting section, so I suppose that no one will read it unless they have "Network connection failures".

Great tip to remember to run `inv docker.down`, that should perhaps be the first simple step.

benjaoming: Oh and I have removed the stuff about `/etc/hosts`, it's confusing and superstitious.

humitos: Yeah, my point is that this is going too deep pretty quickly into a rabbit hole. Telling people to execute `docker network inspect community_readthedocs` which outputs a blob of JSON without telling them what to look at is more confusing than helpful.

Also, I don't think we want to deal with all of this together with the users. It seems a pretty edge case to have it here. As an example, I just executed `route` and I don't even have that command installed 🙃 

Besides, `docker network prune` will prune other Docker networks that are not created by Read the Docs, which could break other Docker setup the user may have. There are more, like telnet to the database without more instructions; but I don't think that writing _all of these instructions is the solution_, I think the solution is to not have this section 😄 

Summarizing, I'm +1 documenting the troubleshooting for the most common cases and guide the user towards the solution. On the other hand, I'm -1 on documenting edge cases like this one and go too deep on the debug. That will cause us more support requests to us than benefits to the users.

benjaoming:
> However, I think the real solution here is to upgrade the instructions/code to use Docker Compose v2 instead of maintaining compatibility with both of them.

I think it would be great to do this, once everyone on the team has updated. This should make it possible to do the update without any issues :+1:

docs/dev/install.rst
benjaoming: Added `inv docker.down` as the first step.

ericholscher: Yea, I'm on board with Manuel that this seems super deep and specific. I'm 99% sure we won't maintain this content as the commands change, and it will just become technical debt. I think documenting the simple case is worth doing, but the rest seems like nobody will ever need it. We can put that in an issue somewhere if you want it to be Googleable or something?

benjaoming: Sure, will remove it :+1: Not sure if `route` will need to be maintained much :)

As a background, I have had several encounters where understanding that Docker subnets can have local issues becomes **super costly** in local debugging. In other encounters, I've had with this, the Docker subnet overlapped with a real subnet. That's why I added it, it was meant as a 101 to quickly debug such issues.

benjaoming: Now it's removed.

benjaoming:
Merging despite test failures, seems more of an issue with Docker Hub.

Issue #9577, opened 2022-08-31
Docs: update downloadable documentation page


agjohnson:
From #9258 

  - "Offline formats" is the nomenclature used in new application templates and
    the website
  - (https://docs.readthedocs.io/en/latest/downloadable-documentation.html)

PR #9686, opened 2022-10-21
Docs: Split up Build Notifications into feature/reference and how-to (Diátaxis)

benjaoming:
This made a lot of sense.

I had to write new text for the old feature description, though. But I think that text was already missing in the former version.

PR #9768, opened 2022-11-30
Refactor downloadable docs

ericholscher:
Refs https://github.com/readthedocs/readthedocs.org/issues/9746
Closes https://github.com/readthedocs/readthedocs.org/issues/9577

---

- User's documentation (`docs`): https://docs--9768.org.readthedocs.build/en/9768/

- Developer's documentation (`dev`): https://dev--9768.org.readthedocs.build/en/9768/


humitos (commented):
When splitting content into explanation and how-to, shouldn't we link them each other somehow? I mean, as a user, after reading the explanation and I will be excited about offline formats and I'd want to know how to enable the in my project.

It works in the other direction too, I guess. Arriving at the "how-to" without knowing exactly what they are, I may be lost. So, a link back to the explanation in the first or second paragraph could work great here.

docs/user/downloadable-documentation.rst
```diff
@@ -1,50 +1,42 @@
...
-you can download the Read the Docs documentation in the following formats:
-
-    * `PDF`_ 
-    * `ePub`_ 
-    * `Zipped HTML`_ 
-    
-.. _PDF: https://docs.readthedocs.io/_/downloads/en/latest/pdf/
-.. _ePub: https://docs.readthedocs.io/_/downloads/en/latest/epub/
-.. _Zipped HTML: https://docs.readthedocs.io/_/downloads/en/latest/htmlzip/
+This means that every commit that you push will automatically update your these offline formats as well as your HTML.
```
benjaoming, line 10:
```suggestion
This means that every commit that you push will automatically update your offline formats as well as your documentation website.
```

docs/user/downloadable-documentation.rst
```diff
@@ -1,50 +1,42 @@
-Downloadable Documentation
-==========================
+Offline Formats
+===============
 
-Read the Docs supports building multiple formats for Sphinx-based projects:
+Read the Docs supports building multiple offline formats:
```
benjaoming, line 4: I guess we should remove "Sphinx-based" because the build process can be configured for any kind of documentation stack?

Should we clarify in a separate section that the easiest integration happens with Sphinx, but we support any kind of build process and making different formats available is a matter of moving build artifacts into the correct location at the end of a build?

humitos:
> When splitting content into explanation and how-to, shouldn't we link them each other somehow? I mean, as a user, after reading the explanation and I will be excited about offline formats and I'd want to know how to enable the in my project.

I think this can be standardized in a simple way to start. Maybe just a `.. seealso::` with some bullets and links at the end of the pages?

```rst

.. seealso::

  * :doc:`downloadable-documentation`
```

benjaoming:
@humitos yup, definitely, we also added them in previous Diátaxis refactors, although the text has varied a bit. I don't think we're far from arriving at some kind of template. I'm trying to compile generic inputs here: https://docs.google.com/document/d/1Ir8UvupznvAH4oF4-DEiXrILnEk2Lqsg4FrPDA-8uXA/edit#

docs/user/downloadable-documentation.rst
```diff
@@ -1,50 +1,42 @@
...
 Use cases
 ---------
 
 This functionality is great for anyone who needs documentation when they aren't connected to the internet.
 Users who are about to get on a plane can grab a PDF and have the entire doc set ready for their trip.
 
-The other value is having the entire docset in a single file.
+The other value is having the entire documentation in a single file.
 You can send a user an email with a single PDF or ePub and they'll have all the docs in one place.
+This is useful for archiving documentation, as well as distributing it in other scenarios.
```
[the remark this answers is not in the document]
ericholscher: Yea, I just didn't want to get too deep into use cases, but I do think it's a core benefit so probably worth expanding 👍

docs/user/downloadable-documentation.rst
ericholscher: > I guess we should remove "Sphinx-based" because the build process can be configured for any kind of documentation stack?

It's still Sphinx-only for now, but we're actively working to remove that limitation.


> Should we clarify in a separate section that the easiest integration happens with Sphinx, but we support any kind of build process and making different formats available is a matter of moving build artifacts into the correct location at the end of a build?

Yes, when it's true :)

ericholscher:
> @humitos yup, definitely, we also used `seealso::` in [other Diátaxis refactors](https://github.com/readthedocs/readthedocs.org/pull/9686/files#diff-42b9bd40dbe7427acedde46addb77852a542418ddfe88ade024d402cc031c83eR19), although the text has varied a bit. I don't think we're far from arriving at some kind of template. I'm trying to compile generic inputs here: https://docs.google.com/document/d/1Ir8UvupznvAH4oF4-DEiXrILnEk2Lqsg4FrPDA-8uXA/edit#

Strong agree on the seealso, I just ran out of energy last night when working on it 😆

docs/user/downloadable-documentation.rst
benjaoming: Ah yeah, referring to the conversation today, the output path of other formats (such as a PDF version) should be accessed through a build environment variable defined in a real build API of sorts?

Is it fine to keep the Sphinx part sort of bit implicit because we're gonna add it very soon?

ericholscher: Now I"m confused, since this comment is on a line where I removed the Sphinx-based part?

benjaoming: I think your original suggestion should stand :+1: 

To clarify what I mean is that it's fine to remove it *because we're gonna add wider support very soon*. Until then, it can keep being implicit or unspoken whether it's only for Sphinx or not.

* Only Sphinx: Well yes, that's the way it's supported and documented
* Wider support: Well it's actually possible, but it's a longer story that we won't document until it's through an official API.

ericholscher:
Updates are ready for review 👍

ericholscher:
@benjaoming Would love a review on this soon, so we can get it merged.

docs/user/downloadable-documentation.rst
```diff
@@ -1,50 +1,66 @@
...
-Use cases
----------
-
-This functionality is great for anyone who needs documentation when they aren't connected to the internet.
-Users who are about to get on a plane can grab a PDF and have the entire doc set ready for their trip.
-
-The other value is having the entire docset in a single file.
-You can send a user an email with a single PDF or ePub and they'll have all the docs in one place.
+Continue learning
+-----------------
```
benjaoming, line 58:
```suggestion
Continue learning
-----------------

Downloadable documentation formats are built by your documentation framework. Therefore, it's also your framework that decides exactly how each output is built and which formats are supported:

Sphinx
   All output formats are built mostly lossless from the documentation source, meaning that your documentation source (reStructuredText or Markdown/MyST) is built from scratch for each output format.

MkDocs and Docsify + more
   The common case for most documentation frameworks is that several alternative extensions exist supporting various output formats. Most of the extensions export the HTML outputs as another format (for instance PDF) through a conversion process.

Because Sphinx supports the generation of Downloadable formats through an official process, we are also able to support it officially. Other alternatives can also work, provided that you identify which extension you want to use and configure the environment for it to run.
```

Elaborating this comment:

* As mentioned earlier, it seemed misleading to _require_ Sphinx in our How-To since there are other ways to generate downloadable formats. So we removed that part.
* We are awaiting official support from an Environment variable to dictate the expected location of supported output formats. So our efforts here could "make space" for this.
* Ultimately, I thought adding something in the explanation here is appropriate? I thought that it would be great to put under "Continue learning" since learning more probably relies on the documentation framework.

I don't know if what I wrote gives away a slight bias towards Sphinx here, but I think that it really explains why it is that these formats are supported and the others are unspecified.

Consider removing the last paragraph :+1:

benjaoming (approved):
I have sync'ed branches, which worked fine.

Have just the one suggestion above to add some kind of explanation about the role of the documentation framework.

Otherwise, looks great, really nice to get started with this, as there are many additional things in the horizon. Noticing both custom build processes, but also suggestions to add DocSets support :+1:

docs/user/downloadable-documentation.rst
ericholscher: Edited to add a small caveat, and shipped 👍

PR #9777, opened 2022-12-05
Treat latest pre-release as stable when no other stable releases

davfsa:
## Quick summary

I recently started using readthedocs to host the documentation of one of my projects and found out that `stable` was pointing to an old stable tag, so I went ahead and deleted the really old stable tag to hopefully get it to detect my latest development release as the newest stable, but got no luck in it, `stable` got stuck in the commit that that old tag was tagged to.

This PR changes the behavior of this to fallback to the newest pre-release as the most "stable" one, which is more in-line to what other services provide (ie, PyPI).

~~*I was unable to get the test suite running locally, so there is a very high chance that this PR will not pass it. I read through the Developer Guide to no luck, so if it does fail, I will make sure to fix it right away :)*~~

### DIFF
--- a/readthedocs/projects/version_handling.py
+++ b/readthedocs/projects/version_handling.py
@@ -138,25 +138,34 @@ def determine_stable_version(version_list):
     :param version_list: list of versions
     :type version_list: list(readthedocs.builds.models.Version)
 
-    :returns: version considered the most recent stable one or ``None`` if there
-              is no stable version in the list
+    :returns: version considered the most recent stable one, the latest
+              pre-release if there is no stable version in the list or
+              ``None`` if there are no versions in the list
 
     :rtype: readthedocs.builds.models.Version
     """
     versions = sort_versions(version_list)
-    versions = [
-        (version_obj, comparable)
+
+    if not versions:
+        # No versions, so no info we can give
+        return None
+
+    stable_versions = [
+        version_obj
         for version_obj, comparable in versions
         if not comparable.is_prerelease
     ]
 
-    if versions:
+    if stable_versions:
         # We take preference for tags over branches. If we don't find any tag,
         # we just return the first branch found.
-        for version_obj, comparable in versions:
+        for version_obj in stable_versions:
             if version_obj.type == TAG:
                 return version_obj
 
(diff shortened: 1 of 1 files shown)


davfsa:
Thanks for the fast response!

> It seems that you are giving to `stable` a new and different interpretation of what it is currently. We don't treat alpha, release candidates, or pre-releases as `stable`. Only `x.y.z` are treated as `stable` for Read the Docs. Pre-releases fit better under what Read the Docs calls `latest`, instead. Take a look at https://docs.readthedocs.io/en/stable/versions.html

I did have a look at `latest`, but ran into some limitations within it. My specific use case is a library for which there are only pre-releases. There used to be a stable version, but we have been working for a long time in only development releases until we finalize it. ReadTheDocs seems to pick up an old stable release as the "stable" one (one for which the tag doesnt exist anymore) and I have no control over changing it. 

My ideal state would be for both `stable` and `latest` to point to the same version (the latest pre-release) until we release the first stable release (as denoted by the sysver), at which point the tags will unsync and the latest will build from the latest release and stable from the stable one.

> It seems that you are giving to `stable` a new and different interpretation of what it is currently.

This would have no effect on any projects that already make use of `stable` and `latest` independently. It would just add an extra `stable` to those that (like ours) only have pre-releases, until they stabilize. This means they can keep development and have urls they link around pointed to `/stable/`, which will point to the newest docs during development and later unsync to the latest release.


This behaviour would be similar to what PyPI does, where it shows the latest release if they are no stables ones (example: https://pypi.org/project/hikari/)

davfsa:
> You can create a `stable` branch/tag pointing to the pre-release you want. In that case, Read the Docs won't create/assign the "automatic `stable`" version and use the one you created. Note that you can do the same with `latest`, in case you want both to point to the same version.

That could be an option I considered (as stated in the docs), but it would be nice to change this behaviour on ReadTheDocs side, as it feels more intuitive (at least for me), specially when compared to other services like PyPI, considering they are in the same environment

Regardless, thank you for taking the time to consider this pr :)

davfsa:
I realized I never closed this PR :sweat_smile:

PR #9783, opened 2022-12-07
Abandoned Projects policy: Relax reachability requirement

benjaoming:
There has been a lot of different suggestions up, but I think this one is the easiest to process for now.

---

- User's documentation (`docs`): https://docs--9783.org.readthedocs.build/en/9783/

- Developer's documentation (`dev`): https://dev--9783.org.readthedocs.build/en/9783/

### DIFF
--- a/docs/user/abandoned-projects.rst
+++ b/docs/user/abandoned-projects.rst
@@ -41,8 +41,8 @@ Reachability
 
 The user of Read the Docs is solely responsible for being reachable by the core team
 for matters concerning projects that the user owns. In every case where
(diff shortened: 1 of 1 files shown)


agjohnson (commented):
This matches what I do by default anyways :+1: 

Most often the GitHub repo is removed, the user has no public contact info in GitHub, and the documentation is our default page or is a fork of an existing doc set anyways. So, I don't see much need to even look at sources other than the email we have on file, but good to have defined here.

ericholscher (approved):

PR #9787, opened 2022-12-08
CI Builds: Conditional builds with path filtering

benjaoming:
Benefits of conditional builds:

1. Tests are over 45 minutes, so they are valuable to skip them when it's easy and risk-free.
1. Documentation PRs (we have a lot of them) will have a much quicker feedback time.
 
Because of the many current documentation PRs, the timing for "conditional builds" will be really good.

* [x] Step 1: Make path filtering work
* [ ] Step 2: Discuss and adjust the actual paths and filters!
* [ ] Step 3: Apply path filters (and other conditions that we might want)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -17,3 +17,6 @@ workflows:
             pytest.* build-code true
             tox.* build-code true
             package.* build-code true
+            \.circleci/* build-code true
+            common build-code true
+            setup.* build-code true
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,67 +1,19 @@
 version: 2.1
 
+setup: true
+
 orbs:
-  codecov: codecov/codecov@3.2.2
-
-jobs:
-  tests:
-    docker:
-      - image: 'cimg/python:3.10'
-        environment:
-          TOX_POSARGS: ''
-      - image: 'docker.elastic.co/elasticsearch/elasticsearch:7.14.0'
-        name: search
-        environment:
-          discovery.type: single-node
-          ES_JAVA_OPTS: -Xms750m -Xmx750m
-    steps:
-      - checkout
-      - run: git submodule sync
-      - run: git submodule update --init
-      - run: pip install --user 'tox<5'
-      - run: tox -e py310
-      - codecov/upload
-
-  tests-embedapi:
-    docker:
-      - image: 'cimg/python:3.10'
-    steps:
-      - checkout
-      - run: git submodule sync
-      - run: git submodule update --init
-      - run: pip install --user 'tox<4'
-      - run: tox -c tox.embedapi.ini
-
-  checks:
-    docker:
-      - image: 'cimg/python:3.10'
-        environment:
-          NODE_VERSION: 10.17.0
-    steps:
-      - checkout
-      - run: git submodule sync
-      - run: git submodule update --init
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -1,67 +1,20 @@
...
-            - tests
+      - path-filtering/filter:
+          base-revision: main
+          config-path: .circleci/config-conditional-builds.yml
+          mapping: |
+            requirements/.* build-code true
+            readthedocs/.* build-code true
+            pytest.* build-code true
+            tox.* build-code true
+            package.* build-code true
```
benjaoming, line 19: These are essential bits to adjust. We want to run this stuff whenever something essential is changed.

2 considerations that we might have for later:

1) Breaking up checks and tests
2) Branch rules: I think we should also consider an alternative pipeline that always runs on the main branch and sets `build-code` to `true` regardless of the path. This requires adding a separate pipeline that runs on `main` (and configuring the other pipeline to NOT run on `main`), calling the [continuation Orb](https://circleci.com/developer/orbs/orb/circleci/continuation#orb-source).

.circleci/config.yml
```diff
@@ -1,67 +1,20 @@
...
+            docs/.* build-docs true
```
benjaoming, line 20: This part can be removed if https://github.com/readthedocs/readthedocs.org/pull/9788 is merged.

benjaoming:
> That said, what are the exact cases where we want to avoid running the tests? I suppose the only one is when only modifying the docs/ directory since they will be run by Read the Docs.

That's what this is mostly about, yes :+1: 

There's 40 min+ test running time for no reason.

.circleci/config.yml
```diff
@@ -1,67 +1,19 @@
...
+  generate-config:
     jobs:
-      - checks
-      - tests
-      - tests-embedapi:
-          requires:
-            - checks
-            - tests
+      - path-filtering/filter:
+          base-revision: main
```
benjaoming, line 12: When we target branches different from `main`, there might be some false positives. I've opened up a feature request: https://github.com/CircleCI-Public/path-filtering-orb/issues/65

benjaoming:
This is now ready for review :+1: 

I imagine the following issues, which IMO are interesting for further study:

* Merge commits might trigger redundant builds. This is low-risk since it means building _too much_.
* A series of commits pushed at the same time aren't compared correctly to the base branch (`main`) AFAICT. This can get annoying. The workarounds are numerous and quite easy. So I wouldn't expect it to be a big problem.

Our safety nets are:

* `main` is always built

We can expand rules as we become more clever.

humitos:
I'm 👍🏼 by closing this PR and continue using the workflow we have. The documentation build is skipped now with the latest changes in our code. On the other hand, I prefer to not deal with edge cases for the test cases and always be sure they are running. That gives me confidence about merging the PR.

Feel free to re-open if you consider it's important to continue with this work and you think there are some room for prioritization here.

benjaoming:
Seems good to close it, since it won't stay ready without a lot of heavy maintenance. It was pretty hard to maintain this PR because it refactored the Circle CI configuration that was constantly changing a while back (new Tox etc).

I wanted this change because when working on documentation, I didn't need the whole platform CI to trigger and it became a huge distraction. Once others realize that the 1 hour build times also annoy them, you know where to find me :smile:

PR #9789, opened 2022-12-08
Fix Circle CI builds: Tox 4 compatibility, add external commands to allowlist

benjaoming:
Since we haven't pinned tox, we will have to make adjustments quite quickly when they introduce breaking changes.

See:

https://tox.wiki/en/latest/faq.html#breaking-changes-in-tox-4

Because it took a long time to get build outputs, I did a small optimization of the order of pipeline jobs so all the node-based stuff is pushed to the end.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -29,7 +29,7 @@ jobs:
       - checkout
       - run: git submodule sync
       - run: git submodule update --init
-      - run: pip install --user tox
+      - run: pip install --user 'tox<4'
       - run: tox -c tox.embedapi.ini
 
   checks:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -41,6 +41,8 @@ jobs:
       - checkout
       - run: git submodule sync
       - run: git submodule update --init
+      - run: git fetch origin main  # needed for comparisson in pre-commit
+      - run: git branch --track main origin/main  # needed for comparisson in pre-commit
       - run: pip install --user tox
       - run: scripts/circle/install_node.sh
       - run:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -44,17 +44,17 @@ jobs:
       - run: git fetch origin main  # needed for comparisson in pre-commit
       - run: git branch --track main origin/main  # needed for comparisson in pre-commit
       - run: pip install --user tox
+      - run: tox -e pre-commit
+      - run: tox -e migrations
+      - run: tox -e lint
+      - run: tox -e docs
+      - run: tox -e docs-dev
       - run: scripts/circle/install_node.sh
       - run:
           name: Add node to the path
           command: |
             echo 'export PATH=~/.nvm/versions/node/v${NODE_VERSION}/bin:$PATH' >> $BASH_ENV
             source $BASH_ENV
-      - run: tox -e migrations
-      - run: tox -e pre-commit
-      - run: tox -e lint
-      - run: tox -e docs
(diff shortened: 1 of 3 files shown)


benjaoming:
I'm not sure what the new behavior here is, might be that we need to change directory when running `pre-commit` since it isn't finding any git data.

.circleci/config.yml
```diff
@@ -41,18 +41,20 @@ jobs:
       - checkout
       - run: git submodule sync
       - run: git submodule update --init
+      - run: git fetch origin main  # needed for comparisson in pre-commit
+      - run: git branch --track main origin/main  # needed for comparisson in pre-commit
```
benjaoming, line 45: This apparently now required.

Before this, I was getting:

```
[INFO] Restored changes from /home/circleci/.cache/pre-commit/patch1670506529-1455.
An unexpected error has occurred: CalledProcessError: command: ('/usr/bin/git', 'diff', '--name-only', '--no-ext-diff', '-z', 'main..HEAD')
return code: 128
expected return code: 0
stdout: (none)
stderr:
    fatal: ambiguous argument 'main..HEAD': unknown revision or path not in the working tree.
    Use '--' to separate paths from revisions, like this:
    'git <command> [<revision>...] -- [<file>...]'
    
Check the log at /home/circleci/.cache/pre-commit/pre-commit.log
pre-commit: exit 3 (15.94 seconds) /home/circleci/project> pre-commit run --from-ref main --to-ref HEAD pid=1455
  pre-commit: FAIL code 3 (19.02=setup[3.08]+cmd[15.94] seconds)
  evaluation failed :( (19.19 seconds)

Exited with code exit status 3
```

.circleci/config.yml
```diff
@@ -41,18 +41,20 @@ jobs:
...
       - run: pip install --user tox
+      - run: tox -e pre-commit
+      - run: tox -e migrations
```
ericholscher, line 48: Any reason we swapped these around?

.circleci/config.yml
```diff
@@ -41,18 +41,20 @@ jobs:
       - checkout
       - run: git submodule sync
       - run: git submodule update --init
+      - run: git fetch origin main  # needed for comparisson in pre-commit
+      - run: git branch --track main origin/main  # needed for comparisson in pre-commit
       - run: pip install --user tox
```
ericholscher, line 46: Shouldn't we pin this too?

ericholscher (approved):
Simple enough... I don't know if we have anything downstream of this depend on the tox naming, but I wouldn't expect so.

.circleci/config.yml
benjaoming: I saw the build times and migrations had a long running time, so in order to have a fail-fast for linting, I put it higher.

I think it should ultimately have its own pipeline.

.circleci/config.yml
benjaoming: It's not currently monitored by anything so I wanted to "cope" with the failures by fixing them. We won't get alerts when there are upstream changes. We could consider `tox<5` to be fairly low-risk.

benjaoming: Pinning is also a good idea because all currently-open PRs now have to merge these fixes.

ericholscher: Yea, I'm not sure why we weren't pinning this before, but it's definitely a best practice. So we should pin each of them.

benjaoming: Added the upper bound :+1:

benjaoming:
Merging :+1:

PR #9675, opened 2022-10-19
Docs: Split and relabel VCS integration as explanation and how-to (Diátaxis)

benjaoming:
* Split article into explanation and how-to
* Add more content to explanation
* Systematic change: We replace "VCS" with "git" and "git platform".

---
:books: Documentation previews :books:

- User's documentation (`docs`): https://docs--9675.org.readthedocs.build/en/9675/

- Developer's documentation (`dev`): https://dev--9675.org.readthedocs.build/en/9675/

PR #9790, opened 2022-12-08
Docs: Relabel Localization as Explanation (Diátaxis)

benjaoming:
Refs: #9746 

---

- User's documentation (`docs`): https://docs--9790.org.readthedocs.build/en/9790/

- Developer's documentation (`dev`): https://dev--9790.org.readthedocs.build/en/9790/


benjaoming:
@ericholscher it's possible to have a lot of improvements here. But I feel like this is a large subject that we might want to push to later iterations?

One of the improvements would be a diagram for the translation workflow. We already have "add a diagram" opened as a suggestion in https://github.com/readthedocs/readthedocs.org/pull/9675 and I think we should be careful about how many times we want diagrams added before we should decide on an approach to easily have Mermaid diagrams added (it eventually wasn't that easy in my first attempt because of small technical reasons).

ericholscher:
Yea, I think a small update here is best. We don't get a ton of users using this feature, so I'm less worried about it's overall quality.

docs/user/localization.rst
```diff
@@ -1,17 +1,25 @@
...
+In this article, we explain high-level approaches to internationalizing and localizing your documentation.
 
-Read the Docs supports hosting your docs in multiple languages.
-There are two different things that we support:
+By default, Read the Docs assumes that your documentation is or *might become* multilingual one day.
+The initial default language is English and
+therefore you often see the initial build of your documentation published at ``/en/latest/``,
+where the ``/en`` denotes that it's in English.
+By having the ``en`` URL component present from the beginning,
+you are ready for the eventuality that you would want a second language.
```
ericholscher, line 11: This is a great intro 👍

docs/user/localization.rst
```diff
@@ -1,17 +1,25 @@
...
 
-* A single project written in another language
-* A project with translations into multiple languages
+Read the Docs supports hosting your documentation in multiple languages.
```
ericholscher, line 13:
```suggestion
Read the Docs supports hosting your documentation in multiple languages.
Read below for the various approaches that we support.
```

A bit short and not fully explained?

docs/user/localization.rst
```diff
@@ -1,17 +1,25 @@
...
 
+.. contents::
+    :local:
```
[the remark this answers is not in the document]
ericholscher: Seems fine.

docs/user/localization.rst
```diff
@@ -22,10 +30,13 @@ The language will be represented in the URL for your project.
 For example,
 a project that is in Spanish will have a default URL of ``/es/latest/`` instead of ``/en/latest/``.
 
-.. note:: You must commit the ``.po`` files for Read the Docs to translate your documentation.
+Projects with multiple translations (Sphinx)
+--------------------------------------------
```
ericholscher, line 34:
```suggestion
Projects with multiple translations (Sphinx-only)
-------------------------------------------------
```

docs/user/localization.rst
```diff
@@ -44,7 +55,7 @@ This has the results of serving:
 * ``phpmyadmin`` at ``http://phpmyadmin.readthedocs.io/en/latest/``
 * ``phpmyadmin-spanish`` at ``http://phpmyadmin.readthedocs.io/es/latest/``
 
-It also gets included in the Read the Docs flyout:
+It also gets included in the Read the Docs :term:`flyout menu`:
```
ericholscher, line 58: :100:

ericholscher (approved):
Looks great 👍

docs/user/localization.rst
benjaoming: Yes, it's nice with the extra sentence here :100:

docs/user/localization.rst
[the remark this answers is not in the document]
benjaoming: Great, thanks for looking, I think it works well like this, too, just really needed a second opinion :+1: :heart:

docs/user/localization.rst
benjaoming: Thanks :blush: 

I hope we can have some time to expand fresh ideas in this section, but I think we should do it in another iteration. Will mark for later :+1:

Issue #9791, opened 2022-12-08
Conditionally skipped builds are "pending" instead of "passed"


benjaoming:
PRs end up having the "pending" state although a build has been correctly and intentionally skipped.

Example:

![image](https://user-images.githubusercontent.com/374612/206479658-7bdcf10d-06cb-4107-8c58-32c75e9d132a.png)

![image](https://user-images.githubusercontent.com/374612/206515172-da26397d-797c-44c5-8dcf-9db5c4d32965.png)

![image](https://user-images.githubusercontent.com/374612/206515256-b2008337-3375-406a-b932-58e178cdb03b.png)

https://github.com/readthedocs/readthedocs.org/pull/9789
https://readthedocs.org/projects/dev/builds/18859457/

In another PR, the build wasn't skipped even though it seems to have contained no changes:

https://github.com/readthedocs/readthedocs.org/pull/9734

humitos:
I've seen the same behavior lately, but I think it's related to some GitHub permission because every time a PR without changes in `docs/` is opened (the build is skipped) I got this notification in my dashboard:

![Screenshot_2022-12-08_19-40-58](https://user-images.githubusercontent.com/244656/206539904-6add59fd-1deb-479a-8d67-0d3c96e0bdbe.png)

benjaoming:
We need to track the GitHub settings for "Required Builds" here.

The PR referred here https://github.com/readthedocs/readthedocs.org/pull/9788 will make it impossible to use required builds together with skipped builds AFAICT. The skipped builds will turn out yellow and the PR will never be merge-able.

ericholscher:
I am debugging this now, and think I've found the issue.

PR #9807, opened 2022-12-13
Fix status reporting on PRs with the magic exit code

ericholscher:
This was because we were using this logic to return a relative URL,
and then sending it to GitHub:

* https://github.com/readthedocs/readthedocs.org/blob/f86b5376ffc70242bdd677cb02f016429ff6bd4b/readthedocs/oauth/services/github.py#L459-L460
* https://github.com/readthedocs/readthedocs.org/blob/f86b5376ffc70242bdd677cb02f016429ff6bd4b/readthedocs/builds/models.py#L343-L349

Fixes https://github.com/readthedocs/readthedocs.org/issues/9791


readthedocs/projects/tasks/builds.py
```diff
@@ -508,6 +508,10 @@ def on_failure(self, exc, task_id, args, kwargs, einfo):
                 build_pk=self.data.build['id'],
                 commit=self.data.build_commit,
                 status=status,
+                # Fix bug where we were generating relative links to the version
+                # with ``get_absolute_url``.
+                # https://github.com/readthedocs/readthedocs.org/issues/9791
```
humitos, line 513: While writing my last comment, I found the `get_absolute_url` is not the problem. It returns full URL "sometimes". Please, read the final comment.

The comment from the code can just be removed since it's not accurate.

humitos (commented):
This looks good.

I think there are is a little more to dig into here, since it seems we are missing some cases. For example, in this part of the code we are not passing `link_to_build=`:

https://github.com/readthedocs/readthedocs.org/blob/90d611df0cfe4bd33d7cf4a9e000e582969dfe74/readthedocs/builds/tasks.py#L453

Also, it seems there is only one case where we want `link_to_build=False` (when the build completes successfully [^1]) and for all the other cases, we want `link_to_build=True`. However, we are using `False` as default argument having to change it in most of the cases. Inverting the default, or even better, removing the default and making it always explicit, would be easier to follow and will avoid the first problem that I found while reviewing this code.

Finally, this part of the workflow seems to be broken

https://github.com/readthedocs/readthedocs.org/blob/90d611df0cfe4bd33d7cf4a9e000e582969dfe74/readthedocs/oauth/services/github.py#L459-L460

We were hitting it and generating relative URLs which GitHub does not accept that. Since we should never send relative URLs, we should delete that code completely to avoid hitting it again.

...

Aha! You linked to the first part of `Version.get_absolute_url`! However, the second part, uses a full absolute URL (not relative). See: https://github.com/readthedocs/readthedocs.org/blob/f86b5376ffc70242bdd677cb02f016429ff6bd4b/readthedocs/builds/models.py#L352-L355

[^1]: in that case we link directly to the built documentation from the PR

readthedocs/projects/tasks/builds.py
ericholscher: @humitos I don't fully understand this comment, can you edit your comment to just have the part that I should address? :)

humitos: Just read https://github.com/readthedocs/readthedocs.org/pull/9807#pullrequestreview-1217124674 which has all the context here 😄

humitos, line 513:
```suggestion
                # Fix bug where we were generating relative links to the version
                # with ``get_absolute_url``.
                # https://github.com/readthedocs/readthedocs.org/issues/9791
```

ericholscher: @humitos there's no change here?

humitos: Weird. I deleted the comment completely because it was not correct IMO

ericholscher:
> Also, it seems there is only one case where we want link_to_build=False (when the build completes successfully [1](https://github.com/readthedocs/readthedocs.org/pull/9807#user-content-fn-1-27a7368cacbffe6db911d1d8387f1c39)) and for all the other cases, we want link_to_build=True. However, we are using False as default argument having to change it in most of the cases. Inverting the default, or even better, removing the default and making it always explicit, would be easier to follow and will avoid the first problem that I found while reviewing this code.

It already already defaulted to `False` in other places in the code, so I wanted a minimal change.

readthedocs/projects/tasks/builds.py
ericholscher: I should disregard everything after `Finally, this part of the workflow seems to be broken`, right?

humitos: Yes

ericholscher:
I think I misunderstood this code, so I made it a bit more explicit.

readthedocs/oauth/services/github.py
```diff
@@ -456,7 +456,7 @@ def send_build_status(self, build, commit, state, link_to_build=False):
         target_url = build.get_full_url()
         statuses_url = f'https://api.github.com/repos/{owner}/{repo}/statuses/{commit}'
 
-        if not link_to_build and state == BUILD_STATUS_SUCCESS:
+        if not link_to_build and not build.version.built and not build.version.uploaded:
```
ericholscher, line 459: @humitos I wonder if this is actually the proper fix after all? Have it match the logic in `get_absolute_url`?

humitos: Hrm, I'm not 100% sure, but I don't think so. `.built` and `.uploaded` are not updated to `False` when a new build is triggered for the same version; so we will be reading an old value here.

IMO, we need this logic:

- Successful build: link to documentation
- Failed build: link to build details' page

As you may noted, the link to send does not depends on the "Commit build status" we sent to GitHub (`state == BUILD_STATUS_SUCCESS`, in that code), but it depends on the "status" of the build itself.

Actually, I'd completely remove `link_to_build` and just use the build's status inside this method. That way, we manage the logic just in one place and we don't have to worry about _where_ and _why_ we are passing `link_to_build=False/True` in different places of the code.

readthedocs/projects/tasks/builds.py
```diff
@@ -483,7 +483,7 @@ def on_failure(self, exc, task_id, args, kwargs, einfo):
             )
 
         # NOTE: why we wouldn't have `self.data.build_commit` here?
-        # This attribute is set when we get it after clonning the repository
+        # This attribute is set when we get it after cloning the repository
```
ericholscher, line 486: 😆

ericholscher (commented):
This is much cleaner and more obvious. I'm 👍 on shipping this!

humitos (approved):

PR #9813, opened 2022-12-14
Break documentation style guide out into its own file

ericholscher:
This is the start of a larger style guide that we should have,
trying to capture places where we make decisions.

Lots of inspiration possible here: https://docs.pylonsproject.org/projects/docs-style-guide/

Closes readthedocs/meta#65

---

- User's documentation (`docs`): https://docs--9813.org.readthedocs.build/en/9813/

- Developer's documentation (`dev`): https://dev--9813.org.readthedocs.build/en/9813/

### DIFF
--- a/docs/dev/style-guide.rst
+++ b/docs/dev/style-guide.rst
@@ -50,7 +50,7 @@ Word List
 We have a specific way that we write common words:
 
 * ``open source`` should be lower case, unless you are definitely referring to `OSI's Open Source Definition <https://opensource.org/osd>`.
-* ``git`` should be lower case.
+* ``Git`` should be upper case, except when referring to the command, when it should be written as `:program:\`git\``.
 
 Glossary
 --------
--- a/docs/dev/style-guide.rst
+++ b/docs/dev/style-guide.rst
@@ -49,6 +49,7 @@ Word List
 
 We have a specific way that we write common words:
 
+* ``Git repository`` for the place that stores Git repos. We used to use ``VCS``, but this is deprecated. 
 * ``open source`` should be lower case, unless you are definitely referring to `OSI's Open Source Definition <https://opensource.org/osd>`.
 * ``Git`` should be upper case, except when referring to the command, when it should be written as `:program:\`git\``.
 
--- a/docs/dev/style-guide.rst
+++ b/docs/dev/style-guide.rst
@@ -21,11 +21,11 @@ Titles
 For page titles we use sentence case.
 This means only proper nouns and the first word are capitalized::
 
-# Good ✅
-How we handle support on Read the Docs.
+    # Good ✅
+    How we handle support on Read the Docs.
 
-# Bad 🔴
-How we Handle Support on Read the Docs
+    # Bad 🔴
+    How we Handle Support on Read the Docs
 
 If the page includes multiple sub-headings (H2, H3),
 we use sentence case there as well.
@@ -49,7 +49,7 @@ Word List
 
 We have a specific way that we write common words:
 
-* ``Git repository`` for the place that stores Git repos. We used to use ``VCS``, but this is deprecated. 
+* ``Git repository`` for the place that stores Git repos. We used to use ``VCS``, but this is deprecated.
 * ``open source`` should be lower case, unless you are definitely referring to `OSI's Open Source Definition <https://opensource.org/osd>`.
 * ``Git`` should be upper case, except when referring to the command, when it should be written as `:program:\`git\``.
 
@@ -57,10 +57,10 @@ Glossary
 --------
 
 Since the above Word List is for internal reference,
-we also maintain a :doc:`/glossary` with terms that have canonical definitions in our docs.
+we also maintain a :doc:`rtd:glossary` with terms that have canonical definitions in our docs.
 Terms that can otherwise have multiple definitions
 *or* have a particular meaning in Read the Docs context
-should always be added to the :doc:`/glossary` and referenced using the ``:term:`` role.
+should always be added to the :doc:`rtd:glossary` and referenced using the ``:term:`` role.
 
 Using a glossary helps us (authors) to have consistent definitions
 but even more importantly,
--- a/docs/dev/style-guide.rst
+++ b/docs/dev/style-guide.rst
@@ -33,6 +33,7 @@ we use sentence case there as well.
 Content
 -------
 
+* Use ``:guilabel:`` when referring to an element on the screen.
 * Do not break the content across multiple lines at 80 characters,
   but rather break them on semantic meaning (e.g. periods or commas).
   Read more about this `here <https://rhodesmill.org/brandon/2012/one-sentence-per-line/>`_.
@@ -48,3 +49,5 @@ We have a specific way that we write common words:
 
 * ``open source`` should be lower case.
 * ``git`` should be lower case.
+
+You can also see the :doc:`/glossary` other words that have canonical definitions in our docs.
--- a/docs/dev/style-guide.rst
+++ b/docs/dev/style-guide.rst
@@ -33,7 +33,9 @@ we use sentence case there as well.
 Content
 -------
 
-* Use ``:guilabel:`` when referring to an element on the screen.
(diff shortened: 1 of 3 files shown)


docs/dev/style-guide.rst
```diff
@@ -0,0 +1,53 @@
...
+* If you are cross-referencing to a section within our website,
+  use the ``ref`` role with the label from the `autosectionlabel extension <http://www.sphinx-doc.org/en/master/usage/extensions/autosectionlabel.html>`__.
+
+Word List
+---------
+
+We have a specific way that we write common words:
+
+* ``open source`` should be lower case.
+* ``git`` should be lower case.
```
agjohnson, line 51: I missed where this came up, but Git is a proper noun and their docs use "Git", not "git":

https://git-scm.com/about

docs/dev/style-guide.rst
```diff
@@ -0,0 +1,53 @@
...
+
+We are called **Read the Docs**.
+The ``the`` is not capitalized.
+
+We do however use the acronym **RTD**.
+
+Titles
+------
+
+For page titles we use sentence case.
```
benjaoming, line 21: Awesome - should we take an approach to use sentence case in all open and upcoming PRs and file an issue for "100% consistency in all titles" that we solve at the end of Iteration 1?

docs/dev/style-guide.rst
```diff
@@ -0,0 +1,53 @@
...
+# Bad 🔴
+How we Handle Support on Read the Docs
+
+If the page includes multiple sub-headings (H2, H3),
+we use sentence case there as well.
+
+Content
+-------
+
+* Use ``:guilabel:`` when referring to an element on the screen.
```
benjaoming, line 36:
```suggestion
* Use ``:menuitem:`` when referring to an item in navigation or the exact text of a link label.
* Use ``:guilabel:`` when referring to an visual element on the screen - such as a button, drop down or input field.
* Use ``**bold text**`` when referring to a non-interactive text element, such as a header.
```

Does this reflect what we want to do?

docs/dev/style-guide.rst
```diff
@@ -0,0 +1,53 @@
...
+
+Word List
+---------
+
+We have a specific way that we write common words:
+
+* ``open source`` should be lower case.
+* ``git`` should be lower case.
+
+You can also see the :doc:`/glossary` other words that have canonical definitions in our docs.
```
benjaoming, line 53:
```suggestion
You can also see the :doc:`/glossary` other words that have canonical definitions in our docs.
Terms that can have multiple definitions
or are strictly defined in a Read the Docs context
should always be added to the :doc:`/glossary` and referenced using the ``:term:`` role.
```

docs/dev/style-guide.rst
humitos: What is `:menuitem:`? I've never seen it and I'm not finding it in the docs.

benjaoming: typo - it's this one: https://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html#role-menuselection

docs/dev/style-guide.rst
humitos: TIL: `:program:` role. Maybe it's useful for these cases where we talk about the command itself. See https://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html#role-program

docs/dev/style-guide.rst
agjohnson: I didn't know about this role! I have used a string of `guilabel` roles, which is more tedious. Assuming the display output is as nice, this seems like a really handy pattern :+1:

docs/dev/style-guide.rst
```diff
@@ -0,0 +1,53 @@
...
+  use the ``doc`` role and not a hyperlink.
+* If you are cross-referencing to a section within our website,
+  use the ``ref`` role with the label from the `autosectionlabel extension <http://www.sphinx-doc.org/en/master/usage/extensions/autosectionlabel.html>`__.
+
+Word List
+---------
+
+We have a specific way that we write common words:
+
+* ``open source`` should be lower case.
```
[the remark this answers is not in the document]
agjohnson: To clarify what I was referring to here, "open source" only needs capitalization if it's part of a proper noun, like "Open Source Initiative".

Here's Google's styleguide on the matter:

https://opensource.google/documentation/reference/comms/style-guide

> Open source: Don’t capitalize or hyphenate open source unless it’s part of a proper noun (like Google Open Source).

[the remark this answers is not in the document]
benjaoming: I have always been under the impression that writing "Open Source" is signalling that you know about the definition of the term and not throwing around lose claims.

I remember a point about countering what MongoDB was doing was to always write Open Source when referring to something that implements the OSI definition, even though they failed to get a US Trademark registered in 1999 :)

https://lwn.net/Articles/888291/
https://drewdevault.com/2022/03/01/Open-source-is-defined-by-the-OSD.html

The latter blog post has a good claim to say that simply writing "open source" should always _morally_ mean a real OSI approved license. And it seems the courts have said so, too, in 2022.

docs/dev/style-guide.rst
benjaoming: Yes, we chose that we wanted this and until the decision was made to actually do it, we would keep consistency with the _old_ Title Case for titles.

I think this is the milestone where we use sentence cases moving forwards :+1:

docs/dev/style-guide.rst
ericholscher: Yea, I just found `menuselection` the other day and thought that was way nicer.

docs/dev/style-guide.rst
ericholscher: Yea, I agree we could be using the Glossary more actively 👍 This is a great addition.

docs/dev/style-guide.rst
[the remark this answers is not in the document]
benjaoming: But just to close this - there are also reasons to say "open source" and care about the term, I see the other POV :+1:

ericholscher:
Addressed feedback 👍

docs/dev/style-guide.rst
```diff
@@ -0,0 +1,68 @@
...
+* ``Git`` should be upper case, except when referring to the command, when it should be written as `:program:\`git\``.
+
+Glossary
+--------
+
+Since the above Word List is for internal reference,
+we also maintain a :doc:`/glossary` with terms that have canonical definitions in our docs.
+Terms that can otherwise have multiple definitions
+*or* have a particular meaning in Read the Docs context
+should always be added to the :doc:`/glossary` and referenced using the ``:term:`` role.
```
ericholscher, line 62: Looks like this needs to be intersphinxed :D

benjaoming (approved):
I think this PR worked really well. We went over several writing style preferences which may have seemed too small for processing in isolation. But by combining it in one PR, we got several things dealt with at once and the whole team's attention was there as well.

Seems like all the feedback is addressed. We could open a follow-up issue for "capturing" small items for the next "round of style guide updates"?

I don't immediately recall anything else "ripe" for the styleguide, but I think there are several conventions that are coming from our Diátaxis work that we can address once we feel that it has settled in:

* How we give titles: How-tos, explanations etc.
* Introduction paragraphs
* `seealso::` usage
* How we use Diátaxis (a super-quick instruction list to new authors)

ericholscher:
@benjaoming Yea, my hope here is to slowly add things here as we discuss them, instead of queueing them up. I think the `seealso` one is a great one to get in here in another PR soon. Similarly, some ideas around how many admonitions to use, because I feel like we all have a bit different views there.

benjaoming:
@ericholscher I really enjoyed how you combined several additions on one PR, I think it's a great service to team collaboration, rather than for instance making 1 PR to say "lets write Git instead of git" and have that whole process separately. We saved a lot of time :100:

ericholscher:
Yea, if we have a bunch of changes that are similar, they should be in the same PR. Otherwise it just gets messy.

PR #9815, opened 2022-12-15
Build details page: normalize/trim command paths

humitos:
Small hack to trim/normalize command paths when displaying them to the user under the build detail's page.

This is a simple hack and there are more we can do here. Probably, there are some edge cases this is not catching. However, the worst it can happens is the path not being trimmed.

## Before 

Screenshot_2022-12-15_18-48-01

## After

Screenshot_2022-12-15_18-47-45

### DIFF
--- a/readthedocs/api/v2/serializers.py
+++ b/readthedocs/api/v2/serializers.py
@@ -1,5 +1,7 @@
 """Defines serializers for each of our models."""
 
+import re
+from django.conf import settings
 from allauth.socialaccount.models import SocialAccount
 from rest_framework import serializers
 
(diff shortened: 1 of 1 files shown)


readthedocs/api/v2/serializers.py
```diff
@@ -127,11 +129,33 @@ class VersionAdminSerializer(VersionSerializer):
...
+        #   $ python
+        project_slug = obj.build.version.project.slug
+        version_slug = obj.build.version.slug
+
+        docroot = settings.DOCROOT
+        # Remove Docker hash from DOCROOT when running it locally
+        if settings.RTD_DOCKER_COMPOSE:
+            docroot = re.sub('/[0-9a-z]+/?$', '', settings.DOCROOT, count=1)
+
+        command = re.sub(f'{docroot}/[0-9a-z]+/{project_slug}/envs/{version_slug}(/bin/)?', '', obj.command, count=1)
```
ericholscher, line 156: I wonder if we want to remove `count=1`, so if we have multiple paths in the command they will be removed? I don't know how often that happens, but seems fine to have this for now, also.

ericholscher (approved):
Simple hacks ftw 👍

readthedocs/api/v2/serializers.py
humitos, line 156:
```suggestion
        command = re.sub(f'{docroot}/{project_slug}/envs/{version_slug}(/bin/)?', '', obj.command, count=1)
```

ericholscher: This is definitely nicer and more explicit.

benjaoming: I'm not sure I understand the change. `[0-9a-z]+/` was removed. But if this regex worked _before_ isn't it broken now?

ericholscher: Oh, I thought we were changing `[0-9a-z]+` to `project_slug`, but I see we just removed it. Not 100% sure.

humitos: I added this suggestion to not forget to review this on Monday, basically.

I added `[0-9a-z]+` to remove the "container hash" we append on the docroot when working locally to avoid collision between builders when using more than 1 builder. However, if you take a look at line 154 and 156 both include that part of the regex when it should only be on line 154 (only when working locally)

humitos: OK, my assumption when I saw this code again was wrong.

Now, I remember why I've done this:

- the docroot _contains_ the containers' hash and we have to remove it because it is different every time spin up Docker again. So, it can't be part of the regex
- in the regex _we need_ that part `[0-9a-z]` because it will match the paths that are already in the commands

So, we need both and the code was correct as it was.

humitos:
Ups, I enabled auto-merge, but I didn't add a test case for this. I will add it in another PR.

PR #9821, opened 2022-12-19
Dependencies: use backtracking pip's resolver

humitos:
Upgrade the resolver to avoid conflicts while resolving dependencies.
The GHA has been failing in the last 2 week because of this issue. Hopefully,
by using the new resolver we can fix it.

Solves: https://github.com/readthedocs/readthedocs.org/actions/runs/3722354255


requirements/debug.txt
```diff
@@ -50,16 +50,13 @@ six==1.16.0
     # via asttokens
 stack-data==0.6.2
     # via ipython
-toml==0.10.2
+tomli==2.0.1
```
ericholscher, line 53: This changes the package?

ericholscher (approved):
This looks good. I'm always a little scared about dependency updates, especially so many at once 🙃

requirements/debug.txt
humitos: It seems `ipdb` installs `toml` or `tomli` depending on other dependencies: https://github.com/gotcha/ipdb/blob/master/setup.py#L66

ericholscher: Interesting.

PR #9824, opened 2022-12-20
Test: path is trimmed when returned by the API

humitos:
This commit contains the test that was missing for my other PR: https://github.com/readthedocs/readthedocs.org/pull/9815

The test checks the commands' PATH are trimmed before being returned by the API.


ericholscher (approved):
This test looks good, along with the test @stsewd is doing to test writing which was also broken.  👍

PR #9825, opened 2022-12-20
CI build: Conditionally create the 'main' branch for pre-commit comparison

benjaoming:


.circleci/config.yml
```diff
@@ -42,7 +42,8 @@ jobs:
       - run: git submodule sync
       - run: git submodule update --init
       - run: git fetch origin main  # needed for comparisson in pre-commit
-      - run: git branch -f --track main origin/main  # needed for comparisson in pre-commit
+      # Meaning of below line: If a "main" branch does not exist, then create one.
+      - run: git rev-parse --verify main 2>/dev/null || git branch -f --track main origin/main
```
humitos, line 46: I don't understand why the `main` branch (the default) wouldn't exist after doing the `checkout` step in the first place. Also, I don't know why `git fetch origin main` is now required here for pre-commit. I seems you added in https://github.com/readthedocs/readthedocs.org/pull/9789/files#r1043369043 and you also weren't sure why this was needed.

Then we added the `-f`, but that seemed to not work as we expected. In this change it seems we are adding another small hack on top of the previous hack, but we still don't understand why this is happening and/or if this is gonna solve the problem.

My suggestion here is to debug this a little more and understand what is the root cause of the problem before adding more extra commands and keep complicating the CI process. Starting by removing all these `git` commands.

humitos (commented):
We don't know if this solution will work before merging it into `main`, since the problem is there. However, I'd suggest to debug this problem a little more to understand what's happening and to find out the root cause of it. IMO, none of these `git` commands should be required since `main` (the default branch) should always exist and be up-to-date due the repository is cloned each time.

.circleci/config.yml
benjaoming: Btw. I have a `main` branch that I can build it on :) => https://app.circleci.com/pipelines/github/benjaoming/readthedocs.org/95/workflows/b98f810a-74e6-4dd9-8129-b788f2df7a58

humitos: > I have a `main` branch that I can build it on

Cool! 💯  Can you try triggering a build on CircleCI for that repository by removing these `git` commands?

benjaoming: main: https://app.circleci.com/pipelines/github/benjaoming/readthedocs.org/98/workflows/f4d47f42-9fe6-4767-882d-19196d614404
other branch: https://app.circleci.com/pipelines/github/benjaoming/readthedocs.org/99/workflows/7f5356eb-1420-4e7c-aeef-e47eea7cdfb3/jobs/231 :x:

benjaoming: (this PR has a couple of weird commits because I had a hard time making Circle CI build the same commit hash on different branches)

benjaoming: The old error is back by removing the git commands.

humitos: Clicking on the "Checkout code" tab and reading the script, it seems these are the commands it executes to clone the repository and checkout the proper branch:

```
git clone --no-checkout git@github.com:benjaoming/readthedocs.org.git
git checkout --force -B ci-fix-for-pre-commit-base-branch e2c03dc3f4fb74659f75410eb58f11565dc4a3d5
```

Then, our `pre-commit` command executes:

```
git diff --name-only --no-ext-diff -z main..HEAD
```

humitos: It seems this only happens in your `github.com:benjaoming/readthedocs.org.git` because it has `master` instead of `main` branch. If you create the PRs directly in the `readthedocs/readthedocs.org` repository you won't face this issue.

benjaoming: Yeah, it's been left like that for a while :)

> This branch is [12455 commits behind](https://github.com/benjaoming/readthedocs.org/compare/master...readthedocs:readthedocs.org:main) readthedocs:main.

humitos: 😉 -- problem solved, then, and we can remove these `git` commands. You have to update your fork or, probably better, use `readthedocs/readthedocs.org` to push your code and open PR from there.

benjaoming:
It seems like we can make this issue go away by ensuring that the origin repository of a PR has `main` as the default branch. Since my repo didn't, the issue started occurring.

Also, it would really be nice to run `pre-commit` in a different way, so there's no need to make this solution more complicated.

I'll open a PR to remove the addition `git` commands, thanks for looking into this @humitos :100:

PR #9785, opened 2022-12-07
Docs: Split Subprojects in Explanation and How-to (Diátaxis)

benjaoming:
This was originally identified as a "relabel", but I think it's best that we split out the How-to part.

There's a potential to do more here. Is it possible to review and finalize the proposed content and postpone further additions to later?

Note also that `subprojects` has been added to the glossary in #9676 

Refs: https://github.com/readthedocs/readthedocs.org/issues/9746

---
:books: Documentation previews :books:

- User's documentation (`docs`): https://docs--9785.org.readthedocs.build/en/9785/

- Developer's documentation (`dev`): https://dev--9785.org.readthedocs.build/en/9785/

PR #9829, opened 2022-12-20
Label for subproject select renamed "Child" => "Subproject" + help text added

benjaoming:
In #9785, we are looking for a more consistent terminology.

A new helpful text is also added to explain what the page is about and where the documentation is.

image


readthedocs/projects/forms.py
```diff
@@ -433,7 +433,8 @@ class ProjectRelationshipForm(forms.ModelForm):
 
     class Meta:
         model = ProjectRelationship
-        fields = '__all__'
+        fields = "__all__"
+        labels = {"child": _("Subproject")}
```
ericholscher, line 437: Seems like we should change this on the model instead, so it doesn't leak out other places? 

https://github.com/readthedocs/readthedocs.org/blob/main/readthedocs/projects/models.py#L85-L90

ericholscher (commented):
💯 on adding actual helptext.

readthedocs/projects/forms.py
benjaoming: In general, I think `verbose_name` is a bad place for fine adjustments of language because it creates a migration each time. It's great to use form labels because they are easier to adjust and their context is clear, while `verbose_name` is confusing because it's applied implicitly to contexts that are really hard to guess.

But for sure, it's also bad to have wrong terminology cemented into the models forever :)

So I chose to do both... i.e. fix the model AND add the form labels :magic_wand:

humitos: Wouldn't that lead to a confusing scenario where changing the `verbose_name=` in the model won't be reflected in the form? This will require us to update the text in both places again.

Also, I think we would have different naming between the UI and the Admin interface when using `labels=` and `verbose_name=` with different values, for example.

benjaoming: > Wouldn't that lead to a confusing scenario where changing the verbose_name= in the model won't be reflected in the form?

That's 100% what I just wrote, but instead of "confusing", I argued for "intentional" and "explicit" and ultimately less confusing ;)

I sense a more general discussion coming up, so maybe raise it if there is a general principle or concept here that you'd like to discuss or decide.

ericholscher: I think I agree with @humitos. I'd prefer to not have something overriding the `verbose_name` in the form that just sets the name the same? It seems redundant, and confusing, so we should definitely remove this and just keep the verbose name. If we want to override it, it should say something different so that it's obvious where to change it in the future.

benjaoming, line 437:
```suggestion
        fields = "__all__"
```

benjaoming: It's gone :+1:

humitos (approved):

readthedocs/projects/forms.py
ericholscher: @benjaoming Thanks! ✨

ericholscher:
Thanks for fixing the code instead of documenting around it @benjaoming. A pro move :)

readthedocs/projects/forms.py
benjaoming: I channeled the "this is not perfect" energy into thinking about how it's actually a Django issue :+1:

ericholscher: Yea, the migration thing irks me too, but I don't think that means we should work around it. Migrations aren't so bad :)

benjaoming: My own observation is that they discourage fine-tuning of `verbose_name` and `help_text` but also that they make the work effort of small positive improvements worsen.. for instance, if you gave a review suggestion to change a `verbose_name` or `help_text`, the author of the PR would have to delete and re-generate migrations _or_ have to migrations in the same PR.

benjaoming: *) that's an observation outside of RTD.

PR #9832, opened 2022-12-21
Docs: Relabel the "Science" page as Explanation

benjaoming:
Refs: https://github.com/readthedocs/readthedocs.org/issues/9746

This raises two questions:

1) Can we limit the amount of changes to this?
2) Are there any follow-up actions that we can already express or is it better to just plan for a revisit

(The reason that I wanted to minimize this effort, is in anticipation of some sort of separation between what lives on https://about.readthedocs.com/ and what lives in the documentation)

---

- User's documentation (`docs`): https://docs--9832.org.readthedocs.build/en/9832/

- Developer's documentation (`dev`): https://dev--9832.org.readthedocs.build/en/9832/

### DIFF
--- a/docs/user/index.rst
+++ b/docs/user/index.rst
@@ -82,6 +82,7 @@ to help you create fantastic documentation for your project.
    /custom-domains
    /pull-requests
    /downloadable-documentation
+   /science
 
 .. toctree::
    :maxdepth: 2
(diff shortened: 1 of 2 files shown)


ericholscher (approved):
I'm not too worried about this one, so I think a move is fine.

PR #9834, opened 2022-12-21
Docs: Relabel and move explanation and how-tos around OAuth (Diátaxis)

benjaoming:
* This splits up the explanation into the usual explanation + how to.
* Some of the texts are tweaked
* [x] Implement decisions from Style Guide: https://github.com/readthedocs/readthedocs.org/pull/9840

Refs: https://github.com/readthedocs/readthedocs.org/issues/9746

---
:books: Documentation previews :books:

- User's documentation (`docs`): https://docs--9834.org.readthedocs.build/en/9834/

- Developer's documentation (`dev`): https://dev--9834.org.readthedocs.build/en/9834/

PR #9840, opened 2022-12-22
Docs: Additions to style guide - placeholders, seealso::, Diátaxis and new word list entry

benjaoming:
CC: @ericholscher these were the 3 things I remember from recent conversations.

Please feel free to suggest more things :+1: 

---

- User's documentation (`docs`): https://docs--9840.org.readthedocs.build/en/9840/

- Developer's documentation (`dev`): https://dev--9840.org.readthedocs.build/en/9840/


humitos:
If you liked my suggestion about com/org differentiation on a note, we could add it here as well

benjaoming:
@humitos I meant to maybe add something about substitutions like `|com_brand|` and `|org_brand|` but can you reference the suggestion?

docs/dev/style-guide.rst
```diff
@@ -43,6 +43,10 @@ Content
   use the ``doc`` role and not a hyperlink.
 * If you are cross-referencing to a section within our website,
   use the ``ref`` role with the label from the `autosectionlabel extension <http://www.sphinx-doc.org/en/master/usage/extensions/autosectionlabel.html>`__.
+* Use ``<abstract concept>`` and ``<variable>`` as placeholders in code and URLs. For instance:
```
ericholscher, line 46: I was thinking `$variable` to differentiate them?

docs/dev/style-guide.rst
```diff
@@ -67,3 +71,98 @@ but even more importantly,
...
+* Consider adding an Examples section
+* Can you add screenshots or diagrams?
+
+How-to Guides
+^^^^^^^^^^^^^
+
+* Title should begin with **“How to ...”**,
+  but navigation titles should not contain the “How to” part.
+* Introduce the scope: **“In this guide, we will…”**
+* Cross-reference related explanation.
```
ericholscher, line 138:
```suggestion
* `Cross-reference <#cross-references>`_ related explanation.
```

:D

docs/dev/style-guide.rst
```diff
@@ -67,3 +71,98 @@ but even more importantly,
...
+   such as an explanation *and* a how-to,
+   you can easily design your content to end up in the right place *Diátaxis-wise*.
+   This is great for the author and the reviewers
+   and it saves coordination work.
+
+   Minimal or isolated changes generally raise more questions and concerns
+   than changes that seek to address a larger perspective.
+
+Explanation
+^^^^^^^^^^^
```
ericholscher, line 120: I think we generally use `~` for third-level headings... should probably go in the style guide :D

docs/dev/style-guide.rst
```diff
@@ -67,3 +71,98 @@ but even more importantly,
...
+We have not started organizing the Reference section yet,
+guidelines pending.
+
+Tutorial
+^^^^^^^^
+
+.. note:: We don’t really have tutorials targeted in the systematic refactor, so this checklist isn’t very important right now.
+
+* Cross-reference related explanation and how-to.
+* Make sure absolutely no explanation
```
ericholscher, line 167:
```suggestion
* Try not to explain things too much, and instead link to the explanation content.
```

This seems a bit strong?

docs/dev/style-guide.rst
```diff
@@ -43,6 +43,10 @@ Content
   use the ``doc`` role and not a hyperlink.
 * If you are cross-referencing to a section within our website,
   use the ``ref`` role with the label from the `autosectionlabel extension <http://www.sphinx-doc.org/en/master/usage/extensions/autosectionlabel.html>`__.
+* Use ``<abstract concept>`` and ``<variable>`` as placeholders in code and URLs. For instance:
+
+  * ``https://<slug>.readthedocs.io``
+  * ``:guilabel:`<your username>` dropdown``
```
ericholscher, line 49:
```suggestion
  * ``:guilabel:`<your username>` dropdown``
* Make sure that **all bullet list items end with a period**, and don't mix periods with no periods.   

```

docs/dev/style-guide.rst
```diff
@@ -67,3 +71,98 @@ but even more importantly,
...
+^^^^^^^^^^^
+
+* Introduce the scope: **“This article introduces ...”**
+  (write this as the very first thing,
+  then consider rewriting it or turning it into an internal comment afterwards).
+* Cross-reference the related How-to Guide.
+  Put a ``seealso::`` somewhere visible.
+  It should likely be placed right after the introduction,
+  and if the article is very short, maybe at the bottom.
+* Consider adding an Examples section
```
ericholscher, line 129:
```suggestion
* Consider adding an Examples section.
```

ericholscher (commented):
Some good first steps here, had a couple additions and one question.

docs/dev/style-guide.rst
benjaoming: Yes! So on that note, would you prefer ``$variable`` and that the example is ``https://$slug.readthedocs.io`` ?

docs/dev/style-guide.rst
benjaoming: Right, haha, the other subsection came after this was written. Should this be 

```
:ref:`Cross-references`
```

?

docs/dev/style-guide.rst
benjaoming: I really don't get Sphinx/reST on this particular thing. You can apparently choose the heading characters and then the parser just magically understands. It made me completely unable to memorize which character I'm supposed to use.

benjaoming: Can you outline what you believe is our convention for outlining headers?

```
Header 1
======== 

Header 2
--------

Header 3
~~~~~~~~

Header 4
^^^^^^^^
```

?

docs/dev/style-guide.rst
agjohnson: Also, the bash syntax is nice as it give variable boundaries:

`https://${slug}.readthedocs.io`

docs/dev/style-guide.rst
agjohnson: reST is funny. You can use almost any symbol character you want, including some unicode I believe.

```
Heading 1
)))))))))

Heading 2
(((((((((

Heading 3
@@@@@@@@@

Heading 4
!!!!!!!!!
```

Heading characters just need to be relative to the first heading character you used and docutils is happy :upside_down_face:

ericholscher: Yea, that's also what Plone uses: https://docs.plone.org/about/contributing/rst-styleguide.html#headings

benjaoming: Do we want to do the Plone version? It's not fully consistent with most of our articles, since we usually don't have this:

```
========
Header 1
========
```

But I like the Plone version, personally I have the easiest time remembering that order... like you say `=, -, ~, ^` feels somehow like how the characters should be ordered :)

humitos:
@benjaoming 

> I meant to maybe add something about substitutions like |com_brand| and |org_brand| but can you reference the suggestion?

https://github.com/readthedocs/readthedocs.org/blob/d8f528263e3fa9485bec694dc1b53a8e4076436d/docs/user/guides/subprojects.rst?plain=1#L25-L32

The version `diataxis/main` is broken on Read the Docs, so I cannot link to the rendered version, but it would appear at https://docs.readthedocs.io/en/diataxis-main/subprojects.html#adding-a-subproject once it builds properly.

docs/dev/style-guide.rst
benjaoming: I think that I prefer just one notation and that it's `<variable name>` and `<abstract concept>`. It's because I have a hard time coming up with examples from our documentation of referenced variable names that are 100% bound and not actually also an abstract concept that we introduced in the docs.

So even if we say `https://$slug.readthedocs.io` or `https://${slug}.readthedocs.io`, there actually doesn't exist a real variable called `slug` but just a concept from the text. So this might as well be `<slug>` and no distinction.

ericholscher (approved):
Header changes look good 👍

docs/dev/style-guide.rst
benjaoming: Fixed the headlines and added a new section about headlines :+1:

docs/dev/style-guide.rst
benjaoming: There is now a cross-reference to cross-references, so the point should be hopefully clear :)

docs/dev/style-guide.rst
```diff
@@ -39,17 +39,25 @@ Content
...
+
 
 Word List
 ---------
 
 We have a specific way that we write common words:
 
 * ``Git repository`` for the place that stores Git repos. We used to use ``VCS``, but this is deprecated.
+* ``Git provider`` for generic references to GitHub/Bitbucket/GitLab/Gitea etc.
+  We avoid "host" and "platform" because they are slightly more ambiguous.
```
benjaoming, line 60: Better suggestions are most welcome :pray:

benjaoming:
@humitos I added a new section `Substitutions` where we can document substitutions and showcase their current value.

docs/dev/style-guide.rst
```diff
@@ -39,20 +39,37 @@ Content
...
+Substitutions
+-------------
+
+The following substitutions are used in our documentation to guarantee consistency and make it easy to apply future changes.
+
+* ``|org_brand|`` is used for mentioning of |org_brand|
+* ``|com_brand|`` is used for mentioning of |com_brand|
+* ``|git_providers_and|`` is used to mention currently support Git providers with "and". Example: |git_providers_and|
+* ``|git_providers_or|`` is used to mention currently support Git providers with "or". Example: |git_providers_or|
+
```
benjaoming, line 72: New section added based on @humitos' [comment](https://github.com/readthedocs/readthedocs.org/pull/9840#issuecomment-1364538348)

I also added two types of substitutions for supported Git providers. They are used quite a lot, so I would expect this to be very applicable in the current refactor. Examples in #9834

humitos: I was suggesting to add a section, similar to "Cross-references" that explains the `.. seealso::`, to explain how we are going to use the `.. note::` to differentiate one platform from another.

benjaoming: That missed me by a mile :) Is this point more generally about how we use admonition boxes? Like "if you want to point out a difference between org and com, use ``.. note::`` etc?

docs/dev/style-guide.rst
ericholscher: I think that seems reasonable, especially since we're using `$` variables to actually be explicit vars in places in the application (eg;. `$rest` in redirects). So probably better to preserve them for that.

docs/dev/style-guide.rst
ericholscher: Yea, we also call them integrations in some places too I think? It would be good to be explicit when we're talking about Git-specific hosts that are actually used for code & auth, so I think Git provider is fine? Though provider is a bit abstract as well, tbh, but so are host & platform 🤷

docs/dev/style-guide.rst
```diff
@@ -39,19 +39,36 @@ Content
...
 We have a specific way that we write common words:
 
 * ``Git repository`` for the place that stores Git repos. We used to use ``VCS``, but this is deprecated.
+* ``Git provider`` for generic references to GitHub/Bitbucket/GitLab/Gitea etc.
+  We avoid "host" and "platform" because they are slightly more ambiguous.
 * ``open source`` should be lower case, unless you are definitely referring to `OSI's Open Source Definition <https://opensource.org/osd>`.
-* ``Git`` should be upper case, except when referring to the command, when it should be written as `:program:\`git\``.
+* ``Git`` should be upper case. Except when referring to the :program:`git` command, then it should be written as `:program:\`git\``.
+
+Substitutions
```
ericholscher, line 63: This is great 💯

docs/dev/style-guide.rst
```diff
@@ -39,19 +39,36 @@ Content
...
+-------------
+
+The following substitutions are used in our documentation to guarantee consistency and make it easy to apply future changes.
+
+* ``|org_brand|`` is used for mentioning of |org_brand|
+* ``|com_brand|`` is used for mentioning of |com_brand|
```
ericholscher, line 69:
```suggestion
* ``|org_brand|`` is used for mentioning of .org: Example: |org_brand|
* ``|com_brand|`` is used for mentioning of .com. Example: |com_brand|
```

docs/dev/style-guide.rst
```diff
@@ -67,3 +84,126 @@ but even more importantly,
...
+   propose the changes in the same PR.
+
+   By multi-tasking on several articles about the same topic,
+   such as an explanation *and* a how-to,
+   you can easily design your content to end up in the right place *Diátaxis-wise*.
+   This is great for the author and the reviewers
+   and it saves coordination work.
+
+   Minimal or isolated changes generally raise more questions and concerns
+   than changes that seek to address a larger perspective.
```
ericholscher, line 158: I think in general it's the opposite with code, but for this refactor, it's more true because we want to fix all the issues around a topic at once, instead of leaving them worse than we found them. So not 100% sure this will prove true after the initial refactor, but we'll see.

ericholscher (approved):
These changes are great. A couple nits, but I think it's ready to merge.

docs/dev/style-guide.rst
benjaoming: Agreed - I hope that the "Warning" format of this will automatically make it seem opinionated enough that we will want to re-evaluate it.

docs/dev/style-guide.rst
benjaoming: As you also noted in the PR adding texts to the Dashboard UI of "Connected Services", we would want to keep changes to the current Dashboard minimal. In the upcoming Dashboard frontend, we will _hopefully_ recall our terminology and find it natural to add some more tweaks both in the Dashboard and in the documentation. The future version of the documentation will also be less complicated to impose changes to - I feel somewhat confident about that :crossed_fingers:

benjaoming:
I hope that it's fine merging this based on @ericholscher's feedback. I don't have too many worries here, but please feel welcome to ping me back with additional inputs to these changes :+1:

docs/dev/style-guide.rst
humitos: Yes and no 🙃 . Let me expand. I wasn't saying to explain about the usage of admonitions _in general_; but the _specific case_ about using `.. note::` to differentiate community from commercial.

benjaoming: Okay, I'll have to find a location for it in the style guide.

PR #9843, opened 2022-12-23
API: add project name/slug filters

agjohnson:
I am trying to use the project API for some search and dropdown filters,
and while I can filter the local results, it's much nicer to use the API
filter for this.

This adds an icontains filter field for project name and slug, and
hopefully some accurate tests.

---

- User's documentation (`docs`): https://docs--9843.org.readthedocs.build/en/9843/

- Developer's documentation (`dev`): https://dev--9843.org.readthedocs.build/en/9843/

### DIFF
--- a/docs/user/api/v3.rst
+++ b/docs/user/api/v3.rst
@@ -186,6 +186,8 @@ Projects list
             }]
         }
 
+    :query string name: return projects with matching name
+    :query string slug: return projects with matching slug
     :query string language: language code as ``en``, ``es``, ``ru``, etc.
     :query string programming_language: programming language code as ``py``, ``js``, etc.
 
--- a/readthedocs/api/v3/tests/test_projects.py
+++ b/readthedocs/api/v3/tests/test_projects.py
@@ -24,46 +24,46 @@ class ProjectsEndpointTests(APIEndpointMixin):
         )
 
     def test_projects_list_filter_full_hit(self):
-        self.client.credentials(HTTP_AUTHORIZATION=f'Token {self.token.key}')
+        self.client.credentials(HTTP_AUTHORIZATION=f"Token {self.token.key}")
         response = self.client.get(
-            reverse('projects-list'),
+            reverse("projects-list"),
             data={
-                'name': self.project.name,
+                "name": self.project.name,
             },
         )
         self.assertEqual(response.status_code, 200)
         self.assertDictEqual(
             response.json(),
-            self._get_response_dict('projects-list'),
+            self._get_response_dict("projects-list"),
         )
 
     def test_projects_list_filter_partial_hit(self):
-        self.client.credentials(HTTP_AUTHORIZATION=f'Token {self.token.key}')
+        self.client.credentials(HTTP_AUTHORIZATION=f"Token {self.token.key}")
         response = self.client.get(
-            reverse('projects-list'),
+            reverse("projects-list"),
             data={
-                'name': self.project.name[0:3],
+                "name": self.project.name[0:3],
             },
         )
         self.assertEqual(response.status_code, 200)
         self.assertDictEqual(
             response.json(),
-            self._get_response_dict('projects-list'),
+            self._get_response_dict("projects-list"),
         )
 
     def test_projects_list_filter_miss(self):
-        self.client.credentials(HTTP_AUTHORIZATION=f'Token {self.token.key}')
+        self.client.credentials(HTTP_AUTHORIZATION=f"Token {self.token.key}")
         response = self.client.get(
-            reverse('projects-list'),
+            reverse("projects-list"),
             data={
-                'name': "63dadecd5323d789cafe09f01cda85fd",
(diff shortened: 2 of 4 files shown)


docs/user/api/v3.rst
```diff
@@ -186,6 +186,8 @@ Projects list
             }]
         }
 
+    :query string name: return projects with matching name
+    :query string slug: return projects with matching slug
```
humitos, line 190: Does "with matching ..." conveys that it could only partially match? I didn't get at first without reading the code.

humitos (approved):
I think this is fine as long as we use this filter only over the projects the user is owner/maintainer. Otherwise, it will kill our servers.

docs/user/api/v3.rst
agjohnson: I copied this from the Version API docs for consistency, but agree both could be clearer. I almost wanted to have fields `name` and `name__icontains`, but just followed what the Version related API pieces did.

agjohnson:
> I think this is fine as long as we use this filter only over the projects the user is owner/maintainer. Otherwise, it will kill our servers.

This is a great point, I hadn't thought of that. The intention is to only use this param to filter owner/user projects. However, if we are still allowing public project listing search, someone could abuse this query.

If you have any thoughts on how to secure this more, this might be a good place for you to jump in on this. Seems we could probably truncate the queryset or validate the request somehow, if we're concerned about querying all projects with this.

agjohnson:
Dev docs failure isn't related to this change.

humitos:
> This is a great point, I hadn't thought of that. The intention is to only use this param to filter owner/user projects. However, if we are still allowing public project listing search, someone could abuse this query.

I think we are fine: https://github.com/readthedocs/readthedocs.org/blob/d8f528263e3fa9485bec694dc1b53a8e4076436d/readthedocs/api/v3/permissions.py#L25-L26

We allow details on all the objects and listing only over the objects the user has access to.

readthedocs/api/v3/tests/test_projects.py
```diff
@@ -23,6 +23,49 @@ def test_projects_list(self):
...
+
+    def test_projects_list_filter_miss(self):
+        self.client.credentials(HTTP_AUTHORIZATION=f"Token {self.token.key}")
+        response = self.client.get(
+            reverse("projects-list"),
+            data={
+                "name": "63dadecd5323d789cafe09f01cda85fd",
+            },
+        )
+        print(response.request)
```
benjaoming, line 62:
```suggestion
        )
```

benjaoming (commented):
LGTM, perhaps when the print statement is removed, this can go in?

readthedocs/api/v3/tests/test_projects.py
benjaoming: I seem to still say "print statement" even for Python 3 :)

docs/user/api/v3.rst
agjohnson: I don't have strong opinions here, both certainly work. I'll defer to you on this one as you are most familiar with the API construction.

humitos: I'd follow your approach (using `__` on filters) because I think it's more flexible and will allow better support of features in the future.

I think it's better to do it now without breaking changes than later.

agjohnson: Makes sense. Should I keep the `name` filter as an exact match too you think? I will be using the fuzzy match for the dashboard, but exact match on name/slug also seems useful for API users.

I won't touch `Version` filters, which will differ slightly in function (`name` is fuzzy match on `Version`)

humitos: > Should I keep the name filter as an exact match too you think?

I'd say "No" here so all the endpoint are consistent. `name` will be treated as an alias of `name__icontains` to keep backward compatibility and a standardized way among all the endpoints.

agjohnson: Hrm, so I suppose this PR is then accurate then -- that is, leave the `name` filter as `name__icontains`? Or are you describing something else here too?

Perhaps it helps to treat my suggestion as a next iteration and perhaps spend some time on a wider solution. Perhaps django-filters is a solution to weight here?

humitos: This PR is fine, yes 👍🏼 

> Perhaps it helps to treat my suggestion as a next iteration and perhaps spend some time on a wider solution. Perhaps django-filters is a solution to weight here?

If you are not going to implement the `__<filter>` in this PR, we should open an issue with the plan we talked here --so, we don't deviate from it into an incompatible direction when we need a new filter.

humitos: Actually, besides the issue, you can add a comment in the code explaining this too 😄

agjohnson: Good plan. I opened https://github.com/readthedocs/readthedocs.org/issues/9862 to track this and updated the filter with a comment for later.

humitos:
This is ready to merge 💯

Issue #9718, opened 2022-11-08
[UX] Filter repositories sidebar should label the account origins


webknjaz:
## Details

I've got GitHub, GitLab and Bitbucket accounts linked to RTD. All of these account have the same username and the profile picture. As a result, on the "Import a Repository" page, the "Filter repositories" sidebar shows 3 exactly identical entries.
I suppose it'd be the same if there'd be an organization I'd have access to across multiple hostings.

* Read the Docs project URL: https://readthedocs.org/dashboard/import/
* Read the Docs username (if applicable): https://readthedocs.org/profiles/webknjaz/

## Expected Result

It should be possible to distinguish between the accounts from different code hostings.
The easiest solution I could think of is to just add icons of the corresponding code hosting next to each entry.
While there, it could also be useful to know if an account is a user or an org.

## Actual Result

It's a guessing game. Here's what it looks like to me:
![rtd-confusing-filtering](https://user-images.githubusercontent.com/578543/200646529-737015af-f8df-4ad1-8b10-864a4838cb86.png)

humitos:
We solved this issue in the new templates. You can search for the specific repository you want now and it auto-completes/auto-suggests the results showing all the organizations/usernames you belong to:

![Screenshot_2023-06-05_12-20-13](https://github.com/readthedocs/readthedocs.org/assets/244656/c6bca388-f098-4656-9f43-2d1af5834239)

Please, give it a try at https://beta.readthedocs.org/ and let us know any feedback you may have 👍🏼

PR #9848, opened 2022-12-29
Import Repository Account Filter Display Provider Name

saadmk11:
closes https://github.com/readthedocs/readthedocs.org/issues/9718

### Output:
Screenshot from 2022-12-29 16-11-52

**Note:** The Remote Organisation API (`/api/v2/remote/org/`) does not return the Provider Name with correct case but the Account API Does. We can either change the Remote Organisation API to return the display name or update `readthedocs/projects/static-src/projects/js/import.js` to change the name depending on the value (e.g.: `github` -> `GitHub`)


benjaoming:
Hi @saadmk11 :wave: 

I think the solution looks like a good a simple way to address this.

When generating `display_name` is there any context available to see how many providers are displayed? Or can that context be added and applied easily?

This could be great to avoid that users with just 1 provider (I'm guessing most users) do not see "GitHub" or "Bitbucket" displayed for no apparent reason.

saadmk11:
> When generating `display_name` is there any context available to see how many providers are displayed? Or can that context be added and applied easily?

I don't think so. It is stored in `ProjectImportView.organizations_raw` / `ProjectImportView.accounts_raw`. I would always show the providers, a not so common case can be that the user forgot which provider they connected and they need to check their account settings to figure it out because the Profile picture and username are same.

benjaoming:
It's also not too bad to always display the provider name :+1:

What about writing some generic short notation like `github/acme` instead of `acme (github Organization)` ? I'm not even sure if both Bitbucket and GitLab call these "organizations". I think on Bitbucket, it's called a Workspace :upside_down_face:

saadmk11:
> It's also not too bad to always display the provider name +1
> 
> What about writing some generic short notation like `github/acme` instead of `acme (github Organization)` ? I'm not even sure if both Bitbucket and GitLab call these "organizations". I think on Bitbucket, it's called a Workspace upside_down_face

Screenshot from 2023-01-08 18-20-30

benjaoming (approved):
LGTM :100: 

This way of writing it makes it easy for me to quickly skim the list when there are multiple providers.

In other cases, it still makes the list look good by clearly indicating what belongs to a private profile and what belongs to an organization.

A secondary effect: Maybe not all users will know that Read the Docs supports connecting multiple accounts. The design of the list hints that this is an option. By making the list appear this way, we could in the future add a link below "connect additional Git providers" or similar.

humitos:
This page has changed considerably in our new templates and I'm not sure it makes sense to merge this work anymore. If anything, we should make the changes in the new templates.

This is an example of how it's going to be:

Peek 2023-03-07 11-02

Please, feel free to comment here if the new idea does not cover the case this PR was covering

saadmk11:
> This page has changed considerably in our new templates and I'm not sure it makes sense to merge this work anymore. If anything, we should make the changes in the new templates.
> 
> This is an example of how it's going to be:
> 
> Peek 2023-03-07 11-02 [ Peek 2023-03-07 11-02 ](https://user-images.githubusercontent.com/244656/223389683-1d25c2f4-4b26-4a9a-971a-85577d61f7c3.gif) [ ](https://user-images.githubusercontent.com/244656/223389683-1d25c2f4-4b26-4a9a-971a-85577d61f7c3.gif)
> 
> Please, feel free to comment here if the new idea does not cover the case this PR was covering

@humitos This looks nice. :100:  Is the new UI available for testing?

humitos:
@saadmk11 not yet, unfortunately. We are working hard to make a public beta release soon, tho 👍🏼

humitos:
Hey @saadmk11! I'm happy to let you know how to jump into the private alpha/beta instance that contains the new templates. Take into account that this instance is not publicly accessible for everybody and, in theory, nobody knows about it. We are using it internally to test the progress on this work.

The URL to connect to it is https://beta.readthedocs.org. You need to manually create a cookie in your side with the key `beta` and the value `true` (I've done this with Firefox by pressing `F12 -> Storage -> Cookies -> "Plus" sign`) . That's all 😅 

Feel free to open an issue in this repository with your feedback in case you have any. I'm happy to receive more general feedback and also specific feedback. However, keep in mind that things can drastically change and nothing is written in stone 😄

saadmk11:
> Hey @saadmk11! I'm happy to let you know how to jump into the private alpha/beta instance that contains the new templates. Take into account that this instance is not publicly accessible for everybody and, in theory, nobody knows about it. We are using it internally to test the progress on this work.
> 
> The URL to connect to it is https://beta.readthedocs.org. You need to manually create a cookie in your side with the key `beta` and the value `true` (I've done this with Firefox by pressing `F12 -> Storage -> Cookies -> "Plus" sign`) . That's all sweat_smile
> 
> Feel free to open an issue in this repository with your feedback in case you have any. I'm happy to receive more general feedback and also specific feedback. However, keep in mind that things can drastically change and nothing is written in stone smile

Thanks will try it out :)

PR #9865, opened 2023-01-05
Account: organize allauth settings

humitos:
- make email verification `mandatory`: users without a verified email will be asked to verify their address before being able to login
- rename `ACCOUNT_ACTIVATION_DAYS` since it was a setting from a Django app we don't use anymore. It's called `ACCOUNT_EMAIL_CONFIRMATION_EXPIRE_DAYS` in Django Allauth.
- email verification is not required for development instance

This PR **requires other PRs** to work as intended:
* https://github.com/readthedocs/readthedocs-ops/pull/1271
* https://github.com/readthedocs/readthedocs-corporate/pull/1529

> NOTE that I review all the settings from https://django-allauth.readthedocs.io/en/latest/configuration.html and I think we are fine moving forward with this combination.

Context: https://github.com/readthedocs/meta/discussions/103
Closes: https://github.com/readthedocs/meta/issues/104

### DIFF
--- a/readthedocs/settings/base.py
+++ b/readthedocs/settings/base.py
@@ -701,11 +701,16 @@ class CommunityBaseSettings(Settings):
     # All auth
     ACCOUNT_ADAPTER = 'readthedocs.core.adapters.AccountAdapter'
     ACCOUNT_EMAIL_REQUIRED = True
+
+    # Make email verification mandatory.
+    # Users won't be able to login until they verify the email address.
     ACCOUNT_EMAIL_VERIFICATION = 'mandatory'
+
     ACCOUNT_AUTHENTICATION_METHOD = 'username_email'
-    ACCOUNT_ACTIVATION_DAYS = 7
+    ACCOUNT_EMAIL_CONFIRMATION_EXPIRE_DAYS = 7
     SOCIALACCOUNT_AUTO_SIGNUP = False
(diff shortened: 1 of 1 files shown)


ericholscher:
I wish there was a way to set this at the DB level or with a feature flag. I'm sure we could find a way to hack it, but I just assume that when we deploy this we'll have some subset of users get locked out of their accounts. Something to be on the lookout for, but I do agree it's a good default worth trying to enable.

agjohnson (approved):
Seems good enough! I also recall some issues with verification in the past but don't remember what the problem was. I'm guessing either the problem is outdated and we're fine, or we'll find out really quick why it's not mandatory.

humitos:
I'm not merging this just yet because we talked about adding an on-site notification to all those users that don't have any email address validated and give them some weeks before deploying this change.

Next step: create a simple script to add the notification to all of them 😄

ericholscher:
I believe we have a couple scripts that already does this: https://github.com/readthedocs/readthedocs.org/blob/main/readthedocs/core/management/commands/contact_owners.py -- if it can easily be modified.

ericholscher:
Ah, I just saw the attached PR 🙃

humitos:
We sent the on-site notification to our users and we waited 3 weeks. We are deploying these changes and making the email verification mandatory for login next week.

PR #9874, opened 2023-01-09
Suggestions/grammar for user-facing texts

Justman10000:


readthedocs/templates/homepage.html
```diff
@@ -147,8 +147,8 @@ <h2>{% trans 'About Read the Docs' %}</h2>
     <!-- Updated: June 2018, pageviews June 2021 -->
     {% blocktrans trimmed %}
     <ul class="donate-about">
-      <li>Serve over <b>55 million pages</b> of documentation a month</li>
-      <li>Serve over <b>40 TB</b> of documentation a month</li>
+      <li>Serve over <b>55 million pages</b> of documentations a month</li>
+      <li>Serve over <b>40 TB</b> of documentations a month</li>
```
benjaoming, line 151: Documentation cannot be pluralized. The original text is correct.

readthedocs/templates/homepage.html
```diff
@@ -55,7 +55,7 @@ <h2>{% trans "Technical documentation lives here" %}</h2>
 
   <section>
     <div class="feature">
-      <h3>{% trans "Free docs hosting for open source" %}</h3>
+      <h3>{% trans "Free docs hosting for open source projects" %}</h3>
```
benjaoming, line 58: While adding "projects" could be slightly more accurate, the former text is also understandable and to my knowledge not incorrect. For instance, if someone asks me what I do, I will often respond "I work with open source".

"Richard Stallman has a long legacy in open source"
"[Read the Docs provides] Free docs hosting for open source"
"OpenSUSE's community love open source"
etc.

benjaoming: I don't think this change to a source translation is worth breaking existing translations (seeing that this would be the outcome)

Justman10000: > I don't think this change to a source translation is worth breaking existing translations (seeing that this would be the outcome)

They would then be updated...

benjaoming: Yes exactly. That's the problem.

benjaoming:
We'll close this because a) no feedback, b) breaking current translations without clear benefits, c) we're in a process of re-writing most of this code.

Thanks for stopping by :+1:

Issue #9869, opened 2023-01-06
Incorrect Sphinx version used during RTD build


sethfischer:
## Details

* Read the Docs project URL: https://readthedocs.org/projects/sphinxcontrib-cadquery/
* Build URL: https://readthedocs.org/projects/sphinxcontrib-cadquery/builds/19088114/
* Read the Docs username: https://readthedocs.org/profiles/sethfischer/

## Expected Result

Build [19088114](https://readthedocs.org/projects/sphinxcontrib-cadquery/builds/19088114/) is expected to use Sphinx 5.3.0.

## Actual Result

Build [19088114](https://readthedocs.org/projects/sphinxcontrib-cadquery/builds/19088114/) uses Sphinx 6.1.1.

## Further information

[rtd-build-19088114.txt](https://github.com/readthedocs/readthedocs.org/files/10356438/rtd-build-19088114.txt)

Build 19088114 uses Poetry to install dependencies with a `build.jobs.post_create_environment` user-defined job, in a manner [recommended by RTD](https://docs.readthedocs.io/en/stable/build-customization.html#install-dependencies-with-poetry).

As expected `poetry install` installs Sphinx 5.3.0. This is confirmed with `poetry show sphinx` in `build.jobs.post_create_environment`.

However, when the following is run by RTD:

```
/home/docs/checkouts/readthedocs.org/user_builds/sphinxcontrib-cadquery/envs/1/bin/python -m sphinx -T -E -b html -d _build/doctrees -D language=en . _build/html
```

The output shows that Sphinx 6.1.1 is being used:

```
Running Sphinx v6.1.1
WARNING: html_static_path entry '_static' does not exist
building [mo]: targets for 0 po files that are out of date
writing output...
⋮
```

`.readthedocs.yaml`:

```yaml
version: 2

build:
  os: ubuntu-20.04
  tools:
    nodejs: "19"
    python: "3.9"
  jobs:
    post_create_environment:
      # install Poetry
      - pip install poetry
      - poetry config virtualenvs.create false
      # install Python dependencies
      - poetry install
      - poetry show sphinx
    post_install:
      # install Node.js dependencies
      - npm clean-install
      # build
      - npm run build

sphinx:
  configuration: docs/conf.py
```

benjaoming:
Hi @sethfischer 

It seems there is a built-in `pip install` which installs the latest sphinx *after* `post_create_environment` is run.

What happens if you move these to `post_install`?

```
    post_install:
[...]

PR #9879, opened 2023-01-10
Docs: update build customization Poetry example

sethfischer:
Move `poetry install` from `post_create_environment` to `post_install`.

This ensures the Sphinx version installed from `poetry.lock` is used to build the docs, rather than the default version of Sphinx.

Closes: #9869

---

- User's documentation (`docs`): https://docs--9879.org.readthedocs.build/en/9879/

- Developer's documentation (`dev`): https://dev--9879.org.readthedocs.build/en/9879/

### DIFF
--- a/docs/user/build-customization.rst
+++ b/docs/user/build-customization.rst
@@ -330,6 +330,7 @@ Take a look at the following example:
          - pip install poetry
          # Tell poetry to not use a virtual environment
          - poetry config virtualenvs.create false
+       post_install:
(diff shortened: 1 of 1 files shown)


benjaoming:
Note: The build failure in `checks` is related to blacken-docs. It's because we are not using correct Python syntax in a `code-block:: Python` -- it's a change that overlapped with the introduction of the blacken-docs pre-commit hook.

PR #9891, opened 2023-01-12
UI updates to Connected Accounts

benjaoming:
Related to additions in #9834 

Also related: The terminology of "Git provider" to be discussed in #9840 

Example:

image

### DIFF
--- a/readthedocs/projects/static-src/projects/css/admin.less
+++ b/readthedocs/projects/static-src/projects/css/admin.less
@@ -38,3 +38,14 @@
         content: "\f063";
     }
 }
+
+.connected-services {
+    ul.socialaccount_providers {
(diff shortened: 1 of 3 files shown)


ericholscher (approved):
This looks like a great change, but I do think we should be minimizing UI changes on the old dashboard and trying to push ahead on the new one. But these smallish fixes are good 👍

benjaoming:
definitely agree to minimalizing changes, but the additional help-text was pushing the "Connect to <Service>" buttons further down... so they looked very "secondary"... that made it nicer to move them to the top, after which the page needed an introduction text and thus: :rabbit: :hole: 

Gonna just read the comments about "Git provider" before merging...

PR #9892, opened 2023-01-12
Docs: add warning that pull requests only build HTML and not other formats

mwtoews:
This adds a warning to the [Configuration File V2, Formats](https://docs.readthedocs.io/en/stable/config-file/v2.html#formats) section, to help users like myself to evaluate why the formats seemed to be ignored while building docs in pull requests.

Rewording suggestions or moving this to (e.g.) the "note" section is welcome.

---

- User's documentation (`docs`): https://docs--9892.org.readthedocs.build/en/9892/

- Developer's documentation (`dev`): https://dev--9892.org.readthedocs.build/en/9892/

### DIFF
--- a/docs/user/config-file/v2.rst
+++ b/docs/user/config-file/v2.rst
@@ -144,6 +144,8 @@ Example:
    At the moment, only Sphinx supports additional formats.
    ``pdf``, ``epub``, and ``htmlzip`` output is not yet supported when using MkDocs.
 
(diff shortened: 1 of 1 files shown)


benjaoming (approved):
Thanks for this :+1: 

I like the change:

* Expands an existing warning with a related message
* Based on user feedback

It's also mentioned in the PR documentation but it's understandable that it can be missed there! https://docs.readthedocs.io/en/stable/pull-requests.html#limitations

PR #9893, opened 2023-01-13
CI: Add option `--show-diff-on-failure`  to pre-commit

benjaoming:
Referencing troubleshooting in https://github.com/readthedocs/readthedocs.org/pull/9411#issuecomment-1381366598, I think it would be really nice to see what pre-commit detects.

Docs: https://pre-commit.com/#pre-commit-run

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -77,7 +77,7 @@ deps =
 commands =
     # FIXME: use `github.event.pull_request.base.sha` in `--from-ref` because if
     # the base branch is different, this won't work as expected
-    pre-commit run --from-ref main --to-ref HEAD
+    pre-commit run --from-ref main --to-ref HEAD --show-diff-on-failure
 
 [testenv:eslint]
 allowlist_externals = npm


benjaoming:
In a one-liner change to documentation, I am now getting [a pre-commit error](https://app.circleci.com/pipelines/github/benjaoming/readthedocs.org/199/workflows/3ed83982-b468-4ff9-96bd-c33eec98df2f/jobs/512) because of a completely unrelated change.

Then I have to re-run pre-commit locally to see the diff.

Result:

```
diff --git a/readthedocs/rtd_tests/tests/test_privacy_urls.py b/readthedocs/rtd_tests/tests/test_privacy_urls.py
index c351e305c..ffebb762f 100644
--- a/readthedocs/rtd_tests/tests/test_privacy_urls.py
+++ b/readthedocs/rtd_tests/tests/test_privacy_urls.py
@@ -48,7 +48,11 @@ class URLAccessMixin:
             path += "/"
         path += "lib.js"
         r = self.client.head(path)
-        self.assertNotEqual(r.status_code, 200, f"Path vulnerable to cache poisoning. path={original_path}")
+        self.assertNotEqual(
+            r.status_code,
+            200,
+            f"Path vulnerable to cache poisoning. path={original_path}",
+        )
 
     def assertResponse(self, path, name=None, method=None, data=None, **kwargs):
         self.login()

```

I would be very happy, if we can get this change in.

I'm not sure, but I think that the above error is because of the comparison to a different (and newer!) `main` branch than what the PR is targeting (`diataxis/main`). We cannot unfortunately fix this because of the same shortcoming in Circle CI. If anyone wants to know, I can elaborate :+1:

ericholscher (approved):
Seems like a simple enough change 👍

PR #9897, opened 2023-01-13
Docs: Relabel Server Side Search

benjaoming:
Relabling these articles. We should separate references and explanations and not gather them under domain-specific subtopics.

Aside from that, I kept changes minimal here as the articles went under a recent review.

Open question: Is "Server Side Search" to be understood as a proper noun? Or should we just say "server-side search"?

I removed the local TOCs as we haven been using that in many other places.

Refs: #9746 

---

- User's documentation (`docs`): https://docs--9897.org.readthedocs.build/en/9897/

- Developer's documentation (`dev`): https://dev--9897.org.readthedocs.build/en/9897/

### DIFF
--- a/docs/user/index.rst
+++ b/docs/user/index.rst
@@ -87,6 +87,7 @@ to help you create fantastic documentation for your project.
    /single_version
    /science
    /commercial/organizations
+   /server-side-search/index
 
 .. toctree::
    :maxdepth: 2
@@ -106,8 +107,10 @@ to help you create fantastic documentation for your project.
    :caption: Reference
    :glob:
 
-   api/index
-   reference/features
+   /api/index
+   /server-side-search/api
+   /reference/features
+   /server-side-search/syntax
 
 
 Read the Docs feature overview
@@ -153,7 +156,6 @@ and some of the core features of Read the Docs.
    /integrations
    /versions
    /hosting
-   /server-side-search/index
    /security-log
 
    /builds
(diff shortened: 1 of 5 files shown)


benjaoming:
There was also a suggestion to change the title of the explanation article to "Fast server-side search" - I do think it sounds better because it puts emphasis on one of the most important aspects of switching from the built-in search of Sphinx (and MkDocs in the future?).

docs/user/index.rst
```diff
@@ -87,6 +87,7 @@ to help you create fantastic documentation for your project.
    /single_version
    /science
    /commercial/organizations
+   /server-side-search/index
```
ericholscher, line 90: This page doesn't seem like explanation, but feature reference to me? It doesn't really talk at all about search at a high level, but about our implementation?

ericholscher (commented):
I agree the local TOC's are not great, and we should likely just have them in the theme like many other themes (eg. Furo)

docs/user/index.rst
benjaoming: Currently, it _explains_ features.

Do you think that it will be better to move into the current feature reference section?

https://docs.readthedocs.io/en/diataxis-main/reference/features.html

ericholscher: Yes, I think that is where we should be talking about our features explicitly, not about higher level concepts like explanation. I imagine the higher level content would be things like "What to think about when implementing search" and "ways to write content to ensure good searchability". Perhaps I'm thinking a bit too abstractly for our explanation, but I think anything that's explaining our product itself seems like a feature reference.

benjaoming: It's a really good definition and distinction. This can also help to avoid the "marketing effect" in documentation.

I'll move it :+1:

ericholscher (approved):

PR #9908, opened 2023-01-17
Docs: Move and update FAQ (Diátaxis)

benjaoming:
Refs: #9746 

The FAQ doesn't fit that well in Diátaxis in the sense that it's a very unstructured way of delivering content. I also invites "dumping" of documentation, which is a bit of an anti-pattern. So I'll consider it "deprecated" but that we shouldn't delete it before the decision is very easy to take.

Placement as How-To: I think it's very much a how-to, since an FAQ requires the user to know the problem they want to solve and need an instruction on how to solve it. Most likely, it's because they follow an internal link.

I'll clean it up with the following pattern:

* Frequently asked questions should each consist of: A question, an answer, a reference to another documentation page

## Potential deprecation path

We could keep the FAQ and strip out all its instructions to actual locations in the docs (replacing instructions in FAQ with cross-references), then remove the FAQ from TOC trees and keep it public to avoid breaking incoming links.

The current FAQ is the 20th most viewed page. I don't think it says a lot. It has 1% of all Top 20 hits (51k total, FAQ had 638 views past 30 days).

---

- User's documentation (`docs`): https://docs--9908.org.readthedocs.build/en/9908/

- Developer's documentation (`dev`): https://dev--9908.org.readthedocs.build/en/9908/

### DIFF
--- a/docs/user/faq.rst
+++ b/docs/user/faq.rst
@@ -9,11 +9,16 @@ Frequently Asked Questions
   Formulate them as a question and an answer.
   Consider that the answer is best as a reference to another place in the documentation.
 
+
+Building and publishing your project
+------------------------------------
+
+
 .. Old reference
 .. _My project isn't building correctly:
 
 Why does my project have status "failing"?
-------------------------------------------
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 Projects have the status "failing" because something in the build process has failed.
 This can be because the project is not correctly configured,
@@ -41,7 +46,7 @@ you can use an important word or message from the error to search for a solution
 .. _Help, my build passed but my documentation page is 404 Not Found!:
 
 My build says "passed" but I get a 404 page - why?
---------------------------------------------------
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 This often happens because you don't have an `index.html` file being generated.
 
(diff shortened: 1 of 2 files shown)


docs/user/index.rst
```diff
@@ -90,14 +90,14 @@ to help you create fantastic documentation for your project.
 
 .. toctree::
    :maxdepth: 2
-   :hidden:
    :caption: How-to Guides
-   :glob:
+   :titlesonly:
```
benjaoming, line 94: Using `:titlesonly:` is an experiment. I'll probably revert it so we can keep a consistent display across the documentation. But the FAQ item is quite a mess with all the FAQ entries in it.

benjaoming:
In order to make this look less messy, I'll introduce some new top-levels in the FAQ

```
Building and publishing your project
------------------------------------

Why is my project “failing”?
My build says “passed” but I get a 404 page - why?
Why do I get import errors from libraries depending on C modules?
Where do I need to put my docs for RTD to find it?
What commit of Read the Docs is in production?

Additional features and configuration
-------------------------------------

How do I add additional software dependencies for my documentation?
Can I have access to additional features or settings?
How do I change behavior when building with Read the Docs?
I want comments in my docs
How can I avoid search results having a deprecated version of my docs?
Can I remove advertising from my documentation?
How do I change my project slug (the URL your docs are served at)?
How do I change the version slug of my project?

Big projects
------------

How do I host multiple projects on one custom domain?
How do I support multiple languages of documentation?

Sphinx
------

Does Read the Docs work well with “legible” docstrings?
I want to use the Blue/Default Sphinx theme
I want to use the Read the Docs theme locally
Image scaling doesn’t work in my documentation

Python
------

Can I document a Python package that is not at the root of my repository?
I need to install a package in a environment with pinned versions
Can I use Anaconda Project and anaconda-project.yml?

Other documentation frameworks
------------------------------

How can I deploy Jupyter Book projects on Read the Docs?
```

benjaoming:
Yes, I think the experiment of adding top levels is a success. This is how the FAQ looks now:

image

benjaoming:
At this stage, I think the FAQ section is okay. I find it a bit "hard" to spend more time updating the current items when the true solution is to find proper locations for them, typically a new how-to. But our how-to structure isn't done yet.

So I'd like to propose a scope like this:

* A lot of improvements have already been added
* Whatever you see that can be quickly improved, feel free :+1: 
* Do not get lost in texts from the past here, they are more likely to be removed and it doesn't make sense to update or add more Q&As at this point.

ericholscher:
Is this ready for review? I got confused by the `Not Ready` at the top.

benjaoming:
YES! https://github.com/readthedocs/readthedocs.org/pull/9908#issuecomment-1385872333 has the latest reflection.

benjaoming:
The git diff is quite horrible because I moved subsections around into new sections. Sorry. I'm not sure how to make it more pleasant for review.

docs/user/faq.rst
```diff
@@ -1,65 +1,71 @@
...
 
+Building and publishing your project
+------------------------------------
 
-Help, my build passed but my documentation page is 404 Not Found!
------------------------------------------------------------------
 
-This often happens because you don't have an `index.html` file being generated.
-Make sure you have one of the following files:
+.. Old reference
```
benjaoming, line 17: When you see this comment, it means that I've edited the subsection title.

ericholscher: Good approach. The goal here is not breaking inbound links?  👍

docs/user/index.rst
```diff
@@ -90,14 +90,13 @@ to help you create fantastic documentation for your project.
 
 .. toctree::
    :maxdepth: 2
-   :hidden:
```
ericholscher, line 93: Seems this is showing on the index page 2x now?

ericholscher (commented):
Hrm, I didn't really want to dive into reviewing all the content here, so not sure the best way to review this. 🤷 

I think in the future changing content, then restructuring in different commits is probably best. But for now, probably fine to just merge it? 🤷

ericholscher (approved):

docs/user/faq.rst
humitos: If the answer is "Yes", this is probably a good recommendation for the "Automatic redirect" page you are working on 👍🏼

docs/user/faq.rst
```diff
@@ -1,65 +1,71 @@
...
 
-If your project or its dependencies rely on C libraries that cannot be installed
-this way, see :ref:`faq:I get import errors on libraries that depend on C modules`.
 
+.. Old reference
+.. _Help, my build passed but my documentation page is 404 Not Found!:
 
-My project requires some additional settings
---------------------------------------------
+My build says "passed" but I get a 404 page - why?
```
humitos, line 48:
```suggestion
Why my build says "passed" but I get a 404 page?
```

Isn't it better/simpler just making it as a question?

docs/user/faq.rst
```diff
@@ -77,32 +83,164 @@ the C-dependent libraries from your ``install_requires`` in the RTD environment.
...
 
 
+I want comments in my docs
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+RTD doesn't have explicit support for this.
+That said, a tool like `Disqus`_ (and the `sphinxcontrib-disqus`_ plugin) can be used for this purpose on RTD.
+
+.. _Disqus: https://disqus.com/
+.. _sphinxcontrib-disqus: https://pypi.python.org/pypi/sphinxcontrib-disqus
```
humitos, line 216: We could update this one to mention https://giscus.app/ that seems that have more users now.

benjaoming: I've seen this before but haven't given it much consideration WRT documentation projects. If you can write a specific suggestion, that's great. I'll probably merge this PR now, so just DM me or open a new PR :+1:

docs/user/faq.rst
benjaoming: Thanks for that @humitos !!

Yes, I did a test yesterday and there's an invisible `<span id="old_ref">` inserted which makes `page.html#anchor` retained, as well as `:ref:`. I didn't test with Intersphinx, but I'm 99.9% sure that'll work as well.

docs/user/index.rst
benjaoming: Not for me, or maybe there's some way I didn't get this. Anyway, I've re-added it now. I don't quite grasp from the documentation why it's needed... but since it's in so many other TOCs, we can tackle it when the time feels right :)

https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-toctree

benjaoming:
Concluding this PR! Thanks for being gentle about it :1st_place_medal:

docs/user/faq.rst
humitos: > That's not an entire sentence

What do you mean by that?

ericholscher: It's not grammatically correct. You'd need something like

`Why does my build says "passed" but I get a 404 page?`

I agree that it's a better phrasing than the awkward `why` at the end tho.

docs/user/index.rst
ericholscher: @benjaoming It makes it so the TOCTree doesn't show up in the page where it's defined. We're managing the TOC's manually in RST, so if we remove this, it will output _another_ TOC. If you remove it and rebuild the index, you can see it.

docs/user/faq.rst
humitos: Ah, it was missing the "does". OK. Makes sense. Thanks for the explanation.

Yeah, I also prefer this "Why does..." form than the "- why?" one 👍🏼

benjaoming: > Why does my build say~s~ "passed" but I get a 404 page?

benjaoming: https://github.com/readthedocs/readthedocs.org/pull/9919

PR #9914, opened 2023-01-18
Docs: Add new section "How-to / Troubleshooting" and move 2 existing troubleshooting pages

benjaoming:
Refs: #9746 

This is a fairly simple change:

* [x] Create a new sublevel in navigation: "How-to guides / Troubleshooting"
* [x] Move "Troubleshooting build errors"
* [x] Move "Troubleshooting slow builds"
* [x] Adds a new type of suggestion to send in comments and ideas (because Troubleshooting can benefit from the additional grounding in direct user feedback)

I imagine that we will be able to write troubleshooting guides much quicker in the future, given that they have an easily-expandable section now.

Some FAQ items might go and live there in the future.

---

- User's documentation (`docs`): https://docs--9914.org.readthedocs.build/en/9914/

- Developer's documentation (`dev`): https://dev--9914.org.readthedocs.build/en/9914/


benjaoming:
I just realized some other Troubleshooting content to be relabled - so launching "How-to guides / Troubleshooting" is already on the table and easier done in one PR :+1:

docs/user/guides/build-troubleshooting.rst
```diff
@@ -1,8 +1,11 @@
...
+Troubleshooting build errors
+============================
 
-This section is very specific and may change over time.
-Please help us keep it updated and contribute your own build errors and resolutions.
+
+.. tip::
+
+   Please help us keep this section updated and contribute your own build errors and resolutions.
+   Send in helpful ideas 💡 to support@readthedocs.org.
```
humitos, line 8: Wouldn't it be better to open an issue/send a pull request instead?

benjaoming: Definitely as a supplement :+1: Added that.

Since "Edit on GitHub" is already a present feature that technical users might know, I think that noting the support email for ideas and suggestions is also fine... to keep the bar low for adding stuff to a page like this.

docs/user/guides/troubleshooting/index.rst
```diff
@@ -0,0 +1,14 @@
...
+you can learn how to fix specific common problems.
+
+.. File locations are a bit odd here because of legacy.
+.. We will move files to new locations and create redirects in a later refactor
+
+.. toctree::
+   :maxdepth: 1
+
+   Build errors <../build-troubleshooting>
+   Slow builds <../build-using-too-many-resources>
```
benjaoming, line 14: I left these pages in their original location intentionally. We will move pages and create redirects in a separate workflow.

docs/user/guides/troubleshooting/index.rst
```diff
@@ -0,0 +1,14 @@
+How to troubleshoot specific issues
+-----------------------------------
```
ericholscher, line 2:
```suggestion
Troubleshooting Guides
----------------------
```

The original title didn't read like an index listing, but instead like a how to article. I think we should be pretty explicit in language around index or listing pages. 

I could even be convinced on something like `Troubleshooting Guide Listing` or similarly explicit.

docs/user/guides/troubleshooting/index.rst
```diff
@@ -0,0 +1,14 @@
...
+
+In the following guides,
+you can learn how to fix specific common problems.
```
ericholscher, line 5:
```suggestion
you can learn how to fix common problems using Read the Docs.
```

docs/user/guides/troubleshooting/index.rst
```diff
@@ -0,0 +1,14 @@
...
+
+.. File locations are a bit odd here because of legacy.
+.. We will move files to new locations and create redirects in a later refactor
```
ericholscher, line 8: This is a great comment, and I suggested somethign similar in another PR 👍

docs/user/guides/troubleshooting/index.rst
ericholscher: This is another place where I think it probably makes sense to keep the original titles, since they are so close to these.

docs/user/index.rst
```diff
@@ -97,6 +97,7 @@ to help you create fantastic documentation for your project.
    /guides/authors
    /guides/administrators
    /guides/developers
+   Troubleshooting </guides/troubleshooting/index>
```
ericholscher, line 100: You already had the proper title for the page here ;)

ericholscher (approved):
Lots of comments for a tiny PR, but I think a new guide section is worth thinking about! Exciting.

docs/user/guides/build-using-too-many-resources.rst
```diff
@@ -1,11 +1,20 @@
...
+.. note::
+
+   Build resources on Read the Docs are limited to make sure that users don't overwhelm our build systems.
+   The current build limits can be found on our :doc:`/builds` page.
+
+.. tip::
+
+   Please help us keep this section updated and contribute your own performance improvements.
+   Send in your helpful comments or ideas 💡 to support@readthedocs.org
+   or contribute directly by clicking :guilabel:`Edit on GitHub` in the top right corner of this page.
```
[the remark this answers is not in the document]
benjaoming: Yup, I made them slightly different to justify _not_ having an include... but that is silly :smile: 

I'll make them the same. As mentioned, it's not that I think we should put a box like this on all documentation pages, but more to highlight how much this particular type of page benefits from everyone's input.

docs/user/guides/troubleshooting/index.rst
benjaoming: Are you referring to changing original page titles or the fact that they are explicitly labeled in the TOC/Navigation?

If for the TOC/Navigation, I wanted to have it look like this:

```
Troubleshooting
  Slow builds
  Build errors
```

And not for instance:

```
Troubleshooting guides
  Troubleshooting slow builds
  Troubleshooting build errors
```


WRT page title changes, I wanted to reuse the pattern from before but got distracted...

```

.. _Old Title:

New title
=========

```

ericholscher:
```
Troubleshooting
  Slow builds
  Build errors
```

Yea.. that looks nicer for sure. I wonder about a feature where the section is included in the h1, as a type of Breadcrumb, so that we don't have to keep overwriting things in TOC's. Probably over-optimizing tho for a smallish use case, and also probably a bunch of places we _don't_ want this :)

benjaoming: I want this, too, and I've been wondering if we could write this type of meta-data for pages before and use it in a template-like way. I think it's also okay to write it directly in the TOC. DRY is still retained.. we write the navigation title exactly where it is used, so even though it's a bit disconnected from its original source, it's still in a context where it seems "understandable".

humitos: We should consider SEO in case we want to automate this somehow. What's better "Troubleshooting: Slow builds" or "Slow builds" in the H1? Depending on the implementation (js/static) this could be a problem or not.

I don't think we are seriously talking about implementing this, tho, but I just want to mention this.

benjaoming: Can you clarify? In this case, we would want the full title in the H1 and a shorter version in the navigation.

humitos: That sounds good to me 👍🏼

docs/user/guides/build-using-too-many-resources.rst
```diff
@@ -1,11 +1,20 @@
...
-there are a couple fixes that you might try.
+In case you are waiting a long time for your builds to finish
+or your builds are terminated by exceeding general resource limits,
+this troubleshooting guide will help you resolve some of the most common issues causing slow builds.
 
-.. note:: The current build limits can be found on our :doc:`/builds` page.
+.. note::
+
+   Build resources on Read the Docs are limited to make sure that users don't overwhelm our build systems.
+   The current build limits can be found on our :doc:`/builds` page.
```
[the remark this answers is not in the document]
benjaoming: Fixed! And started to use "hoverxref'ing" as a verb as often as possible :D

PR #9915, opened 2023-01-18
Docs: Move Main Features and Feature Flags to "Reference/Features" (Diátaxis)

benjaoming:
Refs: #9746 

* [x] Relabels "Feature Flags" as Reference
* [x] Moves "Main Features" to Reference/Features as well (this item is suggested for deletion later, so this is mainly a change to help the new "Reference/Features" section be easier to observe, scope, and define)

---

- User's documentation (`docs`): https://docs--9915.org.readthedocs.build/en/9915/

- Developer's documentation (`dev`): https://dev--9915.org.readthedocs.build/en/9915/


benjaoming:
Oh, actually it should also be noted that "Feature Flags" are also going to be deprecated and ultimately removed, so I think it's fine to not spend too much time here.

docs/user/reference/features.rst
```diff
@@ -6,4 +6,6 @@ Features
 .. toctree::
    :maxdepth: 1
 
+   ../features
    analytics
+   ../feature-flags
```
ericholscher, line 11: This is a bit wonky. Probably could use a comment explaining why these are in the directory above:

```suggestion
   ../features
   analytics
   ../feature-flags

.. These are not moved into `reference/features` yet to avoid renames. 
```

ericholscher (approved):

docs/user/feature-flags.rst
```diff
@@ -1,9 +1,8 @@
 Feature Flags
 =============
```
ericholscher, line 3:
```suggestion

.. **Please don't add more stuff here**. We want to move user-facing options to the config file. 
```

docs/user/reference/features.rst
benjaoming: Added a good bit of commenting here :)

benjaoming:
All inputs addressed, merging :+1:

PR #9916, opened 2023-01-18
Docs: Relabel pages to new top-level "Reference/Policies and legal documents" (Diátaxis)

benjaoming:
Refs: #9746 

This top-level is easiest to introduce and populate in one go.

* [x] Move all articles that are either a policy or a legal document
* [x] Rename all policies to "<topic> policy", i.e. not "Policy for <topic>".
* [x] Move contents from old section "Legal documents and policies" and remove `legal/index.rst` since its replaced by `reference/policies.rst`

---

- User's documentation (`docs`): https://docs--9916.org.readthedocs.build/en/9916/

- Developer's documentation (`dev`): https://dev--9916.org.readthedocs.build/en/9916/


docs/user/security.rst
```diff
@@ -2,8 +2,8 @@
 ..    https://readthedocs.org/.well-known/security.txt
 ..    https://readthedocs.org/security/
 
-Security
-========
+Security practice
+=================
```
ericholscher, line 6: I believe this is actually a page on reporting security issues, I think. So something like:

```suggestion
Security reports
================
```

I think the `Account security` stuff is probably better said in the `Security policy` doc, so we should probably just add a `seealso` there, and remove that section.

ericholscher (approved):
Simple enough, we need to make sure we have a redirect for the rename here though. Are we tracking redirects anywhere?

docs/user/reference/policies.rst
```diff
@@ -1,16 +1,20 @@
-Legal Documents and Policies
+============================
+Policies and legal documents
 ============================
 
 Here is some of the fine print used by |org_brand| and |com_brand|:
 
 .. toctree::
-    :glob:
-    :hidden:
```
ericholscher, line 8: Seems we're repeating the TOC again here by removing hidden? 

https://docs--9916.org.readthedocs.build/en/9916/reference/policies.html
Screenshot 2023-01-18 at 4 26 08 PM

docs/user/security.rst
benjaoming: Good spot! I'll move it! I think it sounds fine to have "Security reports" in this section still -- it's still a procedure implementing the Security Policy and I think 99% of people that want to report security issues will not look in the How-to section of a documentation but rather for a security policy.

benjaoming:
> Simple enough, we need to make sure we have a redirect for the rename here though. 

Oh yes, I was really hesitating to make this rename, but it just felt sooo messy if I put an index of a section somewhere else.

What if we do this:

1. Keep a standard placeholder at the location of the old page, with `orphan` + `this page is moved here: <path>`
2. When we want to launch and create redirects, we look for all the placeholders and eliminate them.

docs/user/reference/policies.rst
benjaoming: You're right, the old version looked sloppy and putting efforts into a proper little "definition list" makes the page much more pleasant to look at.

ericholscher:
> What if we do this:
> 
> 1. Keep a standard placeholder at the location of the old page, with `:orphan:` + `this page is moved here: <path>`
> 2. When we want to launch and create redirects, we look for all the placeholders and eliminate them.

I think we just create the redirects now? We could make them on the distaxis branch, then just migrate them to `latest` & `stable` after merge.

benjaoming:
> I think we just create the redirects now? We could make them on the distaxis branch

How does that work? Aren't user-defined redirects a global thing so we'd end up redirecting things that are currently live on `main`?

benjaoming:
Right `Exact redirects` FTW :100: 

```
Outcome: /en/diataxis-main/legal/index.html -> /en/diataxis-main/reference/policies.html
```

benjaoming:
Alright, should be good to go now :+1: 

Can test if https://docs.readthedocs.io/en/diataxis-main/legal/index.html will redirect once this has been merged and built.

ericholscher:
Looks like it redirected properly 👍

PR #9917, opened 2023-01-18
Allow searching & filtering VersionAutomationRuleAdmin

ericholscher:
Tried to find something for a user, and couldn't search!


stsewd (approved):
Looks like black needs to be run to make CI pass

Issue #9332, opened 2022-06-14
Refactor JSON fixtures to a scripted logic


benjaoming:
## Details

Currently, we maintain fixtures here: https://github.com/readthedocs/readthedocs.org/blob/main/readthedocs/projects/fixtures/test_data.json

Since it's not clear that the intention/coupling of these fixtures are for either

1) Test cases
2) Demoing / manual tests

...we should probably refactor this to a separate script or test fixture decorator:

Suggestion by @humitos for a quick solution:

https://github.com/readthedocs/readthedocs.org/pull/9319#discussion_r895568755

Note from @stsewd that current fixtures are included in tests:

https://github.com/readthedocs/readthedocs.org/issues/8914#issuecomment-1154039496

## Expected Result

Better developer experience maintaining fixtures, better or more fixtures, better test cases

## Actual Result

No bugs related to this

tanvimoharir:
Hi @benjaoming I would like to work on this. Continuing from #8914 . However I would need some guidance. 
Thanks!

adenisthebestaslan:
hi y'all! i know this is old but ive decided to work on it.


First thing i plan to do is find the markup file and edit it. 

i also plan to make a quick script to grab data from the api,(if there is one) and turn it into fixtures to be used.

adenisthebestaslan:
i might need a little assistance. Hoe should I go about compiling the json fizture

humitos:
It has been really hard to prioritize this work and I don't think it's really required -- we have been moving away from using these fixtures lately and we are creating only the required objects per tests.

PR #9926, opened 2023-01-23
Refactor projects fixtures

tanvimoharir:
Refers https://github.com/readthedocs/readthedocs.org/issues/9332
(previously #9411 )


benjaoming:
@tanvimoharir thanks for having patience and helping to throubleshoot a CI issue here :+1: 

Observing that "checks" now says :heavy_check_mark: ...

I'm quite sure the previous issue with "checks" failing at `pre-commit run --from-ref main --to-ref HEAD` was because the forked repo had an out-dated `main` branch. The comparison is intended to be run against the upstream `main` branch of `readthedocs/readthedocs.org` but that's probably not the remote that Circle CI's git checkout points to.

tanvimoharir:
> @tanvimoharir thanks for having patience and helping to throubleshoot a CI issue here 👍
> 
> Observing that "checks" now says ✔️ ...
> 
> I'm quite sure the previous issue with "checks" failing at `pre-commit run --from-ref main --to-ref HEAD` was because the forked repo had an out-dated `main` branch. The comparison is intended to be run against the upstream `main` branch of `readthedocs/readthedocs.org` but that's probably not the remote that Circle CI's git checkout points to.
> 
> My take: I don't think we can expect a forked repo to always maintain the latest `main` branch, we need the CI to not rely on that. There are probably a few ways to fix this.

thanks @benjaoming for helping me! appreciate it.

humitos:
Thanks for your time and work here! Unfortunately, we haven't had the time to prioritize this work and review it. I will close this PR for now, but we can come back in the future to recover this work.

PR #9928, opened 2023-01-24
Logging: log slugs when at least one of their builds was finished

humitos:
This is just to give us more context when debugging.

### DIFF
--- a/readthedocs/projects/tasks/utils.py
+++ b/readthedocs/projects/tasks/utils.py
@@ -105,6 +105,7 @@ def finish_inactive_builds():
         & Q(date__gt=timezone.now() - datetime.timedelta(days=1))
     )
 
+    projects_finished = set()
     builds_finished = 0
     builds = Build.objects.filter(query)[:50]
     for build in builds:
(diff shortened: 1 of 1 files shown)


readthedocs/projects/tasks/utils.py
```diff
@@ -127,10 +128,12 @@ def finish_inactive_builds():
         )
         build.save()
         builds_finished += 1
+        projects_finished.add(build.project.slug)
 
     log.info(
         'Builds marked as "Terminated due inactivity".',
         count=builds_finished,
+        project_slugs=projects_finished,
```
agjohnson, line 136: Might also be helpful to expose build_pk here too, if it isn't already. Usually in debugging, I'll look up the build_pk first to see the flow through our build process.

agjohnson (approved):

readthedocs/projects/tasks/utils.py
humitos: Yeah, I decided to not log them because I thought they could be _a lot_ of pks. However, I don't think it's the common case...

New Relic for the win! https://onenr.io/08woarZJWjx

In the last 7 days, we have 13 log lines with more than 10 builds terminated due activity. There is one log line with 24. Are we OK dumping all those build pks into a log line? It's not _that much_ as I thought.

agjohnson: Roger. We could split this up by project too, but I'm guessing it's fine if the build_pk list, project list, or both, are on the long side.

humitos:
I added the `build_pks=` attribute with all the `pk` in the same log line. I think that's fine for now. We can come back and improve it if we feel it's not enough.

Issue #9931, opened 2023-01-24
Build: output `_readthedocs/` directories feedback after deploy


humitos:
We were able build other formats (PDF, HTMLZip) by outputting the files into `_readthedocs/{pdf,htmlzip}` directories: https://test-builds.readthedocs.io/en/all-formats-build-jobs/

We found some limitations:

* the name of the PDF **must be** `_readthedocs/pdf/<project-slug>.pdf`
* the name of the HTMLZip **must be** `_readthedocs/htmlzip/project-slug>.zip`
* double-check if it happens the same for ePUB 😄 
* we are only checking for the existence of `_readthedocs/<format>` but **we are not checking if it's a directory** at https://github.com/readthedocs/readthedocs.org/blob/e1e7a186232c494161ce03d58f0ccb8dc5465c60/readthedocs/projects/tasks/builds.py#L539

Also, while debugging this we hit #9468

humitos:
I just realized that we had another issue: the build process raised an exception at `Uploading` state while uploading things to the storage backend and the built got stuck. This is because the upload step happens inside `on_success` instead of `execute` as the other things.

I've already moved this to the right place in https://github.com/readthedocs/readthedocs.org/pull/9800

humitos:
> * the name of the PDF **must be** `_readthedocs/pdf/<project-slug>.pdf`
> * the name of the HTMLZip **must be** `_readthedocs/htmlzip/project-slug>.zip`
> * double-check if it happens the same for ePUB smile

I'm not convinced about how to remove this restriction, yet. I don't want to _rename_ the file on behalf the user because that would be incompatible with ["Multiple pdf files for single project"](https://github.com/readthedocs/readthedocs.org/issues/2045) since we don't want to _rename all the user's PDF filenames_ (e.g. if they have `Tutorial-Basic.pdf` and `Tutorial-Advanced.pdf` we want to keep those names).

We may want to prefix/suffix the project/version slug, tho. In any case, **we will need to save the filename in our database** to know what's the PDF the user wants to download.

However, for now, we could do a `listdir` on S3 bucket and serve the first filename found --since it will always be one.

humitos:
I think my initial proposal to solve this problem would be to return a `.pdf` or a `.zip` file depending on the number of files in the output directory:

- one (`1`) file
  - rename the file at upload time (e.g. `Tutorial-Basic.pdf` -> `$READTHEDOCS_PROJECT_SLUG`) --_this is the new part_
[...]

PR #9939, opened 2023-01-25
Build: check if the output directory is a directory

humitos:
It checks for `_readthedocs/<format>` output directory to be a directory. If it's not a directory, it silently skip that format for now. This behavior will change once we move out `store_build_artifacts` from `on_success` and we can communicate errors easily to the user.

Related https://github.com/readthedocs/readthedocs.org/issues/9931

### DIFF
--- a/readthedocs/projects/tasks/builds.py
+++ b/readthedocs/projects/tasks/builds.py
@@ -796,12 +796,22 @@ class UpdateDocsTask(SyncRepositoryMixin, Task):
         types_to_delete = []
 
         for artifact_type in ARTIFACT_TYPES:
-            if os.path.exists(
-                self.data.project.artifact_path(
-                    version=self.data.version.slug,
-                    type_=artifact_type,
+            artifact_directory = self.data.project.artifact_path(
+                version=self.data.version.slug,
+                type_=artifact_type,
+            )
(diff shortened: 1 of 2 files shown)


readthedocs/projects/tasks/builds.py
```diff
@@ -796,12 +796,22 @@ def store_build_artifacts(self):
...
         for artifact_type in ARTIFACT_TYPES:
-            if os.path.exists(
-                self.data.project.artifact_path(
-                    version=self.data.version.slug,
-                    type_=artifact_type,
+            artifact_directory = self.data.project.artifact_path(
+                version=self.data.version.slug,
+                type_=artifact_type,
+            )
+            if os.path.isdir(artifact_directory):
```
ericholscher, line 803: Don't we want this to be a directory? Won't we basically be checking if `_readthedocs/pdf` is a directory, which we want? I think we want to check it's _not_ a directory, and if it has multiple files in it, if it is?

humitos: Yes. You already found that I've already solve this problem in the following PR.

humitos:
I closed this PR. We are continuing the conversation in https://github.com/readthedocs/readthedocs.org/pull/9941

PR #9940, opened 2023-01-25
Build: refactor logic to validate artifacts output paths

humitos:
Wrap all the logic into a class method so it's easier to follow and read.
It also remove some duplicated checks of `os.path.exists`,
making sure we are always using _validated_ paths already.

PR based in https://github.com/readthedocs/readthedocs.org/pull/9939 (we can probably close that one and only review this one)

Related https://github.com/readthedocs/readthedocs.org/issues/9931

### DIFF
--- a/readthedocs/projects/tasks/builds.py
+++ b/readthedocs/projects/tasks/builds.py
@@ -513,48 +513,88 @@ class UpdateDocsTask(SyncRepositoryMixin, Task):
         # Update build object
         self.data.build['success'] = False
 
+    def get_valid_artifact_types(self):
+        """
+        Return a list of all the valid artifact types for this build.
+
+        It performs the following checks on each output format type path:
+         - it exists
+         - it is a directory
+         - does not contains more than 1 files (only PDF, HTMLZip, ePUB)
+
+        TODO: remove the limitation of only 1 file.
+        Add support for multiple PDF files in the output directory and
+        grab them by using glob syntaxt between other files that could be garbage.
+        """
+        for artifact_type in ARTIFACT_TYPES:
+            artifact_directory = self.data.project.artifact_path(
+                version=self.data.version.slug,
+                type_=artifact_type,
+            )
+            if not os.path.exists(artifact_directory):
+                # There is no output directory.
+                # Skip this format.
+                continue
+
+            if not os.path.isdir(artifact_directory):
+                log.exception(
+                    "The output path is not a directory.",
+                    output_format=artifact_type,
+                )
+                # TODO: we should raise an exception here, fail the build,
+                # and communicate the error to the user.
+                # We are just skipping this output format for now.
+                #
+                # raise BuildUserError(BuildUserError.BUILD_OUTPUT_IS_NOT_A_DIRECTORY.format(artifact_type))
+                continue
+
+            # Check if there are multiple files on artifact directories.
+            # These output format does not support multiple files yet.
+            # In case multiple files are found, the upload for this format is not performed.
+            #
+            # TODO: we should fail the build for these cases and clearly communicate this.
+            # to the user. To do this, we need to call this method (``store_build_artifacts``)
+            # since the ``execute`` method.
+            # It will allow us to just `raise BuildUserError` and handle it at
+            # Celery `on_failure` handler.
+            if artifact_type in ("htmlzip", "epub", "pdf"):
+                if len(os.listdir(artifact_directory)) > 1:
+                    log.exception(
+                        "Multiple files are not supported for this format. "
+                        "Skipping this output format.",
+                        output_format=artifact_type,
+                    )
+                    continue
+
+            # If all the conditions were met, the artifact is valid
+            valid_artifacts.append(artifact_type)
+
+        return valid_artifacts
+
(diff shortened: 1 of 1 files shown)


readthedocs/projects/tasks/builds.py
```diff
@@ -513,48 +513,88 @@ def on_failure(self, exc, task_id, args, kwargs, einfo):
...
+            artifact_directory = self.data.project.artifact_path(
+                version=self.data.version.slug,
+                type_=artifact_type,
+            )
+            if not os.path.exists(artifact_directory):
+                # There is no output directory.
+                # Skip this format.
+                continue
+
+            if not os.path.isdir(artifact_directory):
```
ericholscher, line 539: This is the correct logic that I had issue with here: https://github.com/readthedocs/readthedocs.org/pull/9939/files#r1087009013

readthedocs/projects/tasks/builds.py
```diff
@@ -513,48 +513,88 @@ def on_failure(self, exc, task_id, args, kwargs, einfo):
...
+
+            # Check if there are multiple files on artifact directories.
+            # These output format does not support multiple files yet.
+            # In case multiple files are found, the upload for this format is not performed.
+            #
+            # TODO: we should fail the build for these cases and clearly communicate this.
+            # to the user. To do this, we need to call this method (``store_build_artifacts``)
+            # since the ``execute`` method.
+            # It will allow us to just `raise BuildUserError` and handle it at
+            # Celery `on_failure` handler.
```
ericholscher, line 559:
```suggestion
            # Check if there are multiple files in artifact directories.
            # These output formats don't support multiple files yet.
            # In case multiple files are found, the upload for this format is not performed.
            #
            # TODO: we should fail the build for these cases and clearly communicate this
            # to the user. To do this, we need to call this method (``store_build_artifacts``)
            # It will allow us to just `raise BuildUserError` and handle it at
            # in the `on_failure` handler.
```

Wasn't really sure what the second part here was saying with the `since the ``execute`` method.`?

readthedocs/projects/tasks/builds.py
```diff
@@ -513,48 +513,88 @@ def on_failure(self, exc, task_id, args, kwargs, einfo):
...
+            if artifact_type in ("htmlzip", "epub", "pdf"):
```
ericholscher, line 560: Seems this should be using a constant?

readthedocs/projects/tasks/builds.py
```diff
@@ -513,48 +513,88 @@ def on_failure(self, exc, task_id, args, kwargs, einfo):
...
+                if len(os.listdir(artifact_directory)) > 1:
+                    log.exception(
```
ericholscher, line 562: Is there an exception? I think we want error:

```suggestion
                    log.error(
```

readthedocs/projects/tasks/builds.py
```diff
@@ -777,7 +817,7 @@ def set_valid_clone(self):
         self.data.project.has_valid_clone = True
         self.data.version.project.has_valid_clone = True
 
-    def store_build_artifacts(self):
+    def store_build_artifacts(self, artifacts):
```
ericholscher, line 820: Need to add this to the docstring.

ericholscher (commented):
This looks like a cleaner change 👍

readthedocs/projects/tasks/builds.py
humitos: He, yeah, I made a mistake in the other PR and it seems I fixed it here 😅

readthedocs/projects/tasks/builds.py
humitos: Meh, `s/since/from inside/`. So, the correct sentence there should be:

> we should fail the build for these cases and clearly communicate this to the user. To do this, we need to call this method (``store_build_artifacts``) from inside the ``execute`` method.

readthedocs/projects/tasks/builds.py
humitos: In the next iteration of this code, I removed it again because it was not strictly required: https://github.com/readthedocs/readthedocs.org/pull/9941/files#diff-4eae6664ca2124d124dbfe04a8000483c7a23d426c3f390721d6e80fe5891064L820

readthedocs/projects/tasks/builds.py
humitos: I fixed this in the following PR already https://github.com/readthedocs/readthedocs.org/pull/9941/files#diff-4eae6664ca2124d124dbfe04a8000483c7a23d426c3f390721d6e80fe5891064L562

humitos:
I closed this PR. We are continuing the conversation in https://github.com/readthedocs/readthedocs.org/pull/9941. I'll address all the feedback you provided here in that PR. I'm sorry for all these confusions 🙏🏼

readthedocs/projects/tasks/builds.py
humitos: This comment was removed in the next iteration anyways, since we are calling `store_build_artifacts` from inside `execute` now and we can just raise exceptions there.

PR #9941, opened 2023-01-25
Build: communicate uploading errors to users

humitos:
Running `store_build_artifacts` from inside `.execute()` method
instead of from `.on_sucess()` allow us to raise any exception and handle it
properly.

We can raise a `BuildUserError` to clearly communicate the specific error to the
user, or any other exception to show the generic message: "Try again or contact
us".

Besides, this allows us to *fail the build* instead of continuing silently
without the user noticing this:

- there are more than 1 file of formats that don't support multiple files
- the output path is not a directory
- there were errors uploading the artifacts to the storage

PR based in https://github.com/readthedocs/readthedocs.org/pull/9940
Related to https://github.com/readthedocs/readthedocs.org/issues/9931

### DIFF
--- a/readthedocs/projects/tasks/builds.py
+++ b/readthedocs/projects/tasks/builds.py
@@ -494,7 +494,6 @@ class UpdateDocsTask(SyncRepositoryMixin, Task):
             if self.data.version:
                 version_type = self.data.version.type
 
-            # NOTE: autoflake gets confused here. We need the NOQA for now.
             status = BUILD_STATUS_FAILURE
             if isinstance(exc, BuildUserSkip):
                 # The build was skipped by returning the magic exit code,
@@ -526,6 +525,7 @@ class UpdateDocsTask(SyncRepositoryMixin, Task):
         Add support for multiple PDF files in the output directory and
         grab them by using glob syntaxt between other files that could be garbage.
         """
+        valid_artifacts = []
         for artifact_type in ARTIFACT_TYPES:
             artifact_directory = self.data.project.artifact_path(
                 version=self.data.version.slug,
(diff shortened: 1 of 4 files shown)


readthedocs/projects/tasks/builds.py
```diff
@@ -826,17 +823,16 @@ def store_build_artifacts(self, artifacts):
 
         Remove build artifacts of types not included in this build (PDF, ePub, zip only).
         """
+        time_before_store_build_artifacts = timezone.now()
```
ericholscher, line 826: Moving this to the `store_build_artifacts` function makes a lot of sense :)

readthedocs/projects/tasks/builds.py
```diff
@@ -857,14 +853,20 @@ def store_build_artifacts(self, artifacts):
...
-                    'Error copying to storage (not failing build)',
+                    "Error copying to storage",
                     media_type=media_type,
                     from_path=from_path,
                     to_path=to_path,
                 )
+                # Re-raise the exception to fail the build and handle it
+                # automatically at `on_failure`.
+                # It will clearly communicate the error to the user.
+                raise BuildAppError("Error uploading files to the storage.")
```
ericholscher, line 869: I feel like we need more information here. Won't part of the users files already be uploaded, but not all of them? We likely need to make this atomic somehow...

ericholscher (commented):
This looks like a nicer implementation, and I agree that uploading is probably part of the build process. I do think we need to think a bit more about atomic uploads here, and other possible outcomes of failing builds on upload error.

readthedocs/projects/tasks/builds.py
humitos: Yes. I think atomic uploads is _the correct way_. However, I'm not sure how to do that in a simple way. I suppose we could upload the content to one folder and them move it to the final one, in case the "move to the final one" action is atomic. Otherwise, we are in the same situation.

humitos:
I like the idea of atomic uploads and I think we should go in that direction. I can create an issue to discuss it a little more in deep and decide what path to follow.

I'd like to clarify if you are OK merging this PR as-is, meaning that "if there is an error in the upload step, we fail the build and communicate this to the user --leaving _half of the files_ updated in production" or, you prefer to come back to the previous behavior that was "if there is an error, continue the build, don't communicate to the user and leave _half of the files_ updated in production". Basically, it's the same, but in the new one we inform the user about this.

readthedocs/projects/tasks/builds.py
```diff
@@ -513,55 +512,84 @@ def on_failure(self, exc, task_id, args, kwargs, einfo):
...
+                raise BuildUserError(
+                    BuildUserError.BUILD_OUTPUT_IS_NOT_A_DIRECTORY.format(
+                        artifact_type=artifact_type
+                    )
+                )
+
+            # Check if there are multiple files on artifact directories.
+            # These output format does not support multiple files yet.
+            # In case multiple files are found, the upload for this format is not performed.
+            if artifact_type in ("htmlzip", "epub", "pdf"):
```
humitos, line 553: Address Eric's feedback for this https://github.com/readthedocs/readthedocs.org/pull/9940#discussion_r1087017048

humitos:
> I like the idea of atomic uploads and I think we should go in that direction. I can create an issue to discuss it a little more in deep and decide what path to follow.

I opened https://github.com/readthedocs/readthedocs.org/issues/9947 to track this and discuss there. I'm pinging @stsewd since he has been working with `rclone` and maybe knows if this is supported. I'm sure he will have some ideas there 💯

humitos:
We are not at "atomic uploads" yet, but we will be retrying at least several times before failing the build with `rclone`. So, we are reducing the probabilities of failure during the upload step --and, in case of failure, we will be communicating this to the user. This implementation is _better of what we currently have_, so I'm happy to merge it and continue improving it in future PRs.

ericholscher (approved):
This looks good once tests pass. I don't think we need to block on atomic uploads, and incremental progress is all we can hope for :)

PR #9949, opened 2023-01-26
Docs: Make the GSOC page orphaned (Diátaxis)

benjaoming:
We can also remove it completely, but this also works.

Refs: #9746 

---

- User's documentation (`docs`): https://docs--9949.org.readthedocs.build/en/9949/

- Developer's documentation (`dev`): https://dev--9949.org.readthedocs.build/en/9949/

### DIFF
--- a/docs/user/gsoc.rst
+++ b/docs/user/gsoc.rst
@@ -1,3 +1,5 @@
+:orphan:
+
 Google Summer of Code
 =====================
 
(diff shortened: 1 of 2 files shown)


ericholscher (approved):
Good enough for now. It's pretty outdated, so probably worth removing at some point, but not a huge issue.

PR #9953, opened 2023-01-26
Docs: Split "Automation rules" into reference and how-to (Diátaxis)

benjaoming:
Scope of work:

* Reference: "Automation rules" need to have a clear and simple feature reference.
* Explanation: New explanation contents can be avoided, since there is already an explanation article around versioning.
* How-to: It would work well to have a How-to guide that can be referred to, similar to Redirects how-to, which lists out the concepts with hoverxref'ed links to their reference.
* Tutorial: nay

* New how-to: "How to create new versions automatically"

Refs: #9746 

---

- User's documentation (`docs`): https://docs--9953.org.readthedocs.build/en/9953/

- Developer's documentation (`dev`): https://dev--9953.org.readthedocs.build/en/9953/


docs/user/automation-rules.rst
```diff
@@ -82,11 +89,8 @@ Currently, the following actions are available:
 Order
 -----
 
-The order your rules are listed in  :guilabel:`Admin` > :guilabel:`Automation Rules` matters.
-Each action will be performed in that order,
-so first rules have a higher priority.
-
-You can change the order using the up and down arrow buttons.
+When a new webhook event is received from your Git provider,
+All rules with a successful match will have their action triggered in the order they appear on the :guilabel:`Automation Rules` page.
```
benjaoming, line 93: I need some feedback here. In a former section, we state:

> When a rule matches a new version, (...)

We should be more specific about the trigger. What happens when a new rule is added for a tag or branch and then _something new_ shows up? What happens when a Git tag is deleted and re-added?

docs/user/guides/automation-rules.rst
```diff
@@ -0,0 +1,84 @@
...
+
+   Examples of common usage
+     See :ref:`the list of examples <automation-rules:Examples>` for rules that are commonly used.
+
+   Want to test if your rule works?
+     You can test if your rule works in a few minutes without disturbing your readers:
+     If you are using Git tags in order to create new versions,
+     create a Git tag that matches the rule and check if your automation action is triggered.
+     After the experiment,
+     you can delete the Read the Docs version and the Git tag.
```
benjaoming, line 67: I find this tip to be useful, however I don't know if the process is the best way of testing a rule. Are there other ways of checking if a rule triggers?

benjaoming:
Some attention to labels and help texts in the "Add rule" form could be nice, but I think we can skip the scope creep and consider this change separate from further improvements on the Dashboard.

@ericholscher would you like to try to take over this PR instead of reviewing? Just add changes directly and LMK when it looks good to you?

benjaoming:
Also noteworthy: I left a slightly variating repetitions of content between how-to and reference that I thought was simply more useful for the reader than a cross-reference.

docs/user/guides/automation-rules.rst
```diff
@@ -0,0 +1,84 @@
+How to manage versions automatically
+====================================
```
benjaoming, line 2: This title is great to improve with the user's problem in mind.

ericholscher:
Ran out of time again today, keeping this for tomorrow.

docs/user/guides/automation-rules.rst
```diff
@@ -0,0 +1,84 @@
...
+The rule takes effect immediately when a new version is discovered.
+
+.. tip::
+
+   Examples of common usage
+     See :ref:`the list of examples <automation-rules:Examples>` for rules that are commonly used.
+
+   Want to test if your rule works?
+     You can test if your rule works in a few minutes without disturbing your readers:
+     If you are using Git tags in order to create new versions,
```
ericholscher, line 64: I don't love this, but seems useful.

docs/user/automation-rules.rst
```diff
@@ -1,67 +1,87 @@
...
+Read the Docs receives a callback to its Webhook.
+This triggers the creation of a new version.
 
-All rules are evaluated for this version, in the order they are listed.
-If the version matches the version type and the pattern in the rule,
-the specified action is performed on that version.
+All rules are evaluated for this version,
+in the order they are listed.
+If the version **matches** the version type and the pattern in the rule,
+the specified **action** is performed on that version.
```
ericholscher, line 28: A diagram would be great here in the future, but probably too much for this refactor.

ericholscher (commented):
Did a bunch of small edits here, it's ready for another review :+1:

docs/user/automation-rules.rst
```diff
@@ -64,50 +69,42 @@ Activate version
...
 Make version private
   Sets the version's privacy level to private.
-  See :ref:`versions:privacy levels`.
+  See :ref:`versions:Privacy levels`.
 
 Set version as default
-  Sets the version as default,
-  i.e. the version of your project that `/` redirects to.
-  See more in :ref:`automatic-redirects:Root URL`.
+  Sets the version as the :term:`default version`.
```
ericholscher, line 83: I think we define this in another PR, but seemed useful :)

docs/user/guides/automation-rules.rst
```diff
@@ -14,10 +14,12 @@ Automating your versioning on Read the Docs means you only have to handle your v
    :doc:`/automation-rules`
      Reference for all different rules and actions possible with automation.
 
-Adding a new rule
------------------
+Adding a new Automation Rule
```
ericholscher, line 17: We should agree on capitalization of this term, since we use it both ways.

benjaoming: Yes, I find that the prevailing "case" is lower case. I sometimes upper case per reflex because the UI in the Dashboard does that. But it's unintended.

benjaoming: Confirmed that a lot of other places are lowercase and made this lowercase, too.

docs/user/automation-rules.rst
benjaoming: If we can establish a quick way to use Mermaid diagrams, this would definitely be a valid option. Last time, I hit some issues that would need to be solved, so ended up generating Mermaid.js and saving as PNG "by hand" which was a bit too cumbersome for a "quick diagram".

docs/user/automation-rules.rst
benjaoming: YES! And PR has been merged :)

ericholscher (approved):
I think this is close/ready? Merge if you agree.

docs/user/automation-rules.rst
```diff
@@ -1,96 +1,110 @@
-Automation Rules
+Automation rules
```
benjaoming, line 1: haha just read this again and realized that it sounds a lot like "automation rules [verb]"

Maybe that'll make people click it, anyways :)

benjaoming:
Yes, I didn't find anything, but I've also looked at it too many times :upside_down_face:

PR #9956, opened 2023-01-26
Requirements: reduce complexity

humitos:
Remove `debug.in` to reduce the complexity.
Install those 3 packages directly into the `docker.in` file since it's where they are used.


stsewd (commented):
Personally, I don't use those really, but I remember they were meant to be used when running tests, that's why there is the in the toxt.ini file.

humitos:
Yeah. I think I added them there.

I'm not using them either and installing the extra packages inside tox when I need them ☺️

humitos:
@stsewd are you OK merging this PR then?

stsewd (approved):

PR #9967, opened 2023-01-31
Build: remove `pdflatex` support

### LINKED ISSUE #9886: Builder: remove PDF support for `Sphinx<=1.6.1`
We are running a command on each build to detect if `sphinx>=1.6.1` and decide whether or not use `latexmk`.

https://github.com/readthedocs/readthedocs.org/blob/84a149a57ccd9c1b5cd47bb285263a273124a199/readthedocs/doc_builder/backends/sphinx.py#L310-L311

I quickly check our Metabase query and I found that we have 0 projects using such an older versions, https://ethicalads.metabaseapp.com/question/250-projects-using-sphinx-timeserie, so I'm happy to remove this code and always use `latexmk`.

This will simplify the code and remove the execution of one extra command on _all the builds_.

humitos:
`pdflatex` is only called for project using `Sphinx<=1.6.1` --which we haven't had anyone in the last 6 months.

Closes #9886


docs/user/guides/pdf-non-ascii-languages.rst
```diff
@@ -3,8 +3,6 @@ Sphinx PDFs with Unicode
 
 Sphinx offers different `LaTeX engines`_ that have better support for Unicode characters
 and non-European languages like Japanese or Chinese.
-By default Sphinx uses ``pdflatex``,
-which does not have good support for Unicode characters and may make the PDF builder fail.
```
benjaoming, line 7: Can we keep "By default, Sphinx uses `latexmk <https://mg.readthedocs.io/latexmk.html>`_" or something similar?

I'm not sure how it affects Unicode characters, so seems good to delete.

benjaoming, line 7: Can we keep "By default, Sphinx uses `latexmk <https://mg.readthedocs.io/latexmk.html>`_" or something similar?

I'm not sure how it affects Unicode characters, so seems good to delete.

readthedocs/doc_builder/backends/sphinx.py
```diff
@@ -331,38 +331,6 @@ def sphinx_parallel_arg(self):
             return ['-j', 'auto']
         return []
 
-    def venv_sphinx_supports_latexmk(self):
```
benjaoming, line 334: I know that we don't have any recent builds with Sphinx 1.6, but have we also deprecated it?

If we no longer check if `latexmk` is supported, I suppose it will fail on Sphinx<1.6 since it's not supported?

benjaoming (approved):
LGTM :+1: 

Just want your opinion about whether this is definitely breaking super-old environments with Sphinx<1.6 and if that can be avoided by simply not calling latexmk in those environments :arrow_down:

docs/user/guides/pdf-non-ascii-languages.rst
humitos: Yeah, `latexmk` is the good one to deal with Unicode characters. That's why I deleted the sentence completely.

readthedocs/doc_builder/backends/sphinx.py
humitos: In case there is someone building with Sphinx <1.6 (which I doubt it) it would only fail if they have the PDF enabled. Also, I'd assume that it will probably fail first with other different hundred reasons first 😄

benjaoming: That seems like a fair assessment :)

docs/user/guides/pdf-non-ascii-languages.rst
benjaoming: Yeah, I was thinking that it's nice to inform people about the default builder in case they need to troubleshoot something.

humitos:
@agjohnson @ericholscher I'd like your opinions on this because I may not be considering something that you will 😄

ericholscher (approved):
This seems like a positive change and simplification. I don't have a full memory of why this was added, but I think it should be safe to delete at this point.

We should likely have somewhere in our docs that we mention supported versions of Sphinx, and begrudgingly list 1.8.7 or whatever is the one we install still by default on old projects, until we can migrate them. 🙃

PR #9971, opened 2023-02-01
Build: expose `READTHEDOCS_VIRTUALENV_PATH` variable

### LINKED ISSUE #9629: Build: expose location of virtualenv
I'm using the [PDM](https://pdm.fming.dev/latest/) dependency management tool, moving away from `requirements.txt` files. If PDM does not detect an active virtualenv, or is not told about the location of a virtualenv, it will create one and install packages there.

Read the Docs does not activate or expose the path to the virtualenv it creates. It seems that `which python` does report the virtualenv python despite that, so a hack is used to get the path and tell PDM about it. https://github.com/pdm-project/pdm/discussions/1365#discussioncomment-3581356

```
VIRTUAL_ENV=$(dirname $(dirname $(which python))) ~/.local/bin/pdm sync -dG docs
```

It would be helpful if Read the Docs made the virtualenv easier to find. You could set `VIRTUAL_ENV`, or maybe `RTD_VIRTUAL_ENV` if you don't want to affect other tools by default. Alternatively, activate it.

* Read the Docs project URL: https://readthedocs.org/projects/flask-sqlalchemy/
* Build URL (if applicable): https://readthedocs.org/projects/flask-sqlalchemy/builds/18192947/
* Read the Docs username (if applicable): https://readthedocs.org/profiles/davidism/

humitos:
The variable `$READTHEDOCS_VIRTUALENV_PATH` points to where Read the Docs created the virtualenv.

While working on this I found that we can use this variable to run our own commands. This way, we will be avoiding mistakes and potentially having differences path between "the real path" and the content of this variable. I changed our code to use this variable when running `/bin/python` and others.

Besides, I did the same for Conda environments and the variables `CONDA_ENV_PATHS` and `CONDA_DEFAULT_ENV` that were already defined.

This makes the code cleaner and simpler, while more standardized as well. Users can use these variables in the same way Read the Docs does and be sure the content they are expecting is going to be there.

Closes #9629

---

- User's documentation (`docs`): https://docs--9971.org.readthedocs.build/en/9971/

- Developer's documentation (`dev`): https://dev--9971.org.readthedocs.build/en/9971/

### DIFF
--- a/docs/user/environment-variables.rst
+++ b/docs/user/environment-variables.rst
@@ -50,6 +50,12 @@ Read the Docs builders set the following environment variables automatically for
 
     :Examples: ``en``, ``it``, ``de_AT``, ``es``, ``pt_BR``
 
+.. envvar:: READTHEDOCS_VIRTUALENV_PATH
+
+    Path where Read the Docs created the virtualenv for this build.
+
+    :Example: ``/home/docs/checkouts/readthedocs.org/user_builds/project/envs/version``
+
 User-defined environment variables
 ----------------------------------
 
--- a/docs/user/environment-variables.rst
+++ b/docs/user/environment-variables.rst
@@ -52,7 +52,8 @@ Read the Docs builders set the following environment variables automatically for
 
 .. envvar:: READTHEDOCS_VIRTUALENV_PATH
 
-    Path where Read the Docs created the virtualenv for this build.
+    Path for the :ref:`virtualenv that was created for this build <builds:Understanding what's going on>`.
+    Only exists for builds using Virtualenv and not Conda.
 
     :Example: ``/home/docs/checkouts/readthedocs.org/user_builds/project/envs/version``
 
--- a/readthedocs/api/v2/serializers.py
+++ b/readthedocs/api/v2/serializers.py
@@ -172,6 +172,9 @@ class BuildCommandReadOnlySerializer(BuildCommandSerializer):
 
         regex = f"{docroot}{container_hash}{project_slug}/envs/{version_slug}(/bin/)?"
         command = re.sub(regex, "", obj.command, count=1)
+
+        regex = f"^\\$READTHEDOCS_VIRTUALENV_PATH/bin/"
+        command = re.sub(regex, "", obj.command, count=1)
         return command
 
 
--- a/readthedocs/api/v2/serializers.py
+++ b/readthedocs/api/v2/serializers.py
@@ -173,7 +173,10 @@ class BuildCommandReadOnlySerializer(BuildCommandSerializer):
         regex = f"{docroot}{container_hash}{project_slug}/envs/{version_slug}(/bin/)?"
         command = re.sub(regex, "", obj.command, count=1)
 
-        regex = f"^\\$READTHEDOCS_VIRTUALENV_PATH/bin/"
+        regex = "^\\$READTHEDOCS_VIRTUALENV_PATH/bin/"
+        command = re.sub(regex, "", obj.command, count=1)
+
+        regex = "^\\$CONDA_ENVS_PATH/\\$CONDA_DEFAULT_ENV/bin/"
         command = re.sub(regex, "", obj.command, count=1)
         return command
 
--- a/readthedocs/api/v2/serializers.py
+++ b/readthedocs/api/v2/serializers.py
@@ -170,14 +170,15 @@ class BuildCommandReadOnlySerializer(BuildCommandSerializer):
             docroot = re.sub("/[0-9a-z]+/?$", "", settings.DOCROOT, count=1)
             container_hash = "/([0-9a-z]+/)?"
 
+        command = obj.command
         regex = f"{docroot}{container_hash}{project_slug}/envs/{version_slug}(/bin/)?"
-        command = re.sub(regex, "", obj.command, count=1)
+        command = re.sub(regex, "", command, count=1)
 
-        regex = "^\\$READTHEDOCS_VIRTUALENV_PATH/bin/"
-        command = re.sub(regex, "", obj.command, count=1)
+        regex = r"^\$READTHEDOCS_VIRTUALENV_PATH/bin/"
+        command = re.sub(regex, "", command, count=1)
 
-        regex = "^\\$CONDA_ENVS_PATH/\\$CONDA_DEFAULT_ENV/bin/"
-        command = re.sub(regex, "", obj.command, count=1)
+        regex = r"^\$CONDA_ENVS_PATH/\\$CONDA_DEFAULT_ENV/bin/"
+        command = re.sub(regex, "", command, count=1)
         return command
 
 
--- a/readthedocs/api/v2/serializers.py
+++ b/readthedocs/api/v2/serializers.py
@@ -175,7 +175,7 @@ class BuildCommandReadOnlySerializer(BuildCommandSerializer):
         command = re.sub(regex, "", command, count=1)
 
         # Remove explicit variable names we use to run commands,
-        # since users don't care about these. 
+        # since users don't care about these.
         regex = r"^\$READTHEDOCS_VIRTUALENV_PATH/bin/"
         command = re.sub(regex, "", command, count=1)
 
--- a/readthedocs/api/v2/serializers.py
+++ b/readthedocs/api/v2/serializers.py
@@ -174,6 +174,8 @@ class BuildCommandReadOnlySerializer(BuildCommandSerializer):
         regex = f"{docroot}{container_hash}{project_slug}/envs/{version_slug}(/bin/)?"
         command = re.sub(regex, "", command, count=1)
 
+        # Remove explicit variable names we use to run commands,
+        # since users don't care about these. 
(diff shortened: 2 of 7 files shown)


docs/user/environment-variables.rst
```diff
@@ -50,6 +50,12 @@ Read the Docs builders set the following environment variables automatically for
 
     :Examples: ``en``, ``it``, ``de_AT``, ``es``, ``pt_BR``
 
+.. envvar:: READTHEDOCS_VIRTUALENV_PATH
+
+    Path where Read the Docs created the virtualenv for this build.
```
ericholscher, line 55:
```suggestion
    Path for the virtualenv that was created for this build.
    Only exists for builds using Virtualenv and not Conda. 
```

Can we link the virtualenv link somewhere useful? Perhaps https://docs.readthedocs.io/en/stable/builds.html#understanding-what-s-going-on ?

readthedocs/doc_builder/python_environments.py
```diff
@@ -89,14 +89,17 @@ def install_package(self, install):
                 bin_path=self.venv_bin(),
             )
 
-    def venv_bin(self, filename=None):
+    def venv_bin(self, filename=None, conda=False):
```
ericholscher, line 92: Need to add this to the params

readthedocs/doc_builder/python_environments.py
```diff
@@ -89,14 +89,17 @@ def install_package(self, install):
...
         """
         Return path to the virtualenv bin path, or a specific binary.
 
         :param filename: If specified, add this filename to the path return
         :returns: Path to virtualenv bin or filename in virtualenv bin
         """
-        parts = [self.venv_path(), 'bin']
+        parts = ["$READTHEDOCS_VIRTUALENV_PATH", "bin"]
+        if conda:
```
ericholscher, line 100: This seems like an awkward place to introduce conda. Previously this function didn't know about conda at all, and I'm wondering if it's the right place for this logic? Or perhaps we need to rename this function?

readthedocs/doc_builder/python_environments.py
```diff
@@ -318,8 +318,9 @@ class Conda(PythonEnvironment):
     .. _Conda: https://conda.io/docs/
     """
 
-    def venv_path(self):
-        return os.path.join(self.project.doc_path, 'conda', self.version.slug)
+    # pylint: disable=arguments-differ
```
ericholscher, line 321: This also makes me worry this is not the right approach. It seems like having each subclass define this cleanly is a nicer?

readthedocs/doc_builder/environments.py
```diff
@@ -380,7 +380,18 @@ def get_wrapped_command(self):
 
     def _escape_command(self, cmd):
         r"""Escape the command by prefixing suspicious chars with `\`."""
-        return self.bash_escape_re.sub(r'\\\1', cmd)
+        command = self.bash_escape_re.sub(r"\\\1", cmd)
+
+        # HACK: avoid escaping variables that we need to use in the commands
+        not_escape_variables = (
```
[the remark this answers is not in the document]
ericholscher: I just read your comment in https://github.com/readthedocs/readthedocs.org/pull/9913 -- and if we're already not doing this for build.jobs and build.commands, I'm fine with this hack for now 👍

readthedocs/doc_builder/python_environments.py
humitos: I don't have a strong preference here. It's the same code -- The only difference is the prefix (virtualenv or condaenv variable)

I put it in the super-class only to avoid repetition of the same logic.

readthedocs/doc_builder/python_environments.py
humitos: I updated this to make the code more explicit: 66fe0a7d9. Let me know if that looks better 👍🏼

readthedocs/doc_builder/environments.py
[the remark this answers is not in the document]
humitos: Cool! I understand we are OK to move forward with this hack for now, then. Right?

I just saw that we have this documented already in the docstring: https://github.com/readthedocs/readthedocs.org/blob/cc39779565d779782b8cfa891a0df8700e35ec23/readthedocs/doc_builder/director.py#L301 💯

readthedocs/doc_builder/director.py
```diff
@@ -556,6 +556,7 @@ def get_build_env_vars(self):
         if self.data.config.conda is not None:
             env.update(
                 {
+                    # NOTE: should these be prefixed with "READTHEDOCS_"?
```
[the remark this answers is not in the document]
humitos: Yeah, I'm not sure why we defined this in the first place, that's why I wrote this comment here. I'm not sure to understand what these variables are for, but I don't think we need to research or change anything about it in this PR 😄 

I can remove the "Note" comment if it makes it more confusion, tho.

humitos:
This is ready for another round of review. I think I addressed all the feedback provided.

readthedocs/doc_builder/director.py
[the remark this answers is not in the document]
ericholscher: I'm fine leaving it, but it's worth noting that it's also used by Conda, so probably something we added in our initial implementation that something random is relying on 😆

readthedocs/doc_builder/python_environments.py
ericholscher: That is definitely nicer. Though I'd argue the `venv` in the name is misleading for conda envs :) Perhaps it should just be called `env_bin` or similar? Not a huge deal tho.

readthedocs/api/v2/serializers.py
```diff
@@ -170,14 +170,15 @@ def get_command(self, obj):
...
             container_hash = "/([0-9a-z]+/)?"
 
+        command = obj.command
         regex = f"{docroot}{container_hash}{project_slug}/envs/{version_slug}(/bin/)?"
-        command = re.sub(regex, "", obj.command, count=1)
+        command = re.sub(regex, "", command, count=1)
 
-        regex = "^\\$READTHEDOCS_VIRTUALENV_PATH/bin/"
-        command = re.sub(regex, "", obj.command, count=1)
+        regex = r"^\$READTHEDOCS_VIRTUALENV_PATH/bin/"
```
ericholscher, line 177:
```suggestion
        # Remove explicit variable names we use to run commands.
        regex = r"^\$READTHEDOCS_VIRTUALENV_PATH/bin/"
```

readthedocs/doc_builder/python_environments.py
```diff
@@ -89,20 +89,17 @@ def install_package(self, install):
                 bin_path=self.venv_bin(),
             )
 
-    def venv_bin(self, filename=None, conda=False):
+    def venv_bin(self, prefixes, filename=None):
```
ericholscher, line 92: Do we even need this function? It seems like it's just a wrapper around `os.path.join`?

ericholscher (approved):
I'm on board with this as is.

A small bit is the `venv_bin` implementation is still a bit hacky how it different in signature. Now that we've removed some so much logic from it, I wonder if it's even still useful?

readthedocs/doc_builder/python_environments.py
humitos: We don't _need it_, but it's used a lot of times in this file. I guess it was created to avoid repeating these lines that many times only since this function it's not super smart 😄 .

I counted them. It's used 19 times.

humitos:
I added a note in `PythonEnvironment.install_package` about the `venv_bin` and how we are managing the `prefixes=` attribute since I'm skipping that linter check for now. I'm not sure how to solve it without re-structuring the code or duplicating the logic from `venv_bin` (maybe that's the simplest thing to do, tho).

We can come back to this later if we want to. For now, I want to deploy the feature and unblock this to move forward and deploy these changes.

Bilokin:
Hello everyone,

I have a build failure with an error for a project which uses conda:
```
python -mvirtualenv $READTHEDOCS_VIRTUALENV_PATH
SystemExit: 2
usage: virtualenv [--version] [--with-traceback] [-v | -q] [--read-only-app-data] [--app-data APP_DATA] [--reset-app-data] [--upgrade-embed-wheels] [--discovery {builtin}] [-p py] [--try-first-with py_exe]
                  [--creator {builtin,cpython3-posix,venv}] [--seeder {app-data,pip}] [--no-seed] [--activators comma_sep_list] [--clear] [--no-vcs-ignore] [--system-site-packages] [--symlinks | --copies] [--no-download | --download]
                  [--extra-search-dir d [d ...]] [--pip version] [--setuptools version] [--wheel version] [--no-pip] [--no-setuptools] [--no-wheel] [--no-periodic-update] [--symlink-app-data] [--prompt prompt] [-h]
                  dest
virtualenv: error: the following arguments are required: dest
```
I think this error is related to this PR.
Is there workaround for this issue? 
Or do I need to set `READTHEDOCS_VIRTUALENV_PATH` environment variable value from now on?

humitos:
@Bilokin please open a new issue filling the template and linking back to this issue if you think it's related. Thanks 👍

PR #9972, opened 2023-02-01
Management commands: remove unused ones

### LINKED ISSUE #1526: Audit management commands
Refs #1524

We have a bunch of random management commands that have been written over time. I don't think any of them are tested, and most are just wrapped around util functions, or old dead code. We should think through what we actually need to ship with, make them work, document them, and kill the rest.

humitos:
I'm deleting these management commands because we are not using them and probably some of them not even work properly.

To avoid confusions, and also to remove the maintainance of them, I'm deleting them for now. We can always find them in the history if we want to.

Closes #1526


humitos:
Yeah, I know some of these were used when all the Ads logic was integrated into Read the Docs. However, I understand that all of this was moved to EthicalAds and it's not managed by the server itself. cc @ericholscher

davidfischer:
> We do still use the RTD-provided tags, but I don't think we've synced them from GH, and with our ML-based approach it's a lot less important than it used to be. I'm fine deleting this, since we can always find it in the history if we need it. I'll also ping @davidfischer here, but I think we aren't likely to run these again in the future.

I'm fine removing it.

humitos:
Cool! I enabled auto merge for when it gets approved 👍

ericholscher (approved):

PR #9978, opened 2023-02-02
Settings: simplify all the settings removing a whole old layer (`dev`)

humitos:
We are not using `settings/dev.py` since we started using Docker (`settings/docker_compose.py`).
However, we never updated this code.

This commit removes the extra `Dev` class layer by moving the required (and still used settings) into `settings/test.py` and `settings/docker_compose.py` since they were inheriting from `settings/dev.py`.

Related: https://github.com/readthedocs/readthedocs-ops/issues/937


humitos:
I think I missed one setting related because there are 2 test cases failing. I'll check it and push a fix.

humitos:
Documentation is failing because we have a super complex setup 🤯 . I simplified it and removed a lot of unnecessary things in https://github.com/readthedocs/readthedocs.org/pull/9981. I'd like to move forward with this because it simplifies the build _a lot_.

PR #9983, opened 2023-02-03
Proxito: always check `404/index.hmtml`

humitos:
With the introduction of `build.commands` we cannot use `version.documentation_type` anymore since those versions will be `generic` and we can't skip checking for this file location.

Note this commit may add and extra call to S3 API for all the 404 pages where our regular Maze will be shown. However, it removes 2 databsae calls from all the 404 requests.

We could only add this extra check on S3 for
`version.documentation_type='generic'`, but that would make the code a little more complex and we won't be removing these 2 db queries.

Reference: https://github.com/readthedocs/sphinx-notfound-page/issues/215#issuecomment-1415771549

### DIFF
--- a/docs/user/hosting.rst
+++ b/docs/user/hosting.rst
@@ -74,18 +74,18 @@ Custom Not Found (404) Pages
 ----------------------------
 
 If you want your project to use a custom page for not found pages instead of the "Maze Found" default,
-you can put a ``404.html`` at the top level of your project's HTML output.
+you can put a ``404.html`` or ``404/index.html`` in your project's HTML output.
 
 When a 404 is returned,
-Read the Docs checks if there is a ``404.html`` in the root of your project's output
+Read the Docs checks if there is a ``404.html`` or ``404/index.html`` in your project's output
 corresponding to the *current* version
 and uses it if it exists.
-Otherwise, it tries to fall back to the ``404.html`` page
+Otherwise, it tries to fall back to the ``404.html`` or ``404/index.html`` page
 corresponding to the *default* version of the project.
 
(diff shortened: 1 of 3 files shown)


ericholscher (approved):
> Note this commit may add and extra call to S3 API for all the 404 pages where our regular Maze will be shown. However, it removes 2 databsae calls from all the 404 requests.

Can you explain this a bit more? I don't see the 2 removals.

Overall this is the right approach. Let's just serve what is on the filesystem at either reasonable path, and not try to be magical here.

humitos:
> Can you explain this a bit more? I don't see the 2 removals.

We are removing the db queries to know the `documentation_type` of the 2 versions:
* default version: https://github.com/readthedocs/readthedocs.org/pull/9983/files#diff-8a36e82f90e1e725e057497e285a7fb0afe6ad99deef3568db50e1c9b895a994L361

* requested version: https://github.com/readthedocs/readthedocs.org/pull/9983/files#diff-8a36e82f90e1e725e057497e285a7fb0afe6ad99deef3568db50e1c9b895a994L350


> Do we need to update the docs to note that secondary location?

Yes. Good point.

ericholscher:
Also @benjaoming, note the docs update here. Another point for getting our branch merged into main so we can stop diverging.

benjaoming:
oh there's been quite a lot of those already, I merge `main` into `diataxis/main` frequently. Will do that now :+1:

ericholscher:
Two steps ahead :)

benjaoming:
Wow, no conflicts. Very elegant @humitos :heart_eyes:

PR #9989, opened 2023-02-06
Docs: Bump sphinx-rtd-theme 1.2.0rc4

benjaoming:
Checking out https://github.com/readthedocs/sphinx_rtd_theme/pull/1430

Noticing that sphinxcontrib-jquery 2.0.0 loads jQuery from Google CDN... didn't recall it'd do that. 3.0.0 has it bundled in but with a bug, and I've asked for a new release.

@agjohnson we currently run Sphinx 4.5.0 with this comment in `pip.in`.. it'd be interesting to know if we can bump this to some 6.x version to track things a bit closer through our own documentation work.

```
# For intersphinx during builds
# We need these pinned to build the docs properly
Sphinx==4.5.0
docutils==0.17.1
```


humitos:
We removed a lot of requirements from docs. You should update with main and re-run the pip-tools commands.

benjaoming:
Yes, [3 hours ago](https://github.com/readthedocs/readthedocs.org/pull/9978#event-8456098733) :)

humitos (approved):

benjaoming:
Well, we got it tested for a PR build, but no reason to merge, can use 1.2.0 instead :)

PR #10002, opened 2023-02-08
Celery: cleanup `pidbox` keys

humitos:
Simple task to remove `pidbox` keys older than 15 minutes. This is a workaround to avoid Redis OOM for now.

We will need to find out a better solution here. There is an upstream issue opened that we should check in the near future and probably remove this
workaround: https://github.com/celery/celery/issues/6089

Related: https://github.com/readthedocs/readthedocs-ops/issues/1260

### DIFF
--- a/readthedocs/core/tasks.py
+++ b/readthedocs/core/tasks.py
@@ -2,6 +2,7 @@
 
 import math
 
+import redis
 import structlog
 from django.conf import settings
 from django.core.mail import EmailMultiAlternatives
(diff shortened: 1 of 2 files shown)


readthedocs/core/tasks.py
```diff
@@ -56,3 +58,31 @@ def clear_persistent_messages():
...
+
+    This task is executed periodically to remove "pdibox" objects with an
+    idletime bigger than 5 minutes from Redis and free some RAM.
+
+    https://github.com/celery/celery/issues/6089
+    https://github.com/readthedocs/readthedocs-ops/issues/1260
+
+    """
+    total_memory = 0
+    keys = app.backend.client.keys("*reply.celery.pidbox*")
```
humitos, line 79: `app.backend.client` is not working anymore since we are ignoring results and backed is instantiated as `DisabledBackend` now. We have to use `from django.core.cache import cache`, I think.

ericholscher: Yea, just need to make sure we're connecting to the right redis database.

ericholscher (approved):
Seems like a better fix than my tmux session :D

PR #10006, opened 2023-02-08
Builds: ignore cancelling the build at "Uploading" state

humitos:
This protects the user to end up with half of their documentation uploaded.

### DIFF
--- a/readthedocs/projects/tasks/builds.py
+++ b/readthedocs/projects/tasks/builds.py
@@ -316,6 +316,14 @@ class UpdateDocsTask(SyncRepositoryMixin, Task):
 
         def sigint_received(*args, **kwargs):
             log.warning('SIGINT received. Canceling the build running.')
+
+            # Only allow to cancel the build if it's not already uploading the files.
+            # This is to protect our users to end up with half of the documentation uploaded.
+            # TODO: remove this condition once we implement "Atomic Uploads"
+            if self.data.build["state"] == BUILD_STATE_UPLOADING:
+                log.warning('Ignoring cancelling the build at "Uploading" state.')
+                return
+
             raise BuildCancelled
 
         # Do not send the SIGTERM signal to children (pip is automatically killed when


readthedocs/projects/tasks/builds.py
```diff
@@ -316,6 +316,14 @@ def sigterm_received(*args, **kwargs):
 
         def sigint_received(*args, **kwargs):
             log.warning('SIGINT received. Canceling the build running.')
+
+            # Only allow to cancel the build if it's not already uploading the files.
+            # This is to protect our users to end up with half of the documentation uploaded.
+            # TODO: remove this condition once we implement "Atomic Uploads"
+            if self.data.build["state"] == BUILD_STATE_UPLOADING:
+                log.warning('Ignoring cancelling the build at "Uploading" state.')
```
ericholscher, line 324: I think this will just end with the build getting KILL after INT, after an amount of time, but I guess that will take a bit longer?

ericholscher (approved):
I feel like there's probably more we should be doing here, but this seems like a useful first step.

readthedocs/projects/tasks/builds.py
humitos: I'm not sure to understand your comment.

When the build process receive the INT signal, we are just ignoring it here. The build will continue uploading the files normally.

humitos:
> I feel like there's probably more we should be doing here

What other things you have in mind? Are they easy enough to add to this PR? If not, we should open an issue with these ideas so we don't loose them.

readthedocs/projects/tasks/builds.py
ericholscher: Yes, but celery will send a KILL signal after the hard stop limit: https://docs.celeryq.dev/en/stable/userguide/configuration.html?highlight=settings#task-time-limit

ericholscher:
Mostly just trying to have better error handling when the process gets KILL'd. I think the atomic uploads (#9947) captures the useful parts of that conversation though.

readthedocs/projects/tasks/builds.py
humitos: Yes. Hard stop limit is a different thing, tho. I opened a PR some time ago to work on that but it's not trivial: https://github.com/readthedocs/readthedocs.org/pull/9387

PR #10026, opened 2023-02-14
Docs: Navigation reorder (Diátaxis)

benjaoming:
This reorders the navigation, anticipating new top-levels to be created in the next iteration.

---

- User's documentation (`docs`): https://docs--10026.org.readthedocs.build/en/10026/

- Developer's documentation (`dev`): https://dev--10026.org.readthedocs.build/en/10026/


benjaoming:
Maybe @humitos has ideas for Tutorial/How-to/Explanation/Reference emojis :)

docs/user/index.rst
```diff
@@ -1,6 +1,63 @@
...
+   /single_version
+   /environment-variables
+
+.. toctree::
+   :maxdepth: 2
+   :hidden:
+   :caption: Reference 📚
+
+   /reference/features
+   /about/index
```
ericholscher, line 52: This is too high. It should be above changelog probably.

docs/user/index.rst
```diff
@@ -1,6 +1,63 @@
...
+   /api/index
+   /server-side-search/api
+   /server-side-search/syntax
+   /reference/environment-variables
```
ericholscher, line 56: Little confused how this one got here? Seems out of place, like it should be under `features`?

docs/user/index.rst
```diff
@@ -1,6 +1,63 @@
...
+   /guides/administrators
+   /guides/developers
+   Troubleshooting </guides/troubleshooting/index>
+   /examples
+   /faq
+
+.. toctree::
+   :maxdepth: 2
+   :hidden:
+   :caption: Explanation 💡
```
ericholscher, line 29: These articles still seem oddly different in the tone of the title. It seems like eg. `Custom Domains` should be moved to features, or the title should be changed?

ericholscher (commented):
This is a reasonable refactor, but still feels like there's some polish needed. This is a good middle step tho.

docs/user/index.rst
ericholscher: If we're gonna put anything like this in the top-level, config file is the obvious one to start with, I think.

ericholscher:
@benjaoming I'm gonna do a commit on top of this to address a couple ideas.

docs/user/index.rst
```diff
@@ -1,6 +1,63 @@
...
+
+   /reference/features
+   /about/index
+   /api/index
+   /server-side-search/api
+   /server-side-search/syntax
+   /reference/environment-variables
+   changelog
+   /glossary
+   Developer Documentation 🔗 <https://dev.readthedocs.io>
```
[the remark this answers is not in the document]
benjaoming: Do you mean that it's possible to inject HTML here? Or we use some nifty extension for that?

docs/user/index.rst
```diff
@@ -1,6 +1,63 @@
...
+   /subprojects
+   /single_version
+   /environment-variables
+
+.. toctree::
+   :maxdepth: 2
+   :hidden:
+   :caption: Reference 📚
+
+   /reference/features
```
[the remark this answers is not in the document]
benjaoming: It's definitely a heavy lift for now.

The landing page could perhaps accommodate something here?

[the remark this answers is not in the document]
ericholscher: Yea, there's a few places we likely need to address it.

docs/user/index.rst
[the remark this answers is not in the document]
ericholscher: Not sure, but that's the right styling 🤷

ericholscher:
Alright, I did a bunch of refactoring. I think the real takeaway here is that we don't have much in the way of explanation content written yet. I moved a lot of content out of that section because it didn't really belong there as currently written. 

I do worry a bit that we're back in the place of having done a bunch of content improvements here, but have made the docs a bit less usable for normal users by hiding a bunch of content below the top-level TOC. There is always going to be a trade-off though between an overwhelming TOC, and better structure and nesting. I think most people will be finding content via Google though, so I'm not too worried in general, as long as it's possible to find it via navigation. 

I think some of the future index page updates will help there, and building out a few other small indexes for each of the sections.

docs/user/connected-accounts.rst
```diff
@@ -1,5 +1,5 @@
-Connecting your Git repository
-==============================
+How to connect your Git repository
+==================================
```
benjaoming, line 2: This was intended as an explanation since there isn't anything actionable. I can guess at several reasons that a user should read the article, and since some of them are also how-to oriented, I don't see anything bad with the refactor.

We can call it a how-to, that can work as well :+1:

docs/user/index.rst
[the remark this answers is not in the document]
benjaoming: Sounds like we can do it with CSS and `:after` :+1:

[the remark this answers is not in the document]
benjaoming: Added to iteration 2

benjaoming:
> Alright, I did a bunch of refactoring. I think the real takeaway here is that we don't have much in the way of explanation content written yet. I moved a lot of content out of that section because it didn't really belong there as currently written.

That makes the navigation look better and less overwhelming - which was intended as next steps :)

And it also fills up the Feature Reference to look more complete, which I understand is very important.

But it also moves 10 articles where the content was written as explanation. So that'll leave some refactoring debt.

Examples:

* this is not a how-to, it's explanation: https://docs--10026.org.readthedocs.build/en/10026/automatic-redirects.html - I think it should be moved back, but we can invent a format that fits well to cases like this: "Best practices for <property>" or "Steps to make a documentation more <property>"
* This was an explanation: https://docs--10026.org.readthedocs.build/en/10026/canonical-urls.html - the feature reference should also exist, but again we'd need to figure out a good "template" for how we want to represent the a) explanation of a concept and b) the feature.

docs/user/integrations.rst
```diff
@@ -6,8 +6,8 @@
        (Perhaps reuse this: https://about.readthedocs.com/images/homepage.png)
 
 
-Building documentation on every Git commit
-==========================================
+Continuous Documentation Deployment
+===================================
```
benjaoming, line 10: I'm a very big fan of this!

Even shortening it to "Continuous Documentation" can be an interesting way of talking about the workflow and the technical platform at the same time.

docs/user/index.rst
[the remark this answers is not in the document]
benjaoming: I think the work you did to the feature reference page was a great way to quickly improve it using toctrees :+1:

docs/user/index.rst
```diff
@@ -1,6 +1,63 @@
...
+
+   /tutorial/index
+   /intro/getting-started-with-sphinx
+   /intro/getting-started-with-mkdocs
+   /intro/import-guide
+
+.. toctree::
+   :maxdepth: 2
+   :hidden:
+   :caption: How-to guides 🪄
```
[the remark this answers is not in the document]
benjaoming: I'd like to defer to emoji expert @humitos for suggestions to all 4 categories - will gladly update later :)

docs/conf.py
```diff
@@ -61,6 +61,7 @@
 release = version
 exclude_patterns = ["_build", "shared", "_includes"]
 default_role = "obj"
+intersphinx_cache_limit = 14  # cache for 2 weeks
```
benjaoming, line 64: Thanks :)

docs/user/about/index.rst
```diff
@@ -68,7 +53,23 @@ we've brought most of the most important ones below.
 ⏩ :doc:`/support`
    Read this before asking for help: How to get support and where.
 
+⏩ :doc:`/glossary`
+   A useful index of terms used in our docs
```
benjaoming, line 57: This seems like a wrong location for our Glossary?

Is it because you think the discoverability of the page itself isn't valuable? I tend to agree with that. So kind of :broom: 'ing into this corner seems fine until it has so many terms that it's a real resource.

docs/user/commercial/sharing.rst
```diff
@@ -1,5 +1,5 @@
-Sharing
-=======
+Private Documentation Sharing
+=============================
```
benjaoming, line 2: :100:

docs/user/guides/authors.rst
```diff
@@ -16,6 +16,9 @@ and :doc:`/intro/getting-started-with-mkdocs`.
    cross-referencing-with-sphinx
    Link to external projects (Intersphinx) <intersphinx>
    jupyter
+   /guides/technical-docs-seo-guide
    Migrate from rST to MyST <migrate-rest-myst>
    enable-offline-formats
    Using search analytics <search-analytics>
+   /automatic-redirects
+   /science
```
benjaoming, line 24: hahaha, this page will keep moving around because it's such a hack.

BUT it's noted for later :+1:

docs/user/index.rst
```diff
@@ -198,7 +183,7 @@ out of your documentation and Read the Docs.
 
 * **Advanced project configuration**:
   :doc:`subprojects` |
-  :doc:`Single version docs <single_version>` |
+  :doc:`Single version docs <single-version>` |
```
benjaoming, line 186: redirect needed?

ericholscher:
> this is not a how-to, it's explanation: https://docs--10026.org.readthedocs.build/en/10026/automatic-redirects.html - I think it should be moved back, but we can invent a format that fits well to cases like this: "Best practices for " or "Steps to make a documentation more "

I would argue it's much closer to a How To than an explanation, given that the primary content is a set of Do's and Dont's ?

> This was an explanation: https://docs--10026.org.readthedocs.build/en/10026/canonical-urls.html - the feature reference should also exist, but again we'd need to figure out a good "template" for how we want to represent the a) explanation of a concept and b) the feature.

I'm also not sure this is useful enough to be a top-level explanation? Perhaps we need more nesting for "niche" explanations like this. But I don't see this content as something a normal user needs to care about really at all?

docs/user/about/index.rst
ericholscher: I think we need some other reference index to hide this stuff under, rather than in the top-level nav, is the real answer.

docs/user/guides/authors.rst
ericholscher: Yea, I mostly just hid it away. I think it could live under a top-level Tutorials page like `Specific use cases` or something?

docs/user/index.rst
ericholscher: Already created.

benjaoming:
> I would argue it's much closer to a How To than an explanation, given that the primary content is a set of Do's and Dont's ?

I imagined a refactor where the objective is to avoid mixing how-to and explanation by means of:

1) Creating a simple and actionable list of do's and don'ts as you mentioned. But the "executive summary"
2) Keeping the bulk of the current article as the longer explanation, perhaps removing the simplest actionable stuff and referring to the "How-to: Best practices for SEO"-ish article

benjaoming:
> I'm also not sure this is useful enough to be a top-level explanation?

Definitely agreeing that it's not a very important article to figure as a top-level. I think you did a wonderful work on getting the current navigation to be very beautiful and easy to read.

I had an intention to be more pragmatic to accept more overwhelming amounts of items in Explanation because of the next steps to add new sub-levels in Explanation and How-to.

humitos:
> Maybe @humitos has ideas for Tutorial/How-to/Explanation/Reference emojis :)

I'd put the emojis at the beginning of the sentence so they are all aligned:

Screenshot_2023-02-16_08-07-35


Also, I would not use emojis on the inner links because it make them more prominent and differentiate over the others. In particular, "Developer Documentation", is _the least important one_ but it's the most prominent in that section now because of the emoji.

PR #10033, opened 2023-02-15
Docs: Upgrade Sphinx & sphinx_rtd_theme

ericholscher:
This basically removes all pins from ``docs.in``,
and lets it resolve the properly fancy versions :)


humitos (approved):
👏

benjaoming:
harvesting the good stuff :100:

ericholscher:
It's still pinning us to Sphinx 5.3, which I'm guessing is the upper limit on one of the dependencies? But baby steps :)

benjaoming:
Yes, it's because of sphinx-design IIRC

Issue #10036, opened 2023-02-15
Placeholder: build output path issues


ericholscher:
I think we've come across a few edge cases where the build output path changes have introduced bugs. We can probably expect more or perhaps a few fixes here.

Add warning if the old artifact path still exists, or a shim to copy that in with a warning for now.

This one will be partially solved by documentation too, though I'm guessing we probably want some error catching here, if a third party tool writes to `_build/html` etc.

humitos:
@agjohnson do you have an idea about how to show this warning?

Currently, we don't support showing a warning message from the build. We only support showing an error. I'd say that probably the first step here is to `log.info` those projects that are outputting this into `_build/html` and see how many are those. Then, decide what to do and how.

humitos:
@agjohnson also, do you have an example of a project outputting to `_build/html`?

agjohnson:
> do you have an idea about how to show this warning?

I've described needing not a warning, but a build error. Projects that are using a hardcoded path end up with docs missing lots of content right now. This is silent, and only noticeable if you open up your documentation and notice the missing content.

A warning is only a small upgrade here, as it is still silent, and still requires the user to open up their documentation to find the missing content.

So, a hard error makes the most sense here. I'd prefer setting `Build.error` for now, but would like to continue pushing towards better tip/warning/error messages.

agjohnson:
> also, do you have an example of a project outputting to _build/html?

I'd have to dig up the conversations in support, but anyone using Doxygen might be hitting this right now.

humitos:
> We probably want an error, not a warning

OK. I will do that then. Makes sense 👍🏼

agjohnson:
With the logging changes through this morning, I think the volume of these errors is reasonable enough to fail any build outputting to `_build/html`. So far, I only see two hits from commercial, none from community yet.

humitos:
It seems there is only one project in commercial so far: https://onenr.io/0OQMEqG19QG

PR #10038, opened 2023-02-16
Build: log usage of old output directory `_build/html`

humitos:
We have changed the output directory from `_build/html` to `$READTHEDOCS_OUTPUT/html` in the last few weeks.

However, there are some projects that may have `_build/html` directory hardcoded in different places (e.g. `conf.py`). We want to log these projects and contact them to communicate this situation.

Ideally, we should show a warning in the build detail's page, but I don't think we currently have a way to display a warning message. We only support error messages in this page.

```
build_1     | [warning  ] Directory '_build/html' exists. This may lead to unexpected behavior. [readthedocs.telemetry.collectors] build_id=2756 builder=c9b06583941f commit=None ip=10.10.0.100 path_resolved=/usr/src/app/checkouts/readthedocs.org/user_builds/c9b06583941f/test-builds/checkouts/old-output-directory/.readthedocs.yaml project_slug=test-builds request_id=8bfcc743-dbc4-4886-8da0-b8a36709e556 task_id=ecbd1cd2-f352-43e0-be35-9442b1cab77e user_id=1 version_slug=old-output-directory
```

Related #10036


stsewd (approved):

readthedocs/telemetry/collectors.py
```diff
@@ -92,8 +92,27 @@ def collect(self):
...
+        #
+        # This temporal and should be removed in the future once we have
+        # decided what to do with this collected data.
+        self._check_using_old_output_directory()
+
         return data
 
+    def _check_using_old_output_directory(self):
+        code, stdout, stderr = self.run(
+            "ls",
```
agjohnson, line 107: The more standard way of testing for this path would be:

```suggestion
            "test",
            "-d",
```

agjohnson: Or if we're worried it's a symlink:

```suggestion
            "test",
            "-x",
```

readthedocs/telemetry/collectors.py
```diff
@@ -92,8 +92,27 @@ def collect(self):
...
+            "_build/html",
+            cwd=self.project.checkout_path(self.version.slug),
+        )
+        if code == 0:
+            log.warning(
+                "Directory '_build/html' exists. This may lead to unexpected behavior."
```
agjohnson, line 113: We should raise something user facing as well. I'm still a huge +1 on better error/warning patterns, but would it make sense to set the `Build.error` here? If there is something in this path, that almost certainly mean the build will be missing content if we don't fail the build explicitly.

readthedocs/telemetry/collectors.py
```diff
@@ -92,8 +92,27 @@ def collect(self):
             },
         }
         data["doctool"] = self._get_doctool()
+
+        # NOTE: Check if there are files at `_build/html` (the old output directory)
+        # and log these projects so we can communicate with their maintainers
+        #
+        # This temporal and should be removed in the future once we have
+        # decided what to do with this collected data.
```
agjohnson, line 100: We probably aren't as concerned about collecting this data as we are about saving user builds from having bad content. Anyone using doxygen and friends could have `_build/html` hardcoded, and will have silently broken docs.

agjohnson:
> We want to log these projects and contact them to communicate this situation.

It seems easier on us to just emit a build error. Is there a strong reason not to raise a build error instead?

humitos:
@agjohnson the only reason was that I thought we didn't want to fail the build. I will update this PR (or close it and open a new one) to raise an error and clearly communicate this to the users.

humitos:
@agjohnson OK. I updated this PR so it goes out next week and we start collecting this data. I will open a new PR that fails the build completely in those cases, but we may not get in time to deploy it next week tho.

PR #10039, opened 2023-02-16
Docs: emojis in TOC captions, FontAwesome on external links in TOC (Diátaxis)

benjaoming:
Refs: https://github.com/readthedocs/readthedocs.org/pull/10026#issuecomment-1432624078

---

- User's documentation (`docs`): https://docs--10039.org.readthedocs.build/en/10039/

- Developer's documentation (`dev`): https://dev--10039.org.readthedocs.build/en/10039/

### DIFF
--- a/docs/_static/css/custom.css
+++ b/docs/_static/css/custom.css
@@ -9,3 +9,15 @@
 #rtd-stickybox-container {
   text-align: center;
 }
+
+.wy-menu.wy-menu-vertical a.reference.external::after {
+  /* \00a0 = no break space */
+  content: "\00a0\f08e";
+  font-family: FontAwesome;
+  display: inline-block;
+  font-style: normal;
+  font-variant: normal;
+  text-rendering: auto;
+  -webkit-font-smoothing: antialiased;
+
+}
--- a/docs/_static/js/expand_tabs.js
+++ b/docs/_static/js/expand_tabs.js
@@ -21,17 +21,19 @@ $( document ).ready(function() {
 
   // Uses a hash referencing a Sphinx label from the URL page#sphinx-label
   var hash = window.location.hash.substr(1);
-  hash = hash.replace(/[^0-9a-z\-_]/gi, '');
-  // If the hash is inside a tab panel
-  var tab_with_reference = $(".sphinx-tabs-panel #" + hash).parents(".sphinx-tabs-panel");
+    if (hash) {
+    hash = hash.replace(/[^0-9a-z\-_]/gi, '');
+    // If the hash is inside a tab panel
+    var tab_with_reference = $(".sphinx-tabs-panel #" + hash).parents(".sphinx-tabs-panel");
 
-  if (tab_with_reference.length > 0) {
-    // Use the panel's ID to guess the tab's ID
-    var panel_id = tab_with_reference.first().attr("id");
-    var tab_id = panel_id.replace("panel-", "tab-");
-    // Invoke the tab buttons click() method to display it
-    $("button#" + tab_id).click();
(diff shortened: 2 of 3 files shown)


docs/_static/css/custom.css
```diff
@@ -9,3 +9,15 @@
 #rtd-stickybox-container {
   text-align: center;
 }
+
+.wy-menu.wy-menu-vertical a.reference.external::after {
```
ericholscher, line 13: I'd argue for this on _all_ external links, but baby steps :)

Though [@humitos had a good point](https://github.com/readthedocs/readthedocs.org/pull/10026#issuecomment-1432624078) that it makes the link more visible, but I think it's good to be explicit here?

docs/_static/js/expand_tabs.js
```diff
@@ -21,17 +21,19 @@ $( document ).ready(function() {
 
   // Uses a hash referencing a Sphinx label from the URL page#sphinx-label
   var hash = window.location.hash.substr(1);
-  hash = hash.replace(/[^0-9a-z\-_]/gi, '');
-  // If the hash is inside a tab panel
-  var tab_with_reference = $(".sphinx-tabs-panel #" + hash).parents(".sphinx-tabs-panel");
+    if (hash) {
```
ericholscher, line 24: What changed in this file? Seems like it needs a comment.

ericholscher (approved):
Looks 💯 to me

Screenshot 2023-02-16 at 9 18 58 AM

docs/_static/js/expand_tabs.js
benjaoming: The git diff is awful, it was just a matter of adding `if (hash) {` to fix a JS error occurring on all pages.

It's explained in line 22, the following code is pretty verbose:

```
// Uses a hash referencing a Sphinx label from the URL page#sphinx-label
var hash = window.location.hash.substr(1);
if (hash) {
...
```

docs/_static/css/custom.css
benjaoming: image

> Though https://github.com/readthedocs/readthedocs.org/pull/10026#issuecomment-1432624078 that it makes the link more visible, but I think it's good to be explicit here?

The font-awesome icon is much more discreet so I think that it works better.

benjaoming: I get a feeling that this will cause unintended side-effects.. like if the `<a>` is on an `` tag

ericholscher: Yea, we likely need to scope it smartly. I'm sure there's good examples of doing this other places, but we can start w/ the toctree for now.

gwilku:
Is there any chance this feature and usage is documented in the user guide? I'd like to use emoji's in my docs, but can't figure a way to add the sphinxemoji module myself...

benjaoming:
@gwilku you can actually use emojis directly in your code. Using an extension isn't required. Here is the raw code: https://raw.githubusercontent.com/readthedocs/readthedocs.org/main/docs/user/index.rst

Small tip: You should check that your text editor can display emojis, though, since it's really annoying to edit files with emojis that aren't shown in the editor. Most modern text editors do.

Good luck!

gwilku:
Yes, but it crashes PDF generator. these unicode characters are not
supported.

śr., 15 mar 2023, 16:13 użytkownik Benjamin Balder Bach <
***@***.***> napisał:

benjaoming:
@gwilku interesting! You can perhaps use the `only` directive to fix that: https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-only

PR #10042, opened 2023-02-16
Refactor: remove `Version.documentation_type`

humitos:
We are planning to suppose _any_ doctool and we want to get rid of the `documentation_type` attribute and make our logic more generic to work on most scenarios without specific logic per doctool.

### DIFF
--- a/readthedocs/api/v2/views/footer_views.py
+++ b/readthedocs/api/v2/views/footer_views.py
@@ -155,6 +155,7 @@ class BaseFooterHTML(CDNCacheTagsMixin, APIView):
 
         page_slug = self.request.GET.get('page', '')
         path = ''
+        # TODO: review custom logic based con `documentation_type`
         if page_slug and page_slug != 'index':
             if version.documentation_type in {SPHINX_HTMLDIR, MKDOCS}:
                 path = re.sub('/index$', '', page_slug) + '/'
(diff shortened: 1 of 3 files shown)


humitos:
Summarizing the required changes here:

* EmbedAPIv1 needs to be removed since it uses `documentation_type` to decide how to parse the file. I took a look at CF and I didn't find people using it. So, it seems safe to remove.
* Footer API is using it to generate a `path` and remove the `/index` part of the URL when it's Sphinx HTMLDIR or MKDOCS. It may makes sense to review this logic because I'm not 100% sure how it works and why it's needed. Also, this endpoint will be replaced by the new Addons flyout and its `/_/readthedocs-config/` API endpoint.
* The same logic about striping the `index.html` is used in the `PageSearchSerializer` APIv2 serializer.
* Telemetry data collector use this type to decide whether or not collecting Sphinx specific data. This code could easily be adapted by checking for the existence `telemetry.json` file instead.
* We require https://github.com/readthedocs/readthedocs.org/issues/10272 to always use the `GenericParser` to create search indexes.

It seems the only blocker here is "understanding the logic behind the stripping of `index.html` from FooterAPI and `PageSearchSerializer`. Then, we can re-write that logic to not be based on the doctype or remove it completely if it's not required.

I think we are close to be able to move forward here.

humitos:
Related:
* https://github.com/readthedocs/readthedocs.org/issues/10677
* https://github.com/readthedocs/website/pull/308

humitos:
We could probably give it another change to this PR since a lot of things have changed since the last commit. I suppose we are probably in a better place now to be able to remove this field.

Issue #9747, opened 2022-11-21
Docs: Diátaxis refactor - iteration 2 high-level tracking


benjaoming:
Successor of #9746 

This is a high-level issue moved from the [initial planning document](https://docs.google.com/document/d/1LQXMUOAKBZ4RhAew2F4TCDp-QW7czH5Dq6BPxlyA_dw/edit#). Further issues may be opened and linked from here.

* **Acceptance Criteria:** New sub-levels are created in Explanation and How-to, old How-to sub-levels are removed, Any remaining “Features” are moved to Reference and finally delete “Features”. Out of scope: Articles that have to be created from scratch.

* **Outcome:** The menu already starts to have clear entrypoints aligned with Diátaxis (but it is still lengthy).

### Ongoing work from iteration 1

* [x] Relabel "Build process" in Explanation (Title suggestion: "Build process overview") (@ericholscher)
  * Implement inputs from @agjohnson https://github.com/readthedocs/readthedocs.org/issues/9575
  * https://github.com/readthedocs/readthedocs.org/pull/10028
* [x] ~Split "Flyout Menu"~ We can skip this, a new flyout menu is on its way
  * Reference / Features / Flyout Menu - https://docs.readthedocs.io/en/stable/flyout-menu.html
  * How-to / How to customize Read the Docs “Flyout” Menu
    - Example: https://stackoverflow.com/questions/50425740/how-to-find-the-pdf-version-of-a-read-the-docs-project
* [x] Split "Configuration File" https://github.com/readthedocs/readthedocs.org/pull/10416
  * Explanation / Configuration File: Using “Configuration as Code” (.readthedocs.yaml)
  * Reference / Configuration File
  * Bonus: Post an issue that we should perhaps have a How-to for adding a configuration file to a project?
* [x] Split "Build customization" https://github.com/readthedocs/readthedocs.org/pull/10028
  * Discussed on call: Let's focus on reference
  * Explanation / Build process details
  * How-to / How to override the entire default build process (Docsify, Pelican, Hugo and more)
  * It's possible that this will generate more how-to articles.
  * Implement inputs from @agjohnson https://github.com/readthedocs/readthedocs.org/issues/9575
* [x] Split "Reproducible Builds" (@ericholscher) https://github.com/readthedocs/readthedocs.org/pull/10030
  * Explanation / Reproducible Builds
  * How-to / How to write a requirements.txt for Reproducible Builds
[...]

PR #10059, opened 2023-02-22
Docs: Fix undeclared labels after refactor + fix root causes

benjaoming:
* [x] conf.py update so it doesn't happen again
* [x] Fix 33 unresolved labels
* [x] Add documentation about this issue

---

- User's documentation (`docs`): https://docs--10059.org.readthedocs.build/en/10059/

- Developer's documentation (`dev`): https://dev--10059.org.readthedocs.build/en/10059/

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -88,13 +88,12 @@ intersphinx_mapping = {
     "jupyter": ("https://docs.jupyter.org/en/latest/", None),
 }
 
-# Do not try to resolve ANY unresolved labels that aren't explicitly prefixed.
-# Meaning that an unresolved :ref:`label_name` isn't resolved in all of the
-# intersphinx doctrees. Only :ref:`mapping:label_name` will get resolved.
+# Intersphinx: Do not try to resolve unresolved labels that aren't explicitly prefixed.
 # The default setting for intersphinx_disabled_reftypes can cause some pretty bad
 # breakage because we have rtd and rtd-dev stable versions in our mappings.
 # Hence, if we refactor labels, we won't see broken references, since the
 # currently active stable mapping keeps resolving.
+# Recommending doing this on all projects with Intersphinx.
 # https://www.sphinx-doc.org/en/master/usage/extensions/intersphinx.html#confval-intersphinx_disabled_reftypes
 intersphinx_disabled_reftypes = ["*"]
 
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -87,9 +87,17 @@ intersphinx_mapping = {
     "rtd-dev": ("https://dev.readthedocs.io/en/latest/", None),
     "jupyter": ("https://docs.jupyter.org/en/latest/", None),
 }
-# Redundant in Sphinx 5.0
+
+# Do not try to resolve ANY unresolved labels that aren't explicitly prefixed.
+# Meaning that an unresolved :ref:`label_name` isn't resolved in all of the
+# intersphinx doctrees. Only :ref:`mapping:label_name` will get resolved.
+# The default setting for intersphinx_disabled_reftypes can cause some pretty bad
+# breakage because we have rtd and rtd-dev stable versions in our mappings.
+# Hence, if we refactor labels, we won't see broken references, since the
+# currently active stable mapping keeps resolving.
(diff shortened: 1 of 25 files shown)


docs/conf.py
```diff
@@ -87,9 +87,16 @@
...
     "jupyter": ("https://docs.jupyter.org/en/latest/", None),
 }
-# Redundant in Sphinx 5.0
+
+# Intersphinx: Do not try to resolve unresolved labels that aren't explicitly prefixed.
+# The default setting for intersphinx_disabled_reftypes can cause some pretty bad
+# breakage because we have rtd and rtd-dev stable versions in our mappings.
+# Hence, if we refactor labels, we won't see broken references, since the
+# currently active stable mapping keeps resolving.
+# Recommending doing this on all projects with Intersphinx.
```
ericholscher, line 96: Not sure what this line means, but doesn't seem like this is the right place for it?

docs/conf.py
```diff
@@ -87,9 +87,16 @@
     "rtd-dev": ("https://dev.readthedocs.io/en/latest/", None),
     "jupyter": ("https://docs.jupyter.org/en/latest/", None),
 }
-# Redundant in Sphinx 5.0
+
+# Intersphinx: Do not try to resolve unresolved labels that aren't explicitly prefixed.
+# The default setting for intersphinx_disabled_reftypes can cause some pretty bad
+# breakage because we have rtd and rtd-dev stable versions in our mappings.
+# Hence, if we refactor labels, we won't see broken references, since the
+# currently active stable mapping keeps resolving.
```
ericholscher, line 95: 💯 comment.

docs/user/analytics.rst
```diff
@@ -19,7 +19,7 @@ and then click on :guilabel:`Traffic Analytics`.
 
    Traffic analytics demo
 
-You can also access to analytics data from :ref:`search results <server-side-search/index:Search Analytics>`.
```
ericholscher, line 22: I kinda like the refs because they hoverxref. Should we standardize on using them instead of `:doc:`?

docs/user/automation-rules.rst
```diff
@@ -82,7 +82,7 @@ Make version private
 Set version as default
   Sets the version as the :term:`default version`.
   It also activates and builds the version.
-  See :ref:`automatic-redirects:Root URL`.
+  See :ref:`root_url_redirect`.
```
ericholscher, line 85: This is probably something we should have in the style guide as well. Should we use the titles, which are more likely to change, or make manual refs?

Basically: when do we make manual refs for something?

ericholscher (approved):

docs/user/analytics.rst
benjaoming: > I kinda like the refs because they hoverxref.

This is true. I would say that we should add actual labels to headlines so we can freely change headlines and not have to update references as a result (that adds more work to a simple headline improvement).

docs/user/automation-rules.rst
benjaoming: If the headlines seems likely to change, we should use labels... and maybe even add link texts if we think the headline is fragile to include in a sentence structure. There are many places where we use the headline in the sentence structure, so it will render an invalid sentence when the headline changes "Read more in our <Guide to custom domains>" will not look nice if it becomes "Read more in our <Custom domains>".

benjaoming: The best way to reference is of course to NOT use the headline in the sentence structure.

benjaoming: See also: `Custom domains`
See also: `Guide to custom domains`

This is not fragile to headline changes.

benjaoming: (the above text we're commenting on is fine though)

docs/conf.py
benjaoming: Removed

docs/user/analytics.rst
benjaoming: I think we should defer to https://github.com/readthedocs/sphinx-hoverxref/issues/18 because of the widespread use and need of `:doc:`. But would be great to fix this.

docs/user/automation-rules.rst
benjaoming: Will treat this in https://github.com/readthedocs/readthedocs.org/issues/9747

PR #10062, opened 2023-02-22
Add X-Content-Type-Options as a custom domain header

ericholscher:
A user requested this via https://infosec.mozilla.org/guidelines/web_security


dockerfiles/nginx/proxito.conf.template
```diff
@@ -82,6 +82,8 @@ server {
         add_header Access-Control-Allow-Headers $access_control_allow_headers always;
         set $x_frame_options $upstream_http_x_frame_options;
         add_header X-Frame-Options $x_frame_options always;
+        set $x_content_type_options $upstream_http_x_content_type_options;
+        add_header X-Content-Type-Options $x_content_type_options always;
```
humitos, line 86: This change probably requires to be duplicated in corporate and also on -ops.

humitos (approved):
Looks good 👍🏼 

We should have tests for this code eventually 😅

dockerfiles/nginx/proxito.conf.template
ericholscher: Hrm... Good point. I'll check before merging.

ericholscher:
Missed this before last deploy 🙃

PR #10078, opened 2023-02-28
Celery: consider only `PageView` from the last 3 months

humitos:
Similar to https://github.com/readthedocs/readthedocs.org/pull/10077

### DIFF
--- a/readthedocs/analytics/tasks.py
+++ b/readthedocs/analytics/tasks.py
@@ -9,7 +9,6 @@ from readthedocs.worker import app
 from .models import PageView
 from .utils import send_to_analytics
 
-
 DEFAULT_PARAMETERS = {
     'v': '1',  # analytics version (always 1)
     'aip': '1',  # anonymize IP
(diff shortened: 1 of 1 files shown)


readthedocs/analytics/tasks.py
```diff
@@ -81,4 +80,7 @@ def delete_old_page_counts():
     """
     retention_days = settings.RTD_ANALYTICS_DEFAULT_RETENTION_DAYS
     days_ago = timezone.now().date() - timezone.timedelta(days=retention_days)
-    return PageView.objects.filter(date__lt=days_ago).delete()
+    return PageView.objects.filter(
+        date__lt=days_ago,
+        date__gt=days_ago - timezone.timedelta(days=90),
```
ericholscher, line 85: Seems this should be dynamic, and based on `retention_days`?

ericholscher (approved):

readthedocs/analytics/tasks.py
humitos: It's based on `days_ago`, which is based on `retention_days`. I think it's fine.

It will delete `PageViews` older than `retention_days` and newer than `retention_days - 90 days`.

PR #10081, opened 2023-02-28
Migration edit: Allow null values in field "Project.delisted"

benjaoming:
Hotfix for deployment.

If you've done a local deployment of the current `main` branch, you need to:

```
# Before merging main
inv docker.manage "migrate projects 0095"
# After merging main
inv docker.manage "migrate projects"
```


readthedocs/projects/models.py
```diff
@@ -362,7 +362,11 @@ class Project(models.Model):
 
     skip = models.BooleanField(_("Skip (disable) building this project"), default=False)
 
+    # null=True can be removed in a later migration
+    # be careful if adding new queries on this, .filter(delisted=False) does not work
+    # but .exclude(delisted=True) does!
```
ericholscher, line 367: I think this will work fine, except for any projects created during the deploy window. But we can just plan to remove the null in next deploy.

ericholscher (approved):

PR #10082, opened 2023-02-28
Catch the base OAuth2Error error, not a custom one

ericholscher:
An update to https://github.com/readthedocs/readthedocs.org/pull/10022

### DIFF
--- a/readthedocs/oauth/services/gitlab.py
+++ b/readthedocs/oauth/services/gitlab.py
@@ -8,7 +8,7 @@ import structlog
 from allauth.socialaccount.providers.gitlab.views import GitLabOAuth2Adapter
 from django.conf import settings
 from django.urls import reverse
-from oauthlib.oauth2.rfc6749.errors import OAuth2Error
+from oauthlib.oauth2.rfc6749.errors import InvalidGrantError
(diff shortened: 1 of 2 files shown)


readthedocs/oauth/services/gitlab.py
```diff
@@ -602,7 +602,7 @@ def send_build_status(self, build, commit, state):
                 'GitLab commit status creation failed.',
                 debug_data=debug_data,
             )
-        except CustomOAuth2Error as e:
+        except OAuth2Error as e:
```
humitos, line 605: Why not catching directly `InvalidGrantError` and removing the following `if`?

humitos (commented):
Looks good, but I think it's better if we catch the _exact_ exception now we know which one is 😄

readthedocs/oauth/services/gitlab.py
ericholscher: That makes more sense, but I figured there was a reason it was implemented this way 🤷

humitos: Nope, the only reason was that I didn't find the correct/exact exception as you did 😄

humitos (approved):

PR #10086, opened 2023-02-28
Docs: add Plausible

### LINKED ISSUE #9603: Docs: use Plausible for analytics
We started using Plausible on our new website, we should eventually consolidate our own docs site there as well.

This might take a little bit of thought, but isn't a lot of work otherwise. Is there anything we care to preserve from GA or replicate from GA before migrating?

humitos:
Use Plausible for analytics in our documentation.

Closes #9603 

---

- User's documentation (`docs`): https://docs--10086.org.readthedocs.build/en/10086/

- Developer's documentation (`dev`): https://dev--10086.org.readthedocs.build/en/10086/


humitos:
We have this code for Google Analytics, should we have something similar for Plausible? https://github.com/readthedocs/readthedocs.org/blob/c445700d861b2610c89a77ac33a3c5f8a3eacfe8/readthedocs/templates/base.html#L21-L23

humitos:
Hrm, from their documentation I understand that it's not "required", tho -- https://plausible.io/docs/integration-guides#dnt-and-offline-functionality

> Plausible by default doesn't take DNT (Do Not Track) into consideration as we're a privacy first analytics tool and [DNT allows](https://www.eff.org/issues/do-not-track) "data analysis techniques that protect the anonymity of the users".

ericholscher:
> Hrm, from their documentation I understand that it's not "required", tho -- https://plausible.io/docs/integration-guides#dnt-and-offline-functionality
> 
> > Plausible by default doesn't take DNT (Do Not Track) into consideration as we're a privacy first analytics tool and [DNT allows](https://www.eff.org/issues/do-not-track) "data analysis techniques that protect the anonymity of the users".

Good question for @davidfischer, who I think did this work initially.

docs/_templates/layout.html
```diff
@@ -1,5 +1,10 @@
 {% extends "!layout.html" %}
 
+{% block extrahead %}
+  {{ super() }}
+
+  <script defer data-domain="docs.readthedocs.io" src="https://plausible.io/js/script.js"></script>
```
ericholscher, line 6: Does this also get added to `dev.`?

davidfischer:
> Good question for @davidfischer, who I think did this work initially.

I think it's OK to have Plausible despite DNT although it also might be nice to just disable it for those users. I don't think there's any hard requirement so I'll leave it up to you.

At some point we will have to update the privacy policy and our section on GA in our docs although maybe those can wait until we roll it out more widely.

docs/_templates/layout.html
humitos: Good point! 👍🏼 -- This should probably use a Sphinx variable.

benjaoming:
> I think it's OK to have Plausible despite DNT although it also might be nice to just disable it for those users. I don't think there's any hard requirement so I'll leave it up to you.

Oh interesting! Yup, they say so themselves here, stating that DNT says so, too.. https://plausible.io/docs/integration-guides#dnt-and-offline-functionality

docs/_templates/layout.html
humitos: I updated this to use `READTHEDOCS_PROJECT` variable.

We should create a new `dev.readthedocs.io` on Plausible now. Can you do that and give me permissions?

docs/conf.py
```diff
@@ -160,6 +160,8 @@
     # TODO: remove once we support different rtd config
     # files per project.
     "conf_py_path": f"/docs/{docset}/",
+    # Use to generate the Plausible "data-domain" attribute from the template
+    "plausible_domain": f"{os.environ.get('READTHEDOCS_PROJECT')}.readthedocs.io",
```
benjaoming, line 164: Running a normal `PROJECT=user make html` locally, I get:

```
<script defer data-domain="None.readthedocs.io" src="[https://plausible.io/js/script.js](view-source:https://plausible.io/js/script.js)"></script>
```

But it works for me with `docset`

```suggestion
    "plausible_domain": f"{docset}.readthedocs.io",
```

benjaoming (approved):
I like the approach, that's exactly what multiproject is for!! But think we need to do the below suggestion...

docs/conf.py
humitos: I've done that first. The problem with that is that it doesn't match the subdomain. `docset` values are `user` and `dev` and we need `docs` and `dev`. The project slug is the right value.

Locally, this value doesn't make too much sense. We could remove plausible locally if we want to, tho.

humitos: > Locally, this value doesn't make too much sense. We could remove plausible locally if we want to, tho.

I did this by checking for `if READTHEDOCS` in the template.

benjaoming: Oh right, I forgot that we have `user` matching `docs` here :)

The new solution is better :+1:

humitos:
Docs are failing because some unrelated warnings.

benjaoming:
We could merge https://github.com/readthedocs/readthedocs.org/pull/10090 and then update this PR and check the PR build?

PR #10087, opened 2023-02-28
Analytics: add Plausible to our dashboard

### LINKED ISSUE #9604: Application: use Plausible for dashboard analytics
We started using Plausible on our new website, we should eventually consolidate our own application/dashboard analytics there as well. This might aid in correlating website traffic to subscription creation/etc.

I don't believe we're tracking anything of moderate importance currently, so this should be fairly low friction.

humitos:
Use Plausible for dashboard's analytics.

Closes #9604

### DIFF
--- a/readthedocs/templates/base.html
+++ b/readthedocs/templates/base.html
@@ -58,6 +58,8 @@
   <!-- End Google Analytics -->
   {% endif %}
 
+  <script defer data-domain="{{ PRODUCTION_DOMAIN }}" src="https://plausible.io/js/script.js"></script>
+
   <!-- css -->
   <link rel="stylesheet" href="{% static 'css/core.css' %}">
   {% block extra_links %}{% endblock %}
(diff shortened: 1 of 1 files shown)


agjohnson (approved):
:+1: We'll also need this on the new dashboard templates

humitos:
I'm not merging this one before, because @ericholscher said that maybe this generates a lot of traffic and we may want to with commercial-only first. However, the PR is here and we can decide _when_ we want to merge it.

ericholscher:
Yea, I think .org gets a lot of traffic, and I'm not sure how useful it would be. I'd be 👍 on starting with admin pages or only for logged in users, if there's specific questions we want to ask vs. just capturing how many project detail pages get hit each day.

humitos:
I like the idea of logged-in users. I will adapt this PR to follow that.

ericholscher (approved):
Good start 👍

PR #10105, opened 2023-03-05
pre-commit: move `prospector` inside pre-commit

humitos:
- removes `prospector` check from tox
- run `prospector` check inside pre-commit

This allows developers to auto-run prospector _before_ pushing changes to GitHub and be able to fix these problems locally in a quicker way.

Requires: https://github.com/readthedocs/common/pull/166


.circleci/config.yml
```diff
@@ -46,7 +46,6 @@ jobs:
       - run: pip install --user 'tox<5'
       - run: tox -e pre-commit
       - run: tox -e migrations
-      - run: tox -e lint
```
ericholscher, line 49: I think we still want this in tox? What happens if a user doesn't run lint locally?

ericholscher: Ah, ok 👍

benjaoming (approved):
Yes please :heart:

The workflow of waiting for CI to lint things that aren't automatically linted locally is slow. So this will be a relative improvement and most welcome :+1: 

After doing this, I think we should seriously consider getting a much faster linting process, though. Having `https://raw.githubusercontent.com/readthedocs/readthedocs.org/main/requirements/pip.txt` inside pre-commit requirements also seems a bit dramatic.

.circleci/config.yml
benjaoming: pre-commit is super clever, it maintains a virtual env for each linting tool. So in this case, it's an improvement over `tox -e lint` since we don't want a virtualenv for _all_ linting, but one for each tool (so they don't conflict).

humitos:
> Having https://raw.githubusercontent.com/readthedocs/readthedocs.org/main/requirements/pip.txt inside pre-commit requirements also seems a bit dramatic.

This is only executed once when the environment is created. Unfortunately, on CircleCI we have to install these requirements each time :/ --but I didn't find a better way of doing this for now. We can explore other alternatives in the future.

benjaoming:
> This is only executed once when the environment is created. Unfortunately, on CircleCI we have to install these requirements each time :/ --but I didn't find a better way of doing this for now. We can explore other alternatives in the future.

If it makes sense from build times, we could open an issue explore putting `.pre-commit` into a Circle CI cache. AFAIK, pre-commit manages its virtualenvs in a very sophisticated manner so it's quite likely that it could bootstrap from a cache and save to a cache afterwards.

humitos:
It seems all of us are more or less on board with this change and the one in common/. We can tune it a little more as we go while finding things that are not working as we expect. Also, we can revert if we find it does not work at all.

I'm merging this PR and the one in common as well to start trying this out.

PR #10106, opened 2023-03-06
Proxito: use cacheops for 2 more models

humitos:
`Organization` and `PlanFeature` weren't being cached by `cacheops` and they are used when serving documentation.

By caching these, we will be caching all the models used when serving documentation, reducing the db hits and a also a litte the response time.

With this change, we won't hit the db at all when serving documentation 👍🏼 . This has other benefits besides trying to reduce the response time.

### DIFF
--- a/readthedocs/settings/base.py
+++ b/readthedocs/settings/base.py
@@ -1055,4 +1055,14 @@ class CommunityBaseSettings(Settings):
             'ops': CACHEOPS_OPS,
             'timeout': CACHEOPS_TIMEOUT,
         },
+
+        # readthedocs.organizations.*
+        'organizations.organization': {
+            'ops': CACHEOPS_OPS,
+            'timeout': CACHEOPS_TIMEOUT,
+        },
+        'organizations.planfeature': {
+            'ops': CACHEOPS_OPS,
+            'timeout': CACHEOPS_TIMEOUT,
+        },
     }


ericholscher (approved):
@humitos I thought we were pulling back on this experiment? Do you think this will show a significant improvement in doc serving? 

This is interesting for me in another context, which is availability. It would be really nice to be able to continue to serve docs via redis if we're having issues with Postgres.

Another area this is interesting is in distributed nodes around the world. If we're doing real-only doc serving, we could do some read-only proxito notes with `proxito + redis` in Europe, that only hit the US when something isn't already cached. This would be a really interesting experiment in my mind, if we could enable it with something this simple in our codebase.

humitos:
> @humitos I thought we were pulling back on this experiment?

I'm not canceling the experiment just yet because it's not worse than the original approach and I think it could be better with some more tuning. 

> Do you think this will show a significant improvement in doc serving response times?

I don't think so, but I wanted to remove the db requirement completely if possible because of the reasons you mentioned. Serving everything from cache could open other possibilities in the future and it seems that a hybrid method as we have right now is not a problem so far. However, taking a look at the graph, and we can earn ~1ms maybe. 

Due to these, I'll leave one instance running with these settings to keep exploring results when we have some time.

humitos:
> If we're doing redis-only doc serving, we could do some read-only proxito notes with `proxito + redis` in Europe, that only hit the US when something isn't already cached.

Note that it does not serve _the docs_ from Redis, we would still need a S3 replica in Europe as well.

ericholscher:
> > If we're doing redis-only doc serving, we could do some read-only proxito notes with `proxito + redis` in Europe, that only hit the US when something isn't already cached.
> 
> Note that it does not serve _the docs_ from Redis, we would still need a S3 replica in Europe as well.

Useful note. But at least removing Postgres from the picture is interesting as a starting point architecturally. It's also the most likely to have some kind of failure in my mind, and we _could_ proxy S3 to Europe if we wanted to pay for it. But yea, CDN solves most of this in a much cleaner way :)

PR #10113, opened 2023-03-06
Build: fail PDF command (`latexmk`) if exit code != 0

### LINKED ISSUE #10015: Document build failed with the error "Build output directory for format "pdf" contains multiple files and it is not currently supported."
I build my project document successfully before, but recently it failed with the error message "Build output directory for format "pdf" contains multiple files and it is not currently supported. Please, remove all the files but the "pdf" your want to upload. "
image

it seems like error occurred in this step: 
image

I didn't find any method to solve this problem. Is there anyone who knows this problem? Thank u.  T-T

humitos:
We were forcing `exit_code=0` if we detected that `Output written on (.*?)` was found in the standard output of the `latexmk` command. This was producing confusions to our users because the PDF generation failed but we were considering it successful.

Now, we are relying on the exit code itself as we do for all the other commands making sure the command succeeded before proceding with the build. Now, if the PDF command fails, we fail the build completely.

Closes #10015

### DIFF
--- a/readthedocs/doc_builder/backends/sphinx.py
+++ b/readthedocs/doc_builder/backends/sphinx.py
@@ -579,14 +579,18 @@ class PdfBuilder(BaseSphinx):
             '-interaction=nonstopmode',
         ]
 
-        cmd_ret = self.run(
-            *cmd,
-            cwd=self.absolute_host_output_dir,
-        )
+        try:
+            cmd_ret = self.run(
+                *cmd,
+                cwd=self.absolute_host_output_dir,
+            )
+            self.pdf_file_name = f"{self.project.slug}.pdf"
+            return cmd_ret.successful
 
-        self.pdf_file_name = f'{self.project.slug}.pdf'
-
-        return cmd_ret.successful
+        # Catch the exception and re-raise it with a specific message
+        except BuildUserError:
+            raise BuildUserError(BuildUserError.PDF_COMMAND_FAILED)
+        return False
 
     def _post_build(self):
         """Internal post build to cleanup PDF output directory and leave only one .pdf file."""
--- a/readthedocs/doc_builder/backends/sphinx.py
+++ b/readthedocs/doc_builder/backends/sphinx.py
@@ -26,8 +26,6 @@ from readthedocs.projects.models import Feature
 from readthedocs.projects.utils import safe_write
 
 from ..base import BaseBuilder
-from ..constants import PDF_RE
-from ..environments import BuildCommand, DockerBuildCommand
 from ..exceptions import BuildUserError
 from ..signals import finalize_sphinx_context_data
 
@@ -488,30 +486,6 @@ class EpubBuilder(BaseSphinx):
             )
 
 
-class LatexBuildCommand(BuildCommand):
-
-    """Ignore LaTeX exit code if there was file output."""
-
-    def run(self):
-        super().run()
-        # Force LaTeX exit code to be a little more optimistic. If LaTeX
-        # reports an output file, let's just assume we're fine.
-        if PDF_RE.search(self.output):
-            self.exit_code = 0
-
-
-class DockerLatexBuildCommand(DockerBuildCommand):
-
-    """Ignore LaTeX exit code if there was file output."""
-
-    def run(self):
-        super().run()
-        # Force LaTeX exit code to be a little more optimistic. If LaTeX
-        # reports an output file, let's just assume we're fine.
-        if PDF_RE.search(self.output):
-            self.exit_code = 0
-
-
 class PdfBuilder(BaseSphinx):
 
     """Builder to generate PDF documentation."""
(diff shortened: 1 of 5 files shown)


stsewd (approved):
I'm +1 with this change. @ericholscher may want to review this as well.

readthedocs/doc_builder/backends/sphinx.py
```diff
@@ -488,30 +486,6 @@ def _post_build(self):
             )
 
 
-class LatexBuildCommand(BuildCommand):
```
ericholscher, line 491: Love deleting super janky code 😻

readthedocs/doc_builder/backends/sphinx.py
```diff
@@ -610,10 +579,8 @@ def _build_latexmk(self, cwd):
             '-interaction=nonstopmode',
         ]
 
-        cmd_ret = self.build_env.run_command_class(
-            cls=latex_class,
-            cmd=cmd,
-            warn_only=True,
+        cmd_ret = self.run(
```
ericholscher, line 582: My only worry here is builds that were working before and break now. Is there a way to keep our existing `PDF Written` check, but have that modify the build failure code to explain what changed?

ericholscher (approved):
This is an obvious and good thing to do, but I do think we'll hear from users on this. It would be really wonderful if we could message this in the build failure smartly, so they know what happened.

readthedocs/doc_builder/backends/sphinx.py
humitos: I'm not sure to understand your proposal. You want to keep checking for the PDF written in disk and, if not found, fail the build with a custom message instead of the generic saying "You build failed due to a command that failed"?

humitos: OK. It wasn't hard to implement 😄 . I really like our current pattern with the Celery handler 💯

humitos:
> It would be really wonderful if we could message this in the build failure smartly, so they know what happened.

I wrote a simple and quick proposal for the message. I'm not happy with it tho 😅 . Please, suggest a better option if you have something in mind. This is how this page will look like:

Screenshot_2023-03-07_09-48-15

readthedocs/doc_builder/exceptions.py
```diff
@@ -52,6 +52,13 @@ class BuildUserError(BuildBaseException):
         "and it is not currently supported. "
         'Please, remove all the files but the "{artifact_type}" your want to upload.'
     )
+    PDF_COMMAND_FAILED = gettext_noop(
+        "PDF generation failed. "
+        "Please double check the command's output looking for errors and fix them. "
+        "Note it's possible the PDF has been always failing silently in your project, "
+        "but due to latest changes, "
+        "Read the Docs now makes it explicit and fail the whole build."
+    )
```
ericholscher, line 61:
```suggestion
    PDF_COMMAND_FAILED = gettext_noop(
        "PDF generation failed. "
        "The build log below contains information on what errors caused the failure." 
        "Our code has recently changed to fail the entire build on PDF errors, "
        "where we used to pass the build when a PDF was created."
        "Please contact us if you need help understanding this error."
    )
```

I don't love this. I think we probably want a link to a docs article here, but this is much better than a generic build error 👍

ericholscher (approved):
Thanks for implementing this. I was thinking it wouldn't be too difficult, and I think the UX is _much better_. I really like this pattern!

readthedocs/doc_builder/backends/sphinx.py
ericholscher: Thanks for working through my meaning here! I think it makes the code a lot better.

readthedocs/doc_builder/exceptions.py
humitos: Yeah, but it works for now 😄 . Let's move forward with this and update the message as we go.

readthedocs/doc_builder/exceptions.py
```diff
@@ -52,6 +52,13 @@ class BuildUserError(BuildBaseException):
         "and it is not currently supported. "
         'Please, remove all the files but the "{artifact_type}" your want to upload.'
     )
+    PDF_COMMAND_FAILED = gettext_noop(
+        "PDF generation failed. "
+        "The build log below contains information on what errors caused the failure." 
+        "Our code has recently changed to fail the entire build on PDF errors, "
+        "where we used to pass the build when a PDF was created."
+        "Please contact us if you need help understanding this error."
+    )
```
humitos, line 61:
```suggestion
    PDF_COMMAND_FAILED = gettext_noop(
        "PDF generation failed. "
        "The build log below contains information on what errors caused the failure."
        "Our code has recently changed to fail the entire build on PDF errors, "
        "where we used to pass the build when a PDF was created."
        "Please contact us if you need help understanding this error."
    )
```

PR #10114, opened 2023-03-06
Project: only support Git as VCS for new projects

humitos:
This is the continuation of a soft deprecation. We already remove mentions to other VCS different than Git from our documentation.

In this step, we remove the ability to import other repositories other than Git. However, we are keeping the code around to avoid existing projects to break when building. In the future, we will remove this support completely.

Continuation of #8840

### DIFF
--- a/readthedocs/projects/forms.py
+++ b/readthedocs/projects/forms.py
@@ -385,6 +385,7 @@ class UpdateProjectForm(
             # Basics
             'name',
             'repo',
+            "repo_type",
             # Extra
             'description',
             'language',
--- a/readthedocs/projects/forms.py
+++ b/readthedocs/projects/forms.py
@@ -86,7 +86,7 @@ class ProjectBasicsForm(ProjectForm):
 
     class Meta:
         model = Project
-        fields = ('name', 'repo', 'repo_type', 'default_branch')
+        fields = ("name", "repo", "default_branch")
 
     remote_repository = forms.IntegerField(
         widget=forms.HiddenInput(),
@@ -385,7 +385,6 @@ class UpdateProjectForm(
             # Basics
             'name',
             'repo',
-            'repo_type',
             # Extra
             'description',
             'language',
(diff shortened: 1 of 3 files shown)


readthedocs/projects/forms.py
```diff
@@ -86,7 +86,7 @@ class ProjectBasicsForm(ProjectForm):
 
     class Meta:
         model = Project
-        fields = ('name', 'repo', 'repo_type', 'default_branch')
```
ericholscher, line 89: This will also stop people from migrating off of old VCS to new right? I think we want to exclude it for projects using git, but keep it when they aren't, so they can change things? I think we also want to change the help_text to explain the situation to those projects.

benjaoming: I often use a conditional pattern to change a form field's widget to HiddenInput whenever its irrelevant or vice-versa. This happens in the form's `__init__`.

Example:

```
class SomeForm(forms.ModelForm):
    dynamic_field = forms.BooleanField()
        if self.instance and self.instance.blah = True:
            self.fields["dynamic_field"].widget = forms.HiddenInput()
```

By doing this, it's easier to phase out BZR/SVN/Mercurial support and still keep evolving the form for others. Hiding an input field in this way has much less impact on forms/views/models/templates than the alternatives.

humitos: I updated this code to:

- don't allow importing new projects that are not Git
- allow changing repository type on those projects that are already imported

humitos:
> Do we plan to tell them we can't support it, or some hacks around how to do it (eg. import it, then change it? 

I'd tell them we don't support it. Otherwise, we will end up with users doing hacky stuffs that are not standard and break because unknown/unexpected reasons.

> I do think this is a reasonable step, but I think these half-deprecations are also pretty confusing for users... but I also don't have a better idea for moving forward, so let's ship it.

Yeah, I think this is the best we can do at this moment. With this step, at least, we will be reducing the amount of projects that are using "unsupported" VCS backends and, when we decide to do the hard-deprecation and remove the code, we will have less projects to migrate away/communicate/help/etc.

PR #10115, opened 2023-03-06
Downloadable artifacts: make PDF and ePub opt-in by default

### LINKED ISSUE #8359: Make PDF and EPUB opt-in, rather than opt-out?
At the moment, every newly created project on Read the Docs has PDF and EPUB builds enabled in the UI:

Screenshot 2021-07-21 at 13-32-17 Edit Advanced Project Settings Read the Docs

If I understand correctly, this can be overridden by:

1. Unchecking those boxes in the UI, or
2. Using a `.readthedocs.yaml` file, where the default for `formats` is `[]` https://docs.readthedocs.io/en/stable/config-file/v2.html#formats

However, building PDFs is quite costly, error-prone, and unclear if being used by a majority of our users. On a related note, PDF errors are now passing silently #7884.

Given that the config file (2) in itself is opt-in, I wonder if we should make PDF and EPUB off by default on the UI.

Some questions that might help making this decision, although I don't think we should hard block the decision on those:

- What percentage of active\* projects use a config file on their default version?
- What percentage of active\* projects have zero downloads on their PDFs and EPUBs?

(*active = with X builds in the past Y months? with NNNN visits in the past Y months?)

Thoughts @readthedocs/core ?

humitos:
Newly created projects will have PDF and ePub disabled by default. This matches the default value of `formats` in the YAML file v2.

Building PDF and ePub is not a trivial task and many projects start failing because of this or, if succeeding, they are just building pretty low quality PDF and ePub files --which is a waste of resources.

This commit disables this by default for new projects, keeping the old projects working as they are currently.

Closes #8359


ericholscher (approved):
I'm 👍 on this. We definitely want folks opting into this if they want it. I wish there was an easy way to turn it off for projects that don't care, but stopping it defaulting on is step 1 :)

humitos:
> our build infra definitely is our least tested area, I think thinking

Yeah. Testing build workflow is _hard_. I've done lot of it when migrating our code to Celery handlers. However, it's super fragile and hard to follow/read due to the amount of mocks we are using 😞

PR #10119, opened 2023-03-07
Build: pass shell commands directly (`build.jobs` / `build.commands)`

### LINKED ISSUE #10103: Build: `build.jobs.post_build` produces mysterious shell syntax errors
## Details

* Read the Docs project URL: https://readthedocs.org/projects/mmref/
* Build URL (if applicable): https://readthedocs.org/projects/mmref/builds/19671469/ and https://readthedocs.org/projects/mmref/builds/19671433/ and most previous  builds
* Read the Docs username (if applicable): https://readthedocs.org/profiles/rptb1/

## Expected Result

A post_build script `if test -n "A"; then echo "B"; fi` should not produce a syntax error.
A post_build script with `sh -c "foo"` in it should run the shell with the argument.

## Actual Result

`/bin/sh: 1: Syntax error: "then" unexpected`
`sh: 0: -c requires an argument`

I'm following the manual at ["Extend the build process"](https://docs.readthedocs.io/en/stable/build-customization.html#extend-the-build-process) and trying to insert a simple shell script at the post_build step.

A script that works locally (Ubuntu 22) is somehow mangled and causes syntax errors when inserted into the .readthedocs.yml at the post_build stage.  I have tried many ways of writing the script and quoting it in YAML, to try to eliminate YAML as the culprit:

1. https://github.com/Ravenbrook/mps/pull/187/commits/4146715a1253a9ba7dcb8e528997545d52bb7ab2
2. https://github.com/Ravenbrook/mps/pull/187/commits/9fbc6224dbe1b645568fd32a938c77f84ef67d3a
3. https://github.com/Ravenbrook/mps/pull/187/commits/2a642594a525a3cbcd0d9c10f686bf6b6e43307c
4. https://github.com/Ravenbrook/mps/pull/187/commits/7fdbbd3ff01544484ba5a518b46c96dc9c98f1f3
5. https://github.com/Ravenbrook/mps/pull/187/commits/82f0d180a1bb0a7909d94eee22ab4a7b71a7dea3
6. https://github.com/Ravenbrook/mps/pull/187/commits/044e790f24217ad709dbc8ced9c35febaf02f0e9
7. https://github.com/Ravenbrook/mps/pull/187/commits/4a07b8418a4e5b1036dd50f13552098308972fb2
8. https://github.com/Ravenbrook/mps/pull/187/commits/1106c9cd066c9638ccf2f7772083384bb526673f

The [last one](https://github.com/Ravenbrook/mps/pull/187/commits/1106c9cd066c9638ccf2f7772083384bb526673f) is especially mysterious.

Is there some sort of mangling of the script going on, or am I just going blind?

None of these scripts produce syntax errors when I feed them to /bin/sh on my local Ubuntu.  I have written similar scripts for multiple CI systems without coming across this problem before.

Thanks.

humitos:
I'm not sure about _why_ we split the commands before sending it out our executor. The command is finally wrapper properly before being executed.

I did some locally QA after removing the split of the command and it seems these scripts work fine. So, it seems safe to move forward with this, explanding the flexibility of our config file.

The example provided by the user works fine.

Screenshot_2023-03-07_16-46-39

> **Note**
> This example is interpreted just as a comment (since it's just _one line_ after the split) without this PR and nothing is executed. With this changes, the comment is ignored and the code is executed. However, the script is properly shown as multi-line in the UI as well.

Closes #10103

### DIFF
--- a/readthedocs/doc_builder/director.py
+++ b/readthedocs/doc_builder/director.py
@@ -330,7 +330,7 @@ class BuildDirector:
 
         commands = getattr(self.data.config.build.jobs, job, [])
         for command in commands:
-            environment.run(*command.split(), escape_command=False, cwd=cwd)
+            environment.run(command, escape_command=False, cwd=cwd)
 
     def run_build_commands(self):
         reshim_commands = (
@@ -344,7 +344,7 @@ class BuildDirector:
         cwd = self.data.project.checkout_path(self.data.version.slug)
         environment = self.build_environment
         for command in self.data.config.build.commands:
-            environment.run(*command.split(), escape_command=False, cwd=cwd)
+            environment.run(command, escape_command=False, cwd=cwd)
 
             # Execute ``asdf reshim python`` if the user is installing a
             # package since the package may contain an executable
--- a/readthedocs/projects/tests/test_build_tasks.py
+++ b/readthedocs/projects/tests/test_build_tasks.py
@@ -916,10 +916,11 @@ class TestBuildTask(BuildEnvironmentBase):
 
         self.mocker.mocks["environment.run"].assert_has_calls(
             [
-                mock.call(
-                    "git", "fetch", "--unshallow", escape_command=False, cwd=mock.ANY
-                ),
-                mock.call("echo", "`date`", escape_command=False, cwd=mock.ANY),
+                # NOTE: when running commands from `build.jobs` or
+                # `build.commands` they are not split to allow multi-line
+                # scripts
+                mock.call("git fetch --unshallow", escape_command=False, cwd=mock.ANY),
+                mock.call("echo `date`", escape_command=False, cwd=mock.ANY),
             ],
             any_order=True,
         )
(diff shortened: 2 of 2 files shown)


stsewd (approved):

agjohnson (approved):
This seems okay I think, though I am also not 100% on why splitting would have been required. Normally, this is to pass arguments to subprocess.

I wouldn't be surprised if there are some projects negatively affected by this, but we should try this change either way. Command execution is tricky to get right universally.

rptb1:
> This seems okay I think, though I am also not 100% on why splitting would have been required. Normally, this is to pass arguments to subprocess.

I also do not see any reason for splitting.  By splitting commands with `string.split()` you are implementing a new tiny shell language where commands are parsed by splitting at spaces before being passed to [exec](https://linux.die.net/man/3/exec).  This violates the [principle of least astonishment](https://en.wikipedia.org/wiki/Principle_of_least_astonishment).  It definitely astonished me in #10103 :smile:  I was expecting text that looked like shell scripts to be shell scripts.

> I wouldn't be surprised if there are some projects negatively affected by this, but we should try this change either way.

I've been trying to think of a way you could check and provide some sort of warning.  You could perhaps run commands through [shlex](https://docs.python.org/3/library/shlex.html) *and* `split()`, compare the results, and print a warning if they're different. Not foolproof, but it might help people.

> Command execution is tricky to get right universally.

Leave it to the shell, is my advice.  Thoroughly well understood and debugged.

PR #10121, opened 2023-03-07
Proxito: add another model to cacheops

humitos:
We are hitting `projects_feature` table now. So, I'm adding this one to be cached as well.


ericholscher (approved):
Heh, funny that we keep finding additional queries :)

humitos:
I think this one was introduced in today's deploy since we are checking if the project should use the newer proxito implementation or not.

PR #10126, opened 2023-03-08
Build: check for `_build/html` directory and fail if exists

### LINKED ISSUE #10036: Placeholder: build output path issues

humitos:
Example of a failed build:

Screenshot_2023-03-08_10-32-09

Continuation of #10038
Closes #10036

### DIFF
--- a/readthedocs/doc_builder/exceptions.py
+++ b/readthedocs/doc_builder/exceptions.py
@@ -52,6 +52,13 @@ class BuildUserError(BuildBaseException):
         "and it is not currently supported. "
         'Please, remove all the files but the "{artifact_type}" your want to upload.'
     )
+    BUILD_OUTPUT_OLD_DIRECTORY_USED = gettext_noop(
+        "Your project is outputting files to an old directory ('_build/html') "
+        "when building the docs. "
+        "Note this directory is not used anymore and the result may be unexpected. "
+        "Please, check your build process looking for this directory and replace it by "
+        "the path '$READTHEDOCS_OUTPUT/html' instead."
+    )
     PDF_COMMAND_FAILED = gettext_noop(
         "PDF generation failed. "
         "The build log below contains information on what errors caused the failure."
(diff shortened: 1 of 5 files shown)


readthedocs/doc_builder/exceptions.py
```diff
@@ -52,6 +52,13 @@ class BuildUserError(BuildBaseException):
         "and it is not currently supported. "
         'Please, remove all the files but the "{artifact_type}" your want to upload.'
     )
+    BUILD_OUTPUT_OLD_DIRECTORY_USED = gettext_noop(
+        "Your project is outputting files to an old directory ('_build/html') "
+        "when building the docs. "
+        "Note this directory is not used anymore and the result may be unexpected. "
+        "Please, check your build process looking for this directory and replace it by "
+        "the path '$READTHEDOCS_OUTPUT/html' instead."
+    )
```
humitos, line 61: @agjohnson feel free to suggest a better message here.

benjaoming, line 61: Hi @humitos :wave: 

This will be a great help for anyone who is stuck here!

```suggestion
    BUILD_OUTPUT_OLD_DIRECTORY_USED = gettext_noop(
        "Your project is writing files to an unsupported directory ('_build/html')."
        "This directory is not used anymore and as a result, your documentation "
        "website will not be found and uploaded."
        "Please check your build commands and use "
        "the path '$READTHEDOCS_OUTPUT/html' instead of '_build/html'."
    )
```

What do you think about this? I was wondering that the result isn't merely "unexpected" but actually expected? :) I.e. the entire site won't be found and uploaded?

humitos:
Thanks for your suggestions. Both were better than mine 😅 -- I merged Anthony's one and I'm moving forward with this.

Issue #9423, opened 2022-07-10
Build: expose VCS-related environment variables


pradyunsg:
So... with `sphinx-basic-ng`, a bunch of the theme developers in the Sphinx ecosystem are trying to get a shared base theme that we all build upon.

Notably, this has reusable implementations of various design "components" (https://sphinx-basic-ng.readthedocs.io/en/latest/usage/components/). In a few of them[^1], 

```py
html_theme_options = {
    "source_repository": "https://github.com/pradyunsg/sphinx-basic-ng/",  # you can skip the trailing /
    "source_branch": "main",
    "source_directory": "docs",  # or "docs/"
}
```

It would be *great* if builds on Read the Docs could auto-populate these theme variables, when they're not provided by the user.

That would make the community-maintained themes that are using these components "just work" on Read the Docs, similar to how the Read the Docs theme works! Since this is designed to be a shared implementation across themes (Furo and Lutra already use it, there's efforts underway to get this for pydata-sphinx-theme and sphinx-book-theme as well), it's somewhat theme-agnostic and would be a nice user experience improvement!

Notably, it would allow themes to remove custom code for supporting VCS information conditionally on Read the Docs (Furo does this!).

[^1]: [edit-this-page](https://sphinx-basic-ng.readthedocs.io/en/latest/usage/components/edit-this-page/) and [view-this-page](https://sphinx-basic-ng.readthedocs.io/en/latest/usage/components/view-this-page/)

pradyunsg:
FWIW, this is (in effect) a replacement for https://github.com/readthedocs/readthedocs.org/blob/main/docs/user/guides/edit-source-links-sphinx.rst .

If you'd prefer to provide a fully-formatted link instead, that's coming in https://github.com/pradyunsg/sphinx-basic-ng/pull/34 -- which has the relevant documentation updates too. :)

humitos:
> It would be _great_ if builds on Read the Docs could auto-populate these theme variables, when they're not provided by the user.

Would it be useful if Read the Docs passes these as environment variables (e.g. `READTHEDOCS_SOURCE_REPOSITORY`, etc) and it's the theme itself that defines them to the Sphinx context? We are moving out of defining things automagically behind the scenes for the users because it creates confusing scenarios.

pradyunsg:
[...]

Issue #9063, opened 2022-03-31
Build: support build time injected logic via generic HTML injection


agjohnson:
In #8190 we described generalizing our build process with generalized HTML manipulation instead of Sphinx-specific solutions for injection.

I can't summarize here, so might leave this to @humitos to expand more. TBD

- [x] Flyout, CSS, and JS injection in HTML files (most of these are done in `readthedocs-sphinx-ext` at this file https://github.com/rtfd/readthedocs-sphinx-ext/blob/7cc1e60f7dcdeb7af35e3479509a621d5bac0976/readthedocs_ext/readthedocs.py)
- [x] Search indexing of HTML
- [x] Version warnings (https://github.com/rtfd/readthedocs-sphinx-ext/blob/7cc1e60f7dcdeb7af35e3479509a621d5bac0976/readthedocs_ext/versionwarning.py)
- [x] External version warnings (https://github.com/rtfd/readthedocs-sphinx-ext/blob/7cc1e60f7dcdeb7af35e3479509a621d5bac0976/readthedocs_ext/external_version_warning.py)
- [ ] Search as you type
- [ ] ???

This sounds complex enough that we should have two issues to track probably.

Refs #9062

agjohnson:
A few notes regarding this issue from a meeting today:

- We're implementing build.commands without a user contract file initially. The implementation will use convention for now
- We might still need a contract file for HTML upload to work
- We'll be discussing this more after an initial implementation of build.commands is usable, and we have an idea of what limitations users are hitting with a convention based approach.

humitos:
@ericholscher yes, I'd say that's the first step. We can later improve the integration if that makes sense, but communicating to people how to get the same result they get with the regular builder is a good step that gives users a solution without involving us in writing a lot of that integration yet.

I think we have two options here:

1. expose the `readthedocs-sphinx-ext` to users
   * this seems to be the simplest way
   * it only works with projects building with Sphinx
   * it will force us to maintain the extension to be used in multiple different ways that may not be prepared
2. document the "behind the scene" and let users handle/implement them
   * document what exactly each of the extension does (eg. add an HTML block at the top with this content, includes a particular JS and CSS for the flyuout, another for the Ads, etc)
   * this will help us to create the final input/output contract we want for builders
   * let users integrate this as they prefer, using the tools they feel comfortable with
   * work will multiple doctools

agjohnson:
[...]

PR #10127, opened 2023-03-08
Hosting: manual integrations via build contract

humitos:
Integration and design for the future of the Read the Docs features integrations.

Requires generating the `readthedocs-client.js` from (https://github.com/humitos/readthedocs-client) and copying to this repository and then collecting the statics:

```bash
cp ../readthedocs-client/dist/readthedocs-client.js readthedocs/core/static/core/js/readthedocs-client.js
inv docker.manage 'collectstatic --noinput'
```

## Goals

* Integrate all the Read the Docs readers' features (e.g. flyout, doc-diff, warning banners, etc) with any doctool.
* Remove the requirement to install _extra_ Python packages to integrate with us (e.g. `readthedocs-sphinx-ext`, `readthedocs-sphinx-search`)
* Generate the exact same output when running the doctool locally than on Read the Docs (except for conditionals using `READTHEDOCS` environment variables)
* Reduce complexity from the current hosting integrations (some happen inside the builder, others inside the Sphinx extension, others in the Sphinx theme, and others in the frontend)
* Move all the logic for these integrations to the frontend 
* Use modern JavaScript patterns to write all these integrations (see https://github.com/readthedocs/meta/discussions/114)

## Workflow

1. User triggers a build with any doctool (common process, `build.jobs` or custom commands via `build.commands`) [^1]
2. _(Optional)_ The doctool creates a [`$READTHEDOCS_OUTPUT/html/readthedocs-build.yaml`](https://github.com/rtfd/test-builds/blob/manual-integrations/readthedocs-build.yaml) file at build time with a pretty known structure (the "build contract" that we need to define, but we have an idea of what fields are required already) [^2]
3. Before finishing the build process, [the builder reads `readthedocs-build.yaml` and stores its data into the `Version` object](https://github.com/readthedocs/readthedocs.org/blob/53366aa370a4b916bb9b14bb51a991f6c5846685/readthedocs/doc_builder/director.py#L632-L660). This will be used to communicate data from the builder to the frontend.
4. El Proxito injects, at serving time [^3], the Read the Docs javascript client (`readthedocs-client.js`) 
5. The [`readthedocs-client.js`](https://github.com/humitos/readthedocs-client/tree/main/src/index.js) file loads all the extra required JS and CSS files dynamically.
6. These javascript files use an API call (`/_/readthedocs-config.json`) to get all the config required to perform the hosting integrations. The data generated at build time via `readthedocs-build.yaml` is also returned by this endpoint. Note that some of this "hosting integrations" are already implemented in our Sphinx extensions and they have to be re-written into Javascript. I migrated some basic one for now and put them into [the client repository](https://github.com/humitos/readthedocs-client/tree/main/src) as an example.
7. These integrations are: version warning, pull request warning, search, doc-diff, hoverxref, ethicalads, search/traffic analytics, etc

## Integration points from a user's perspective

Nothing.

_Optionally_, the user could configure the build process to generate the `readthedocs-build.yaml` at build time with a specific structure (_to be defined_) to communicate some extra data that's _only known at build time_.

## v2 and future projections

* Make each hosting feature to be enabled/disabled at build time
* Make each hosting feature to be enabled/disabled by the reader
* Allow override the HTML templates injected (e.g. those used to generate the warning banners)
* Allow override the _logic_ used to inject those HTML blobs (e.g. use a different function to decide if the warning banner has to be injected or not --useful when not using `semver`)
* Validate `readthedocs-build.yaml` and enforce a specific structure

----

### Related
* https://github.com/readthedocs/test-builds/tree/manual-integrations
* https://github.com/readthedocs/test-builds/tree/manual-integrations-docsify
* https://github.com/readthedocs/readthedocs.org/pull/9755
* https://github.com/readthedocs/readthedocs.org/issues/9063
* https://github.com/readthedocs/meta/issues/71
* https://github.com/readthedocs/readthedocs.org/pull/8052
* https://github.com/readthedocs/meta/issues/96
* https://github.com/readthedocs/readthedocs.org/issues/7748
* https://github.com/readthedocs/meta/issues/85
* https://github.com/readthedocs/sphinx_rtd_theme/pull/1445

### External related issues
* https://github.com/python/docs-community/issues/4
* https://github.com/pradyunsg/sphinx-basic-ng/issues/12
* https://github.com/open-contracting/deploy/issues/266

[^1]: Read the Docs pass data to the doctool via environment variables. Exactly as we are doing now. However, we may want to add some extra variables (see #9423)
[^2]: This YAML file is the way for the doctool to pass back data to Read the Docs and configure its integrations
[^3]: [It's done at NGINX for now](https://github.com/readthedocs/readthedocs.org/blob/humitos/hosting-integrations/dockerfiles/nginx/proxito.conf.template#L92), but it can be feature flagged and done at CloudFlare


dockerfiles/nginx/proxito.conf.template
```diff
@@ -88,6 +88,13 @@ server {
...
         set $content_security_policy $upstream_http_content_security_policy;
         add_header Content-Security-Policy $content_security_policy always;
+
+        # Inject our own script dynamically
+        # TODO: decide where is the best place to do this
+        sub_filter '</head>' '<script language="javascript" src="http://devthedocs.org/static/core/js/integrations.js"></script>\n</head>';
+        # sub_filter_types text/html;
+        sub_filter_last_modified on;
+        sub_filter_once on;
     }
```
humitos, line 98: This code automatically injects our `integrations.js` file.

ericholscher:
This is a really interesting example. I like how it basically just duct tapes the pieces together, and doesn't require any user interaction. There's a bunch of design discussions downstream of this, but just having an example of how this might work is great. It's a lot less hacky if we ship `integrations.js` to the users and themes, and they optionally integrate it. 

Let's chat more about this tomorrow, since I don't know what level of feedback is particularly useful at this stage, but I think we can get pretty close to something we could share with users as a demo at least.

humitos:
> It's a lot less hacky if we ship `integrations.js` to the users and themes, and they optionally integrate it.

We touched on this in our call already, but I want to make my position clear about this here. IMO, we have to _force_ the injection of our own JavaScript file. Otherwise, many projects won't include our JavaScript and most of the good features we offer won't be present in their documentation. Even people wanting to integrate with us, will have a bigger barrier to do it. In particular, commercial customers. Besides, on the community side, most projects won't have ads --which is not a minor.

I think if we don't inject the JavaScript automatically, we won't differentiate ourselves from GitHub Pages or any other plain-text hosting, I'd say.

humitos:
I updated the description of this PR with a lot more context and a more polished idea of how all these pieces fit together.

agjohnson:
Overall, this implementation looks great, especially considering the boundaries on goals. I think it's helpful to get some experimentation here on the backend implementation, and have some direction at a high level on _how we'll inject JS_.

*All* of my questions, and all the places I want to put a bit more thought on patterns and API design are in your v2 list. The goals seem great, and for that, I'd 100% skip the V2 parts of the JS implementation -- HTML templates, API overrides, etc. This is where I'll have the most technical input, there are features from Webpack we'll use here, and extra libraries we'll rely on.

----

It's still fuzzy to me *why* a `readthedocs-build.yaml` file is required for this. I'm not against this design, it's just not clear to me. I probably haven't moved much here since https://github.com/readthedocs/readthedocs.org/issues/9088

Do you have a better idea now of what configuration options we need to be in an contract file?

The features list in the contract file stand out still -- and I'm just talking in context of a contract file, not feature enabling/disabling API design or patterns for the next iteration of this work. A contract file may help with reproducible builds somewhat, but there's also very real cases where configuring features at the request level will be much better UX:

- How would an author disable search/docdiff/etc across multiple versions? They update an option in the admin UI and rebuild all of their versions?
- If this is controlled in a version controlled file, how does an author disable search/docdiff/etc on tagged versions? They have to alter the tags and rebuild all tag versions?

I see some benefit to controlling features at the request level -- that is, the same data structure, but included in the documentation via a `https://docs.example.com/_/features.json` type file that is dynamically generated.

> IMO, we have to force the injection of our own JavaScript file

I'd agree, and I think we can very easily keep the footprint of this file very minimal as well.

readthedocs/core/static/core/js/integrations.js
```diff
@@ -0,0 +1,44 @@
...
+
+        let analytics = document.createElement("script");
+        analytics.setAttribute("async", "async");
+        analytics.setAttribute("src", "/_/static/javascript/readthedocs-analytics.js");
+        document.head.appendChild(analytics);
+
+        let hosting = document.createElement("script");
+        hosting.setAttribute("async", "async");
+        hosting.setAttribute("src", "/_/static/core/js/readthedocs-hosting.js");
+        document.head.appendChild(hosting);
```
agjohnson, line 35: For later, but just so you're aware, this import pattern will change a good deal with this library constructed as a proper module and using Webpack. For experimentation it's absolutely fine though.

But just so you're in the loop, all features should be versioned Node packages, and this package will install these through standard NPM tooling. Webpack resolves these directly for us in code, and loads them asynchronously.

Here is an example:

- [The code that loads ansijs and sanitizehtml packages](https://github.com/readthedocs/ext-theme/blob/main/src/js/build/detail.js#L150-L180)
- [Where the packages are actually installed as dependencies](https://github.com/readthedocs/ext-theme/blob/main/package.json#L24-L44)

readthedocs/doc_builder/templates/doc_builder/readthedocs-data.html
```diff
@@ -0,0 +1,6 @@
+{{ data|json_script:"READTHEDOCS_DATA" }}
```
agjohnson, line 1: If this file is just JSON, and we're consuming it by a request, does it need to be HTML? I'll note above what I'm a bit confused by

readthedocs/core/static/core/js/integrations.js
```diff
@@ -0,0 +1,44 @@
+// Unique entry point that our servers will inject
+
+// Load Read the Docs data first
+// This data is generated at build time by the doctool
+fetch("/en/manual-integration-docsify/readthedocs-data.html", {method: 'GET'})
```
agjohnson, line 5: Noted below, but we could probably just use JSON here.

Also, [to expand on below](https://github.com/readthedocs/readthedocs.org/pull/10127/files#r1132952089), ultimately, these features won't be loaded as _global scripts_ [^1], but as modules in this package.

To illustrate, without getting into asynchronous imports:

```js
import { DocDiff } from 'readthedocs-docdiff';
import { DocEmbed } from 'readthedocs-embed';

fetch("/blah/readthedocs-data.json")
   ...
  .then((readthedocs_data) => {
    const doc_diff = new DocDiff(readthedocs_data);
    doc_diff.enable();
    const doc_embed = new DocEmbed(readthedocs_data);
    doc_embed.enable();
    ...
  }
}
```
   
[^1]: Global scripts are loaded on to window/global, and we have less control over what these are. Instead, our package should isolate itself _and all dependencies_ in a bundle.

readthedocs/doc_builder/templates/doc_builder/readthedocs-data.html
humitos: No, it does not need to be HTML. I'm doing it like this here just because I quickly found that `json_script` was the easiest way to generate a JSON object. However, that template filter wraps the object with a `<script>` HTML tag.

The only thing I wanted to show with this file was the concept of "how to consume data generated at build time from the frontend".

readthedocs/core/static/core/js/integrations.js
humitos: Yeah, I assumed that something like that would be the right pattern to do. However, I didn't want to jump into those concepts because I have no knowledge on how to do it correctly. I just wanted to put everything together here to have a PoC that I can preset to the team to discuss over something more concrete than abstracts ideas in the air.

Follow the pattern you are described is part of my goals, but I wanted to start there yet without having the exact workflow and data flow defined yet.

humitos:
@agjohnson 

> It's still fuzzy to me _why_ a `readthedocs-build.yaml` file is required for this. I'm not against this design, it's just not clear to me. I probably haven't moved much here since #9088
> Do you have a better idea now of what configuration options we need to be in an contract file?

This space is reserved for all the configuration that _only the doctool knows at build time_ (or that needs to be generated dynamically). I understand why you are not finding _required_ and that's probably because all the data that I put as example there is not _mandatory_.

We need more feedback from other doctools here to find out if there are things we strictly require from them. I don't have that information yet and we should keep trying out other doctools as I was doing previously and keeping this in mind.

For now, I refactored a little the `readthedocs-build.yaml` proposed to have only two main keys:

1. `telemetry`: contains all the data the doctool can share with us for telemetry purposes [^1]
2. `features`: communicate us what are the features the author wants to integrate with us [^2]

> I see some benefit to controlling features at the request level and configuring them in the admin dashboard. We'd use the same data structure, but included in the documentation via a `https://docs.example.com/_/features.json` type file that is dynamically generated.


@ericholscher and I briefly touched on this idea on Thursday. I thought it could be a good idea to have the ability to decide things dynamically at request time. However, I didn't come up with good arguments while talking about this. Also, Eric mentioned it was simpler to just serve a `.json` file directly from S3, which was a good argument and we just moved on 😄 .

I think the questions you are rising here are important and may make us to use an endpoint as you are suggesting. If that's the case tho, all the `feature` key to decide the integrations don't make sense to be in the YAML file anymore and they would become db fields (at `Project` or `Version` level). That would be another _long conversation_ to have that I'm not sure if it makes a lot of sense at this immediate time. I'm happy to have it, tho, but I don't want to detour too much from the other bigger concepts and be able to start moving forward making the first step of the plan.

----

That said, I understand that you agree with 80% of this proposal _--without considering the specific JavaScript technical aspects_. That missing 20% would be:

- reduce complexity by skipping the `readthedocs-build.yaml` for now until we find it's strictly required
- make the data coming dynamically from an endpoint under `/_/` at docs domain instead of just a file in S3
- decide _how_ we will inject the JS (at build time on each HTML, NGINX, CloudFlare, etc)

[^1]: I know you said "we'd have to consider if we want community provided telemetry even" in https://github.com/readthedocs/readthedocs.org/issues/9088#issuecomment-1331246443. We can skip it for now if we want to, but probably worth to start having that conversation at least.

[^2]: you said this could be moved directly into the `.readthedocs.yaml`; which I see it totally doable

humitos:
@agjohnson 

> I see some benefit to controlling features at the request level and configuring them in the admin dashboard. We'd use the same data structure, but included in the documentation via a `https://docs.example.com/_/features.json` type file that is dynamically generated.

Thinking a bit more about this and after writing a comment in https://github.com/pradyunsg/sphinx-basic-ng/issues/12#issuecomment-1464907404, this endpoint could probably return some data (e.g. versions, downloads, urls on Read the Docs, etc) that will be used in the future to generate/customize/integrate the flyout exactly as our users want.

humitos:
I pushed some changes (see https://github.com/readthedocs/readthedocs.org/pull/10127/commits/d5130cc19dd8b7823532d592dc934674006477c4) that:
- make usage of `/_/readthedocs-config.json` API endpoint instead of a `.json` file in S3
- use the [`readthedocs-client.js` library](https://github.com/humitos/readthedocs-client) that I started working on following some of the new JavaScript patterns

The initial test that injects the "external version warning" worked great! 💪🏼 . Honestly, I think this approach is pretty clean and I'm happy to continue working on it.

readthedocs/core/static/core/js/integrations.js
agjohnson: Yup, absolutely agree. What you have here is totally fine for experimenting. Altering the loading pattern will change some code, but not too drastically.

agjohnson:
> we should keep trying out other doctools as I was doing previously

Yes yes yes! I was really excited to see experimentation here and I feel we may have knocked you off course a bit on this work when we got to unfamiliar tooling. I still feel that getting experience from the user perspective is really important for us here, even if it leads to some dead ends.

> That said, I understand that you agree with 80% of this proposal

Probably a bit more than that even, I was just unsure on the contract file specifics. I'd agree that if we don't yet know what we need from this file, we just keep the implementation minimal and probably treat it more as a proof of concept. This is totally reasonable.

> this endpoint could probably return some data

Hrm, great point. I didn't put it together that I was basically describing this API. :thinking:

humitos:
I updated this PR to make usage of:

- `/_/readthedocs-config.json` API endpoint
- inject [`readthedocs-client.js`](https://github.com/humitos/readthedocs-client) file. This file is now generated using webpack and all the nice features from JavaScript 😄

humitos:
Today I expanded this work with:

- use our EthicalAd client directly by importing it as a module instead of loading as a `<script>` [^1]
- record `readthedocs-build.yaml` at build time in the database and use it to update the data returned at `/_/readthedocs-config.json`. This allows the doctool to communicate data to the frontend in case it's required. This still needs some conversation and decisions.
- generate the flyout on the client by using a HTML template [^2] and all the data returned by `/_/readthedocs-config.json`. This would allow users to opt-out from our flyout integration and generate their own.

[^1]: our own `ethica-ad-client` library needs some work to allow to be used as a module. There are some work done there already, but is has some bugs that I'm not sure how to solve yet 😅 
[^2]: currently follows what we have documented at https://docs.readthedocs.io/en/stable/flyout-menu.html

ericholscher:
> > I see some benefit to controlling features at the request level and configuring them in the admin dashboard. We'd use the same data structure, but included in the documentation via a `https://docs.example.com/_/features.json` type file that is dynamically generated.
> 
> @ericholscher and I briefly touched on this idea on Thursday. I thought it could be a good idea to have the ability to decide things dynamically at request time. However, I didn't come up with good arguments while talking about this. Also, Eric mentioned it was simpler to just serve a `.json` file directly from S3, which was a good argument and we just moved on 😄 .
> 
> I think the questions you are rising here are important and may make us to use an endpoint as you are suggesting. If that's the case tho, all the `feature` key to decide the integrations don't make sense to be in the YAML file anymore and they would become db fields (at `Project` or `Version` level). That would be another _long conversation_ to have that I'm not sure if it makes a lot of sense at this immediate time. I'm happy to have it, tho, but I don't want to detour too much from the other bigger concepts and be able to start moving forward making the first step of the plan.
> 
> That said, I understand that you agree with 80% of this proposal _--without considering the specific JavaScript technical aspects_. That missing 20% would be:
> 
> * reduce complexity by skipping the `readthedocs-build.yaml` for now until we find it's strictly required
> * make the data coming dynamically from an endpoint under `/_/` at docs domain instead of just a file in S3
> * decide _how_ we will inject the JS (at build time on each HTML, NGINX, CloudFlare, etc)
> 
> ## Footnotes
> 1. I know you said "we'd have to consider if we want community provided telemetry even" in https://github.com/readthedocs/readthedocs.org/issues/9088#issuecomment-1331246443. We can skip it for now if we want to, but probably worth to start having that conversation at least. [↩](#user-content-fnref-1-09386f6f8ef56942e59adb105211e800)
> 2. you said this could be moved directly into the `.readthedocs.yaml`; which I see it totally doable [↩](#user-content-fnref-2-09386f6f8ef56942e59adb105211e800)

I think this is a really interesting line of conversation. I think about it this way:

There are a couple different things we're trying to define here:

* When are features enabled/disabled?
* When are defaults defined?

The *reader* should have ultimate control over *enabling features*. The flyout or custom theme integration is where real-time enable/disable functionality should live, and give the reader full control over most features. We will need to consider if there are some features that users can't disable (eg. version warning, since authors really want users to know they are on an unsupported version)

For an example here, I want a little toggle I can enable/disable to show doc diffing on a PR. By default this would go into the flyout, but themes could make something nicer to integrate it into the UX itself.

The *project admin/author* should have control over *default on/off configuration*. This means that doc diffing should default on/off based on what the author thinks is best, and has configured on the project, but the reader should be able to still enable them if they want to be able to use them. This seemingly could be done at both *build* and *serve* time.

The *theme authors/integrators* should be able to *support* various features, but not have control over defaults or enabling. If something isn't supported by the theme, it should be integrated via the flyout.

I'm not 100% sure this is correct, because we need to make sure doc diffing, in this example, doesn't break the UX/design of the page. Taking control away from the authors, and giving it to the readers, makes ourselves as the platform more responsible to ensure that these features work well across all pages. Is this a responsibility that we want? 

I think providing a consistent set of valuable features to readers is the most important way we can grow and provide value as a platform, and not require authors to integrate/enable features that would be valuable to users. If something is off by default, we still need to do the work via the flyout to educate the user on the feature. That is going to be really hard, but once users know about it they will value it. 

I think the understanding of how these features are configured, enabled, and shipped is going to be super important around our design of the architecture here, which is why I'm trying to consider them here.

humitos:
@ericholscher 

> The _reader_ should have ultimate control over _enabling features_.

Yes. I think the same.

> The _project admin/author_ should have control over _default on/off configuration_.

Agreed.

> The _theme authors/integrators_ should be able to _support_ and integrate various features, but not have control over defaults or enabling. If something isn't supported by the theme, it should be integrated via the flyout

Exactly, but note theme authors will always have full control over defaults and/or enabling, because they run code on behalf you 😄 

All of this is how I've been thinking about these integrations. That's why I decided to use an API endpoint and generate all the injected HTML code in the client via templates. These changes are going to be possible without a ton of work:

* change the style of the flyout via CSS
* replace completely the HTML template for the flyout [^1]
* decide where to inject the flyout by defining `#readthedocs-embed-flyout` (https://docs.readthedocs.io/en/stable/flyout-menu.html)
* remove a chunk of the HTML template to remove an integration from it (e.g. doc-diff) and natively integrate it in the theme
* disable the flyout completely and natively integrate it in the theme [^1]

People wanting to integrate _natively_ with us will have the exact same data we are using to generate our flyout. All this data comes from the `/_/readthedocs-config.json` API endpoint. Theme authors will be able to do what they consider best to integrate with us.

---

> If something isn't supported by the theme, it should be integrated via the flyout. If the doc diff UX is broken on a theme, who's responsibility is it to fix it? 
> I'm not 100% sure this is correct, because we need to make sure doc diffing, in this example, doesn't break the UX/design of the page. Taking control away from the authors, and giving it to the readers, makes ourselves as the platform more responsible to ensure that these features work well across all pages. Is this a responsibility that we want?

I don't fully agree here. I don't think we are 100% responsible for all the features we are exposing to work flawlessly with all the type of HTML pages there are out there. We are _defining a contract_ for authors/theme developers _to integrate with us_, not the other way around. **We cannot guess the HTML structure of every page and integrate properly with that**.

On the other hand, I think doc-diff is a bad example at this point because it's a beta project that we are not exposing to anyone yet. So, it has 0 exposure to real projects, other than our docs.

IMO, _we are responsible of the technical integration itself_ (e.g. hitting this endpoint will return this data, calling this JavaScript function will perform this DOM diff and add these CSS classes, this HTML code will be prepended when the warning version is executed, etc)  --but _we can't be responsible for the style integration of it_.

If we want to handle cases similar to "As a reader, when I enable doc-diff in this project, it completely breaks" we should consider giving authors an extra power on this decision: disable it completely and disallow readers to enable it. In that case, we will have 3 options [^2]:

* enabled by default (readers can turn it off)
* disabled by default (readers can turn it on)
* _disabled completely_ (readers can't turn it on) -- the author already knows the project is not integrated properly with this feature and don't want to expose it to their users.


I think this conversation _is important_, but I don't think it will dictate the technical implementation. I'd like to not get blocked on this if possible. I think we should move forward replicating the integrations we already have and expanding to those projects that are not currently supported yet (those using `build.commands`, for example). Currently, we have Flyout, Search As You Type, non-latest version warning, external version warning, Analytics and EthicalAds already integrated there 👍🏼 

[^1]: we will need some policy here to avoid misuse of our platform --but this conversation is outside the scope of this implementation for now.
[^2]: we can think on a 4th one as well: "completely enabled", for features like the version warning as you mentioned.

humitos:
This is the progress I have so far, which supports:

* Non-latest version warning banner
* Flyout
* search as you type
* hoverxref
* EthicalAds

hosting-integrations


At this point, it emulates what happens on Read the Docs, but it requires nothing from the user. Everything is integrated automatically without having to install any external dependency at build time or similar. All the features are migrated to the frontend.

humitos:
I'm marking this ready for review because my exploration here has finished. We have everything in place and it's already working. The next steps is to get some feedback from reviewers and continue the discussions to polish this implementation.

agjohnson:
I agree with most of the above. Echoing what we've talked about in search, I'm leaning into a pattern where readers have granular control over features, but authors dictate the default configuration and state. Sounds like we all agree here.

> If something isn't supported by the theme, it should be integrated via the flyout.

I'd probably go a bit further and say that the flyout should always have the same default UI/UX, respecting author configuration of course. Theme/tooling maintainers wouldn't move UI out of the flyout, but would implement it in addition to the flyout UI. This is mostly because we want to expose UI to readers (like toggling doc-diff), and that should have some consistency.

There is a conversation about adding more floating UI elements outside the flyout too though.

> The theme authors/integrators should be able to support and integrate various features, but not have control over defaults or enabling.

However, we probably do want to consider theme/tooling maintainer input as well.

I do see how theme/tooling maintainers might want to define configuration defaults for features though. For example on `doc-diff`, tool/theme maintainers have a better idea of what selector to use for the main content pane used for diffing. We probably want to allow this configuration to come from theme/tooling maintainers, overridden by authors, and the feature to be toggled by readers.

----

It's worth considering this when we get to constructing API override and configuration patterns in the next iteration of this work. Obviously for a private beta we won't get much tool maintainer input, but I doubt we'll see interest in this level of collaboration from other tool maintainers for a while. We're most likely to be working with Sphinx theme maintainers on these patterns next.

We'll have some more collaboration to do with tool maintainers, and I think it's probably wise to assume we'll be the ones maintaining tool-specific implementations/configuration while we're seeding interest.

ericholscher:
Cool -- glad we're all mostly aligned here. I think it does make sense to discuss responsibility while we're building the data flows, since where the data comes from is part of who is responsible for configuring it. Given that it seems we're all 👍 on the high-level, we can move on. 🚀

dockerfiles/nginx/proxito.conf.template
```diff
@@ -88,6 +88,18 @@ server {
         proxy_hide_header Content-Security-Policy;
         set $content_security_policy $upstream_http_content_security_policy;
         add_header Content-Security-Policy $content_security_policy always;
+
+        # This header allows us to decide whether or not inject the script at CloudFlare level
+        # Now, I'm injecting it in all the NGINX responses because `sub_filter` is not allowed inside an `if` statement.
+        set $rtd_hosting_integrations $upstream_http_x_rtd_hosting_integrations;
+        add_header X-RTD-Hosting-Integrations $rtd_hosting_integrations always;
```
ericholscher, line 95: I think we need this on `ops` as well, when we deploy this.

readthedocs/core/static/core/js/readthedocs-client.js
ericholscher, line 1: Do we want to ship this with the app, or version it directly on S3, like we do with the ad client? I think keeping it deployable outside of the application seems right to me. 

https://github.com/readthedocs/ethical-ad-client/tags

https://media.ethicalads.io/media/client/v1.4.0/ethicalads.min.js

readthedocs/doc_builder/director.py
```diff
@@ -625,3 +628,33 @@ def get_build_env_vars(self):
...
+                version=self.data.version.slug, type_="html"
+            ),
+            "readthedocs-build.yaml",
+        )
+
+        try:
+            with open(yaml_path, "r") as f:
+                data = yaml.safe_load(f)
+        except Exception:
+            # NOTE: skip this work for now until we decide whether or not this
```
ericholscher, line 645: I think we want to check if the file exists, instead of catching exceptions on read? 

I think to start, we definitely don't want it to be required, so let's just change this logic to only read it when it exists.

readthedocs/doc_builder/director.py
```diff
@@ -625,3 +628,33 @@ def get_build_env_vars(self):
...
+
+        log.info("readthedocs-build.yaml loaded.", path=yaml_path)
+
+        # TODO: validate the YAML generated by the user
+        # self._validate_readthedocs_build_yaml(data)
+
+        # Copy the YAML data into `Version.build_data`.
+        # It will be saved when the API is hit.
+        # This data will be used by the `/_/readthedocs-config.json` API endpoint.
+        self.data.version.build_data = data
```
ericholscher, line 660: Do we want to actually parse the data into memory, or just store the file contents directly as a string? I think we just want to store a string to start?

readthedocs/proxito/middleware.py
```diff
@@ -278,9 +279,19 @@ def process_request(self, request):  # noqa
...
         return None
 
+    def add_hosting_integrations_headers(self, request, response):
+        project_slug = getattr(request, "path_project_slug", "")
+        if project_slug:
+            project = Project.objects.get(slug=project_slug)
+            if project.has_feature(Feature.HOSTING_INTEGRATIONS):
+                response["X-RTD-Hosting-Integrations"] = "true"
+            else:
+                response["X-RTD-Hosting-Integrations"] = "false"
```
ericholscher, line 289:
```suggestion
            if project.has_feature(Feature.HOSTING_INTEGRATIONS):
                response["X-RTD-Hosting-Integrations"] = "true"
```

I think we just don't want the header when it's off?

readthedocs/proxito/views/hosting.py
```diff
@@ -0,0 +1,103 @@
...
+class ReadTheDocsConfigJson(CDNCacheControlMixin, View):
+    def get(self, request):
+
+        unresolved_domain = request.unresolved_domain
+        project = unresolved_domain.project
+
+        # TODO: why the UnresolvedURL object is not injected in the `request` by the middleware.
+        # Is is fine to calculate it here?
+        unresolved_url = unresolver.unresolve_url(request.headers.get("Referer"))
+        version = unresolved_url.version
```
ericholscher, line 24: @stsewd seems like one for you. 

@humitos I see we're using the referer here. Is this because we don't want to pass in the domain explicitly as a GET arg? I think we likely want to actually use the `request.host` here, instead of a header that the browser is setting.

readthedocs/proxito/views/hosting.py
```diff
@@ -0,0 +1,103 @@
...
+
+        # TODO: use Referrer header or GET arguments for Version / Build
+        project.get_default_version()
+        build = version.builds.last()
+
+        # TODO: define how it will be the exact JSON object returned here
+        # NOTE: we could use the APIv3 serializers for some of these objects if we want to keep consistency.
+        # However, those may require some extra db calls that we probably want to avoid.
+        data = {
```
ericholscher, line 33: It seems like we're re-creating the footer API, but with JSON instead of HTML. That seems reasonable, but I do think we'll want to think more about what this API is returning.

It seems like we actually want the client to be able to tell us exactly what data points they want? Something similar to what we're doing with apiv3 with the `expand` and `fields` [parameters](https://docs.readthedocs.io/en/stable/api/v3.html#resources)? That way we give the caller control over how much data they want...

That might be an over-optimization at this stage, but I think we want to design this with expandability and performance in mind.

ericholscher (commented):
This is a really simple, beautiful implementation. I think we can trim down the amount in this PR already, and get something shipped.

dockerfiles/nginx/proxito.conf.template
humitos: Good point. Opened at https://github.com/readthedocs/readthedocs-ops/pull/1319

readthedocs/core/static/core/js/readthedocs-client.js
humitos: I haven't thought too much about this and I don't have experience deploying the script with a different process than the normal one. I'm not sure about the pros/cons here.

I put this file here because we need it for development as well. We could just put this file in the local MinIO S3, tho, as well.

Note this file is generated by running `npm run build` _from another repository_.

humitos: We need to also think about _what version_ are we going to serve by default. The latest one? Would people be able to pin to a particular version? Are we going to support multiple versions at the same time? How do we deploy new features to those that are pinned to an older version? Do we care? Too many questions 🤷🏼

readthedocs/doc_builder/director.py
humitos: > Do we want to actually parse the data into memory, or just store the file contents directly as a string? I think we just want to store a string to start? I'd like to avoid as much YAML parsing as possible...

I think we will need to parse it so we can validate it at some point, anyways.

This also allows us to use a JSON field in the database that we can query in the future, looking for answers.

> I also wonder if we should make this file JSON, instead of YAML? If the goal is for it to be aligned with the JSON data returned via the API, I think that makes more sense. But if it's closer to our .readthedocs.yaml config, then YAML makes sense thinking

I decide to use YAML here on purpose. It's a lot easier to write than JSON, a lot less nit picking (e.g. requires no trailing comma in the last element of a list), supports comments, works better with more data types, and others. The structure is going to be just a dictionary, YAML is going to be just the representation/serialization of it

In particular, being able to put comments in `.readthedocs.yaml` is important for ourselves and our users as well. That was one of the reasons why I picked YAML for this file as well. Otherwise, you end up with things like this in JSON: https://github.com/readthedocs/readthedocs.org/blob/48de5977f1626315cd6431cad4f3f9af1c346ed9/readthedocs/proxito/views/hosting.py#L36-L39

readthedocs/proxito/middleware.py
humitos: Makes sense 👍🏼

readthedocs/proxito/views/hosting.py
humitos: > @humitos I see we're using the referer here. Is this because we don't want to pass in the domain explicitly as a GET arg? I think we likely want to actually use the request.host here, instead of a header that the browser is setting.

Not strong preference. We can use something different here. I used the `Referer` because it was what I had at hand and easy to get.

A `url=` GET attribute probably seems fine as well, since it's what we are using in other API endpoints (analytics and embed api, for example). Referer also makes testing harder since "copying and pasting a URL doesn't work" --I've ended up hardcoding the URL in the code because of this 😅 

I think we should combine both, `url=` GET attribute with a validation with `request.host` on that URL.

readthedocs/proxito/views/hosting.py
humitos: > I do think we'll want to think more about what this API is returning.

Yes. This is one part of the design decisions I mentioned we have to sit down, talk together, think about and decide what's the best structure.

> It seems like we actually want the client to be able to tell us exactly what data points they want? Something similar to what we're doing with apiv3 with the expand and fields [parameters](https://docs.readthedocs.io/en/stable/api/v3.html#resources)? That way we give the caller control over how much data they want...

This is a good point to start considering and thinking about, yeah. However, I don't think we will require that complexity because:

* the API endpoint will be cached
* most of the values are static
* there are a few db queries only (`Project` and `Project.versions(active=True)`)
* we could decide whether or not return some specific data based on the feature being enabled. If `non_latest_version_warning` is _disabled_, we don't calculate/query nor return the list of active versions this feature consumes in the client.

In any case, I think we will arrive at the best answer by exploration, experimentation and brainstorming. Also, once we deploy this, we will have some data in NR to know how much time this endpoint takes on a real production db.

readthedocs/core/static/core/js/readthedocs-client.js
ericholscher: I think we always deploy the latest, but if the goal is for _other people_ to integrate the library, then it can be versioned based on their needs. The versioning is primarily valuable for strict SRI protection, like PyPI does, to validate the hash of the library hasn't changed for security reasons.

I think we should definitely deploy the client outside of our application. We don't need to decide on a proper deployment pattern yet though, but I think we should keep it out of the application from the start though.

readthedocs/doc_builder/director.py
ericholscher: Yea.. let's use YAML for anything human-writable, and JSON for machine-writable 👍

readthedocs/proxito/views/hosting.py
ericholscher: To be clear, I'm fine shipping this as a test for now. 

But I do think over time if this is the "one endpoint to get all the data you need in the page", there will be demands on this functionality that are resource intensive, vary by user, etc. That is where we'll start to get value from the API approach.

humitos: Good point about the .json name in the URL endpoint. We should definitely change that as first step.

I will think a little more about the expand argument. I don't have a strong opinion on that yet 😅

readthedocs/doc_builder/director.py
```diff
@@ -638,6 +638,10 @@ def store_readthedocs_build_yaml(self):
             "readthedocs-build.yaml",
         )
 
+        if not os.path.exists(yaml_path):
+            log.debug("Build output YAML file (readtehdocs-build.yaml) does not exist.")
+            return
```
ericholscher, line 643: 💯

ericholscher (approved):
I'm good to ship this initial experimentation next week so we can play with it in production.

Alternatively, we could just use defaults in the `readthedocs-client`, and not require a server response for initial testing? We could also just return some JSON data via a CF Worker to make it faster to iterate with this, instead of requiring a full Django deploy.

readthedocs/proxito/views/hosting.py
```diff
@@ -68,12 +77,28 @@ def get(self, request):
...
-                "doc_diff": True,
+                "doc_diff": {
+                    "enabled": True,
+                    # "http://test-builds-local.devthedocs.org/en/latest/index.html"
+                    "base_url": f"""{resolver.resolve(
+                        project=project,
+                        version_slug=project.get_default_version(),
+                        language=project.language,
+                        filename=unresolved_url.filename,
+                    )}""",
```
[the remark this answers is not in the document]
humitos: he he, yeah. This is terrible code 😝 -- It was on a f-string because it used to have other replacement in the string itself, but now it's just one function. So, the f-string is not required.

I think I will create helper methods to calculate these stuffs so it's more clear where they come from and easy to read. There is lot to iterate over in this code 😅

humitos:
> Alternatively, we could just use defaults in the `readthedocs-client`, and not require a server response for initial testing? 

I'd prefer if we can get this from the server itself. This will also help me to understand how much time this endpoint takes to return via NR and also confirm that everything is on it's place.

> We could also just return some JSON data via a CF Worker to make it faster to iterate with this, instead of requiring a full Django deploy.

This is a good option in case we need it, yeah. I'm not sure how many times we will be changing this response. It's definitely gonna change, and _big_, probably. But maybe it changes once, it's done. The main changes are gonna happen in the js client, without requiring too much changes in the API response, I'd guess.

In any case, we can keep doing all the test of these stuffs locally while developing. We can also deploy more often if required 😉

ericholscher:
> In any case, we can keep doing all the test of these stuffs locally while developing. We can also deploy more often if required 😉

But then I can't play with it :) I guess there's always ngrok, or I just get it running locally :)

humitos:
He he. Note that you can add/update data from the build process itself, by creating a YAML file 😉. That will give us some extra control to perform changes in the API response without deploying 😏

humitos:
So, we can change the js client by just uploading a new js file to S3.

We can change the API response by generating a YAML file in the build process

humitos:
Latest updates:

- rename the endpoint to remove `.json` from it
- fix the serializer to save `Version.build_data`
- remove `readthedocs-client.js` from this repository
- use development URL for js client to work locally

I'm going to merge this PR once all tests pass.

humitos:
I don't understand why I'm receiving `eslint` issues now 🤷🏼 . I'm merging this PR for now because those errors look pretty unrelated to my work here. We can solve the issue in another PR.

readthedocs/doc_builder/director.py
```diff
@@ -625,3 +628,37 @@ def get_build_env_vars(self):
...
+            ),
+            "readthedocs-build.yaml",
+        )
+
+        if not os.path.exists(yaml_path):
+            log.debug("Build output YAML file (readtehdocs-build.yaml) does not exist.")
+            return
+
+        try:
+            with open(yaml_path, "r") as f:
```
stsewd, line 646: We should use `readthedocs.core.utils.filesystem.safe_open` here

readthedocs/proxito/views/hosting.py
```diff
@@ -0,0 +1,128 @@
...
+            return JsonResponse(
+                {"error": "'url' GET attribute is required"},
+                status=400,
+            )
+
+        unresolved_domain = request.unresolved_domain
+        project = unresolved_domain.project
+
+        # TODO: why the UnresolvedURL object is not injected in the `request` by the middleware.
+        # Is is fine to calculate it here?
```
stsewd, line 30: Only the unresolved domain is injected for all proxito views, the unresolved URL is needed only when serving docs, so you need to call it manually only when you need it. And you are parsing the URL from a query param, so you need to call it manually anyway.

PR #10128, opened 2023-03-08
Search: index <dl>s as sections and remove Sphinx domain logic

benjaoming:
WIP - need to test this, just want to know that this is what you had in mind. I'll test and make it ready for review then. CC: @humitos @stsewd 

* [x] Update test data
* [x] Implement known Sphinx behavior in a generic way
* [x] Figure out how much of SphinxDomain logic to disable
* [ ] Open up follow-up issue with a plan for old SphinxDomain data + deprecating/removing the sphinx_domain app ?

Fixes: https://github.com/readthedocs/readthedocs.org/issues/9571


readthedocs/search/parsers.py
```diff
@@ -435,21 +465,23 @@ def _process_fjson(self, fjson_path):
...
+            "domain_data": {},  # domain_data,
         }
 
     def _get_sphinx_domains(self, body):
         """
+        REMOVING THIS
+
         Get all nodes that are a sphinx domain.
 
         A Sphinx domain is a <dl> tag which contains <dt> tags with an 'id' attribute,
```
benjaoming, line 487: We'll have to introduce a bit of Sphinx-specific logic (skipping `footnote` class), but maybe that's fine?

readthedocs/search/parsers.py
```diff
@@ -153,9 +153,40 @@ def _parse_sections(self, title, body):
                         'title': title,
                         'content': content,
                     }
+                except Exception as e:
+                    log.info("Unable to index section.", section=str(e))
+
+        dls = body.css("dl")
+        for dl in dls:
+            dts = dl.css("dt")
```
stsewd, line 161: We only want to index `dt`s that have an ID, otherwise we won't be able to link to them, results would be distracting (they will be indexed in the section, so we aren't losing any data).

readthedocs/search/parsers.py
```diff
@@ -153,9 +153,40 @@ def _parse_sections(self, title, body):
...
+
+            for dt in dts:
+                try:
+                    title, id = self._parse_dt(tag)
+                    next_element = dt.next
+                    # We only index a dt with an accompanying dd
+                    if next_element.tag != "dd":
+                        continue
+                    content, _ = self._parse_section_content(next_element, depth=2)
```
stsewd, line 170: I think we want to index everything that's inside, not till we find another section, but we may be able to re-use this function.

readthedocs/search/parsers.py
```diff
@@ -153,9 +153,40 @@ def _parse_sections(self, title, body):
...
                     log.info('Unable to index section.', section=str(e))
 
+    def _parse_dt(self, tag):
+        """
+        Parses a definition term <dt>
+        """
+        section_id = tag.attributes.get("id", "")
+        if not section_id:
+            parent = tag.parent
+            section_id = parent.attributes.get("id", "")
```
stsewd, line 186: we shouldn't look for the parent here, it should be assumed that the id is in the dt element itself

readthedocs/search/parsers.py
stsewd: I think we should be fine without that footnote logic, since the code will now index all of them.

stsewd (commented):
This is going in a great direction!

Sorry, you said to not review the code, but kind of had to look at the code p: so just left some notes in case is useful

readthedocs/search/parsers.py
benjaoming: That sounds good, that's very simple :+1:

readthedocs/search/parsers.py
stsewd: yeah, but I mean that we need to index everything that's inside the `dd` element, currently `_parse_section_content` will index things inside the dd element till it finds a section inside it.

readthedocs/search/parsers.py
benjaoming: The logic will now pick up any `<dl>` containing `<dt>` elements that have an `id` attribute.

readthedocs/search/parsers.py
benjaoming: Isolated example is now:

```
from selectolax.parser import HTMLParser
dls=HTMLParser("<dl><dt>term a</dt><dd>description a</dd><dt>term b</dt><dt>term c</dt><dd>description b</dd></dl>").css("dl")

for dl in dls:
    dts = dl.css("dt")

    for dt in dts:
        print(dt.text())
        next_element = dt.css_first("dt ~ dd")
        # We only index a dt with an accompanying dd
        if not next_element or next_element.tag != "dd":
            continue
        print(next_element.text())
```

benjaoming:
@stsewd I wasn't able to find any real-life Sphinx HTML output that I was sure would match the intention of this change. The implementation seems fine and solid, but I just wonder which `<dl>` we are looking with `<dt id="foobar">`? Can you quickly point to an example?

readthedocs/search/tests/data/generic/in/basic.html
```diff
@@ -6,5 +6,14 @@
...
   <body>
     Content of the body.
+    <dl>
+      <dt id="love">Love</dt>
+      <dd>To code is to love</dd>
+      <dt>No section for me</dt>
+      <dd>This term does not have an ID so it's not gonna get its own section</dd>
+      <dt id="code">Code</dt>
+      <dt id="docs">Docs</dt>
+      <dd>Code and docs are like love, they are 4-letter words.</dd>
```
benjaoming, line 16: Notice the two `<dt>` elements one after another: This is allowed, you can have several elements like that. Also in a Sphinx glossary.

stsewd:
> @stsewd I wasn't able to find any real-life Sphinx HTML output that I was sure would match the intention of this change. The implementation seems fine and solid, but I just wonder which `<dl>` we are looking with `<dt id="foobar">`? Can you quickly point to an example?
> 
> Edit: Maybe this is it? https://docs.readthedocs.io/en/stable/glossary.html

http domains https://docs.readthedocs.io/en/stable/api/v3.html#get--api-v3-projects-, and autodoc output

https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#directive-automodule

stsewd:
https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#directive-automodule is kind of an interesting case, it has several aliases, we may do some research if that's the standard way of doing that, and see how we can handle those cases, but also, something that can be done in another PR.

benjaoming:
Great! Thanks! I'll brew up some test cases based on this :+1:

benjaoming:
@stsewd I added httpdomain and autodoc test data.

Could you give it a review again? The most important part is to understand if the JSON data pertaining sections is how you'd expect from the HTML.

readthedocs/search/parsers.py
```diff
@@ -145,16 +145,54 @@ def _parse_sections(self, title, body):
...
+                try:
+                    title, _id = self._parse_dt(dt)
+                    # https://developer.mozilla.org/en-US/docs/Web/CSS/General_sibling_combinator
+                    dd = dt.css_first("dt ~ dd")
+                    # We only index a dt with an id attribute and an accompanying dd
+                    if not dd or not _id:
+                        continue
+
+                    # The content of the <dt> section is the content of the accompanying <dd>
+                    content = dd.text()
```
stsewd, line 178: This should still run some of the logic from `_parse_section_content` (but checking for the next definition list, since we can have nested definitions https://requests.readthedocs.io/en/latest/api/#requests.Response), since anything can be inside a dd, in case that there is a numbered code block we want to run the logic from `_parse_code_section`.

readthedocs/search/parsers.py
```diff
@@ -435,34 +472,19 @@ def _process_fjson(self, fjson_path):
                 try:
                     # Create a new html object, since the previous one could have been modified.
                     body = HTMLParser(data["body"])
-                    domain_data = self._generate_domains_data(body)
```
stsewd, line 438: doesn't look like we need this whole try block anymore.

readthedocs/search/tests/data/sphinx/out/autodoc.json
```diff
@@ -0,0 +1,222 @@
...
+    },
+    {
+      "id": "directive-autoexception",
+      "title": ".. autoexception::",
```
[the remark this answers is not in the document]
stsewd: I'm good with indexing the same content multiple time per ID. Maybe @ericholscher has some opinions there?

readthedocs/search/tests/data/sphinx/out/autodoc.json
```diff
@@ -0,0 +1,222 @@
...
...e following class definition, all attributes can be autodocumented: class Foo: \"\"\"Docstring for class Foo.\"\"\" #: Doc comment for class attribute Foo.bar. #: It can have multiple lines. bar = 1 flox = 1.5 #: Doc comment for Foo.flox. One line only. baz = 2 \"\"\"Docstring for class attribute Foo.baz.\"\"\" def __init__(self): #: Doc comment for instance attribute qux. self.qux = 3 self.spam = 4 \"\"\"Docstring for instance attribute spam.\"\"\" Changed in version 0.6: autodata and autoattribute can now extract docstrings. Changed in version 1.1: Comment docs are now allowed on the same line after an assignment. Changed in version 1.2: autodata and autoattribute have an annotation option. Changed in version 2.0: autodecorator added. Changed in version 2.1: autoproperty added. Changed in version 3.4: autodata and autoattribute now have a no-value option. Note If you document decorated functions or methods, keep in mind that autodoc retrieves its docstrings by importing the module and inspecting the __doc__ attribute of the given function or method. That means that if a decorator replaces the decorated function with another, it must copy the original __doc__ to the new function."
```
stsewd, line 13: I usually delete anything that isn't related to the test and cut the paragraphs, so they are easier to read/change (somehow, since JSON doesn't support multiline strings)

readthedocs/search/tests/data/sphinx/out/autodoc.json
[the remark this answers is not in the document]
ericholscher: Seems fine -- definitely an edge case, so not too worried about it.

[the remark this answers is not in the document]
benjaoming: Great! Even if we worried, this is the only solution I can think of :)

readthedocs/search/parsers.py
benjaoming: The whole block in fact!

benjaoming: 612c68772

readthedocs/search/parsers.py
benjaoming: We shouldn't need any recursive logic here since it's handled with `dls = body.css("dl")` ? I looked at `_parse_section_content` and it calls `is_section` which is not exactly up2date and thinks that only headers are sections --- so it won't find numbered code blocks. I think we should detect "code block sections" independently from how `dl` parsing works.

It's interesting that we are creating a flat structure from a nested structure. And that will almost certainly lead to decisions that we have to make. I think the current implementation is a nice and simple addition. But I would want to slightly rewrite things so it's easier to quickly grasp how the call pattern inside the parser works.

stsewd: > We shouldn't need any recursive logic here since it's handled with dls = body.css("dl") ? 

The problem is that it will duplicate the content of nested lists, giving users duplicate results, so I'm suggesting doing something similar to what we did with nested sections.

> I looked at _parse_section_content and it calls is_section which is not exactly up2date 

Yes, that's why I'm suggesting using something similar, to detect definition lists instead of sections for this case.

stsewd (commented):
Two things:

- We need to better handle nested lists https://requests.readthedocs.io/en/latest/api/#requests.Response, otherwise we will be indexing duplicate content.
- First index the lists and delete those nodes before parsing sections, that way we don't index duplicate content.

readthedocs/search/parsers.py
benjaoming: > giving users duplicate results

but isn't that desirable, given that a nested text node is also inside the outer nesting section? I don't see that it's such a big problem.

I find it easy to come up with an example where it seems desirable?

```html
<!-- Look at how generic and uninteresting this section's title is... if we skip the <dl>, there's nothing to index -->
<h2 id="...">Reference</h2>

<dl>
<dt id="...">Very unique and great search term</dt>
<dd>
   <dl>
     <dt id="...">Definition 1></dt>
     <dd>Oh here is some really important text that won't be indexed by neither of the two nesting sections?</dd>
   </dl>
</dd>
</dl>
```

stsewd: The result of B is what we want, we transform nested sections/lists into a flat sequence of lists/sections.

benjaoming:
Thanks @stsewd - can you give more detail on what you mean here?

> First index the lists and delete those nodes before parsing sections, that way we don't index duplicate content.

Is what you mean that you think all the contents of definition lists should be excluded from for instance the contents of a section that follows an `<h2>Title</h2>` ?

readthedocs/search/parsers.py
benjaoming: Check, but then I'm kind of fine with the duplicate outputs.

When you removed `<dl>` lists from the previous sections, was it because that typical Sphinx domains were too much content to have in those sections?

stsewd: > but isn't that desirable, given that a nested text node is also inside the outer nesting section? I don't see that it's such a big problem.

In your example, that text will be returned 3 times when searching, 3 results for the same block of text.

benjaoming: @stsewd something about this feels like it would be a classic search index problem :)

But can you also see the point that when the contents are removed from the containing/nesting element, then the indexing of that element is "worsened" or "less representative" ?

stsewd: Another way of solving this without having to use recursion to detect nested sections/lists would be to start indexing from the child nodes up to the parent nodes and delete the children after parsing, but I haven't explored this solution.

stsewd: > But can you also see the point that when the contents are removed from the containing/nesting element, then the indexing of that element is "worsened" or "less representative" ?

It can trim the context, yeah, but I think that's okay compared to have duplicate results of the same section/text.

stsewd:
> But I found out about running with --no-migrations which saves me A LOT of time now while developing this, since the tests require a database

yeah, our test suite is slow to run, I usually just open a shell and test the changes manually to debug something.

stsewd:
Oh, we should also update our docs at https://dev.readthedocs.io/en/latest/search-integration.html

readthedocs/search/parsers.py
stsewd: > Another way of solving this without having to use recursion to detect nested sections/lists would be to start indexing from the child nodes up to the parent nodes and delete the children after parsing, but I haven't explored this solution.

Opened https://github.com/readthedocs/readthedocs.org/issues/10184

benjaoming: Thanks @stsewd - it is really helpful that you broke it out into another issue :+1: 

We should definitely figure this out, but I feel like it's a more general design decision about how indexing, searching and presenting results should work and that it might affect more elements which we want to index.

I'm also not saying that I'm an expert on this, I barely know how the search works.. I just know how to generate the JSON for the index now :)

benjaoming:
@stsewd I think we can open an issue now about getting rid of the non-generic parsers for MkDocs and Sphinx and build everything into one parser now? That can help about with getting rid of the `documentation_type` field in https://github.com/readthedocs/readthedocs.org/pull/10042

docs/dev/search-integration.rst
```diff
@@ -117,6 +117,11 @@ Tags to be ignored:
 
 - ``nav``
 
+Special Sphinx and MkDocs heuristics:
```
stsewd, line 120: These are removed for all documentation types, not just mkdocs and sphinx.

readthedocs/search/parsers.py
```diff
@@ -140,21 +140,68 @@ def _parse_sections(self, title, body):
...
+                try:
+                    title, _id = self._parse_dt(dt)
+                    # https://developer.mozilla.org/en-US/docs/Web/CSS/General_sibling_combinator
+                    dd = dt.css_first("dt ~ dd")
+                    # We only index a dt with an id attribute and an accompanying dd
+                    if not dd or not _id:
+                        continue
+
+                    # The content of the <dt> section is the content of the accompanying <dd>
+                    content = dd.text()
```
stsewd, line 167:
```suggestion
                    content = self._parse_content(dd.text())
```

readthedocs/search/parsers.py
```diff
@@ -140,21 +140,68 @@ def _parse_sections(self, title, body):
...
+                    yield {
+                        "id": _id,
+                        "title": title,
+                        "content": content,
+                    }
+                except Exception as e:
+                    log.info("Unable to index dt section.", section=str(e))
```
stsewd, line 174: If you want to include the information about the exception in a log, you can use `exc_info=True`.

```suggestion
                    log.info("Unable to index dt section.", exc_info=True)
```

readthedocs/search/parsers.py
```diff
@@ -140,21 +140,68 @@ def _parse_sections(self, title, body):
...
-                        'id': id,
-                        'title': title,
-                        'content': content,
+                        "id": _id,
+                        "title": title,
+                        "content": content,
                     }
                 except Exception as e:
-                    log.info('Unable to index section.', section=str(e))
+                    log.info("Unable to index section.", section=str(e))
```
stsewd, line 194:
```suggestion
                    log.info("Unable to index section.", exc_info=True)
```

readthedocs/search/parsers.py
```diff
@@ -177,6 +224,9 @@ def _clean_body(self, body):
         """
         Removes nodes with irrelevant content before parsing its sections.
 
+        This method is documented here:
+        https://dev.readthedocs.io/en/latest/search-integration.html#irrelevant-content
```
stsewd, line 228:
```suggestion
        https://dev.readthedocs.io/page/search-integration.html#irrelevant-content
```

readthedocs/search/parsers.py
```diff
@@ -188,6 +238,9 @@ def _clean_body(self, body):
             body.css('[role=search]'),
             # Permalinks
             body.css('.headerlink'),
+            # Sphinx code blocks contain line numbers and they are very noisy in contents
```
stsewd, line 241: We are doing this for all doctools, not just Sphinx.

```suggestion
            # Line numbers from code blocks, they are very noisy in contents.
            # This convention is popular in Sphinx.
```

readthedocs/search/parsers.py
```diff
@@ -424,137 +476,45 @@ def _process_fjson(self, fjson_path):
...
 
         return {
-            'path': path,
-            'title': title,
-            'sections': sections,
-            'domain_data': domain_data,
+            "path": path,
+            "title": title,
+            "sections": sections,
+            # this used to contain content from <dl> but this is now handled in a generic parser
```
stsewd, line 488:
```suggestion
            # This used to contain content from <dl> nodes, but they are now handled in the generic parser as sections.
```

readthedocs/search/parsers.py
```diff
@@ -424,137 +476,45 @@ def _process_fjson(self, fjson_path):
...
-
-    def _parse_domain_tag(self, tag):
-        """Returns the text from the description tag of the domain."""
-
-        # Remove all nested domains,
-        # since they are already parsed separately.
-        nested_domains = self._get_sphinx_domains(tag)
-        for node in nested_domains:
-            if tag != node:
-                node.decompose()
```
stsewd, line 553: Looks like this is where our current code handled removing the duplicate nested nodes, so we do need to keep this behavior in the new implementation. This is, index the content from nested description lists just once.

stsewd (commented):
Looks like the current implementation doesn't duplicate the content of nested domains/desc lists, so we need to keep doing the same with this new implementation.

docs/dev/search-integration.rst
benjaoming: Yes, I know, the point was more to say that its not really "generic" behavior but behavior that's in the parser because that's how Sphinx does :) I'll try to spell it out. After I wrote it, I didn't find anything special to MkDocs.

This is also in anticipation of having just one generic parser.

readthedocs/search/parsers.py
```diff
@@ -140,21 +140,68 @@ def _parse_sections(self, title, body):
...
+                        "id": _id,
+                        "title": title,
+                        "content": content,
+                    }
+                except Exception as e:
+                    log.info("Unable to index dt section.", section=str(e))
+
+        # Remove the <dl> nodes that were found to have indexed content
+        for node in dl_nodes_to_be_removed:
+            node.decompose()
```
benjaoming, line 178: @stsewd I suspect that you didn't see this? This is where indexed `<dl>` nodes are removed.

readthedocs/search/parsers.py
benjaoming: thanks, it's probably better to print out the entire exception for sure. I was wondering if the previous pattern was there to silence a lot of anticipated exceptions, but this is really an anti-pattern.

readthedocs/search/parsers.py
benjaoming: Is it pygments that has this line numbering pattern? Or it's a more widespread thing?

readthedocs/search/parsers.py
benjaoming: yes, it has been moved, I mentioned you now in a comment where it's now located.

docs/dev/search-integration.rst
```diff
@@ -117,6 +117,11 @@ Tags to be ignored:
...
+
+- ``.linenos``, ``.lineno`` (line numbers in code-blocks)
```
benjaoming, line 122: So as I understand, this is pygments generated code? @stsewd I'll add it in a comment and in the docs... let's track the origins of these patterns, we might end up with more...

docs/dev/search-integration.rst
```diff
@@ -117,6 +117,11 @@ Tags to be ignored:
...
+- ``.headerlink``
```
benjaoming, line 123: This one I copied from the Sphinx parser, do you know what it is @stsewd ?

docs/dev/search-integration.rst
stsewd: I think they are more from Sphinx and MKDocs material

https://squidfunk.github.io/mkdocs-material/reference/code-blocks/#adding-line-numbers

Not sure if pygments does the line number generation

readthedocs/search/parsers.py
benjaoming: This is now handled based on some assumptions. I added https://requests.readthedocs.io/en/latest/api/#requests.Response as raw test data so we can track any changes in behavior closely. The test runs very fast, so I'm not concerned about the length of this HTML file.

readthedocs/search/parsers.py
benjaoming: This didn't work, it gives me a bunch of attribute errors.. I think it doesn't work on strings.

I'm not sure, but I think that the previous method is fine, it returns all nested text nodes.

readthedocs/search/tests/data/generic/out/basic.json
```diff
@@ -6,7 +6,22 @@
       {
         "id": "",
         "title": "Title of the page",
-        "content": "Content of the body."
+        "content": "Content of the body. Love To code is to love No section for me This term does not have an ID so it's not gonna get its own section Code Docs Code and docs are like love, they are 4-letter words."
```
stsewd, line 9: This doesn't look correct, we are indexing the content from the dls here again.

readthedocs/search/parsers.py
```diff
@@ -140,21 +140,70 @@ def _parse_sections(self, title, body):
...
+                        "id": _id,
+                        "title": title,
+                        "content": content,
+                    }
+                except Exception:
+                    log.info("Unable to index dt section.", exc_info=True)
+
+            # Remove the <dl> node
+            # There isn't a clear indication if this behavior is DFS or BFS
+            dl.decompose()
```
stsewd, line 175: If we aren't sure about the order here, we may just want to duplicate the old logic of making a copy of the node and remove all the children on each iteration.

readthedocs/search/parsers.py
```diff
@@ -140,21 +140,70 @@ def _parse_sections(self, title, body):
...
+
+        # This is pattern is intended for a future generic scenario
+        # We could include the <dl> pattern in _is_section etc. instead of having the above
+        # <dl> block
+        outer_sections = []
         # Index content from h1 to h6 headers.
         for head_level in range(1, 7):
-            tags = body.css(f'h{head_level}')
-            for tag in tags:
+            outer_sections.append(body.css(f"h{head_level}"))
```
stsewd, line 183: Looks like we are looping twice for no reason?

readthedocs/search/tests/data/generic/in/basic.html
```diff
@@ -6,5 +6,14 @@
...
+    </dl>
```
stsewd, line 17: Can we add a nested list here? I love how short this is, it's easy to see if the code is really working.

readthedocs/search/parsers.py
benjaoming: See the new comment above. This is just to accomodate the pattern. I think we can be more generic. So I left it open for additions. I was thinking that the addition of the `<dl>` block is waiting for the next chance to be improved and generalized and merged together with this block.

benjaoming: Turned it into a 1-liner :)

readthedocs/search/parsers.py
stsewd: The previous code is doing a copy of the tag before removing its children, so isn't the same as the current code. The current code is relying on the order of the returned elements to remove the correct elements. This order, we don't know if it's DFS or random.

readthedocs/search/parsers.py
stsewd: Doesn't look like that we need it now, and it's iterating the nodes twice. We should add this code only it's needed now.

readthedocs/search/parsers.py
```diff
@@ -140,21 +140,68 @@ def _parse_sections(self, title, body):
...
+
+        # This is pattern is intended for a future generic scenario
+        # We could include the <dl> pattern in _is_section etc. instead of having the above
+        # <dl> block
         # Index content from h1 to h6 headers.
-        for head_level in range(1, 7):
-            tags = body.css(f'h{head_level}')
-            for tag in tags:
+        outer_sections = [body.css(f"h{h}") for h in range(1, 7)]
+        for section in outer_sections:
```
stsewd, line 182: This is still iterating the nodes twice, but even if we use a generator expression, I'd prefer if we don't add code that we don't know yet if we are going to need.

readthedocs/search/tests/data/generic/in/basic.html
benjaoming: Absolutely! I don't know if there was a test from before that showed the behavior, but all tests are updated, you can check if they are how they should be. I think stuff works like BFS.

readthedocs/search/tests/data/generic/out/basic.json
benjaoming: Oh interesting, actually this block should remove the stuff that's indexed, right?

But a quick hack would be to move the `<dl>` stuff above since it already removes indexed nodes :+1:

benjaoming: nice find!

benjaoming:
Quick update:

The current implementation is producing BFS results, i.e. it takes the outermost `<dl>` and then we can either duplicate all the nested `<dl>`s or remove nested `<dl>`s from the DOM before continuing indexing. But what we want is DFS, so we take the deepest nodes and index first, then removing them from the DOM and indexing higher up `<dl>`s.

I'm going to quickly test how the previous implementation works on the new test data, i.e. does it do DFS like we want.

If it does DFS, I'm going to figure out how to merge it with the new CSS selector `~` that works with adjacent `<dt>` fields.

benjaoming:
Update - I saw the light @stsewd :) results are similar to the old style and emulate DFS... just redoing expected test outputs.

benjaoming:
A small note about leaving the test data in this `tests/data/sphinx` folder - this is not to say that it's data for the Sphinx-specific parser.. this is more to say that this is test data that's been generated _by_ Sphinx. I think we should continue that structure even though everything is handled in a generic parser :+1:

readthedocs/search/tests/data/generic/out/basic.json
benjaoming: Shuffled the order of the parsing, this looks good now :+1:

readthedocs/search/tests/data/generic/out/basic.json
```diff
@@ -1,11 +1,46 @@
...
+      },
+      {
+        "id": "ergo",
+        "title": "Ergo",
+        "content": "Mother is a stone"
+      },
       {
         "id": "",
-        "title": "Title of the page",
+        "title": "Ergo",
```
benjaoming, line 43: Oh this is actually quite strange, it looks like a variable from a previous loop is spilling over

readthedocs/search/parsers.py
```diff
@@ -123,6 +123,59 @@ def _parse_sections(self, title, body):
...
+                    if not dd or not _id:
+                        continue
+
+                    # Create a copy of the node to avoid manipulating the
+                    # data structure that we're iterating over
+                    dd_copy = HTMLParser(dd.html).body.child
+
+                    # Remove all nested domains from dd_copy.
+                    # They are already parsed separately.
+                    nested_domains = dd_copy.css("dl")
```
stsewd, line 159: We need to remove only nested elements that will be indexed (items from description lists that don't have an id won't be indexed).

readthedocs/search/parsers.py
```diff
@@ -123,6 +123,59 @@ def _parse_sections(self, title, body):
...
+        # All terms in dls are treated as sections.
+        # After indexing them, we remove them from the body such that their contents aren't
+        # repeated in the following indexing
+        dls = body.css("dl")
+        seen_dls = []
+        for dl in dls:
+            if dl in seen_dls:
+                continue
+            for inner_dl in dl.css("dl"):
+                seen_dls.append(inner_dl)
```
stsewd, line 136: Is there a reason why we are adding the nested dls here? We want to index those too separately

readthedocs/search/parsers.py
```diff
@@ -123,6 +123,59 @@ def _parse_sections(self, title, body):
...
+                    yield {
+                        "id": _id,
+                        "title": title,
+                        "content": content,
+                    }
+                except Exception:
+                    log.info("Unable to index dt section.", exc_info=True)
+
+            # Remove the <dl> node
+            # There isn't a clear indication if this behavior is DFS or BFS
```
stsewd, line 176: Didn't you figure out that this is BFS? Probably worth updating the comment.

But also, if this is BFS, we will be deleting the nested lists here too, I think.

readthedocs/search/tests/data/generic/in/basic.html
stsewd: We should also add nested elements that don't have an ID to test with

stsewd (commented):
Just a note about removing elements that won't be indexed. And removing those print statements :)

readthedocs/search/parsers.py
benjaoming: I _experience_ that it traverses like BFS... however I did not read the implementation :) So I still think we're safest in not assuming the order here?

stsewd: Yeah, I'm +1 on not assuming the order, so we need to delete all indexed nodes at the end of the loop, not inside.

benjaoming: Yes, :100: 

I've changed it locally and it actually _does_ affect indexing on test data :+1:

readthedocs/search/parsers.py
benjaoming: Thanks! This was from the former approach and I developed a blind spot for it... removing :+1:

readthedocs/search/tests/data/generic/out/basic.json
benjaoming: Fixed

readthedocs/search/tests/data/generic/in/basic.html
```diff
@@ -6,5 +6,29 @@
...
+          <dd>Mother cannot fly</dd>
+          <dt id="stones">All stones</dt>
+          <dd>All stones cannot fly</dd>
+          <dt id="ergo">Ergo</dt>
+          <dd>Mother is a stone</dd>
+        </dl>
+        <dl>
+          <dt>Sub-term of complexity</dt>
+          <dd>I am indexed in the #complexity section</dd>
+        </dl>
```
benjaoming, line 30: This worked well, it's now indexed with its parent `#complexity` section.

readthedocs/search/parsers.py
```diff
@@ -133,28 +148,110 @@ def _parse_sections(self, title, body):
...
-                except Exception as e:
-                    log.info('Unable to index section.', section=str(e))
+                except Exception:
+                    log.info("Unable to index section.", exc_info=True)
+
+    def _parse_dls(self, body):
+
+        # All terms in <dl>s are treated as sections.
+        # We traverse by <dl> - traversing by <dt> has shown in experiments to render a
+        # different traversal order, which could make the tests more unstable.
```
benjaoming, line 177: @stsewd understandable if you think it seems strange with this approach, but at least here we know that we are just traversing 1 <dl> at a time. If we directly iterated on <dt>s with `dl > dt[id]:not([id=""])`, then I would think the ordering is even worse. I did try to swap and the order changed.. so I just left it like this.

benjaoming:
> Just a note about removing elements that won't be indexed. And removing those print statements :)

As you can see, a lot of changes went in :disappointed: 

Kudos to your objection to the removal of `<dl>`s inside the loop: It turned out that by chance, when I was _wrongly_ removing `<dl>`s inside the loop to avoid later indexing in the below sections, it made all the first results look good. But nodes got removed too early and never indexed... I had to look closer at the bigger test sets to realize this.

That also meant that parts of the approach wasn't accurate... I found out that "pre-removing" content in `<dd>` nodes before indexing them was the sensitive part. I wrestled with CSS selectors of selectolax/Modest. They don't cover the full subset of possible CSS selectors and aren't AFAICT really documented.. it made me go back and forth between different approaches, getting a bit superstitious at times :)

Finally, I split everything into a new `_parse_dls` method. Looking at it, it does things how I'd like to do _read_ them now: "for each <dl>, do this".

readthedocs/search/parsers.py
benjaoming: This is now finally done in the most accurate manner by removing exactly the `<dt>, <dd>` elements that are indexed :+1:

readthedocs/search/parsers.py
benjaoming: This turned out to be where all the inaccuracy came from :100: Accurately removing everything that will be indexed later should obviously be done in a correct way.

I've gone over a bunch of the test data now and former duplicates, missing texts etc. now seemed to be fixed :+1: 

Thanks for the patience!!

readthedocs/search/parsers.py
benjaoming: Splitting the whole thing into a method and yielding indexed nodes seems to work well :+1:

readthedocs/search/parsers.py
```diff
@@ -133,28 +149,102 @@ def _parse_sections(self, title, body):
...
+            # We would perhaps like to have written '> dt' but the syntax isn't allowed.
+            # We also cannot use :host.
+            # So in order to ensure that we are only selecting the <dt> elements that are direct
+            # descendents of the traversed <dl>, we give it a unique ID.
+            random_uuid = uuid4().hex
+            # Here, it's okay to manipulate the ID of the <dl> because we're not going to use it.
+            # Note: It didn't work to use custom node attributes nor :has().
+            dl.attrs["id"] = random_uuid
+            # Select all dts with id defined
+            dts = dl.css(f'dl#{random_uuid} > dt[id]:not([id=""])')
```
stsewd, line 193: The `>` selector is basically doing all immediate children, so you can just use `dl.iter()` to iterate over all immediate children, and manually doing the filter of dts with ID. Not as fancy as the CSS selector, but I think it's more explicit than assigning the random ID.

```python
dts = (node for node in dl.iter() if node.tag == 'dt' and node.id)
```

stsewd (commented):
I think we can avoid assigning the random UUID. Everything else looks good.

There is also one thing we haven't taken into consideration, this is domain titles are currently indexed using the simple analyzer

https://github.com/readthedocs/readthedocs.org/blob/32ffebeafcd3d8893ce7ec86bf3669da9ff7380a/readthedocs/search/documents.py#L119-L123

And sections use the default (standard) analyzer (https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-analyzers.html). We are using the simple analyzer because mostly of the domains are in the form of `foo.bar` (and the simple analyzer splits that into `foo` and `bar`, and the default analyzer makes it a whole word).

I think I'm fine with that (maybe we should invest in partial matches more, like experiment with https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-edgengram-tokenizer.html?), but maybe @ericholscher has some opinions here.

readthedocs/search/parsers.py
benjaoming: That works + and yes definitely much less code, the css selector dogma was silly :+1:

ericholscher:
> And sections use the default (standard) analyzer (https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-analyzers.html). We are using the simple analyzer because mostly of the domains are in the form of `foo.bar` (and the simple analyzer splits that into `foo` and `bar`, and the default analyzer makes it a whole word).
> 
> I think I'm fine with that (maybe we should invest in partial matches more, like experiment with https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-edgengram-tokenizer.html?), but maybe @ericholscher has some opinions here.

I don't have a great sense for the trade-offs here. I do know we had users who would complain because they couldn't search for things that include periods and get the function results. I think as long as we have a way to ensure the search query is mapped reasonably onto the results, it should be fine.

stsewd:
> If we index django.db.models as django db models, will a query for django.db.models from a user return it properly, or will it get messed up?

Good question, so analyzers are applied to both, the indexed content and the query (by default the same analyzer used for index is used for search). So searching for `django.db.models` would be the same as searching for `django db models`. I can do a quick test to confirm.

stsewd:
Yep, confirmed, the only thing is that the highlight would be broken into three matches.

Screenshot 2023-04-10 at 17-14-26 Search conf py project test-builds_search-special-chars Read the Docs\

(this was searching `conf.py`)

stsewd (approved):
Just waiting for the analyzer stuff, otherwise everything else is good

benjaoming:
@stsewd added on `title` and `content`, supposing that the issue of searching for `conf.py` will affect both fields.

stsewd:
I think it's probably worth a little more discussion before changing the analyzer, I'm fine merging this without the analyzer changes. Changing the analyzer will require a re-index.

benjaoming:
> I think it's probably worth a little more discussion before changing the analyzer, I'm fine merging this without the analyzer changes. Changing the analyzer will require a re-index.

Check :+1: Reverted it.

So this should be done now, I mentioned the analyzer in https://github.com/readthedocs/readthedocs.org/issues/10184 for now, but we can also open a separate issue.

PR #10129, opened 2023-03-08
Lint: one step forward through linting our code

humitos:
I ran `pre-commit` through all the files to see what happened. This is the result after some manual tweaking. Each commit should contain more or less self-contained changes.

---

- User's documentation (`docs`): https://docs--10129.org.readthedocs.build/en/10129/

- Developer's documentation (`dev`): https://dev--10129.org.readthedocs.build/en/10129/


humitos:
The only one failing right now is `prospector` because we have a lot of docstring that are missing. Those would be probably good to solve as we go instead of in a one-time chunk of work --because that's never gonna happen 😄

readthedocs/api/v3/renderers.py
```diff
@@ -25,7 +23,7 @@ def render(self, data, accepted_media_type=None, renderer_context=None):
           - sort_keys=True on json.dumps
           - use str instead of six.text_types
 
-        https://github.com/encode/django-rest-framework/blob/master/rest_framework/renderers.py#L89
+        https://github.com/encode/django-rest-framework/blob/b7523f4/rest_framework/renderers.py#L84
```
benjaoming, line 26: :1st_place_medal:

benjaoming (approved):
Great initiative to get this done.

This is the kind of PR that can't sit around for long.

I read from top to bottom and couldn't find anything beyond very very trivial.

If someone else can give it a look, they could maybe read from bottom to top so we have tired eyes at opposite ends :)

agjohnson:
I only glanced at this, but I'm happy to see some incremental progress. I'm still a strong +1 on chipping away on linting/formatting until we have the entire repo automatically linted. Being able to rely on `prettier`/`black` for formatting in my editor is a big upgrade.

> If someone else can give it a look, they could maybe read from bottom to top so we have tired eyes at opposite ends :)

This is an excellent strategy :+1:

I feel like we've tried similar in other big refactor/linting PRs. Chunking up review makes it way less intimidating

PR #10131, opened 2023-03-08
Docs: New how-to sublevels (Diátaxis)

benjaoming:
* [x] Add new How-to subleves and distribute existing pages
* [x] Make nice intro pages
* [x] Order pages in each level
* [x] Entries are just in 1 location, might wanna add some entries in multiple locations
* [x] Add redirects for old administrator/developer/author index pages
* [x] Is an index page possible or useful at this stage? Move to separate task
* [x] Fix up the tense of how-to guide titles
* [x] Use only absolute paths (remove relative paths)

Refs: #9747 

## Reflections to consider for review

CC: @ericholscher 

* Writing a lot of similarly-worded texts makes the brain a bit tired, so please feel free to make quick suggestions, no darlings here :)
* I don't think these texts are super sensitive, but they shouldn't look bad.
* The variation is on purpose! I tried NOT to use a formula for every introduction sentence because it'd make the list very boring to the reader's eye.

I wrote an introduction for the very first sub-level and for "best practice", but for all of the other sub-leves, I didn't see that an introduction paragraph contributed anything... it was better without one. Feature reference is the same.

---

- User's documentation (`docs`): https://docs--10131.org.readthedocs.build/en/10131/

- Developer's documentation (`dev`): https://dev--10131.org.readthedocs.build/en/10131/


benjaoming:
I had hoped that I didn't need to write too much new content here. But a lot of the current how-to guides don't have very straight-forward introduction paragraphs. Some don't even have any :)

docs/user/guides/access/index.rst
```diff
@@ -0,0 +1,39 @@
...
+==================================
+
+⏩️ :doc:`Setup Single Sign-On (SSO) with GitHub, GitLab, or Bitbucket </guides/setup-single-sign-on-github-gitlab-bitbucket>`
+    When using an :doc:`organization </commercial/organizations>` on  |com_brand|,
+    you can configure SSO for your users to authenticate to Read the Docs.
+
+⏩️ :doc:`Setup Single Sign-On (SSO) with Google Workspace </guides/setup-single-sign-on-google-email>`
+    When using an :doc:`organization </commercial/organizations>` on  |com_brand|,
+    you can configure SSO for your users to authenticate to Read the Docs.
+    **Separate version for Google Workspace**.
```
ericholscher, line 11: Not sure what this means?

docs/user/guides/access/index.rst
```diff
@@ -0,0 +1,39 @@
...
+
+⏩️ :doc:`Manage Read the Docs teams </guides/manage-read-the-docs-teams>`
+    When using an :doc:`organization </commercial/organizations>` on  |com_brand|,
+    it's possible to create different teams with custom access levels.
+
+⏩️ :doc:`Import private repositories </guides/importing-private-repositories>`
+    You can grant access to private Git repositories using |com_brand|.
```
ericholscher, line 18:
```suggestion
⏩️ :doc:`Manually import private repositories </guides/importing-private-repositories>`
    You can grant access to private Git repositories using |com_brand| using a custom process if required.
```

docs/user/guides/best-practice/index.rst
```diff
@@ -0,0 +1,36 @@
+How-to guides: best practice
+============================
```
ericholscher, line 2:
```suggestion
How-to guides: best practices
=============================
```

docs/user/guides/best-practice/index.rst
```diff
@@ -0,0 +1,36 @@
...
+
+Over the years,
+we have become familiar with a number of methods that work well and which we consider **best practice**.
```
ericholscher, line 5:
```suggestion
These guides share best practices that we've learned for managing documentation.
If you need ideas for how to get started in various areas of documentation,
this is a great place.
```

docs/user/guides/best-practice/index.rst
```diff
@@ -0,0 +1,36 @@
...
+
+⏩️ :doc:`Best practices for linking to your documentation </automatic-redirects>`
```
ericholscher, line 7: This is still really weird that this is the title of our automatic redirects page. Definitely something odd there.

docs/user/guides/best-practice/index.rst
```diff
@@ -0,0 +1,36 @@
...
+    Documentation changes over time,
+    and links and cross-references can become challenging manage for various reasons.
+    Here is a set of best practices explaining and addressing these challenges.
+
+⏩️ :doc:`Deprecating content </guides/deprecating-content>`
+    Best practice for removing or deprecating documentation content.
+
+⏩️ :doc:`Create reproducible builds </guides/reproducible-builds>`
```
ericholscher, line 15:
```suggestion
⏩️ :doc:`Creating reproducible builds </guides/reproducible-builds>`
```

We should choose a tense and stick to it.

docs/user/guides/best-practice/index.rst
```diff
@@ -0,0 +1,36 @@
...
+    Using Sphinx, themes and extensions means that your documentation project needs to fetch a set of dependencies,
+    each with a special version.
+    Over time, using an unspecified version means that documentation projects suddenly start breaking.
+    In this guide,
+    you learn how to secure your project against sudden breakage.
+    **This is one of our most popular guides!**
```
ericholscher, line 21:
```suggestion
    Every documentation project has dependencies that are required to build it.
    Using an unspecified versions of these dependencies means that your project can start breaking.
    In this guide,
    learn how to protect your project against breaking randomly.
    **This is one of our most popular guides!**
```

docs/user/guides/best-practice/index.rst
```diff
@@ -0,0 +1,36 @@
...
+
+⏩️ :doc:`Search Engine Optimization (SEO) for documentation projects </guides/technical-docs-seo-guide>`
+    This article explains how documentation can be optimized to appear in search results,
+    ultimately increasing traffic to your docs.
```
ericholscher, line 25:
```suggestion
    increasing traffic to your docs.
```

docs/user/guides/best-practice/index.rst
```diff
@@ -0,0 +1,36 @@
...
+
+⏩️ :doc:`Hiding a version and keep its documentation online </guides/hiding-a-version>`
+    Learn how you can keep your entire version history online without overwhelming the reader with version choices in the flyout menu.
```
ericholscher, line 28:
```suggestion
⏩️ :doc:`Hiding a version </guides/hiding-a-version>`
    Learn how you can keep your entire version history online without overwhelming the reader with version choices.
```

docs/user/guides/best-practice/index.rst
```diff
@@ -0,0 +1,36 @@
...
+
+.. toctree::
+   :maxdepth: 1
+   :hidden:
+
+   Deprecating content </guides/deprecating-content>
+   Best practices for linking to your documentation </automatic-redirects>
+   Hide a version and keep its documentation online </guides/hiding-a-version>
```
ericholscher, line 36: Confused why only a subset is here?

docs/user/guides/build-troubleshooting.rst
```diff
@@ -1,8 +1,9 @@
-How to troubleshoote build errors
-=================================
+Troubleshooting build errors
+============================
 
 .. include:: /shared/contribute-to-troubleshooting.rst
 
+The following is a list of common errors and resolutions encountered in the :doc:`build process </builds>`.
```
ericholscher, line 6:
```suggestion
This guide provides some common errors and resolutions encountered in the :doc:`build process </builds>`.
```

docs/user/guides/build/index.rst
```diff
@@ -0,0 +1,34 @@
...
+.. seealso:
+
+   :doc:`/builds`
+      An overview and introduction to the build process.
+
+.. toctree::
+   :maxdepth: 1
+   :hidden:
+
+   Setup build notifications and webhooks </guides/build-notifications>
```
ericholscher, line 31: I still think we should just make these the titles of the pages, instead of having to hack them everywhere...

This just seems like a massive amount of unnecessary work to rewrite them every time we use them in both the listing and TOCTree view. 

I think users will understand they are how to guides, even if they don't start with `how to`.

docs/user/guides/content/index.rst
```diff
@@ -0,0 +1,82 @@
+How-to guides: content, themes and SEO
```
ericholscher, line 1: This is a weird title, and seems unrelated?

docs/user/guides/index.rst
```diff
@@ -1,16 +1,12 @@
 :orphan:
 
-Guides
-======
+How-to guides A-Z
```
ericholscher, line 3: Not sure what A-Z means? What is this page doing, do we need it if it isn't including the TOCTrees?

docs/user/guides/setup/index.rst
```diff
@@ -0,0 +1,45 @@
...
+The following how-to guides help you solve common tasks and challenges in the setup and configuration stages.
+
+⏩️ :doc:`Connect your Read the Docs account to your Git provider </guides/connecting-git-account>`
+    Steps to connect an account on |git_providers_or| with your Read the Docs account
+
+⏩️ :doc:`Connect your Git repository automatically </connected-accounts>`
+    Connecting your Read the Docs account to one of the supported Git providers, |git_providers_or|.
+
+⏩️ :doc:`Connect your Git repository manually </guides/git-integrations>`
+    Setting up and connecting your Git repository manually, supports most Git providers with webhooks.
```
ericholscher, line 13: It's really not clear to me how these 3 pages differ, as a reader?

docs/user/guides/troubleshooting/index.rst
```diff
@@ -1,14 +1,22 @@
-Guides for troubleshooting problems
------------------------------------
+How-to guides: troubleshooting problems
+---------------------------------------
 
 In the following guides,
 you can learn how to fix common problems using Read the Docs.
 
+⏩️ :doc:`../build-troubleshooting`
```
ericholscher, line 7:
```suggestion
⏩️ :doc:`/guides/build-troubleshooting`
```
This is so much cleaner 😍 

I like using full paths, it will break much less often.

docs/user/guides/build/index.rst
```diff
@@ -0,0 +1,34 @@
+How-to guides: build process
+============================
+
+⏩️ :doc:`Setup build notifications and webhooks </guides/build-notifications>`
+    Build notifications can alert you when your builds fail so you can take immediate action.
+    In this guide,
+    you will learn how to get notified through various available channel integrations,
+    including email and chat.
```
ericholscher, line 8: I feel like we could also just pull this content from the top of the guide, instead of writing it a bunch of different places?

```suggestion
⏩️ :doc:`/guides/build-notifications`
   .. include:: /guides/build-notifications.rst
      :start-line: 3
      :number-lines: 3
```

https://docutils.sourceforge.io/docs/ref/rst/directives.html#toc-entry-37

ericholscher (commented):
This is great, I didn't do a full copy edit of all the listings pages, but I do think there's likely ways we can be smarter here, rather than copy/pasting the same strings 2-3 places just to remove a couple words. I'd like to find ways to repeat ourselves less, while also providing a good UX to users.

docs/user/guides/index.rst
benjaoming: I think that `A-Z` is a good way to say "here is just a full listing".

There's a separate task for this, though. I found it too hard to jump into without a good structure for meta data - titles and intro paragraphs.

> Is an index page possible or useful at this stage? Move to separate task

docs/user/guides/build/index.rst
ericholscher: If this makes sense to you, I can futz with it to get it working. I think we could even parse out the `How to` probably from the heading with some magic, which seems worth doing if we're having to repeat ourselves 75 times.

docs/user/guides/troubleshooting/index.rst
benjaoming: I was in the business of copy-pasting, thanks for fixing :)

The old layout used relative paths, so I didn't get around to fixing all of them.

docs/user/guides/build/index.rst
benjaoming: Oh nice idea! Do you think we can use this "at scale"?

benjaoming: It'd be very interesting to know if this can work. If it's just 5% better than repeating the same text, I don't think we need it... but if there's something here that can also be used in the Feature Reference, then I think we're on to something.

I haven't been able to find a good solution, so I started dreaming up a Sphinx extension that can handle meta data similar to how Hugo does [Front Matter](https://gohugo.io/content-management/front-matter/).

ericholscher: Sphinx should be able to support metadata by default. I think we can definitely come up with a better approach here.

ericholscher: If you can show me how you'd like it to work, I'll see if I can make it happen tomorrow.

benjaoming: I haven't found a way to insert the metadata? Do you know?

```
:short_howto_title: This is a shorter title
:howto_intro: Here we show you a way to achieve something
```

benjaoming: It needs to be possible to insert the metadata into the contents of the page itself but also in other pages.

ericholscher: Other pages is going to be hard. How does Hugo handle referencing other pages metadata?

benjaoming: The short answer is shortcodes:

https://stackoverflow.com/a/62425260

It's pretty close to what we would want a sphinx-extension providing a role or directive for doing.

docs/user/guides/access/index.rst
benjaoming:
```suggestion
    This guide is written for Google Workspace.
```

It's written so it's easy to spot that the second item isn't a duplicate, otherwise the two how-tos look very identical.

docs/user/guides/best-practice/index.rst
benjaoming: Agreed, the form of writing "Create" was from when we removed "[How to] Create ...", so it came from a navigation style that's no longer relevant.

docs/user/guides/best-practice/index.rst
benjaoming: ultimately, you always remove that word :smile:  :+1:

docs/user/guides/best-practice/index.rst
benjaoming: oh yeah, I added 2 entries from other sections, adding them here, too...

docs/user/guides/content/index.rst
benjaoming: You mean than "content, themes and SEO" is a bad category?

docs/user/guides/setup/index.rst
benjaoming: It's the same as before. At least now, there's a sentence accompanying each guide :) I think we should solve it separately in this case.

I don't think we can improve much without restructuring the guides? I could imagine that merging the first 2 guides would be beneficial, since "Connect your Read the Docs account to your Git provider" is a precondition for "Connect your Git repository automatically".

Having 2 guides: "Doing things automatically" vs. "Doing things manually" will be less work for the reader to digest.

docs/user/guides/best-practice/index.rst
benjaoming: Can you update the suggestion? I can't quite make out what the intention was with the change.

docs/user/guides/best-practice/index.rst
benjaoming: Do you want an explanation or are you saying that it'd be nice to have it fixed because it's getting too strange? :)

I've moved it now. There are more file names that are misleading, I think we need to deal with this separately.

docs/user/guides/build/index.rst
benjaoming: > I still think we should just make these the titles of the pages, instead of having to hack them everywhere...

I disagree with "massive" and "hack", I feel quite relaxed about it :)

Now it's also done for all these titles and put into the Style Guide. So I'd consider it job done. The convention is repeated a lot of times, so I hope we'll remember it automatically :)

The thing is that we _could_ compromise in this one case with titles and navigation labels. But that doesn't fix our need to have a pattern for introduction sentences and to repeat similar structures for our feature reference.

Static site generators are great at this, it's just because Sphinx can't do it yet it seems.

docs/user/guides/content/index.rst
benjaoming: I'd rather represent each of these nouns than the alternatives:

1. Represent important nouns that the reader will recognize.
2. Hide everything in an umbrella term like "content" or "writing documentation", I don't really like that.. it's ambiguous.
3. Split up into separate sub-levels. Not to add more sub-levels at this stage IMO.

But if there's better wording for this, that'd be great!

docs/user/guides/index.rst
benjaoming: Leaving this for now, it was an orphan before, it still is.

benjaoming:
@ericholscher thanks for the great review, I went through the input and updated everything. It's ready for another review, I haven't found anything further actionable.

Redirects are up for removed pages, too :+1:

docs/user/guides/best-practice/index.rst
ericholscher: Not hugely important, just trying to clarify what the content is a bit more. But it's not my best writing, for sure.

docs/user/guides/best-practice/index.rst
ericholscher: I think it's just debt we need to fix. I don't think the automatic redirects page should talk about linking to your docs, but instead mention automatic redirects?

docs/user/guides/content/index.rst
ericholscher: I just don't understand how these 3 things are related, and why only these things? It's a different style than all the others, so we should make it similar.

Perhaps:

```suggestion
How-to guides: Improving user experience
```

Or something?

docs/user/guides/setup/index.rst
ericholscher: I just think users want to do the thing, right? The manual & automatic approaches should presumably be in the same place? I just feel like we broke up our guides in a way that makes them much harder for users to use, and we need to undo that if we're finding it confusing, since users will be even more confused.

docs/user/guides/build/index.rst
ericholscher: We should definitely be able to make Sphinx do this without a ton of work. I really dislike all this repeating and substringing of content. It's much better for our tooling to do this for us, instead of making each of us do a bunch of extra work.

docs/dev/style-guide.rst
```diff
@@ -200,9 +202,17 @@ Explanation
 How-to guides
 ~~~~~~~~~~~~~
 
-* Title should begin with **“How to ...”**,
-  but navigation titles should not contain the “How to” part.
+* Title should begin with **“How to ...”**.
+  If the how-to guide is specific for Sphinx, make sure to note it in the title.
```
ericholscher, line 206:
```suggestion
  If the how-to guide is specific for a tool, make sure to note it in the title.
```

docs/dev/style-guide.rst
```diff
@@ -187,9 +187,11 @@ This means that *both content and navigation path* for all sections should fit a
 Explanation
 ~~~~~~~~~~~
 
-* Introduce the scope: **“This article introduces ...”**
-  (write this as the very first thing,
-  then consider rewriting it or turning it into an internal comment afterwards).
+* Title convention: Use words indicating explanation in the title.
+  Like **Understanding <subject>**, **Dive into <subject>**, **Introduction to <subject>** etc.
+* Introduce the scope in the first paragraph: **“This article introduces ...”**.
```
ericholscher, line 192: 💯

ericholscher (approved):
This looks close to me. I really want to fix the issue where we keep repeating subsets of titles and page intros, but we don't need to solve that now. I do think it's important for longer term maintenance, as those will almost certainly get out of sync, but again, not a blocker here.

docs/user/guides/build/index.rst
benjaoming: Yes, haha, totally agree, I just feel like all paths took me towards "write your own extension" :)

benjaoming:
> This looks close to me. I really want to fix the issue where we keep repeating subsets of titles and page intros, but we don't need to solve that now. I do think it's important for longer term maintenance, as those will almost certainly get out of sync, but again, not a blocker here.

Does it seem good to open a separate issue on researching what to do? Would be nice to hear from others, so I'm thinking we could open it in this very repo.

a) If there's an existing way to do it: We do it!
b) If it requires to write an extension, we need to decide if its worth it..

docs/dev/style-guide.rst
benjaoming: Nice one!

I have a weird love when the style guide doesn't follow the style guide, it so perfectly fits that old saying of "do as I say not as I do" :)

ericholscher:
Yea, 👍 on opening a follow-up issue.

docs/user/guides/build/index.rst
benjaoming: Hoping that we will learn more about potential solutions in [the new issue](https://github.com/readthedocs/readthedocs.org/issues/10157) :)

docs/user/guides/content/index.rst
benjaoming: I've noted to get back to this!

docs/user/guides/setup/index.rst
benjaoming: I think that the manual and automatic approaches are different howtos because they solve different problems. I'm not totally happy with it, but I think it improved now by adding much clearer scope on what the manual guide is for:

image

PR #10134, opened 2023-03-09
Build: remove all `reshim` commands

humitos:
They are not needed anymore because the `reshim` happens automatically inside each of the `asdf` plugins now.

The new version of the `asdf` and its plugin handle this in a better way than how we are handling it here.

Requires: https://github.com/readthedocs/readthedocs-docker-images/pull/189

### DIFF
--- a/readthedocs/doc_builder/director.py
+++ b/readthedocs/doc_builder/director.py
@@ -1,3 +1,5 @@
+"""Build process director."""
+
 import os
 import tarfile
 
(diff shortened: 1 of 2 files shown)


stsewd (approved):

humitos:
Before merging this PR, we need to finish the deploy for the builders we started but we had to revert. Moving back to "In progress".

humitos:
I'm closing this for now.

PR #10138, opened 2023-03-10
Automation rules: model text changes for UI

agjohnson:
This is used in the UI to show a message like:

    Version activated by rule <a>Add new versions</a>

Where "Version activated" comes from the model instead of a template string.

It could exist in the templates as well, but it's much easier here and the
field values are defined here as well.

Release note: this is adding a migration, but it only changes the choice verbose text.


agjohnson:
@stsewd what do you think about a pattern like this? These strings are only used in the templates, but it's hard to maintain these there.

This gives me a clean template:

https://github.com/readthedocs/ext-theme/commit/091f1c864e5bda9fe1159f198cac6afebb68789d#diff-6d8c2c2b68ab62e21bd1a3285333ecc4e47845f8ab6d008ad726a37c3dfc8eb3R45-R47

As opposed the logic above:

https://github.com/readthedocs/ext-theme/commit/091f1c864e5bda9fe1159f198cac6afebb68789d#diff-6d8c2c2b68ab62e21bd1a3285333ecc4e47845f8ab6d008ad726a37c3dfc8eb3R21-R39

In theory, this is more of a Django-y pattern, but it's quite hard to balance isolating front end work and sticking to normal Django patterns.

readthedocs/builds/models.py
```diff
@@ -1083,12 +1083,12 @@ class VersionAutomationRule(PolymorphicModel, TimeStampedModel):
...
-        (MAKE_VERSION_PUBLIC_ACTION, _('Make version public')),
-        (MAKE_VERSION_PRIVATE_ACTION, _('Make version private')),
-        (SET_DEFAULT_VERSION_ACTION, _('Set version as default')),
-        (DELETE_VERSION_ACTION, _('Delete version (on branch/tag deletion)')),
+        (ACTIVATE_VERSION_ACTION, _("Activate version")),
+        (HIDE_VERSION_ACTION, _("Hide version")),
+        (MAKE_VERSION_PUBLIC_ACTION, _("Make version public")),
+        (MAKE_VERSION_PRIVATE_ACTION, _("Make version private")),
+        (SET_DEFAULT_VERSION_ACTION, _("Set version as default")),
+        (DELETE_VERSION_ACTION, _("Delete version")),
```
agjohnson, line 1091: Noted in https://github.com/readthedocs/ext-theme/issues/41, this note on the option would be a good UI addition, and the text can be reduced here.

stsewd (approved):
Makes sense

agjohnson:
Mostly certain tests are failing because of broken linting, I'll rebase to see if it resolves.

agjohnson:
Rebased, will see if that helps :shrug:

agjohnson:
It did not. I'm probably running out of steam wrestling with linting here, but I'll give one more rebase a try before I give up and just merge this. :axe:

agjohnson:
Bah, shit, I forgot migrations. Thanks CI!

agjohnson:
Pinging review again as I struggled a little bit on the migration part. It looks correct to me though, but eyes here are good idea.

stsewd (approved):

PR #10202, opened 2023-03-28
Fix filter positional arguments

agjohnson:
These are always passed as positional arguments, never keyword
arguments. The positional split at `*` caused:

```
BuildListFilter.get_type() takes 2 positional arguments but 4 were given
```

### DIFF
--- a/readthedocs/builds/filters.py
+++ b/readthedocs/builds/filters.py
@@ -50,7 +50,7 @@ class BuildListFilter(FilterSet):
         method="get_type",
     )
 
-    def get_state(self, queryset, *, value):
+    def get_state(self, queryset, _, value):
         if value == self.STATE_ACTIVE:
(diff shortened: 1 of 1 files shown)


humitos (approved):
I haven't tested, but I suppose this is fine. Shouldn't we have a test case for this?

agjohnson:
Probably, but I wasn't being super thorough with this. I'd say if anyone wants to jump in and add a test case here, go for it.

Application code is where I'd love to be able to hand off some work from the dashboard project, though this would have required some of my time anyways, as there is some fiddling to get the filters to display the way I want anyways.

agjohnson:
Moved this over to https://github.com/readthedocs/readthedocs.org/issues/10209 in case anyone wants a small task. I'm fairly certain there are some tests already, but I didn't spend much time looking.

PR #10216, opened 2023-04-03
API: hosting integrations endpoint versioning/structure

### LINKED ISSUE #10211: API: review `/_/readthedocs-config/` endpoint
Take a look at the structure of this response and think about how we want to adapt it to support the initial phase of the js client. See https://github.com/readthedocs/readthedocs-client/issues/26

https://github.com/readthedocs/readthedocs.org/blob/f141a8f931ffe78d6aad90a4710b7ec566994dcd/readthedocs/proxito/views/hosting.py#L35-L136

Besides, consider making the initial changes that will allow us use our current APIv3 serializers so we can standardize the way we are communicating our data in a more generic way through all our endpoints. 

The idea is to use "the same structure that would be returned by the APIv3 serializers, but without using them for now" due to https://github.com/readthedocs/readthedocs-ops/issues/1323

humitos:
Review initial JSON response to keep consistency with APIv3 resources. It uses a small modified version of the APIv3 serializers for known resources, removing some URL fields that can't be resolved from El Proxito.

Sharing the same serializers that our APIv3 make our response a lot more consitent between the endpoints and allow us to expand the behavior by adding more fields in a structured way.

Besides, these changes include a minimal checking of the version coming from `X-RTD-Hosting-Integrations-Version` header to decide which JSON structure return, allowing us to support multiple version at the same time in case we require to do some breaking changes.

The JavaScript client is already sending this header in all of its requests. See https://github.com/readthedocs/readthedocs-client/issues/28

Closes #10211

### DIFF
--- a/readthedocs/proxito/tests/responses/v0.json
+++ b/readthedocs/proxito/tests/responses/v0.json
@@ -1,120 +1,118 @@
 {
-    "comment": "THIS RESPONSE IS IN ALPHA FOR TEST PURPOSES ONLY AND IT'S GOING TO CHANGE COMPLETELY -- DO NOT USE IT!",
-            "project": {
-                "created": "2019-04-29T10:00:00Z",
-                "default_branch": "master",
-                "default_version": "latest",
-                "id": 1,
-                "language": {
-                    "code": "en",
-                    "name": "English"
-                },
-                "modified": "2019-04-29T12:00:00Z",
-                "name": "project",
-                "programming_language": {
-                    "code": "words",
-                    "name": "Only Words"
-                },
-                "homepage": "http://project.com",
-                "repository": {
-                    "type": "git",
-                    "url": "https://github.com/readthedocs/project"
-                },
-                "slug": "project",
-                "subproject_of": null,
-                "tags": [
-                    "project",
-                    "tag",
-                    "test"
-                ],
-                "translation_of": null,
-                "users": [
-                    {
-                    "username": "testuser"
-                }
-                ]
-            },
-            "version": {
-                "active": true,
-                "hidden": false,
-                "built": true,
-                "downloads": {},
-                "id": 1,
-                "identifier": "a1b2c3",
-                "ref": null,
-                "slug": "latest",
-                "type": "tag",
-                "verbose_name": "latest"
-            },
-            "build": {
-                "commit": "a1b2c3",
-                "created": "2019-04-29T10:00:00Z",
-                "duration": 60,
-                "error": "",
-                "finished": "2019-04-29T10:01:00Z",
-                "id": 1,
-                "project": "project",
-                "state": {
-                    "code": "finished",
-                    "name": "Finished"
-                },
-                "success": true,
-                "version": "latest"
-            },
-            "domains": {
-                "dashboard": "readthedocs.org"
-            },
-            "readthedocs": {
-                "analytics": {
-                    "code": null
-                }
-            },
-            "features": {
-                "analytics": {
-                    "code": null
-                },
-                "external_version_warning": {
-                    "enabled": true,
-                    "query_selector": "[role=main]"
-                },
-                "non_latest_version_warning": {
-                    "enabled": true,
-                    "query_selector": "[role=main]",
-                    "versions": [
-                        "latest"
-                    ]
-                },
-                "doc_diff": {
-                    "enabled": true,
-                    "base_url": "https://project.dev.readthedocs.io/en/latest/index.html",
-                    "root_selector": "[role=main]",
-                    "inject_styles": true,
-                    "base_host": "",
-                    "base_page": ""
-                },
-                "flyout": {
-                    "translations": [],
-                    "versions": [
-                        {"slug": "latest", "url": "/en/latest/"}
-                    ],
-                    "downloads": [],
-                    "vcs": {
-                        "url": "https://github.com",
-                        "username": "readthedocs",
-                        "repository": "test-builds",
-                        "branch": "a1b2c3",
-                        "filepath": "/docs/index.rst"
-                    }
-                },
-                "search": {
-                    "api_endpoint": "/_/api/v3/search/",
-                    "default_filter": "subprojects:project/latest",
-                    "filters": [
-                        ["Search only in this project", "project:project/latest"],
-                        ["Search subprojects", "subprojects:project/latest"]
-                    ],
-                    "project": "project",
-                    "version": "latest"
-                }
-            }
+  "comment": "THIS RESPONSE IS IN ALPHA FOR TEST PURPOSES ONLY AND IT'S GOING TO CHANGE COMPLETELY -- DO NOT USE IT!",
+  "projects": {
+    "current": {
+      "created": "2019-04-29T10:00:00Z",
+      "default_branch": "master",
+      "default_version": "latest",
+      "id": 1,
+      "language": {
+        "code": "en",
+        "name": "English"
+      },
+      "modified": "2019-04-29T12:00:00Z",
+      "name": "project",
+      "programming_language": {
+        "code": "words",
+        "name": "Only Words"
+      },
+      "homepage": "http://project.com",
+      "repository": {
+        "type": "git",
+        "url": "https://github.com/readthedocs/project"
+      },
+      "slug": "project",
+      "subproject_of": null,
+      "tags": ["project", "tag", "test"],
+      "translation_of": null,
+      "users": [
+        {
+          "username": "testuser"
         }
+      ]
+    }
+  },
+  "versions": {
+    "current": {
+      "active": true,
+      "hidden": false,
+      "built": true,
+      "downloads": {},
+      "id": 1,
+      "identifier": "a1b2c3",
+      "ref": null,
+      "slug": "latest",
+      "type": "tag",
+      "verbose_name": "latest"
+    }
+  },
+  "builds": {
+    "current": {
+      "commit": "a1b2c3",
+      "created": "2019-04-29T10:00:00Z",
+      "duration": 60,
+      "error": "",
+      "finished": "2019-04-29T10:01:00Z",
+      "id": 1,
+      "project": "project",
+      "state": {
+        "code": "finished",
+        "name": "Finished"
+      },
+      "success": true,
+      "version": "latest"
+    }
+  },
+  "domains": {
+    "dashboard": "readthedocs.org"
+  },
+  "readthedocs": {
+    "analytics": {
+      "code": null
+    }
+  },
+  "features": {
+    "analytics": {
+      "code": null
+    },
+    "external_version_warning": {
+      "enabled": true,
+      "query_selector": "[role=main]"
+    },
+    "non_latest_version_warning": {
+      "enabled": true,
+      "query_selector": "[role=main]",
+      "versions": ["latest"]
+    },
+    "doc_diff": {
+      "enabled": true,
+      "base_url": "https://project.dev.readthedocs.io/en/latest/index.html",
+      "root_selector": "[role=main]",
+      "inject_styles": true,
+      "base_host": "",
+      "base_page": ""
+    },
+    "flyout": {
+      "translations": [],
+      "versions": [{ "slug": "latest", "url": "/en/latest/" }],
+      "downloads": [],
+      "vcs": {
+        "url": "https://github.com",
+        "username": "readthedocs",
+        "repository": "test-builds",
+        "branch": "a1b2c3",
+        "filepath": "/docs/index.rst"
+      }
+    },
+    "search": {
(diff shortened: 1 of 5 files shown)


agjohnson:
I feel like we may have had similar conversations in the past, but are we using any versioning patterns like this anywhere else? I think we maybe talked about what this looks like for APIv3 or our configuration file in the past

humitos:
@agjohnson yeah, we talked about this pattern multiple times. We are using "something similar" in EmbedAPIv3 where we are also getting a HTTP header with the doctool and the version of it. However, we have only v1 response there for now and we haven't started performing the check yet.

agjohnson:
Yeah agreed that versioning is a great place to start from. On the implementation side, I could see how maybe a class based approach could eventually work better, especially if we anticipate changing parts of the response slightly.

For example if we wanted to alter projects response slightly in v1, I could see this pattern helping:

```python
class AddonsV0:
    def get_projects(self):
        # Return v0 projects structure

class AddonsV1:
    def get_projects(self):
        # Return v1 projects structure
```

This helps avoid rewriting everything else in that data structure.

Just my immediate thought though, it certainly can shift over time and doesn't need to start this mutable of a structure.

readthedocs/proxito/tests/responses/v0.json
```diff
@@ -0,0 +1,120 @@
...
+                    "base_url": "https://project.dev.readthedocs.io/en/latest/index.html",
+                    "root_selector": "[role=main]",
+                    "inject_styles": true,
+                    "base_host": "",
+                    "base_page": ""
+                },
+                "flyout": {
+                    "translations": [],
+                    "versions": [
+                        {"slug": "latest", "url": "/en/latest/"}
```
agjohnson, line 98: We probably also want the name/verbose name here too, as we'll use that instead of slug for display purposes in the flyout menu.

readthedocs/proxito/tests/responses/v0.json
```diff
@@ -0,0 +1,120 @@
...
+                        "branch": "a1b2c3",
+                        "filepath": "/docs/index.rst"
+                    }
+                },
+                "search": {
+                    "api_endpoint": "/_/api/v3/search/",
+                    "default_filter": "subprojects:project/latest",
+                    "filters": [
+                        ["Search only in this project", "project:project/latest"],
+                        ["Search subprojects", "subprojects:project/latest"]
```
agjohnson, line 114: I like this pattern. It also helps push the translations into the application which I like. Translations will be A Thing though, so we can punt on that conversation either way.

humitos:
That pattern looks good. I understand the idea would be to have "one method per main key" in the JSON, right?

humitos:
I just wanted to note that this will add a non-trivial overhead from the start, if we want to go in that direction. Maybe this can be a later refactor?

agjohnson:
> I understand the idea would be to have "one method per main key" in the JSON, right?

Something like that, probably similar to Django model views or similar.

> Maybe this can be a later refactor?

Yup, absolutely, it's not a now thing. Just raised here in case we immediately anticipate problems maintaining one method per response structure. It's not even a pattern to work about until after we have users relying on the API response.

readthedocs/proxito/tests/responses/v0.json
```diff
@@ -0,0 +1,120 @@
...
+                "created": "2019-04-29T10:00:00Z",
+                "duration": 60,
+                "error": "",
+                "finished": "2019-04-29T10:01:00Z",
+                "id": 1,
+                "project": "project",
+                "state": {
+                    "code": "finished",
+                    "name": "Finished"
+                },
```
agjohnson, line 60: This is probably commentary on the build serializer and not this PR, but I found myself wanting some helpers here around the build state data. The data is all here, but I had to replicate a good deal of backend logic to determine things like `is_finished`, `is_cancelled`, `is_active`, etc.

Consider this a side request I supposed :slightly_smiling_face:

readthedocs/proxito/tests/responses/v0.json
```diff
@@ -0,0 +1,120 @@
+{
+    "comment": "THIS RESPONSE IS IN ALPHA FOR TEST PURPOSES ONLY AND IT'S GOING TO CHANGE COMPLETELY -- DO NOT USE IT!",
+            "project": {
+                "created": "2019-04-29T10:00:00Z",
+                "default_branch": "master",
+                "default_version": "latest",
+                "id": 1,
+                "language": {
+                    "code": "en",
+                    "name": "English"
```
agjohnson, line 10: `project.language.name` here is an example of where we could support a translated API response, probably based on accept headers. If the addition of a language aware API endpoint changes the implementation here drastically, maybe this is a now conversation. But I'm assuming it's not and we can come back to this as we give users more access into this response structure.

readthedocs/proxito/views/hosting.py
```diff
@@ -23,37 +55,120 @@ def get(self, request):
...
+                del self.fields[field]
+
+            if field in self.Meta.fields:
+                del self.Meta.fields[self.Meta.fields.index(field)]
+
+
+# NOTE: the following serializers are required only to remove some fields we
+# can't expose yet in this API endpoint because it running under El Proxito
+# which cannot resolve some dashboard URLs because they are not defined on El
+# Proxito.
```
agjohnson, line 120: I'm guessing this is also probably helpful to not point users of this response structure towards resources on APIv3, where they will likely need to use an authenticated API client.

readthedocs/proxito/views/hosting.py
```diff
@@ -23,37 +55,120 @@ def get(self, request):
...
+
+class VersionSerializerNoLinks(NoLinksMixin, VersionSerializer):
+    pass
+
+
+class BuildSerializerNoLinks(NoLinksMixin, BuildSerializer):
+    pass
+
+
+class ClientResponse:
```
agjohnson, line 135: Do we want to do this?

```suggestion
class AddonsResponse:
```

readthedocs/proxito/views/hosting.py
```diff
@@ -64,6 +179,7 @@ def get(self, request):
             },
             "features": {
                 "analytics": {
+                    # TODO: consider adding this field into the ProjectSerializer itself.
```
agjohnson, line 182: :+1:

agjohnson (commented):
With the client -> addons conversation, there are some naming changes here, but otherwise this is looking great.

Most of my notes are just future ideas, the only immediate change I see is that we likely want the version name in the flyout versions structure. All the other pieces are later pieces, after we have some features public and usable.

readthedocs/proxito/tests/responses/v0.json
agjohnson: For example, I had to sort through a lot of build logic with the API response this the build detail view, for example: https://github.com/readthedocs/ext-theme/blob/main/src/js/build/detail.js#L236-L283

humitos:
> the only immediate change I see is that we likely want the version name in the flyout versions structure

I haven't touched the structure of `features` because we are not going to use it for the initial test. In any case, any Project, Version or Build object that's gonna be rendered inside `features` will use the same serializer we are using at the root keys. So, they will have all those fields as well.

I didn't change those "just yet" because it will also require a change in the client to keep working. We can coordinate both changes in a following up PR if it's not urgent.

readthedocs/proxito/tests/responses/v0.json
humitos: Yes. We will be using the `VersionSerializer` here for all of these versions.

readthedocs/proxito/tests/responses/v0.json
humitos: Well, this is just a hard coded/default example.

In theory, what I understood from the work that @stsewd did, is that _the author of the docs_ is whom will decide the names and the values of the filters. So, we need to keep talking how that data will be passed (see https://github.com/readthedocs/readthedocs-client/issues/13)

readthedocs/proxito/tests/responses/v0.json
humitos: > The data is all here, but I had to replicate a good deal of backend logic to determine things like is_finished, is_cancelled, is_active, etc.

This is a pretty common discussion between backend and frontend teams 😄 . IMHO, the data is there and it's the FE who should create "the helpers" required to present the data as they need --but I guess this is kind of a philosophical discussion. We can continue talking about this outside this PR.

In any case, if we decide changing anything here, it should be done in the `BuildSerializer`.

readthedocs/proxito/views/hosting.py
humitos: You have a good point here, yes.

However, there are _other URLs_ that El Proxito is not able to resolve currently;

- project's dashboard page
- dashboard build's detail page
- downloads
- etc

So, we still need to find a way to resolve those and/or hard code them if we think that's fine for now.

readthedocs/proxito/views/hosting.py
humitos: I rename everything called "client" to "addons" 👍🏼

agjohnson:
That makes sense. Maybe a todo on that response code would be a good addition then. We can return when we want to break our own usage of the API :smile:

humitos:
I'm thinking a little more ahead in time here and just thought that we could have better top keys as well. Mainly for `project`, `version` and `build`. I think it will be helpful to group all the required objects under a more general key. This is the proposal:

```javascript
{
  "projects": {
    "current": { ... },
    "translations": [{ ... }],  // list of all translations for this project
    ...
  },
  "versions": {
    "current": { ... },
    "active": [{ ... }],  // list of all active versions for this project
    ...
  },
  "builds": {
    "current": { ... },
    ...
  }
}
```

This will allow us to expand these objects in the future by returning more categorized results. "Active versions" and "Translations" are the most clear here, since they will be used for the flyout and with the current structure from this PR I feel we will end up having `version` and `active_versions` at the same level, where I think that `versions.current` and `versions.active` is clearer and a nicer expandable structure. @agjohnson @ericholscher what do you think?

agjohnson:
I think it's a solid improvement. I'd agree that the version instances don't need to be nested in the addon configuration section, which feels more like it should be user data and defaults. I could see us referencing versions from an `addons.flyout.version_ids: [1, 2, 3]` though

humitos:
@agjohnson 

> I could see us referencing versions from an addons.flyout.version_ids: [1, 2, 3] though

I'm not sure to follow this part. Can you expand why we would need these version `id`s on the flyout? What they will be useful for here?

agjohnson:
I could see this being useful if we want to include a complete list of versions in the top level of the data response, and only want to reference a few versions for the flyout.

We might want to return all active versions at the top level listing (including hidden versions and/or pull request versions), but the version selector only shows active, non-pull request versions.

humitos:
> We might want to return all active versions at the top level listing (including hidden versions and/or pull request versions), but the version selector only shows active, non-pull request versions.

In that case, we should have `versions.active`, `versions.hidden` and `versions.external`. That would be a lot more clear, and will avoid "the linking versions" for the flyout; IMO, that only adds unnecessary complexity.

In fact, this was my whole point when suggesting this new structure with another level: "We can add more _filtered/categorized data_ for the top keys in the future without breaking the structure"

humitos:
BTW, we cannot return "all the versions" because that will destroy our database. So, there is going to be always an implicit filter.

agjohnson:
> Have the ability to put "all Project objects" under a top key projects; and the same for Version and Build

Agreed, I like this part of the pattern you outlined.

> I'll move forward with the structure <object>.current and later we can decide the other names.

I also like this part of the pattern :+1: 

It was just sublists like `version.active` or `version.hidden` that seemed odd to me

> BTW, we cannot return "all the versions" because that will destroy our database. So, there is going to be always an implicit filter.

I think I understand this and probably agree here. But to clarify, I wouldn't return inactive versions or versions we don't expect to use in display of the version selector/etc. This is where users/authors should have to use our API directly.

> We can figure it out the naming of those sub-keys later if we want to.

Also agreed here, we can come back to this decision/discussion later. My issue was probably mostly with building out the API too much to start, we only really need a single list of projects to continue immediately.

humitos:
> I wouldn't return inactive versions or versions we don't expect to use in display of the version selector/etc. This is where users/authors should have to use our API directly.

Yes. I gave `versions.active` as example because those are the ones we are currently shown in the flyout and I think we will keep the same behavior in the new client. 

However, that `versions.active` also implies:

- only versions the current user hitting the API has permissions on
- exclude `hidden` versions
- exclude external/PR versions

and I think there was where the confusion came. It's not "just `active=True` versions" -- and that's why I said the naming was properly not 100% accurate.

In any case, we can move forward with the latest changes. However, I didn't get an approve 🤷🏼

agjohnson (approved):

PR #10224, opened 2023-04-05
Docs: Experiment with canonical url using READTHEDOCS_CANONICAL_URL

benjaoming:
This is a quick experiment: I noticed that all the canonical  meta tags on [all versions of the Jupyter Notebook](https://jupyter-notebook.readthedocs.io/en/stable/) are almost definitely wrong.. they point to /stable meaning that you cannot generate accurate link previews.

Take for instance this build: https://readthedocs.org/projects/jupyter-notebook/builds/19683327/

The solution is to use our new environment variable :tada:

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -70,7 +70,7 @@ templates_path = ["_templates"]
 # This will ensure that we use the correctly set environment for canonical URLs
 # Old Read the Docs injections makes it point only to the default version,
 # for instance /en/stable/
-html_baseurl = os.environment.get("READTHEDOCS_CANONICAL_URL")
+html_baseurl = os.environ.get("READTHEDOCS_CANONICAL_URL", "/")
 
 master_doc = "index"
 copyright = "Read the Docs, Inc & contributors"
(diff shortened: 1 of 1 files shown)


humitos (approved):
Great!

> This is a quick experiment: I noticed that all the canonical meta tags on [all versions of the Jupyter Notebook](https://jupyter-notebook.readthedocs.io/en/stable/) are almost definitely wrong.. they point to /stable meaning that you cannot generate accurate link previews.

Does it make sense to open a PR in their repository with these changes?

benjaoming:
> Does it make sense to open a PR in their repository with these changes?

It does :+1: I shared a pointer to this PR, hoping that some Jupyter people would be interested in adding this to their various projects :+1: 

If we have a couple of successful implementations of this, we can start tracking them in an issue to use `READTHEDOCS_CANONICAL_URL` for the Sphinx builder conf.py injection - I'm not a big fan of changing the legacy code, but it's misbehaving and perhaps aligning it with the correct/recommended pattern is actually a simplification :+1:

PR #10239, opened 2023-04-12
APIv3: add `analytics` field to `Project` object

humitos:
New `analytics` field that exposes:

- the Google code if the project has it configured
- whether or not the global Google/Plausible Read the Docs analytics is enabled

This field is going to be useful for the new "Analytics Addon".

Related #10216

### DIFF
--- a/docs/user/api/v3.rst
+++ b/docs/user/api/v3.rst
@@ -174,6 +174,14 @@ Projects list
                     "latest": "{VERSION}",
                     "19.0.2": "{VERSION}"
                 },
+                "analytics": {
+                    "google": {
+                         "code": "UA-XXXXX"
+                    },
+                    "readthedocs": {
+                         "enabled": true
+                    }
+                }
                 "_links": {
                     "_self": "/api/v3/projects/pip/",
                     "versions": "/api/v3/projects/pip/versions/",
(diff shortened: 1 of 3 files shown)


humitos:
We are deprecating the GA analytics in https://github.com/readthedocs/readthedocs.org/issues/9530, so may not make sense to expose this here?

benjaoming:
Yeah, given the path laid out in https://github.com/readthedocs/readthedocs.org/issues/9530, it doesn't make sense to expose the field so some new integration would end up using it.

Are you okay with closing this @humitos ?

humitos:
@benjaoming 

> whether or not the global Google/Plausible Read the Docs analytics is enabled

We need to decide what to do with this field, tho. We have ~1900 projects disabling Read the Docs analytics. I think I'm fine removing this config and always enforce our own analytics to projects, https://docs.readthedocs.io/en/stable/analytics.html#disabling-google-analytics-on-your-project -- I'm not sure if it adds too much value to our platform 🤷🏼

benjaoming (approved):
The change is pretty straightforward. And the dictionary structure with `google` and `readthedocs` is great!

There are some weird test failures, but maybe merging main can fix that?

benjaoming:
@humitos You mean that it's better to expose it in the API and take a separate decision about deprecating / keeping it? I think it's fine - actually, it's not bad that someone would want to use this part of the API, just means that it's relevant.

humitos:
I'm saying **we should not merge this PR yet**. We need to define how people will/if decide whether or not they want us to include our own analytics in their docs.

benjaoming:
Okay, I'm confused about how a change that exposes a field in the API contributes to an actual change in configuration and documentation rendering.

humitos:
I'm closing this PR because we are moving forward with the plan from https://github.com/readthedocs/readthedocs.org/issues/9530#issuecomment-1506591104

PR #10240, opened 2023-04-12
API: analytics return 400 when there is an error

humitos:
If `absolute_uri` is not passed, we return 400 with a message. Besides, if the request was success, we return 204 no content.


stsewd (approved):
Test need to be updated

humitos:
I'm not sure to understand why `prospector` check is failing 🤷🏼

benjaoming:
@humitos it's failing because you edited a file that already didn't lint. So it's not your changes that caused the lint failure, but you have to fix other parts of the same files that you edited to get linting to pass.

We lint full files with pre-commit once the file shows up in a git diff.

I'd love to improve our approach to linting because of the accumulated disruption that it causes. I don't recall if I've raised this issue elsewhere.. I just had a look around and couldn't find any discussion on the topic. I recently had an issue with something that mostly looked like a prospector bug: https://github.com/readthedocs/common/pull/174

benjaoming:
@humitos Could be because the file is called `readthedocs/analytics/tests.py` and we normally ignore `readthedocs/analytics/tests/*` ?

humitos:
This should now pass because `common/` was updated and fixes that issue.

PR #10245, opened 2023-04-17
Scripts: Add export statements and instruction to fetch awscli (compile_version_upload_s3.sh)

benjaoming:
Just quickly updating the script...

1) I received some error "credentials not found" until the environment variables got exported. But otherwise it worked fine :+1: 

https://github.com/readthedocs/readthedocs.org/blob/main/requirements/pip.in#L37

2) awscli has been removed from our requirements so needs to be added manually to the environment. Since it's not terribly important, I added a note that we MIGHT use mc (as the docker containers also do) in case there are issues with it.

https://github.com/readthedocs/readthedocs.org/blob/main/requirements/pip.in#L37

### DIFF
--- a/scripts/compile_version_upload_s3.sh
+++ b/scripts/compile_version_upload_s3.sh
@@ -19,6 +19,13 @@
 #
 #   inv docker.up
 #
+# You also need the "awscli" PyPi package in your local Python environment.
+#
+#   pip install awscli
+#
+# Note: The version that used is currently simply the latest. If there are issues in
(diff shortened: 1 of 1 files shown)


humitos (approved):

benjaoming:
"checks" job went fine (but somehow Circle didn't report it to GH)

"tests" seems to be failing with the recent addition of `privacy_level` to APIs, but I'm pretty sure that's unrelated :+1:

PR #10219, opened 2023-04-03
Proxito: inject hosting integration header for `build.commands`

humitos:
Versions using `build.commands` will start using the new Read the Docs JavaScript client.

This is still under beta testing, but we want to move forward slowly and start exposing these features to projects that don't have access currently.

Related https://github.com/readthedocs/readthedocs-client/issues/26

PR #10259, opened 2023-04-19
Revert "Proxito: inject hosting integration header for `build.commands` (#10219)"

ericholscher:
This reverts commit 27215d29b372369e1103d4b480dace77b69ad655.


humitos (approved):
LGTM 👍🏼 

Note: we are reverting this because it queries `builds_build` table which takes a lot of time to be done on each request 😢 

We still need to find out a way to do this in a more efficient way.

PR #10265, opened 2023-04-21
Dev environment: Add settings necessary for https and reverse proxy

benjaoming:
## Why

Depends on: https://github.com/readthedocs/common/pull/176

This setup enables a developer to:

* Run development environments behind HTTPS
* Use a custom production domain and subdomains
* Test developer's own GitHub OAuth applications
* Finally set `DEBUG=False` via an option for `invoke docker.up` :tada: 

## What

Changes to docker-compose setup:

* [x] Storage HTTPS compatibility: Docker container for storage needs additional configuration
* [x] Custom domain w/ https possible: Additional settings to traverse from `inv` to `settings.docker_compose`
* [x] New socialauth options in `settings.docker_compose` to enable https
* [x] New setting to disable a suspicious host check in new unresolver

## Changes

This PR was originally intended to add https compatibility to the storage container, but turned out a few more things beyond the `inv docker.up --ngrok=custom.dev.example` have been necessary

### DIFF
--- a/docs/dev/install.rst
+++ b/docs/dev/install.rst
@@ -114,7 +114,9 @@ save some work while typing docker compose commands. This section explains these
     * ``--init`` is used the first time this command is ran to run initial migrations, create an admin user, etc
     * ``--no-reload`` makes all celery processes and django runserver
       to use no reload and do not watch for files changes
-    * ``--ngrok`` is useful when it's required to access the local instance from outside (e.g. GitHub webhook)
+    * ``--no-django-debug`` runs all containers with ``DEBUG=False``
+    * ``--http-domain`` configures an external domain for the environment (useful for Ngrok or other https proxy)
+    * ``--https`` if using an HTTPS proxy, you may need to force the ``https://`` protocol for settings that otherwise automatically detect it as ``http://``
     * ``--ext-theme`` to use the new dashboard templates
     * ``--webpack`` to start the Webpack dev server for the new dashboard templates
 
--- a/dockerfiles/settings/proxito.py
+++ b/dockerfiles/settings/proxito.py
@@ -1,3 +1,5 @@
+import os
+
 from readthedocs.settings.proxito.base import CommunityProxitoSettingsMixin
 
 from .docker_compose import DockerBaseSettings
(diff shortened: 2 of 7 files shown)


benjaoming:
I think we can remove the mixin since storages supports what we need. Another version of the quoted PR was merged in 2019, seems to only support Azure and then CDN has been added, too.

https://github.com/readthedocs/readthedocs.org/blob/9324142ee57d4be6ef3f3a6e0640d38d9936a53d/readthedocs/storage/mixins.py#L8-L14

I can make it work by removing the mixin and adding these settings:

```
AWS_S3_CUSTOM_DOMAIN = f"{PRODUCTION_DOMAIN}"
AWS_S3_URL_PROTOCOL = "https:"
```

https://django-storages.readthedocs.io/en/latest/backends/amazon-S3.html?highlight=AWS_S3_URL_PROTOCOL#cloudfront

benjaoming:
(deleted a comment about using `AWS_S3_CUSTOM_DOMAIN` instead - yes, that worked - but of course we have different storage classes, so can't use a global setting)

docs/dev/install.rst
```diff
@@ -114,7 +114,9 @@ save some work while typing docker compose commands. This section explains these
     * ``--init`` is used the first time this command is ran to run initial migrations, create an admin user, etc
     * ``--no-reload`` makes all celery processes and django runserver
       to use no reload and do not watch for files changes
-    * ``--ngrok`` is useful when it's required to access the local instance from outside (e.g. GitHub webhook)
+    * ``--no-django-debug`` runs all containers with ``DEBUG=False``
+    * ``--http-domain`` configures an external domain for the environment (useful for Ngrok or other https proxy)
```
humitos, line 118: Doesn't this interfere with the `--ngrok` option from https://github.com/readthedocs/common/pull/151?

humitos: Ah, you removed that option... I'm still sleepy it seems. I'll need to come back to this when I'm 100% awake 😴

humitos:
What's the best way to run this PR locally and check that everything works as expected? What's the `inv` command I should run to spin up the instance?

humitos:
> Run development environments behind HTTPS 

Is this _required_ for any particular reason? This is the chunk of code that makes me more noise from this PR (touches production code and adds complexity) and I'd try to avoid it unless it's strictly needed for a particular task. I've never required to run our development instance over HTTPS, even when working with GitHub webhooks and other stuffs.

I'd be happy if we split this PR in two and remove from this one the `--https` option so we can keep discussing that in another PR and we don't block this one.

benjaoming:
Note to self: We have areas in the code-base using `protocol = 'http' if settings.DEBUG else 'https'`, these should also be fixed.

benjaoming:
New PRs:

* https://github.com/readthedocs/readthedocs.org/pull/10384
* https://github.com/readthedocs/readthedocs.org/pull/10385

Issue #9280, opened 2022-06-01
Config file: make `build.tools` not required when using `build.commands`


humitos:
When using `build.commands`, it's possible to build documentation without installing any of the tools we support on `build.tools` as shown in the example using `echo "Hola mundo!" > _readthedocs/html/index.html`. However, `build.tools` is _required_ currently when using `build.os`. We should remove this restriction and make the `Config` classes code to differentiate legacy over the new pattern by the usage of `build.os` instead of `build.tools.`

Context https://github.com/readthedocs/readthedocs.org/pull/9150#discussion_r882850819

PR #10281, opened 2023-04-30
Allow build.commands without build.tools

benjaoming:
* [x] update tests
* [x] update validate
* [x] check everything else that depends on some assumption

Fixes https://github.com/readthedocs/readthedocs.org/issues/9280

### DIFF
--- a/docs/user/build-customization.rst
+++ b/docs/user/build-customization.rst
@@ -360,7 +360,7 @@ Where to put files
 ~~~~~~~~~~~~~~~~~~
 
 It is your responsibility to generate HTML and other formats of your documentation using :ref:`config-file/v2:build.commands`.
-The contents of the ``_readthedocs/<format>/`` directory will be hosted as part of your documentation.
+The contents of the ``$READTHEDOCS_OUTPUT/<format>/`` directory will be hosted as part of your documentation.
 
 We store the the base folder name ``_readthedocs/`` in the environment variable ``$READTHEDOCS_OUTPUT`` and encourage that you use this to generate paths.
 
--- a/docs/user/build-customization.rst
+++ b/docs/user/build-customization.rst
@@ -369,6 +369,15 @@ Supported :ref:`formats <downloadable-documentation:accessing offline formats>` 
 * ``_readthedocs/pdf/``
 * ``_readthedocs/epub/``
 
+.. note::
+
+   Remember to create the folders before adding content to them.
+   For instance, you can add a folder with:
+
+   .. code-block:: console
+
+       mkdir -p _readthedocs/html/
+
 Search support
 ~~~~~~~~~~~~~~
 
--- a/docs/user/build-customization.rst
+++ b/docs/user/build-customization.rst
@@ -362,12 +362,14 @@ Where to put files
 It is your responsibility to generate HTML and other formats of your documentation using :ref:`config-file/v2:build.commands`.
 The contents of the ``_readthedocs/<format>/`` directory will be hosted as part of your documentation.
 
+We store the the base folder name ``_readthedocs/`` in the environment variable ``$READTHEDOCS_OUTPUT`` and encourage that you use this to generate paths.
+
 Supported :ref:`formats <downloadable-documentation:accessing offline formats>` are published if they exist in the following directories:
 
-* ``_readthedocs/html/`` (required)
-* ``_readthedocs/htmlzip/``
-* ``_readthedocs/pdf/``
-* ``_readthedocs/epub/``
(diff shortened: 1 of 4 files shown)


benjaoming:
Tested on https://github.com/readthedocs/test-builds/tree/bare-commands

benjaoming:
It works! But maybe we should add more documentation about the fact that the directory needs to be created?

https://docs.readthedocs.io/en/stable/build-customization.html#where-to-put-files

image

benjaoming:
@humitos I made a "positive" test (manually) to see that a build like this works. But is there something else to be aware of?

humitos:
> It works! But maybe we should add more documentation about the fact that the directory needs to be created?

I think adding just one sentence in the link you pasted mentioning this in the documentation is enough.

humitos:
> But is there something else to be aware of?

When working with the build process, I recommend you to always test locally a bunch of use-cases/branches from the `test-builds` repository -- not only the exact case you want to make it work. It can be tricky multiple times 😄

readthedocs/config/tests/test_config.py
```diff
@@ -1135,9 +1159,9 @@ def test_commands_build_config_invalid_no_tools(self):
                 },
             },
         )
-        with raises(InvalidConfig) as excinfo:
-            build.validate()
-        assert excinfo.value.key == "build.tools"
+        build.validate()
+        assert isinstance(build.build, BuildWithTools)
+        assert build.build.commands == ["pip install pelican", "pelican content"]
```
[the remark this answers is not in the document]
benjaoming: Ah yeah, too much copy-paste here, should make the test case make more sense :+1:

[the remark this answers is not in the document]
benjaoming: Changed the commands to something that makes sense :+1:

benjaoming:
> I think adding just one sentence in the link you pasted mentioning this in the documentation is enough.

Added as a note box. I think it's great to add it in this PR :)

image

benjaoming:
> > But is there something else to be aware of?

> When working with the build process, I recommend you to always test locally a bunch of use-cases/branches from the `test-builds` repository -- not only the exact case you want to make it work. It can be tricky multiple times 😄 

I'll be trying out different branches that look relevant and get back :+1:

docs/user/build-customization.rst
```diff
@@ -369,6 +369,15 @@ Supported :ref:`formats <downloadable-documentation:accessing offline formats>`
...
 * ``_readthedocs/epub/``
 
+.. note::
+
+   Remember to create the folders before adding content to them.
+   You can ensure that the output folder exists by adding the following command:
+
+   .. code-block:: console
+
+       mkdir -p _readthedocs/html/
```
humitos, line 379: Ideally, the whole page should use `$READTHEDOCS_OUTPUT/` instead of the hardcoded `_readthedocs/`. The environment variable is the absolute path and it will always create the correct directory.

I'm saying the whole page, because it's not just this note, but where the folder are listed in the previous bullet list, and the first paragraph as well.

benjaoming: I was looking for an issue or PR about this yesterday but didn't find anything and forgot about it. Might as well do it here then :)

benjaoming: We decided not to have env vars like `$READTHEDOCS_OUTPUT_HTML`, right?

benjaoming: Updated examples in docs :+1:

docs/user/build-customization.rst
```diff
@@ -362,12 +362,14 @@ Where to put files
 It is your responsibility to generate HTML and other formats of your documentation using :ref:`config-file/v2:build.commands`.
 The contents of the ``_readthedocs/<format>/`` directory will be hosted as part of your documentation.
```
humitos, line 363:
```suggestion
The contents of the ``$READTHEDOCS_OUTPUT/<format>/`` directory will be hosted as part of your documentation.
```

benjaoming: I was kind of wondering if leaving this exposed would be helpful somehow. In case people need to hard-code it. But let's just remove it then :+1:

humitos: Yeah, hardcoding it is exactly what we want to avoid. We may need to change this path in the future and it will break all the projects that are hardcoding this value.

benjaoming:
I did a bunch of builds searching for branch names containing "jobs" and "tools" and "commands" in test-builds and didn't find any positive negatives. The failing branches are failing later in the build.

image

humitos (approved):

PR #10301, opened 2023-05-07
Docs: Configuration file how-to guide

nikblanchet:
This is a how-to guide covering how to create and set up the configuration file. The goal is a narrative document explaining to both the Markos and Riika personae…

1. How to create the ``.readthedocs.yaml`` file
2. What the minimum required settings need to be in the file

It will link to the full configuration file reference page for those who want to go into more detail.

---

- User's documentation (`docs`): https://docs--10301.org.readthedocs.build/en/10301/

- Developer's documentation (`dev`): https://dev--10301.org.readthedocs.build/en/10301/


agjohnson:
We had a conversation around the configuration file being not not correct, I opened up https://github.com/readthedocs/readthedocs.org/issues/10302
nn

agjohnson:
I put up #10303 to update the wrongly tagged required fields. We decided that it would be best to call these three fields required:

- `version`
- `build.os`
- `build.tools` -- this does require a subkey, like `python`, but none of the subkeys are required, just *one* subkey is required

humitos:
> `build.tools` -- this does require a subkey, like `python`, but none of the subkeys are required, just _one_ subkey is required

To add a little more complexity here, when using `build.commands`, the `build.tools` key is not required 😄

humitos:
Thinking a little more about this, it seems users will benefit a lot from having "template use cases" defined with a working YAML example together:

- run Pelican with `build.jobs`
- run Docusaurus with `build.commands`
- recommended way to setup Sphinx on Read the Docs
- config file to use MkDocs without `build.jobs` nor `build.commands`
- and more examples

This will probably allow people to just copy and paste these config files to use in their project and make some minor adjustments (probably file paths) to make them work on their projects.

docs/user/guides/creating-config-file.rst
```diff
@@ -0,0 +1,100 @@
...
+
+####################################
+Where to put your configuration file
+####################################
+
+The ``.readthedocs.yaml`` file should be placed in the **root directory**. The 
+root directory is the top-most location of your repository.
+
+To make the file, navigate to your root directory and create a new 
+file with the name ``.readthedocs.yaml``.
```
benjaoming, line 33:
```suggestion
Add a new file with the exact name ``.readthedocs.yaml`` in the repository's root directory.
We will get to the contents of the file in a moment.
```

I think people have preliminary knowledge of adding new files to a repository.

I wonder if we can skip the _explanation_ of why the config file is hidden? It breaks apart the creation of the file from the content, it would perhaps be easier for the user to have both instructions at the same time.

docs/user/guides/creating-config-file.rst
```diff
@@ -0,0 +1,100 @@
...
+The config file is hidden
+*************************
+
+The ``.`` at the beginning of the file indicates that the file is hidden. If 
+the file browser you are using is not configured to show hidden files, you may 
+not be able to see the file after you create it!
+
+If you cannot see ``.readthedocs.yaml`` after creating it, use your favorite 
+search engine to find instructions on showing hidden files for your specific 
+operating system.
```
benjaoming, line 45:
```suggestion
```

I would suggest to remove all this so the instructions of the how-to appear much clearer.

A way to include the tip about finding a hidden file could perhaps be added as a Troubleshooting section at the end of the how-to?

docs/user/guides/creating-config-file.rst
```diff
@@ -0,0 +1,100 @@
...
+
+#############################################
+The required parts of your configuration file
+#############################################
```
benjaoming, line 49: It would be great if the how-to showed some minimal copy-paste friendly examples in a tabbed interface.

I think it's important to really trim out all explanation. It could be enough to explain in one paragraph what an example file does, but users should lookup configuration values in the reference to know more.

For a start, tabs could be: Sphinx, MkDocs, Custom builder

I agree with @humitos's [comment](https://github.com/readthedocs/readthedocs.org/pull/10301#issuecomment-1544576764) that it will be useful to have even more templates. Where and how to maintain them over time is important, but I think we might as well handle that separately (New issue: "Factor out .readthedocs.yaml examples from documentation and test that they work in real builds" - and guess what? we now support monorepo structures with multiple .readthedocs.yaml files, so we can store them in different folders in our docs and setup a live RTD project for each of them!)

```
tabs::

    tab:: Sphinx

        Here is a tab with a Sphinx config file example

    tab:: MkDocs

        Here is a tab with a Sphinx config file example

    tab:: Custom builder

        Example using build.commands

```

benjaoming (commented):
Great initiative @nikblanchet :1st_place_medal: 

I could see the goal of the how-to as _being able to go from "not having a config file" to "having a minimal config file and being ready to read the reference"_ - I think that reflects the intro of your PR description?

I like the narrative voice! But yeah, I guess narrating too much will quickly move the scope of the article into _explanation_, which Diátaxis has a few words to say about :)

I'll try to provide 3 example configuration files and push them into this branch for everyone to review? I'll leave it up to you to include them, but I think the tabbed interface will work really well :innocent:

benjaoming:
@nikblanchet I was wondering if you are still around? I would like to propose to the core team to move forward with this initiative very soon in relation to https://github.com/readthedocs/readthedocs.org/issues/10342.

Since https://github.com/readthedocs/readthedocs.org/pull/10356, we added example code for a .readthedocs.yaml file to the Import Wizard. I would like that code to be maintained in a single location so we maintain all our example code in a single location.

docs/user/guides/creating-config-file.rst
```diff
@@ -0,0 +1,100 @@
...
+   The complete list of all possible ``.readthedocs.yaml`` settings, including 
+   the optional settings not covered in on this page, is found on 
+   :doc:`the configuration file reference page </docs/user/config-file/index.rst>`.
+
+####################################
+Where to put your configuration file
+####################################
+
+The ``.readthedocs.yaml`` file should be placed in the **root directory**. The 
+root directory is the top-most location of your repository.
```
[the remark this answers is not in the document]
agjohnson: The original is closer to what I'd suggest actually. I would add the mention of repository though:

```suggestion
The ``.readthedocs.yaml`` file should be placed in the **root directory** of your project's repository. The 
root directory is the top-most location of your repository.
```

docs/user/guides/creating-config-file.rst
agjohnson: I'd agree here on removal. This topic is fairly secondary to writing the configuration file, and more a GitHub or OS note. I don't think it's important to cover this in a troubleshooting section or anything, we should retain that section just for problems with our configuration file/build process.

docs/user/guides/creating-config-file.rst
agjohnson: We talked about this for our configuration file deprecation as well. No need to hold this PR up with that though, but I think we will end up with some example here as we start to push users towards the configuration file more.

docs/user/guides/creating-config-file.rst
```diff
@@ -0,0 +1,100 @@
...
+**configuration file** called ``.readthedocs.yaml``.
+
+The configuration file like a personalized guidebook that tells the platform 
+what specific settings for your project. Every project is unique, and by using 
+a configuration file, you can tailor the behavior of Read the Docs to match 
+your specific needs.
+
+This how-to guide covers
+#. Where to put your configuration file
+#. The required parts of your configuration file
```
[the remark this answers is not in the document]
agjohnson: "Choosing a minimal example and copy-pasting it" sounds odd to me. I'd aim for clearer language here.

Something more simple would do:

"What to add to your configuration file"

docs/user/guides/creating-config-file.rst
benjaoming: Would you also think that these 3 examples are the most important ones?

I like the approach of having some commented-out configuration options showcased in the examples, too (as the new Import Wizard step does)

agjohnson: Sphinx/Mkdocs are obvious, but we should cover specific doc tools instead of a generic "Custom builder". Again though, not a priority for this initial pass. There is a lot of information we need to cover on this page.

I'm not even sure this is the page we should cover these examples on. The example for each tool is more reference than anything. I would still suggest punting on this concept.

benjaoming: Sounds good. We can use the tab to get people started with another tool and give them the pointers they need to continue using the reference.

nikblanchet:
I'd lost track of this project. Mind if I take this up this upcoming weekend?

benjaoming:
That sounds great @nikblanchet :+1:  I'll do a bit more work on the PR and perhaps it will be ready for review. But we don't have to rush it - it would be nice to wrap it up on Monday though, since we release on Tuesdays. Not sure if you noticed, but we are making the configuration file mandatory from September, so the how-to became a lot more valuable: https://blog.readthedocs.com/migrate-configuration-v2/

docs/user/guides/creating-config-file.rst
benjaoming: The how-to now has a structure where the user is encouraged to start with one of the templates selected in a tabbed interface.

Examples are maintained in separate files that can be reused across our docs and platform.

docs/user/guides/setup/configuration-file.rst
```diff
@@ -0,0 +1,141 @@
...
+
+        You will need to create the output directories and direct your documentation tool to write its outputs into those directories.
+
+        .. literalinclude:: /config-file/examples/custom-commands-echo/.readthedocs.yaml
+           :language: yaml
+           :linenos:
+
+
+Editing the template
+--------------------
```
benjaoming, line 72: We should decide how this section should work.

Two alternatives:

1. Provide a list of sections that are of special interest for the new user (how it is now)
2. Simply provide guidance about the fact that the file has to be edited and then reference _the reference_.

benjaoming: If we go with option 1), we should add some more sections. Currently, we are just explaining two parts :) We should definitely also explain the `build.os` section.

agjohnson:
Is this a draft or ready for review?

benjaoming:
> Is this a draft or ready for review?

I would like to have a second opinion about the question in https://github.com/readthedocs/readthedocs.org/pull/10301#discussion_r1222198421 before a full review :+1:

docs/user/guides/setup/configuration-file.rst
agjohnson: Well, what I think we need to make our reference more usable is some better content _summarizing_ the configuration file.

This page will serve as a good place for the user to jump off into deeper content, and so I might still lean towards being more thorough with the number of configuration options we mention here, but not going into as much detail as we have.

For example, we describe the python key and what it is, and the python.install.requirements key and what it is below. Instead, I think I'd find the most use out of this page if it summarized the Python install options, but pointed to the relevant config reference for all details.

So, instead of the level of detail we have now, something closer to:

```rst
There are several configuration file options for installing your Python dependencies:

- Use :ref:`python.install.package` to install a package using pip
- Use :ref:`python.install.requirements` to install packages from a requirements file
- Use :ref:`build.jobs` to install packages using Poetry or PDM
```

benjaoming: Yup, I totally agree with that.

At the point of reading that section, the user should be familiar with changing content in the config file, so it should just be a question of introducing some important reference items.

benjaoming: Thanks for the copy-paste friendly suggestion :heart:

agjohnson: Oops, well I was writing in pseudo documentation, so I don't know how well it will paste :upside_down_face:

benjaoming: Yes but it's really nice with a clear example, copy-pasting and then adapting the example will be easier :)

benjaoming: Added a structure now. I found it hard to simply condense all possible adjustments in a simple list and opted for headings and `seealso::` boxes.

docs/user/guides/pull-requests.rst
```diff
@@ -14,6 +14,12 @@ If your account is already connected:
 #. Enable the :guilabel:`Build pull requests for this project` option
 #. Click on :guilabel:`Save`
 
+.. tip::
+
+   Pull requests that existed before enabling pull request builds will not be detected immediately.
+   But once you push a new commit to an existing pull request,
+   it will start building.
```
agjohnson, line 21: Good to note this! To make this a little more clear/exact:

```suggestion
   Pull requests opened before enabling pull request builds will not trigger new builds automatically.
   Push a new commit to the pull request to trigger its first build.
```

docs/user/guides/setup/configuration-file.rst
```diff
@@ -0,0 +1,223 @@
+How to add a configuration file
+===============================
+
+As part of the initial setup for your Read the Docs site,
```
agjohnson, line 4:
```suggestion
As part of the initial set up for your Read the Docs site,
```

docs/user/guides/setup/configuration-file.rst
```diff
@@ -0,0 +1,223 @@
...
+you need to create a **configuration file** called ``.readthedocs.yaml``.
+The configuration file tells Read the Docs what specific settings to use for your project.
+
+This how-to guide covers:
+
+#. Where to put your configuration file.
+#. Templates to setup your configuration file from scratch.
+#. The most basic sections in the configuration file that can help you get started.
+#. Next steps...
```
agjohnson, line 13: This summary could use some condensing/clarification, so that it's more user-centric and the user knows exactly what they are getting from this guide:

```suggestion
#. Where to put your configuration file.
#. What to put in the configuration file.
#. How to customize the configuration for your project.
```

There is nuance there from our side of course, but this is what I'd care about as a user reading this guide

docs/user/guides/setup/configuration-file.rst
```diff
@@ -0,0 +1,223 @@
...
+.. note::
+
+    Why is it called .yaml? What is that?
+      The configuration file is a `YAML`_ file. YAML files are a "map": a collection of
+      key-value pairs that can be nested. This is not unlike a JSON file or ``dict``
+      object in Python. This page won't explain the structure of YAML files, but many resources exist
+      online.
+
+
+.. _YAML: https://en.wikipedia.org/wiki/YAML
```
agjohnson, line 40: This section felt too low level for our docs. I think it's good to call out the fact that this is a YAML file, but we shouldn't need to explain YAML as a concept. If the user isn't familiar, they likely aren't familiar with most of the concepts/terms in this block anyways -- map, key-value paris, JSON, dict, etc.

To retain describing the structure without getting too deep into technical jargon, I think something close to this is best:

```suggestion
.. note::

    The Read the Docs configuration file is a `YAML`_ file. You can read more on the structure of this file and all of our configuration options :doc:`in our configuration file reference <>`.
```

docs/user/guides/setup/configuration-file.rst
```diff
@@ -0,0 +1,223 @@
...
+
+.. _howto_templates:
+
+Getting started with a template
+-------------------------------
+
+Here are some variations of the configuration file that can help you get started.
```
agjohnson, line 47: For some more definitive language and clarity:

```suggestion
Here are some configuration file examples to help you get started.
```

docs/user/guides/setup/configuration-file.rst
```diff
@@ -0,0 +1,223 @@
...
+Pick the one that resembles your project the most,
```
agjohnson, line 48: Some more exact language:

```suggestion
Pick an example based on the tool your project is using,
```

docs/user/guides/setup/configuration-file.rst
```diff
@@ -0,0 +1,223 @@
...
+.. tabs::
+
+    .. tab:: Sphinx
+
+        If your project uses Sphinx,
+        we offer a special builder optimized for Sphinx projects.
+
+        .. literalinclude:: /config-file/examples/sphinx/.readthedocs.yaml
+           :language: yaml
+           :linenos:
```
agjohnson, line 60:
```suggestion
           :linenos:
           :caption: .readthedocs.yaml
```

docs/user/guides/setup/configuration-file.rst
```diff
@@ -0,0 +1,223 @@
...
+
+
+    .. tab:: MkDocs
+
+        If your project uses MkDocs,
+        we offer a special builder optimized for MkDocs projects.
+
+        .. literalinclude:: /config-file/examples/mkdocs/.readthedocs.yaml
+           :language: yaml
+           :linenos:
```
agjohnson, line 70:
```suggestion
           :linenos:
           :caption: .readthedocs.yaml
```

docs/user/guides/setup/configuration-file.rst
```diff
@@ -0,0 +1,223 @@
...
+        .. literalinclude:: /config-file/examples/custom-commands-echo/.readthedocs.yaml
+           :language: yaml
+           :linenos:
+
+
+Editing the template
+--------------------
+
+Now that you have a ``.readthedocs.yaml`` file added to your Git repository,
+you should see Read the Docs trying to build your project with the configuration file.
```
agjohnson, line 88: We actually don't make it obvious to users that the build is using a configuration file. We do surface this in the beta dashboard via the debug command output, but that's also the effective configuration file, not the actual file. It might not be clear to the user that the build is doing anything different.

We've talked a number of times of adding a debug flag to commands so we could output this level of information to users without it being noise, but that is still a ways out and probably only a beta dashboard thing still.

I'm not sure exactly what to say here though, other than "you should see the build doing something different" :thinking:

docs/user/guides/setup/configuration-file.rst
```diff
@@ -0,0 +1,223 @@
...
+   If you have added the file in a pull request,
+   you should enable :doc:`pull request builds </guides/pull-requests>`.
+
+Skip: file header and comments
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+There are some parts of the templates that you can leave in place:
+
+Comments
+  We added a lot of comments that explain stuff. They begin with a ``#``.
```
agjohnson, line 105: To reduce and use more exact language:

```suggestion
  We added comments that explain the configuration options and optional features.
  These lines begin with a ``#``.
```

docs/user/guides/setup/configuration-file.rst
```diff
@@ -0,0 +1,223 @@
...
+
+Commented out features
+  Similar to comments, we also use the ``#`` in front of some configuration options that may or may not be relevant.
+  To enable something that's commented out,
+  simply delete the ``#`` and a blank space in front of the YAML code that you want to enable.
```
agjohnson, line 110: This feels similar to the YAML content above, where it feels like too much detail for our user personas. We're expecting our users to know Git and command line tooling, so I feel it's okay to assume they understand commenting.

It's good to explain the _why_ behind what we commented this out though. That does help the user I feel.

docs/user/guides/setup/configuration-file.rst
```diff
@@ -0,0 +1,223 @@
...
+  The current and only supported version is **version 2**.
+
+
+Adjust: ``build.os``
+~~~~~~~~~~~~~~~~~~~~
+
+In our examples,
+we are using Read the Docs' custom image based on the latest Ubuntu LTS release.
+LTS means long-term support,
+meaning that your builds should not break within next many years.
```
agjohnson, line 123: Builds not breaking might be a lofty statement :smile: 

LTS might be the wrong focus here, I'm not sure. We're trying to tell the user that they won't need to change anything for several years by selecting an operating system version. We could skip on describing LTS here entirely, as this is more a security concept perhaps, and Ubuntu package freezing is the main benefit.

docs/user/guides/setup/configuration-file.rst
```diff
@@ -0,0 +1,223 @@
...
+
+However,
+you should pay attention to this field if your current project needs to build on an older version of Ubuntu,
```
agjohnson, line 126:
```suggestion
you should pay attention to this field if your project needs to build on an older version of Ubuntu,
```

docs/user/guides/setup/configuration-file.rst
```diff
@@ -0,0 +1,223 @@
...
+   :ref:`config-file/v2:sphinx`
+     Configuration file reference for configuring the Sphinx builder.
+
+   :ref:`config-file/v2:mkdocs`
+     Configuration file reference for configuring the MkDocs builder.
+
+.. _config_howto_build.tools:
+
+Adjust: Additional build environments?
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
agjohnson, line 181: To keep nomenclature consistent, probably "build tools" here, even though that is a term specific to RTD

```suggestion
Adjust: Additional build tools
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

docs/user/guides/setup/configuration-file.rst
```diff
@@ -0,0 +1,223 @@
...
+
+If you need software from additional platforms (nodejs, rust, golang),
+you can add and enable such an environment so you can install the packages you need.
```
agjohnson, line 184: More nomenclature consolidation:

```suggestion
If you need additional build tools (Node.js, Rust, Go),
you can define the versions of these tools to install as well, using the ``build.tools`` key.
```

docs/user/guides/setup/index.rst
```diff
@@ -14,6 +14,11 @@ The following how-to guides help you solve common tasks and challenges in the se
     If you are connecting a Git repository from another provider (for instance Gitea or Codeberg),
     this guide tells you how to add and configure the webhook **manually**.
 
+⏩ :doc:`Adding a configuration file </guides/setup/configuration-file>`
+    When setting up a new project,
```
agjohnson, line 18: Just for clarity to the user, describe this as all projects, not just new projects:

```suggestion
    To build your documentation project,
```

docs/user/guides/setup/configuration-file.rst
```diff
@@ -0,0 +1,223 @@
...
+The :ref:`Other tools <howto_templates>` template shows an example of how to do that.
+
+.. seealso::
+
+   :doc:`/build-customization`
+     Jump to a full guide explaining how to setup (almost) any documentation tool.
+
+   :ref:`config-file/v2:build.commands`
+     Configuration file reference for custom build commands.
+
```
agjohnson, line 207: I probably have the same feeling here, `build.commands` is not a good next step for novice users. The progression should try to balance giving the user control and us retaining some control, so: supported tool (sphinx/mkdocs) -> build jobs customization (intermediate, configuration file still works) -> build commands (expert, you're on your own).

Perhaps instead of jumping right to "custom tools" we should be pointing towards "custom jobs" and customizing the build process with new jobs. This at least serves us to promote advanced usage and make the user aware of things without through them into the gauntlet of build.commands.

docs/user/guides/setup/configuration-file.rst
```diff
@@ -0,0 +1,223 @@
...
+
+Next steps
+----------
+
+The configuration options that we mentioned in this guide aren't all.
+There are more.
+
+Once you have your configuration file added to your Git repository,
```
agjohnson, line 215: To fix verb agreement with the rest of the sentence, and simplify a bit more:

```suggestion
After you add a configuration file your Git repository,
```

docs/user/config-file/examples/README
```diff
@@ -0,0 +1,12 @@
+README
+
+What is config-file/examples for?
+
+We are collecting all examples of configurations used in documentation and on
+the website here.
+
+We are putting examples into subfolders and calling them `.readthedocs.yaml`
+because each example might turn into a fully runnable example at some point.
```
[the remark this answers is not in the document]
humitos: He, yeah, I commented something pretty similar in https://github.com/readthedocs/readthedocs.org/issues/10406#issuecomment-1584170855

docs/user/config-file/v2.rst
```diff
@@ -12,62 +12,16 @@ Below is an example YAML file which shows the most common configuration options:
...
+        .. literalinclude:: /config-file/examples/sphinx/.readthedocs.yaml
+              :language: yaml
+              :linenos:
 
    .. tab:: MkDocs
 
-      .. code:: yaml
+        .. literalinclude:: /config-file/examples/mkdocs/.readthedocs.yaml
+           :language: yaml
+           :linenos:
```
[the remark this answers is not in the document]
humitos: If we are using the `:caption:` we may not need the `mkdocs/` subfolder and we can name the file just `/config-file/examples/mkdocs.yaml`, right?

docs/user/config-file/examples/README
```diff
@@ -0,0 +1,12 @@
...
+
+Examples should ultimately be possible to install as Read the Docs projects
+using a custom configuration file path.
```
benjaoming, line 12:
```suggestion
```

docs/user/config-file/examples/README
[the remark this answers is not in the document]
benjaoming: Alright, deleting :+1: I wanted to flag this intention because the reason for calling it for instance `examples/sphinx/.readthedocs.yaml` and not say `examples/sphinx.yaml` was to maintain a structure that someone can easily integrate and build.

I'll keep the structure for now and delete what I wrote in the README.

docs/user/guides/setup/configuration-file.rst
benjaoming: Sure - the previous list was a reflection of the section structure. Not a big loss, though.

docs/user/guides/setup/configuration-file.rst
benjaoming: I felt like this was a strong point in @nikblanchet's original howto: Cluing in the user so they don't feel like the guide is over their head if they haven't seen YAML before. You are not expected to have previous experience with YAML in order to follow this guide and copy the examples.

That's why the headline was: "Why is it called .yaml? What is that?" - to meet the user at their level.

I'm not sure about the change here: Do you still want to keep the reference to the Wikipedia article? I added that _again_ because I think it's great to flag that: No, you do not need to have met the concept of YAML before reading this guide, but here is a Wikipedia article explaining it.

benjaoming: If the note is just gonna be a cross-reference to the configuration file reference, I think we should delete it completely, already plenty of those :)

docs/user/guides/setup/configuration-file.rst
benjaoming: I added: Pick an example based on the tool **that** your project is using

docs/user/guides/setup/configuration-file.rst
benjaoming: OH! Great point!! This is actually _a problem_. I just looked at a build, and no not even the raw log has this info. Can't we fix that in a separate issue very quickly? I'm not sure what's the best approach, but I imagine that this shouldn't have astronomic complexity? :)

docs/user/guides/setup/configuration-file.rst
benjaoming:
```suggestion
Commented out features
  We use the ``#`` in front of some popular configuration options.
  They are there as examples,
  which you can choose to enable, delete or save for later.
```

docs/user/guides/setup/configuration-file.rst
benjaoming: Oh haha well, interesting timing you had for that input: https://github.com/readthedocs/readthedocs.org/pull/10413

benjaoming: Since this PR is already open, I think we can resolve this part of the how-to and try to align these two efforts.

docs/user/config-file/examples/mkdocs/.readthedocs.yaml
```diff
@@ -0,0 +1,20 @@
+# .readthedocs.yaml
```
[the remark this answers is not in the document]
benjaoming: removed :fire:

docs/user/config-file/examples/custom-commands-echo/.readthedocs.yaml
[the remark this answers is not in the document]
benjaoming: I think I found a good way forward!

I've replaced the "Other tools" template with a Docusarus template. So it's pretty obvious what to use the template for. The "Other tools" tab now contains more information about how to setup other tools, rather than being a dead-end.

I think this can work well, looking forwards to the next suggestions :+1:

docs/user/guides/setup/configuration-file.rst
benjaoming: I killed this section now!

I think the method that works is to have it all in a tab "Other tools" further up. So all this information should only be seen by users that aren't using Sphinx, MkDocs nor Docusaurus. It might be more _advanced_, but I think that it's better than leading them to a dead-end with no information available.

benjaoming:
Update! I think this will work well:

To the point of avoiding too much complicated information about custom build commands, we hide all that in "Other tools".

We also show a proper example for a tool that's setup with custom build commands. I picked Docusarus to show that Node.js is supported.

image

docs/user/config-file/examples/README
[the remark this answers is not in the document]
agjohnson: That seems fine. I assumed the structure might be redundant without extra files, but perhaps there will be multiple examples per tool or something.

docs/user/config-file/v2.rst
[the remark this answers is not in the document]
agjohnson: I'm assuming we'll probably end up with a structure like this, and I bet we'll end up moving these examples out of the docs path in into the application or something -- once we're reusing the example configuration files directly in our project creation form wizard templates.

docs/user/guides/setup/configuration-file.rst
agjohnson: I do agree with the point about trying to be inclusive with this page, but I also feel familiarity with YAML is part of our base user persona. So, the copy/voice felt like it wasn't talking to a true user, to talking past our normal user -- it would feel the same as having a section explaining what Linux or what a terminal is to me.

I do agree on the cross reference, it isn't horribly important. I added it there to make the note more meaningful and pad out the content. I just updated my suggestion to remove this and link out to better resources. I meant to find a better link than Wikipedia, but agree that linking out is a good resource for this block.

For some comparison, here is the tutorial from CircleCI:
https://circleci.com/docs/config-intro/

It doesn't mention YAML in any detail, but does link out to a separate doc just for explaining YAML in more detail:
https://circleci.com/docs/introduction-to-yaml-configurations/

This is maybe a good addition for later for our docs, though I might argue we'd have trouble maintaining this if we're _just now_ adding a tutorial on our own configuration file :smile: 

From this page, and also from the YAML spec website, YAML is described as:

> YAML is a human-friendly data serialization language for all programming languages.

That's a good sweet spot for our description :dart:

I've updated the suggestion with this and linked to the YAML website. There is a link to the YAML spec, _but_ the overview and introduction are approachable and not frighteningly technical like specs normally are.

docs/user/guides/setup/configuration-file.rst
agjohnson: Yeah, that makes sense, this is indeed a problem we should resolve. I think it's fine to mention this in our docs even if our UI doesn't exactly match.

docs/user/guides/setup/configuration-file.rst
agjohnson: That sounds great :+1:

docs/user/guides/setup/configuration-file.rst
agjohnson: That's fair for now too. We do link out to the build customization page already, so perhaps that is what I am describing as well.

agjohnson:
Well, so I absolutely do want to eventually cover many tools here, or at least a subset of the tools and cover _all_ the tools we can in the application version of this UI.

However, my opinion has not changed on `build.commands`, and it is confusing to point novice users towards a feature where the rest of our configuration file stops working. This feels like a timing issue: both options are still beta features and `build.commands` is an expert level feature. It's great that we can point users towards this solution now, but it's not something I'd point *all* users towards.

*If we are prioritizing the override of `build.jobs.build.{html,pdf,epub}`*, then this is okay temporarily. But otherwise, I don't think we should accumulate novice users using an advanced feature like `build.commands`. We don't have control of what this user is doing and they will be more likely to be lost as they are using a configuration file where they have to do everything.

benjaoming:
> However, my opinion has not changed on build.commands, and it is confusing to point novice users towards a feature where the rest of our configuration file stops working. This feels like a timing issue: both options are still beta features and build.commands is an expert level feature. It's great that we can point users towards this solution now, but it's not something I'd point all users towards.

Hi @agjohnson !

I was just pulling in the branch from another computer. After pulling it in and building, the tabs shown in the screenshot didn't show for me locally. I had to run `make clean` firstly to get a fresh version of the article. I'm not sure what mechanism in Sphinx that causes this, normally articles are refreshed (but TOC-indexed elements might not refresh).

The current state of the docs are supposed to eliminate _expert level_ features, so maybe it's because you'd like a `seealso` to be removed or that I missed something while trying to eliminate the remnants of the previous approach?

agjohnson:
> The current state of the docs are supposed to eliminate expert level features

I still see `build.commands` used in two examples: Docusaurus and the custom tool example.

I am calling `build.commands` expert as:

- Most of the configuration file doesn't function when using build tools
- The user is responsible for every part of the build process
- And we can't offer any control when projects use build.commands
- These features are beta still

I agree it's still helpful to point users towards an example like the Docusaurus example, but not in a guide recommended for novice users.

benjaoming:
I had this in an early comment...

> To give some context:
>
> * On many occasions, I've seen it noted that we need to promote using other documentation tools than Sphinx (and MkDocs) - so people know they are there. This can only really happen by somehow exposing build.commands.
> * NOT communicating this possibility will seem like we ONLY support Sphinx and MkDocs. This leads to fundamental misunderstandings of what the platform can do.

It would be great if you can suggest a way to include people who read this guide and want to use a different documentation tool :+1: 

I apologize in advance for potentially misunderstanding something around the prioritization of custom tools (See also: "Generalize build process patterns to support more tools"). More specifically that I thought having an easy-to-copy-paste template for Docusaurus (or another tool) aligned _very very well_ with high-level priorities.

agjohnson:
> It would be great if you can suggest a way to include people who read this guide and want to use a different documentation tool 

I'm suggesting we wait on this until the tool is somewhat supported. We already have information on unsupported tools in our build customization docs if the user chooses, but it's under a warning that `build.commands` is beta, might break, and _our hosting features won't work_ [^1]. It's a place where novice users are going to encounter *all* of the rough edges of our platform, so not a great impression.

Right now, we're solving a different problem with this page: we don't have any introductory content on our configuration file for novice users.

It's fine to keep the scope of this page to the tools we currently support. Having examples for Sphinx and Mkdocs is a 100% improvement over what we had. We'll add more as we go.

[^1]: Some of our hosting features are partially working in a beta test of the addons JS, but our flyout is missing still.

benjaoming:
Alright, those guides have been removed :+1:

agjohnson:
I've moved further examples off to an issue we can remember to come back to when we're ready:
https://github.com/readthedocs/readthedocs.org/issues/10418

I'm merging this for now, as it's close and all feedback looks to be addressed now. Thanks for the help everyone!

agjohnson (approved):

benjaoming:
Awesome, I look forward to accommodating all the new build features in future work. Thanks for explaining the position :+1:

benjaoming:
Also big thanks to @nikblanchet for the initiative :+1: It went a bit hectic at the end, but the result is super positive -- if you have any additional feedback please do add it, and we can still pick up on matters.

Minor note: We got sphinx-copybutton added, that was really nice, I've seen it play out on different documentation pages and it Just Works® :)

PR #9921, opened 2023-01-19
Docs: Configuration file pages as explanation and reference (Diátaxis)

benjaoming:
**Not ready for review**

Depends on: https://github.com/readthedocs/readthedocs.org/pull/10071

This flags changes that are incoming

Refs: #9746 
Fixes: #8648

---
:books: Documentation previews :books:

- User's documentation (`docs`): https://docs--9921.org.readthedocs.build/en/9921/

- Developer's documentation (`dev`): https://dev--9921.org.readthedocs.build/en/9921/

PR #10303, opened 2023-05-07
Change a few configuration file options from required to not required

### LINKED ISSUE #10302: Docs: configuration file schema required fields are wrong
The configuration file docs note a number of fields, such as
`python.install.requirements` as `required=True`. These configuration fields
are not actually required though, we should audit these.

agjohnson:
These are not required in our configuration specification

Resolves #10302

---

- User's documentation (`docs`): https://docs--10303.org.readthedocs.build/en/10303/

- Developer's documentation (`dev`): https://dev--10303.org.readthedocs.build/en/10303/

### DIFF
--- a/docs/user/config-file/v2.rst
+++ b/docs/user/config-file/v2.rst
@@ -194,7 +194,7 @@ The path to the requirements file, relative to the root of the project.
 
 :Key: ``requirements``
 :Type: ``path``
-:Required: ``true``
+:Required: ``false``
 
 Example:
 
@@ -222,7 +222,7 @@ The path to the package, relative to the root of the project.
 
 :Key: ``path``
 :Type: ``path``
-:Required: ``true``
+:Required: ``false``
 
 The installation method.
 
@@ -297,7 +297,7 @@ conda.environment
 The path to the Conda `environment file <https://conda.io/projects/conda/en/latest/user-guide/tasks/manage-environments.html>`_, relative to the root of the project.
 
 :Type: ``path``
-:Required: ``true``
+:Required: ``false``
 
 build
 ~~~~~


benjaoming (approved):
LGTM :+1: 

This is a very good case for having more automatic apidocs-like generation of the configuration file schema. But we'd probably have to rework a lot of our schema parsing for that to happen :)

Btw some more changes are in the pipeline in #9921

humitos (commented):
I think the way the documentation is written makes this to be tricky.

For example, `conda` is not required at all. However, if you use the `conda` key, then `conda. enviroment` becomes _required_. I think this is what the documentation tries to communicate, but it's not clear.

agjohnson:
@humitos aye, this is what I figured as well. I think this PR is at least more correct up front, but do agree we need a different way of describing when fields conditionally become required.

I could make this instead `Required: when using conda key` or something similar perhaps? Any preference?

humitos:
I don't have any preference currently. I think we need to come back to this page anyways for a bigger refactor since it's one of the most linked page and we recommend it a lot. So, I think we should do some small research and se how other projects are presenting a config file reference and probably copy the one we like the most.

This page has really good information on it, but it just needs to be re-structured in a more clearer way only.

benjaoming:
> I could make this instead Required: when using conda key or something similar perhaps? Any preference?

I think that sounds good to get this PR in and maintain correctness :+1: 

It would be super nice to do a refactor of this page! What I currently like about it is that everything has an anchor link so you can reference all options directly - and that everything is in the same page.

What I don't like is that the schema's nested structures are unclear and it lacks more copy-paste friendly examples.

docs/user/config-file/v2.rst
```diff
@@ -297,7 +297,7 @@ conda.environment
 The path to the Conda `environment file <https://conda.io/projects/conda/en/latest/user-guide/tasks/manage-environments.html>`_, relative to the root of the project.
 
 :Type: ``path``
-:Required: ``true``
+:Required: ``false``
```
benjaoming, line 300:
```suggestion
:Required: when using conda key
```

agjohnson: I'll leave this change for later for now, as we have a number of fields we should do this for.

Issue #7084, opened 2020-05-15
Provide Github (etc) commit status for builds on branches / tags


Cadair:
This is not in relation to PR builds.

It would be really nice if after a build is triggered on a commit (i.e. a PR is merged or a tag pushed) if RTD sent a commit status back to the VCS site that triggered it. This would allow you to see the status of the RTD build along side all the other CI checks on that commit.

A bonus would be supporting the GitHub checks API to allow restarting the build etc.

benjaoming:
Another point: A push can contain several commits. We always build the last commit.

benjaoming:
New implementation looking good!

![image](https://github.com/readthedocs/readthedocs.org/assets/374612/95893048-58f0-4f4a-a5aa-1a288dd61d94)

![image](https://github.com/readthedocs/readthedocs.org/assets/374612/10e7e46d-a220-4c98-9885-d02667639aff)

Test with a build failure:

![image](https://github.com/readthedocs/readthedocs.org/assets/374612/ee0d868b-f049-4d69-9f34-08d06e63b2be)

Test with build success:

![image](https://github.com/readthedocs/readthedocs.org/assets/374612/4c142af7-cceb-4718-a0ea-eae6c1eb03fa)

humitos:
I'm going to close this issue because we weren't able to prioritize in the past year and we are not planning to work on it soon. We can revisit in the future when working on expanding GitHub features.

PR #10320, opened 2023-05-15
Build: Handle commit hash on incoming webhooks and send commit status on all builds

benjaoming:
Fixes https://github.com/readthedocs/readthedocs.org/issues/7084

## Notes for reviewers

**_I hope you're okay about starting the review a bit high-level:_**

* I wrote my notes for the implementation, starting here: https://github.com/readthedocs/readthedocs.org/issues/7084#issuecomment-1547809333

* Please make sure to read this comment about issues that I suggest to handle separately https://github.com/readthedocs/readthedocs.org/pull/10320#issuecomment-1550324581

* The implementation should maintain fallback support for previous behavior, i.e. "no commit ID supplied": If there for some reason exists a scenario where the commit ID isn't provided in the webhook event (tag pushes or we don't detect it correctly?), then we will just log the event on warning level but continue as before. I prefer that we go this way, since it's hard to guarantee that we have understood all potential events that are pushed to the webhook.

* Noting that the risks of out-of-order webhook events and builds are very low and already exist to some extend, I prefer an approach where we will discuss and solve this separately.

## Work log

* [x] Pre-analysis to figure out the correct solution
* [x] Implementation and local Q&A with GitHub OAuth App
* [x] Updates to tests w/ some new scenarios
* [x] Understand a bit more potential cases we need to investigate.
  * Out of order webhook events
  * Force pushes and deleted commits
* [x] Understand if auto-cancel feature works on existing PR push events where commit IDs were already used and how/if that worked.

image

image

Test with a build failure:

image

Test with build success:

image

### DIFF
--- a/readthedocs/api/v2/views/integrations.py
+++ b/readthedocs/api/v2/views/integrations.py
@@ -732,14 +732,16 @@ class BitbucketWebhookView(WebhookMixin, APIView):
                 changes = data['push']['changes']
                 branches_and_commits = []
                 for change in changes:
-                    old = change['old']
                     new = change['new']
-                    # Fix the 'target', which should map to the latest commit pushed
-                    # if it's available.
-                    # See: https://support.atlassian.com/bitbucket-cloud/docs/event-payloads/
-                    commit = change.get("new", {}).get("target", {}).get("hash", None)
-                    # Normal push to master
-                    if old is not None and new is not None:
+
+                    # Consider changes only in the 'new' part.
+                    # Builds are only triggered when a branch is created and a version
+                    # matches the branch name.
+                    if new:
+                        # push.changes.new.target:
+                        # The details of the most recent commit after the push
+                        # See: https://support.atlassian.com/bitbucket-cloud/docs/event-payloads/
+                        commit = new.get("target", {}).get("hash", None)
                         branches_and_commits.append((new["name"], commit))
                 # BitBuck returns an array of changes rather than
                 # one webhook per change. If we have at least one normal push
@@ -750,6 +752,8 @@ class BitbucketWebhookView(WebhookMixin, APIView):
                         self.project,
                         branches_and_commits,
                     )
+                # No matching versions found for the push event, might be because of
+                # branch creation, hence we will update versions.
                 log.debug("Triggered sync_versions.")
                 return self.sync_versions_response(self.project)
             except KeyError:
--- a/readthedocs/api/v2/views/integrations.py
+++ b/readthedocs/api/v2/views/integrations.py
@@ -644,14 +644,10 @@ class GitLabWebhookView(WebhookMixin, APIView):
             except KeyError:
                 raise ParseError('Parameter "ref" is required')
 
-            try:
+            commit = None
+            # Commits are not present in GITLAB_TAG_PUSH
+            if "commits" in self.data and self.data["commits"]:
                 commit = self.data["commits"][0]["id"]
-            except IndexError:
-                raise ParseError("No commits found in push event.")
-            except KeyError:
-                raise ParseError(
-                    'Push event received with either no "commits" or no "id" key for commit.'
-                )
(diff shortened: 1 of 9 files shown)


benjaoming:
# Some new theoretical issues

All of these are theoretically possible, but they require that either webhook events are received _out of order_ or concurrent builds finish _out of order_.

## Out of order webhook events

Our current build process in `doc_builder.director` will checkout the commit that's stored on the build, and now we have started setting commit ids on builds received from webhook events so this behavior will be the new default. It's necessary because it's the same commit that we are going to send a status back to, and we need to send a "pending" status already before the build is triggered.

## Race conditions in builds

If we start several builds, an older process can succeed in checking out an older commit and finish before a newer build. This is mostly averted by correctly canceling old builds (see next section).

## Canceling old builds

In https://github.com/readthedocs/readthedocs.org/pull/9549, we started to cancel currently running builds when a new build is received based on the RTD version. That's okay since these builds weren't sensitive to the commit they were building - they were unaware, so any new build process would just checkout the latest branch/tag identifier.

The previous approach probably only had one shortcoming: Assuming that concurrent builds won't end up in race conditions. So the new behavior isn't too different, it's just a bit more sensitive to webhook event order or in case our event queue processes events out of order.

## Force pushes and deleted commits

If a commit disappears between the time that a webhook event is received and a build is launched, the build process should fail since the commit doesn't exist. That's fine. It will just deliver an invalid status and fail.

We could anticipate that a new webhook event would show up. It ideally won't be canceled since it should arrive after the previously failed webhook event.

# Potential solutions

There are two solutions that I think are important to address the above. I think they can both be solved separately from this feature, since they are already well-known issues, and to some degree also theoretically possible in our current logic.

## Feature 1: Discard webhook push events based on timestamp

Most webhook events have timestamps from the Git provider so we don't need to look at commit history. Instead, we can store the timestamp, check if we have processed a newer event and discard the event in that case.

## Feature 2: Auto-cancel current builds on the same version: Use commit hash, too

When we auto-cancel currently running builds, we should check if the commit hashes of all active builds are part of the new build's commit history.

**Currently**, active builds are canceled in the `prepare_build` task.

**Instead**, a build process could potentially receive the commit ids of all concurrently running builds with the same version so it can emit a cancel request if they are included in the new build's git history. This is more accurate.

humitos:
I just saw that I was requested a review on this PR and tried to quickly review it. However, I saw it touches multiple parts of the build process to pass the commit around and there are some potential issues and solutions written down. I'm a little concern regarding the amount of code we are adding here together with these potential issues compared with the value this feature adds to the platform at this moment. I think it's a "nice to have" at this point and I find it hard to prioritize currently.

benjaoming:
@humitos there's a lot of work that went into preparing this change. I'll be happy to go over it with you on our call tomorrow, if we can extend it a bit. But please read the PR description and related comments.

I'm especially interested in going over this part with you:

> Understand if auto-cancel feature works on existing PR push events where commit IDs were already used and how/if that worked.

benjaoming:
> Understand if auto-cancel feature works on existing PR push events where commit IDs were already used and how/if that worked.

I looked into this now, and we should have exactly the same functionality preserved, including the out-of-order issues :)

Auto-canceling will continue to work pr version. So if a new push event adds a commit to the `main` branch, then any existing builds for `latest` will be cancelled. This seems fine since the result of the newer commit will always overwrite (upload) the `latest` version, so there's no need to finish the running build.

benjaoming:
While discussing this, I think we came to a conclusion that we need some more factors to prioritize this change:

* Reasons to track the commit id on builds (perhaps integrity/correctness)
* Additional features that need the commit id

humitos:
This idea is great, but considering the amount of work required here and the complexity added to the code base, I think we won't be able to make it for now. I'm closing this PR, but we can recover this work once we can prioritize it.

PR #10325, opened 2023-05-18
Organizations: Organization chooser page

benjaoming:
Fixes https://github.com/readthedocs/readthedocs-corporate/issues/938

If the user has exactly `1` organization, the _organization chooser_ will automatically redirect to the requested page with the `slug` path argument replaced with the slug of the user's single organization.

If the user has `>1` or zero organizations, visiting the _organization chooser_ displays a list of the requested view name with URLs for each organization's slug. Notice how the organization chooser page takes the urlconf name as a URL parameter `next_name`. It can also handle `next_querystring`.

image

### TODO

* [x] Check up on URL path kwargs and ensure they are included if relevant. **UPDATE**: kwargs in URL paths are incompatible with unspecific URLs, so we just raise a 404 and log a warning.
* [x] Write tests
* [x] Remove generic URL scheme

### Removed

This PR is no longer titled "Support URL scheme for unspecific organization w/ chooser page or automatic redirect"

This change allows URLs to be specified without knowledge of an organization's slug, moreover in cases where a user has more than 1 organization.

Allowing such URLs means easy reuse of email templates.

A reusable URL for organization-specific pages uses `-` as the slug.

The URL for different organization pages look like this:

* Organization detail: `devthedocs.com/organizations/-/`
* Members admin: `devthedocs.com/organizations/-/members/`
* Subscription page: `devthedocs.com/organizations/-/subscription/`


benjaoming:
Writing some more test cases but almost done...

benjaoming:
Added more testing.

readthedocs/organizations/views/base.py
```diff
@@ -20,27 +18,14 @@
     Team,
     TeamMember,
 )
-
-
-class CheckOrganizationsEnabled:
```
ericholscher, line 25: I'm confused why this was reimplemented as a decorator instead of keeping it at a mixin?

readthedocs/organizations/templates/organizations/organization_choose.html
```diff
@@ -0,0 +1,51 @@
+{% extends "base.html" %}
+
+{% load i18n %}
+{% load gravatar %}
+{% load organizations %}
+
+{% block title %}{% trans "Before continuing, choose an organization..." %}{% endblock %}
```
agjohnson, line 7: The additional text isn't 
I'd say this sounds better without the passive clause in front, here and in the heading.

```suggestion
{% block title %}{% trans "Choose an organization" %}{% endblock %}
```

readthedocs/organizations/templates/organizations/organization_choose.html
```diff
@@ -0,0 +1,51 @@
...
+{% block content %}
+  <div class="col-major organization-major">
+    <div class="module organizations organizations-list">
+      <div class="module-wrapper">
+
+        <h2>{% trans "Before continuing, choose an organization..." %}</h2>
+
+        <p>
+          {% trans "You are a member of several organizations." %}<br>
+          {% trans "Click the organization that you want to use:" %}
```
agjohnson, line 20: A few notes here:

- This should be a single, trimmed blocktrans block
- I'd drop the additional `<br>`, it looks unnecessary in display
- Also "click" -> "select"

readthedocs/organizations/templates/organizations/organization_choose.html
```diff
@@ -0,0 +1,51 @@
...
+                      </a>
+                    </li>
+                  {% endfor %}
+                </ul>
+
+              </div>
+            </div>
+            <!-- END organization list -->
+          {% else %}
+            <p>You aren't currently a member of any organizations</p>
```
agjohnson, line 44: Not translated here, but this should never show anyways. Perhaps the if statement isn't important?

readthedocs/organizations/tests/test_views.py
```diff
@@ -376,3 +376,129 @@ def test_organization_signup(self):
...
+                "organization_choose", kwargs={"next_name": "organization_members"}
+            ),
+        )
+
+    def test_choose_organization_edit(self):
+        self.assertEqual(Organization.objects.count(), 0)
+        resp = self.client.get(
+            reverse("organization_choose", kwargs={"next_name": "organization_edit"})
+        )
+        self.assertContains(resp, "You aren't currently a member of any organizations")
```
agjohnson, line 504: :100: on the test coverage!

readthedocs/organizations/views/decorators.py
```diff
@@ -0,0 +1,105 @@
...
+from django.urls import NoReverseMatch
+from django.utils.http import urlencode
+from structlog import get_logger
+
+from readthedocs.organizations.models import Organization
+
+log = get_logger(__name__)
+
+
+def redirect_if_organization_unspecified(dispatch):
```
agjohnson, line 15: What is the purpose of these two decorators? On #938 we discussed just adding a single view for an organization selection (and later a project selection) page. It doesn't seem necessary to decorate any other views or wrap them with this logic.

We don't need this on all of our URLs, only specific ones that we link to commonly, like the subscription edit URL. So, to me, this approach seems like it adds more complexity than problems it is solving.

Is there some other reason that this was implemented?

readthedocs/organizations/views/public.py
```diff
@@ -8,21 +8,27 @@
...
 
 
-class OrganizationTemplateView(CheckOrganizationsEnabled, TemplateView):
+class OrganizationTemplateView(TemplateView):
 
     """Wrapper around `TemplateView` to check if organizations are enabled."""
 
+    @redirect_if_organizations_disabled
+    def dispatch(self, *args, **kwargs):
+        return super().dispatch(*args, **kwargs)
```
benjaoming, line 30: @ericholscher for instance here, there is only the `redirect_if_organizations_disabled` active, that's why it wasn't possible to expand the `dispatch` method of the former `CheckOrganizationsEnabled` without adding additional mixin classes or conditional statements and such.

Authentication is hard, it gets harder by adding inheritance complexities. For instance, an additional superclass with a `dispatch` method could silently disable intended logic.

agjohnson (commented):
I'm quite excited we're taking this change on, I've wanted it for a while! I feel like this is close, but would be very close for merge without the decorator change.

readthedocs/organizations/views/base.py
agjohnson: To me, it doesn't really seem like a decorator change here is necessary at all. The view that we wanted to implement was just the redirect view. With that implemented, the magic `-` URLs all seem redundant. I'd stick to using the redirect view too, it's nice an explicit where a magic URL is not.

readthedocs/organizations/views/decorators.py
benjaoming: With a decorator, you can decorate the `dispatch` method on exactly the CBVs that requires the functionality.

With a mixin, you need to somehow understand whateever dispatch functions that are used in the class hierarchy and modify those. We already have a mixin that modifies the `dispatch` method :exploding_head: 

This is where view decorators are perfect.

benjaoming: Also, you're more than welcome to suggest exactly the views that should have this functionality, then I can add the decorator on exactly those views :100:

readthedocs/settings/base.py
```diff
@@ -46,7 +46,7 @@ class CommunityBaseSettings(Settings):
     DEFAULT_AUTO_FIELD = 'django.db.models.AutoField'
 
     # Debug settings
-    DEBUG = True
+    DEBUG = False
```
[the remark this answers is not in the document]
benjaoming: Ay, keeps happening :disappointed: 

But I've added an invoke option for it so I can stop doing it :) => https://github.com/readthedocs/readthedocs.org/pull/10265

readthedocs/organizations/views/decorators.py
agjohnson: > With a decorator, you can decorate the dispatch method on exactly the CBVs that requires the functionality.

I wasn't asking about the code mechanics here of a mixin vs decorator, I was asking more about why we're adding this feature to every view.

> Also, you're more than welcome to suggest exactly the views that should have this functionality

Well, what I described above is that I feel we don't need this pattern at all. We only talked about implementing the redirect view, and this is enough for our use case.

readthedocs/settings/base.py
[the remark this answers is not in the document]
agjohnson: What's the purpose for setting debug off? Our dev environments should all be about equal, and though I haven't needed debug disabled for any changes recently, some of the dashboard work does expect this for local development.

readthedocs/organizations/views/base.py
benjaoming: It's another way to implement the same thing. But I wanted to have real URL paths instead of query string arguments prone to typos.

With this pattern, you simply swap the slug for a `-` whenever you want to make it generic/sharable. With the other pattern, you have more manual work and a new margin for error.

readthedocs/organizations/views/decorators.py
benjaoming: > I wasn't asking about the code mechanics here of a mixin vs decorator, I was asking more about why we're adding this feature to every view.

I looked at the views and their URL patterns and it seemed to be possible to add to all the views. It seems super nice that we can now create a URL for any organization-based settings page when we need to do support and point people to a specific page?

readthedocs/organizations/templates/organizations/organization_choose.html
benjaoming: It's copied from another template, I'll mark both for translation. I'm confused by the inconsistency of marking strings for translation in these templates :)

readthedocs/organizations/templates/organizations/organization_choose.html
benjaoming: image

benjaoming: I was unsure if you wanted to also shorten the headline here?

readthedocs/organizations/views/base.py
agjohnson: We're not talking about using these URLs in many places, so I don't think that needs to a concern in our implementation. The use case for this so we can add a link to some of the pages from documentation and support response templates.

readthedocs/organizations/templates/organizations/organization_choose.html
```diff
@@ -0,0 +1,49 @@
...
+
+          {% if organization_list %}
+            <!-- BEGIN organization list -->
+            <div class="module-list">
+              <div class="module-list-wrapper">
+
+                <ul>
+                  {% for org in organization_list %}
+                    <li>
+                      <a href="{% url next_name slug=org.slug %}{% if next_querystring %}?{{ next_querystring }}{% endif %}">
```
benjaoming, line 31: @agjohnson I'm not sure if this went unnoticed? But by parameterizing the chooser page with the url name of the page we want to redirect to, we can generate all the individual links using `next_name` and `org.slug`.

agjohnson: Hrm, not sure I follow, is there an issue with this approach? I think using `url` like this was I think the intention with this view from the underlying discussion. If there was a question here, I think this looks good :+1:

readthedocs/organizations/views/decorators.py
agjohnson: That was the intention with the redirect view, so I'm not sure we need two different ways of doing this. The redirect view added here is explicit and isolated and doesn't involve altering all our views, so I think dialing this work back is still the correct call.

readthedocs/organizations/templates/organizations/organization_choose.html
agjohnson: Yeah, noted above that the title and heading here could both be shortened.

readthedocs/organizations/templates/organizations/organization_choose.html
benjaoming: If you are hand-crafting a `/redirect/?next=view_name`, wouldn't you find it cumbersome to have to have knowledge about the Django urlconf's view name? And if you think about alternatives, aren't they just as annoying (manipulating URL paths either manually or with some automated logic to inject slugs)?

agjohnson: I would not find this cumbersome as we are not talking about doing this frequently. We are only discussing doing this a few times, for docs and support templates.

I feel like I'm looping though. I think @ericholscher and I have both picked up on the same issue with this PR and without the decorator change this PR would be about ready to merge.

benjaoming: I don't think that's a wise choice. You throw away a perfect generic solution that will immediately solve the practicality of "how to create a correct redirect link".

It can also be used for many other purposes, such as adapting it to "the organization does not exist or you don't have access to this organization, please choose from the list of organizations that you are a member of"

benjaoming: Note that I wasn't trying to scope creep this, I was merely trying to solve it in a way that I believe is most feasible for the immediate use case.

readthedocs/organizations/templates/organizations/organization_choose.html
benjaoming: Alright, done

readthedocs/organizations/views/decorators.py
ericholscher: > With a mixin, you need to somehow understand whateever dispatch functions that are used in the class hierarchy and modify those. We already have a mixin that modifies the dispatch method 🤯

This is my point. We already have an established pattern in the codebase of using mixins to achieve this result. Adding a new way of doing it is bad, regardless of whether there is a small technical benefit to doing it. My thought is that we should just implement the new redirect logic as a mixin.

The question around whether to implement this functionality as a mixin vs. a dedicated view is another question entirely. I tend to agree this implementation adds a lot of complexity compared to the redirect view we discussed in the issue, and was my understanding of what we were going to implement.

readthedocs/organizations/views/base.py
ericholscher: I agree with @agjohnson that this PR adds a lot of complexity to save a small amount of time when generating one of these URLs? It seems like a poor trade off. I'd much prefer an explicit redirect view that is scalable and easy to understand, than a quite complex decorator implementation that introduces ambiguity to our URL patterns that aren't consistent across various types of URLs (eg. the magic `-` slug wouldn't work for many places where we use slugs).

readthedocs/organizations/views/decorators.py
benjaoming: I think that the URL rewriting that's required in order to inject slugs of organizations has been underestimated in the "let's just have a redirect page" scheme. This change includes the redirect page but leverages existing views with slugs to generate the redirect page's parameters.

benjaoming: In the example with `?view=organization_edit`, this requires someone wanting to share a URL to look up the exact view name in the source code's Django urlconf, which I thought was very very nice to avoid.

benjaoming: But yes, we can peel off the layer with the decorator function and write the view name manually for the redirect page. I just thought it'd be nice to not have to do that :)

benjaoming: That also means sharing URLs directly to the redirect page rather than the more convenient URLs with just `-` as slug.

readthedocs/organizations/templates/organizations/organization_choose.html
humitos: As a general rule, they should always be translated. If they are not, it because we forgot by mistake.

readthedocs/organizations/views/decorators.py
humitos: I think this is a case where we value complexity/usability ratio. Considering this was originally thought as an internal view that we will only use while doing support, I'm fine reducing the complexity as much as possible at expenses of keeping the usability a little more complex (e.g. we need to know the internal Django view name to put in the URL query string)

I'm fine considering an easier to use and more complex implementation in the future if we find it's required or we find that the initial redirect view is not covering all the use cases; but I think we are going to far here for the initial implementation.

Even if this implementation would have been better than the one discussed in the original issue, it makes me some noise that something different than what we agreed on was implemented. I'd like to have discussed the new idea in advance before implementing it and having this discussion after that.

readthedocs/organizations/templates/organizations/organization_choose.html
benjaoming: I removed the feature. I want to move on from this discussion. I think it was a nice feature but I'm not willing to defend its usefulness more than what I already did. Seems there's a "popular" demand to not have it so that's fine :+1:

agjohnson (approved):
Looks great! :+1:

readthedocs/organizations/views/private.py
```diff
@@ -69,6 +71,37 @@ def get_queryset(self):
...
 
 
+class ChooseOrganization(ListOrganization):
+    template_name = "organizations/organization_choose.html"
+
+    def dispatch(self, request, *args, **kwargs):
+
+        # We're overwriting a mixin dispatch()
+        if not settings.RTD_ALLOW_ORGANIZATIONS:
+            raise Http404
```
stsewd, line 81: You should be able to override the `get` method instead of `dispatch`, that way you don't need to duplicate this.

benjaoming: this is true :+1: fixed so we don't have duplication :+1: 

will add this to my lightning talk draft "decorators are friends of CBVs" -- i.e. not having to pick and choose between dispatch, get, get_context_data etc in order to add functionality when class inheritance is complicated :)

stsewd: yeah, I also don't like much using mixins, most of the time they can be just functions.

PR #10340, opened 2023-05-23
Requirements: upgrade DDT to avoid an issue

humitos:
See https://github.com/jazzband/django-debug-toolbar/pull/1770

We hit this while doing a deploy.


humitos:
I'm merging this. It's already deployed.

